PR #4, opened 2019-07-24
Fix `ActiveSupport::Deprecation::DeprecatedConstantProxy` instances

paracycle:
The usage of `ActiveSupport::Deprecation::DeprecatedConstantProxy` results in:
```
[1] pry(main)> class B; end
=> nil
[2] pry(main)> A = ActiveSupport::Deprecation::DeprecatedConstantProxy.new("A", "B")
=> B
[3] pry(main)> A.name
DEPRECATION WARNING: A is deprecated! Use B instead. (called from <main> at (pry):3)
=> "B"
[4] pry(main)> Module.instance_method(:name).bind(A).call
=> "A"
```
thus breaking our assumptions about treating aliases as constants which are modules whose name do not match the constant name.

So this PR adds a specific check for `ActiveSupport::Deprecation::DeprecatedConstantProxy` usages and gets the `name` of the `target` class as the "real" name of the constant.

### DIFF
--- a/lib/tapioca/compilers/symbol_table/symbol_generator.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_generator.rb
@@ -541,12 +541,34 @@ module Tapioca
         end
 
         sig { params(constant: Module).returns(T.nilable(String)) }
+        def raw_name_of(constant)
+          Module.instance_method(:name).bind(constant).call
+        end
+
+        sig { params(constant: Module).returns(T.nilable(String)) }
         def name_of(constant)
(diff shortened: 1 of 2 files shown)


paracycle:
I've also added a test-case with pre-Rails 6.1 implementation of the proxy class. It seems https://github.com/rails/rails/commit/48c0abb4748c30adeecc17058f0220e87bcbe84a#diff-7372a3e3a3d4981c75cdd55345a1b13a changed the behaviour on Rails `master` recently.

PR #5, opened 2019-08-14
Generate no errors after running `tapioca` on a fresh Rails project

paracycle:
The current version of `tapioca` does not try to load the Rails project and misses a lot of definitions/types coming from Rails. Those would have to be added by had to the post-require file which is not ideal.

This PR adds the ability to require the Rails, Rake and the `config/application` file to make sure all those things are loaded before we generate RBI files.


lib/tapioca/generator.rb
```diff
@@ -9,6 +9,11 @@ class Generator < ::Thor::Shell::Color
     extend(T::Sig)
 
     DEFAULT_OUTDIR = "sorbet/rbi/gems"
+    DEFAULT_OVERRIDES = T.let({
+      # ActiveSupport overrides some core methods with different signatures
+      # so we generate a typed: false RBI for it to suppress errors
+      "activesupport" => "false",
+    }, T::Hash[String, String])
```
peterzhu2118: I think this hash should be frozen

lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -131,11 +131,7 @@ def compile_object(name, value)
         def type_name_of(value)
           klass = class_of(value)
 
-          type_name = public_module?(klass) && name_of(klass) || "T.untyped"
-          # Range needs to be processed separately to be put in the T::Range[] form
-          type_name = "T::Range[#{type_name_of(T.cast(value, T::Range[T.untyped]).first)}]" if klass == Range
```
peterzhu2118: Is there some change to Sorbet that prefers `Range` over `T::Range`? It still says to use `T::Range` in the [docs](https://sorbet.org/docs/stdlib-generics#docsNav).

Morriar: I guess this fix comes from https://github.com/sorbet/sorbet/pull/1309 ?

paracycle: Ideally we would like to generate the untyped version of these `Enumerable` generic types since we can't consistently be able to guess the type of the type-members. (Here we were looking at the type of the first entry and assuming it was the same for the whole, which makes sense for `Range` but not for other `Enumerable` types)

So the plan was to always use the "Ruby" type names, but Sorbet didn't support it for `Range`. We've since changed that in https://github.com/sorbet/sorbet/pull/1309 and as you can see from the comments on that PR, it is perfectly fine to use the naked types in files that are `typed: true`.

This change removes complexity from the generator so I prefer it. However, it does not strictly belong in this PR. But I'll let it slide for now. :)

PR #6, opened 2019-08-28
Make file matching more accepting

paracycle:
It is sometimes the case that `pry` returns a match that does not live in the gem folder, but our `ConstantLocator` can locate it to the correct file. A case in point is `ActiveStorage::Current`, which, as far as `pry` is concerned is defined in the file `activesupport/lib/active_support/current_attributes.rb` but is actually defined in `activestorage/app/models/active_storage/current.rb`, which is properly located by our `ConstantLocator` (thanks to `TracePoint`).

Thus, instead of falling back from one to the other, we need to consider the whole set of results as one and see if any of them match a file under the current gem folder.


paracycle:
This change is tested on core. We do end up generating some spurious constants because of this increased permissivity, but we are able to generate more valid constants.

PR #7, opened 2019-09-10
Pre-release minor refinements

paracycle:
- Make the default post-require script `sorbet/tapioca/require.rb`
- Add an `init` command to create a basic `sorbet` folder structure with:
   - `sorbet/config` with contents `--dir=.`
   - `sorbet/tapioca/require.rb` with empty contents

### DIFF
--- a/lib/tapioca/cli.rb
+++ b/lib/tapioca/cli.rb
@@ -5,6 +5,8 @@ require 'thor'
 
 module Tapioca
   class Cli < Thor
+    include(Thor::Actions)
+
     class_option :prerequire,
                   aliases: ["--pre", "-b"],
                   banner: "file",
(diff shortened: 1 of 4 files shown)


lib/tapioca/cli.rb
```diff
@@ -29,6 +32,16 @@ class Cli < Thor
                   banner: "gem:level",
                   desc: "Overrides for typed sigils for generated gem RBIs"
 
+    desc "init", "initalizes folder structure"
```
Morriar: You should update the README with this new action.

lib/tapioca/cli.rb
```diff
@@ -29,6 +32,16 @@ class Cli < Thor
...
+    def init
+      create_file(Generator::SORBET_CONFIG, skip: true) do
+        "--dir\n."
+      end
+      create_file(Generator::DEFAULT_POSTREQUIRE, skip: true) do
+        "# Add your extra requires here"
```
Morriar: This should also be explained in the README, furthermore since it's not something *required* by the official `srb` tool.

paracycle: Indeed. I'll write a more detailed documentation after this PR.

PR #8, opened 2019-09-16
Fix typo in README.md

bugtender:
Hi, just a quick typo fix in README.md

And thanks for building this awesome gem.👍


rafaelfranca:
Thank you for the pull request. I think we also need to change here https://github.com/Shopify/tapioca/blob/c22c2215623939ee67000beef25764d9c5b0abe4/lib/tapioca/cli.rb#L35

bugtender:
@rafaelfranca Thanks. I fixed it and force-pushed with new commit.

Issue #10, opened 2019-09-25
tapioca generate creates extend(::T::Private::Methods::SingletonMethodHooks) lines in rbi files


timgaleckas:
This breaks srb and they should be filtered out.

PR #11, opened 2019-09-25
filter out T::Private modules

timgaleckas:
fixes https://github.com/Shopify/tapioca/issues/10


paracycle:
Thank you! I added some tests to make sure this doesn't regress and merged it.

timgaleckas:
Thanks. Perfect.

PR #12, opened 2019-09-25
Convert pre/post-require paths to absolute paths

paracycle:
We had a problem where the default post-require script `sorbet/tapioca/require.rb` was creating a problem with `require` since `require("sorbet/tapioca/require.rb")` was searching for a file in the include path. Converting the require path to an absolute path fixes this problem.

The tests have been updated to perform a `tapioca init` before each run to ensure a `sorbet/tapioca/require.rb` file exists so that we actually do test this case and catch the error.

### DIFF
--- a/lib/tapioca/loader.rb
+++ b/lib/tapioca/loader.rb
@@ -12,14 +12,14 @@ module Tapioca
 
     sig { params(initialize_file: T.nilable(String), require_file: T.nilable(String)).void }
     def load_bundle(initialize_file, require_file)
-      require(initialize_file) if initialize_file && File.exist?(initialize_file)
+      require_helper(initialize_file)
 
       load_rails
       load_rake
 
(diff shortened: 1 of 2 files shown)


lib/tapioca/loader.rb
```diff
@@ -29,6 +29,15 @@ def load_bundle(initialize_file, require_file)
     sig { returns(Tapioca::Gemfile) }
     attr_reader :gemfile
 
+    sig { params(file: T.nilable(String)).void }
```
Morriar: Nilable for client convenience?

paracycle: Yeah. mostly. This method checks if the passed argument is `nil` and short-circuits so that the call-site can just ignore that detail.

Morriar:
LGTM :shipit:

PR #13, opened 2019-09-25
These are the changes I needed to make to get it to work on 2.4.6

timgaleckas:
This fixes API differences between ruby 2.4 and >= 2.5. I'm not too happy about the hack to remove the self.inspect differences between 2.4 and others but it seemed like a small contained hack.

Let me know what you think.

### DIFF
--- a/spec/tapioca/compilers/symbol_table_compiler_spec.rb
+++ b/spec/tapioca/compilers/symbol_table_compiler_spec.rb
@@ -5,6 +5,19 @@ require "spec_helper"
 require "pathname"
 require "tmpdir"
 
+RSpec.configure do |config|
+  # Some tests are not compatible with different Ruby versions.
+  # You can add `ruby: "X.Y.Z"` on a spec to specify which version should run it.
+  #
+  # For example:
+  #   it("tests something with Ruby 2.5 or greater", ruby: ">= 2.5.0") do
+  #     # ...
+  #   end
+  config.filter_run_excluding(ruby: ->(v) do
+    !Gem::Requirement.new(v).satisfied_by?(Gem::Version.new(RUBY_VERSION))
+  end)
+end
+
 RSpec.describe(Tapioca::Compilers::SymbolTableCompiler) do
   describe("compile") do
     def run_in_child
@@ -608,7 +621,7 @@ RSpec.describe(Tapioca::Compilers::SymbolTableCompiler) do
       )
     end
 
-    it("compiles Structs, Classes, and Modules") do
+    it("compiles Structs, Classes, and Modules for Ruby >= 2.5", ruby: ">= 2.5.0") do
       expect(
         compile(<<~RUBY)
           class S1 < Struct.new(:foo)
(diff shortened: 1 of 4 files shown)


spec/tapioca/compilers/symbol_table_compiler_spec.rb
```diff
@@ -636,7 +636,9 @@ def foo
           M3 = Module.new
         RUBY
       ).to(
-        eq(<<~RUBY.chomp)
+        # this is a low impact hack to remove differences in Struct behaviour from
+        # ruby 2.4 testing.
+        eq(<<~RUBY.chomp.lines.reject { |line| RUBY_VERSION =~ /^2.4/ && line =~ /self.inspect/ }.join)
```
paracycle: Agreed, this is the only bit that feels weird. How about using ERB templates for this kind of output matching?

We can do something like:

```ruby
eq(render(<<~RUBY)
  class S1 < ::Struct
  def self.[](*_); end
<% unless Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.5.0") %>
  def self.inspect; end
<% end %>
  def self.members; end
  def self.new(*_); end
end
```

and `render` could be a method that takes a string and evaluates it as an ERB template.

Thoughts @Morriar?

Morriar: I definitely prefer this: https://github.com/Shopify/tapioca/pull/18/commits/f6b1129ebd19d4bb67bf3bf595e153ec85289c51

paracycle: The reason I propose this, btw, is because we have a similar situation in another test where the output changes depending on if you have Ruby compiled with GMP or not. 🤦‍♂

paracycle: @Morriar I saw that commit and that also looks good, but again, it only solves it for the ruby version problem and it ends up duplicating the test output. If we need to change one of the tests the change has to be duplicated etc, which is brittle.

Morriar: > ends up duplicating the test output

IMHO it will be easier to read / maintain than a test output riddled with ERB conditions.

spec/spec_helper.rb
```diff
@@ -26,4 +26,6 @@
   if config.files_to_run.one?
     config.default_formatter = "doc"
   end
+
+  config.order = :random
```
Morriar: Do we need this? It's already defined at [line 23](https://github.com/Shopify/tapioca/pull/13/files#diff-93830fa29d616f7c87903d08b5b1b29aR23).

timgaleckas: Removed.

timgaleckas:
Rebased and cherry-picked https://github.com/Shopify/tapioca/commit/f6b1129ebd19d4bb67bf3bf595e153ec85289c51

paracycle:
This looks great @timgaleckas, thank you for your contribution again.

timgaleckas:
Thank you all.

PR #15, opened 2019-10-01
Remove support artefacts before and after all tests

Morriar:
While testing locally with the `rake` command, I noticed that some artefacts were not cleaned after the tests:

* `spec/support/repo/sorbet/config`
* `spec/support/repo/sorbet/tapioca/require.rb`

This PR removes them after each test.

### DIFF
--- a/spec/tapioca/cli_spec.rb
+++ b/spec/tapioca/cli_spec.rb
@@ -73,21 +73,15 @@ RSpec.describe(Tapioca::Cli) do
   end
 
   around(:each) do |example|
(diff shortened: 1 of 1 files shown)


Morriar:
The tests break on Ruby 2.4 because `glob` is only defined as a singleton method.

Was the CI pipeline changed since last PR?

Morriar:
> Feel free to merge it despite the CI failing.

Actually I'm on the process to add support for 2.3.7 as needed by other projects. Keeping this one open until I fixed everything.

PR #16, opened 2019-10-01
Bump sorbet to 0.4.4813

Morriar:
All is in the title.

I fixed the version on `sorbet` instead of fixing it independently for `static` and `runtime`. Was this intentional? If so, why?

### DIFF
--- a/tapioca.gemspec
+++ b/tapioca.gemspec
@@ -26,7 +26,7 @@ Gem::Specification.new do |spec|
   )
 
   spec.add_dependency("pry", ">= 0.12.2")
-  spec.add_dependency("sorbet-static", "~> 0.4.4471")
+  spec.add_dependency("sorbet-static")
   spec.add_dependency("sorbet-runtime")
   spec.add_dependency("thor", "~> 0.20.3")
   spec.add_dependency("zeitwerk", "~> 2.1")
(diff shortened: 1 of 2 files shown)


tapioca.gemspec
```diff
@@ -35,7 +35,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency("pry-byebug")
   spec.add_development_dependency("rspec", "~> 3.7")
   spec.add_development_dependency("rubocop", "~> 0.70.0")
-  spec.add_development_dependency("sorbet")
+  spec.add_development_dependency("sorbet", "~> 0.4.4813")
```
rafaelfranca: This is only a development dependency, while sorbet-static was a runtime dependency. They behave in different ways. I think what we want is to lock the runtime dependency version. That way users can't install this gem with a version below the one we require.

Also, theoretically you don't need to change the gemspec on this commit, unless this gem only works now with `sorbet >= 0.4.4813`. Since I see no changes in the code on this PR, I think that is not the case.

Morriar:
Closing since not needed. Thanks @rafaelfranca!

paracycle:
Thanks @rafaelfranca for chiming in. This was exactly what I came here to say.

@Morriar for reference, `tapioca` explicitly depends on `sorbet-runtime` to find the "symbols" in gems. Other than that, it is a coincidence that it actually uses Sorbet signatures, etc.

Also, we have a minimum version for that dependency, since that was the first version to add support for `symbol-table-full-json` in `sorbet-static`.

PR #18, opened 2019-10-02
Add support for Ruby 2.3(.7)

Morriar:
This PR contains the few tweaks needed to support Ruby 2.3.x.

Please disregard the first commit as it's coming from #19.

### DIFF
--- a/spec/tapioca/compilers/symbol_table_compiler_spec.rb
+++ b/spec/tapioca/compilers/symbol_table_compiler_spec.rb
@@ -115,13 +115,21 @@ RSpec.describe(Tapioca::Compilers::SymbolTableCompiler) do
 
           ::RUBY_ENGINE_VERSION = T.let(T.unsafe(nil), String)
 
+          <% if ruby_version(">= 2.4.0") %>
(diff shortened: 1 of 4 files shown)


rafaelfranca:
Why? Ruby 2.3 is not supported anymore even by the Ruby team.

rafaelfranca:
Oh. We still need to support system Ruby on MacOS.

spec/tapioca/compilers/symbol_table_compiler_spec.rb
```diff
@@ -1380,7 +1398,7 @@ def self.new(*args, &block)
                   super
                 end
 
-                instance_methods.each { |m| undef_method m unless /^__|^object_id$/.match?(m) }
+                instance_methods.each { |m| undef_method m unless /^__|^object_id$/.match(m) != nil }
```
rafaelfranca: don't need `!= nil` here. `unless /^__|^object_id$/.match(m)` will work.

Morriar: 🤦‍♂ fixed

Morriar:
🎉

PR #19, opened 2019-10-02
Remove dependency to Zeitwerk so we can support Ruby 2.3

Morriar:
From [Zeitwerk documentation:](https://github.com/fxn/zeitwerk#supported-ruby-versions)

> Zeitwerk works with MRI 2.4.4 and above.

This make it difficult to use `tapioca` on Ruby 2.3.

After discussing with @paracycle, seems like Zeitwerk is not absolutely mandatory for this project and we can do manual requires since there are only a few files. This is the goal of this PR.

### DIFF
--- a/lib/tapioca.rb
+++ b/lib/tapioca.rb
@@ -1,10 +1,7 @@
 # typed: true
 # frozen_string_literal: true
-require "sorbet-runtime"
-require "zeitwerk"
 
-loader = Zeitwerk::Loader.for_gem
-loader.setup
+require "sorbet-runtime"
 
 module Tapioca
   def self.silence_warnings
@@ -31,4 +28,12 @@ rescue
   nil
 end
 
-loader.eager_load
+require_relative "tapioca/loader"
+require_relative "tapioca/constant_locator"
+require_relative "tapioca/generator"
+require_relative "tapioca/cli"
+require_relative "tapioca/gemfile"
+require_relative "tapioca/compilers/symbol_table_compiler"
+require_relative "tapioca/compilers/symbol_table/symbol_generator"
+require_relative "tapioca/compilers/symbol_table/symbol_loader"
+require_relative "tapioca/version"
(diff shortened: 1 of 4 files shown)


lib/tapioca.rb
```diff
@@ -31,4 +28,12 @@ class Error < StandardError; end
...
-loader.eager_load
+require_relative "tapioca/cli"
+require_relative "tapioca/loader"
+require_relative "tapioca/compilers/symbol_table/symbol_generator"
+require_relative "tapioca/compilers/symbol_table/symbol_loader"
+require_relative "tapioca/compilers/symbol_table_compiler"
+require_relative "tapioca/constant_locator"
+require_relative "tapioca/gemfile"
+require_relative "tapioca/loader"
+require_relative "tapioca/version"
```
paracycle: I think this order would be more correct:
```ruby
require_relative "tapioca/loader"
require_relative "tapioca/constant_locator"
require_relative "tapioca/generator"
require_relative "tapioca/cli"
require_relative "tapioca/gemfile"
require_relative "tapioca/compilers/symbol_table_compiler"
require_relative "tapioca/compilers/symbol_table/symbol_generator"
require_relative "tapioca/compilers/symbol_table/symbol_loader"
require_relative "tapioca/version"
```

lib/tapioca/cli.rb
```diff
@@ -3,6 +3,8 @@
 
 require 'thor'
 
+require_relative "generator"
```
paracycle: With the above change, this shouldn't be needed.

Morriar: This made me think, shouldn't the tests for `cli` and `compilers` only import their respective module instead of `tapioca` all in one?

paracycle: Yeah, not sure about that either. But from what I see almost all gems end up doing all their `require`s in their entrypoint and don't have any other `require` calls anywhere else. Thus, the best practice ends up being to `require` the entrypoint file as it would be in real usage.

lib/tapioca.rb
Morriar: Fixed

lib/tapioca/cli.rb
Morriar: Fixed too

PR #20, opened 2019-10-03
Fixed static lib issue on older Macs; upgraded 11 gems; Fixed warning

jasnow:
* Changed Integer to String in symbol_table_compiler_spec.rb since there is a problem with static library on older Macs. (Been working with @paracycle on this item)
 * Added back 2 temp files. They seen to come and go.
 * Fixed ERB.new keyword args warning.


jasnow:
New PR - "green" TravisCI

spec/tapioca/compilers/symbol_table_compiler_spec.rb
```diff
@@ -57,7 +57,11 @@ def ruby_version(selector)
     end
 
     def template(src)
-      ERB.new(src, nil, ">").result(binding).chomp
+      if RUBY_VERSION >= '2.6'
```
paracycle: Thanks for catching this. Can we do the check like [here](https://github.com/rtomayko/tilt/commit/4aff2d124e684c40240baf2d79d0ca1cd4cfc0f0) instead of relying on Ruby version?

jasnow: Did the 2 changes as requested. TravisCI is green.

paracycle:
@jasnow Thank you for this PR. The `ERB` logic was moved, so I've fixed this on top of your branch and squash-merged these changes. Unfortunately, it didn't mark this PR as merged, so I'll have to close it. Just making sure that you don't think we've rejected your changes.

PR #22, opened 2019-10-03
Move templating features to `RSpec::Matchers` so it can be used from any test file

Morriar:
We also need it from `cli_spec.rb` for variations with Ruby 2.3.


paracycle:
LGTM

Issue #23, opened 2019-10-23
Generating rbi for gem fails when `singleton_class` overridden


eightbitraptor:
Attempting to generate rbi's for gems in Services DB using `tapioca generate gems` is failing with the following exception:

```
  Compiling rack-mini-profiler, this may take a few seconds... bundler: failed to load command: tapioca (/Users/mattvh/.gem/ruby/2.6.5/bin/tapioca)
NoMethodError: private method `singleton_class' called for Rack::MiniProfiler:Class
Did you mean?  singleton_class?
  /Users/mattvh/.gem/ruby/2.6.5/gems/tapioca-0.2.3/lib/tapioca/compilers/symbol_table/symbol_generator.rb:324:in `compile_methods'
  /Users/mattvh/.gem/ruby/2.6.5/gems/tapioca-0.2.3/lib/tapioca/compilers/symbol_table/symbol_generator.rb:164:in `block in compile_body'
  /Users/mattvh/.gem/ruby/2.6.5/gems/tapioca-0.2.3/lib/tapioca/compilers/symbol_table/symbol_generator.rb:437:in `with_indentation'
  /Users/mattvh/.gem/ruby/2.6.5/gems/tapioca-0.2.3/lib/tapioca/compilers/symbol_table/symbol_generator.rb:163:in `compile_body'
```

It looks like I can't call `singleton_class` on `Rack::MiniProfiler` because a private method with a different signature but the same name is being defined here: https://github.com/MiniProfiler/rack-mini-profiler/blob/master/lib/mini_profiler/profiling_methods.rb#L147

```
[21] pry(main)> Rack::MiniProfiler.singleton_class
NoMethodError: private method `singleton_class' called for Rack::MiniProfiler:Class
Did you mean?  singleton_class?
from (pry):15:in `<main>'
[22] pry(main)> Rack::MiniProfiler.send(:singleton_class)
ArgumentError: wrong number of arguments (given 0, expected 1)
from /Users/mattvh/.gem/ruby/2.6.5/gems/rack-mini-profiler-1.1.1/lib/mini_profiler/profiling_methods.rb:147:in `singleton_class'
[23] pry(main)> Rack::MiniProfiler.method(:singleton_class).source_location
=> ["/Users/mattvh/.gem/ruby/2.6.5/gems/rack-mini-profiler-1.1.1/lib/mini_profiler/profiling_methods.rb", 147]
```

PR #24, opened 2019-10-23
Call `singleton_class` on Object explicitly

eightbitraptor:
This addresses https://github.com/Shopify/tapioca/issues/23

This works around the case where a gem has defined a `singleton_class`
method with a different method signature (and potenitally different visibility) to the one that's defined on `Object`

This is an approach that is sometimes used to allow access to the singleton class of an object in Rubies pre 2.1.0

When generating `rbi` files we always want to use the result of `Object#singleton_class`

### DIFF
--- a/lib/tapioca/compilers/symbol_table/symbol_generator.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_generator.rb
@@ -257,7 +257,7 @@ module Tapioca
 
           inherited_singleton_class_ancestors =
             if constant.is_a?(Class)
-              Set.new(constant.superclass.singleton_class.ancestors)
+              Set.new(singleton_class_of(constant.superclass).ancestors)
             else
               Module.ancestors
             end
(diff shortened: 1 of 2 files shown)


lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -321,7 +321,9 @@ def compile_methods(name, constant)
           )
 
           instance_methods = compile_directly_owned_methods(name, constant)
-          singleton_methods = compile_directly_owned_methods(name, constant.singleton_class, [:public])
+
+          constant_singleton = Object.instance_method(:singleton_class).bind(constant).call
```
Morriar: There are a couple of other `singleton_class` uses in this same file, I wonder if we should fix them too?

eightbitraptor: @Morriar I have just spotted those! Working on it now. Thank you.

paracycle: Thanks, good catch. While you are working through the other instances, can you wrap this in a method named `singleton_class_of` like how we've done with similar concerns (like `name_of`, `constants_of` etc at the end of the same file)?

Morriar:
@eightbitraptor: as discussed on slack I pushed a test showing the crash on your branch. Also I'm sorry to inform you that the test is still failing with your patch because we are now calling a private method :(

Maybe more luck with `send` or `instance_eval`?

lib/tapioca/compilers/symbol_table/symbol_generator.rb
paracycle: Actually I went ahead a did this, which also fixed @Morriar's added test.

Morriar:
Good teamwork, thanks everybody :D

PR #27, opened 2019-12-16
Remove thor dependency requirement upperbound:

Edouard-chin:
Remove thor dependency requirement upperbound:

- ### Problem

  Rails made thor 1.0 a dependency requirement. Application
  using Rails HEAD and taopica can't have their dependencies resolved
  since there is a version conflict.

  ### Solution

  I couldn't find a reason in the history on why we added a upperbound
  requirement on thor so I'm just removing it. It will make things
  easier for applications next time thor release a new version.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -5,7 +5,7 @@ PATH
       pry (>= 0.12.2)
       sorbet-runtime
       sorbet-static (~> 0.4.4471)
-      thor (~> 0.20.3)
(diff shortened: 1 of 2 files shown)


paracycle:
Hahah, didn't look that closely to the repo name. This _is_ relaxing it in `tapioca`. 🤦‍♂️ All good, we didn't have a specific reason for the upperbound.

PR #34, opened 2020-01-24
Stop ignoring gems if they are vendored into the application folder

paracycle:
Tapioca ignores gems that live under the application folder since those gems should get type-checked along with other application code. However, that assumption breaks if developers have installed gems into a `vendor` folder inside application directory.

In that case, Sorbet would not (or should not) be type-checking those gem files but would still need their RBIs. Thus, we should only ignore gems that live in the application folder but are not in the Bundler install path.

This PR implements that new behaviour.

### DIFF
--- a/lib/tapioca/gemfile.rb
+++ b/lib/tapioca/gemfile.rb
@@ -33,8 +33,8 @@ module Tapioca
         definition
           .resolve
           .materialize(specs)
-          .reject { |spec| ignore_gem_spec?(spec) }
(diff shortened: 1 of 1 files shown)


paracycle:
@morriar More useful than the current behaviour? I don't think so, since, on Core, for example, we woukd have to manually maintain a list of in-repo gems for such an ignore list. 

However, I see that such a flag could be useful on its own.

PR #36, opened 2020-02-12
Bump Sorbet version

paracycle:
Tapioca had been using a pretty dated version of Sorbet. This PR updates it and makes all the necessary code changes.


lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -74,14 +74,18 @@ def generate_from_symbol(symbol)
...
           Object.const_get(symbol, false)
         rescue NameError, LoadError, RuntimeError, ArgumentError, TypeError
           nil
         end
 
-        sig { params(name: T.nilable(String), constant: BasicObject).returns(T.nilable(String)) }
+        sig do
+          params(name: T.nilable(String), constant: BasicObject)
+            .returns(T.nilable(String))
+            .checked(:never)
```
Morriar: Why?

paracycle: So, the problem is that Sorbet tries to match the type using `is_a?` but that method does not exist on `BasicObject`, so any method that accepts or returns a value that actually is a `BasicObject` ends up failing. Until we have a better solution, this will have to do.

paracycle:
This is ready for review @Morriar

lib/tapioca/compilers/symbol_table/symbol_generator.rb
Morriar: Should we raise this in sorbet/sorbet then add a SmartTODO here?

paracycle: We could do that, but tbh, I do intend to replace the runtime checker with the stub in this project, so not sure if it will be helpful.

The actual problem is that this was actually a bigger problem in the earlier versions of `sorbet-runtime` which prevented me from working on `tapioca` at the time. So I had stubbed out the runtime and was using that. Then, it somehow got better for a bit (I have no idea how) but now it is broken again.

paracycle: And there is already an issue: https://github.com/sorbet/sorbet/issues/908

PR #41, opened 2020-03-25
Add centralized configuration with optional config file override

paracycle:
Common usage of `tapioca` could involve setting many option flags for every run, especially for `typed-override` flags. This necessitates the ability to load options from a configuration file. This PR adds that support.

The implementation introduces a new `Tapioca::Config` value object that gets built by the `Tapioca::ConfigBuilder` class from supplied options, with a fallback to configuration file and default values.

The change ended up simplifying a lot of the codebase as well as adding a new feature.


.travis.yml
```diff
@@ -3,7 +3,6 @@ sudo: false
 language: ruby
 cache: bundler
 rvm:
-  - 2.3
```
Morriar: Are we discreetly loosing support for 2.3 or not testing it anymore?

paracycle: Since we started using `T::Struct` in `Config`, we lost 2.3 support. `sorbet-runtime` seems to be using `String#match?` in their implementation which is not supported in 2.3, so I am dropping it, since anybody using `sorbet-runtime` cannot be running 2.3 anyway.

PR #42, opened 2020-03-26
Add `--exclude` option

paracycle:
It is sometimes desirable to exclude certain gems from the RBI generation process. This PR adds the `--exclude` flag to both `generate` and `sync` commands to cater for this use case.

### DIFF
--- a/lib/tapioca/generator.rb
+++ b/lib/tapioca/generator.rb
@@ -30,13 +30,15 @@ module Tapioca
     def build_gem_rbis(gem_names)
       require_gem_file
 
-      gems_to_generate(gem_names).map do |gem|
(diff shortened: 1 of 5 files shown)


lib/tapioca/generator.rb
```diff
@@ -94,6 +96,7 @@ def existing_rbis
     sig { returns(T::Hash[String, String]) }
     def expected_rbis
       @expected_rbis ||= bundle.dependencies
+        .reject { |gem| config.exclude.include?(gem.name) }
```
rafaelfranca: `filter_map`?

paracycle: Ideally yes, but we aim to support all the way back to 2.4 for now, so that won't work. 😞

PR #45, opened 2020-04-03
Better `mixes_in_class_methods` detection

paracycle:
The current state of `mixes_in_class_methods` discovery is subpar and can only detect ones coming from the usage of `ActiveSupport::Concern`. However, there is a chance to do better and this PR is an attempt to implement that.

The basic idea is to have a temporary empty class template that we include any `module` we find. The empty class template does two things:

1. It defines `method_missing` both on the instance and class level so that we can ignore calls to some expected methods on the including class coming from some modules as they are being includes/extended. For example, `IdentityCache` just plain assumes that it will be able to call some AR methods in the `self.included` block. Implementing an empty `method_missing` allows those calls to happen in a no-op manner.
2. It hooks the singleton `include` method to keep track of what modules are being included to the test class and what `extends` are they pulling into the class as they are being included. This is for the cases where a single top-level include pulls in more includes each of which does an `extend` on the base class. Again, `IdentityCache` is a case-in-point, where `include IdentityCache` ends up including about 5-6 other modules each of which is an `ActiveSupport::Concern` and end up mixing their own class methods. The singleton hook is able to tell us exactly which extra includes are being done and which class methods those are bringing along with them.

This _does_ add extra complexity for the `tapioca` runtime and slows RBI generation down by about 10%-15% (based on my tests on Core, where the total gem RBI generation went from ~5m30s to ~6m10s) which I think is acceptable in return for better type generation.

### DIFF
--- a/lib/tapioca/compilers/symbol_table/symbol_generator.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_generator.rb
@@ -317,16 +317,54 @@ module Tapioca
         def compile_mixes_in_class_methods(constant)
           return "" if constant.is_a?(Class)
 
-          temp = Class.new
-          ancestors_before = temp.singleton_class.ancestors
-          temp.include(constant)
-          ancestors_after = temp.singleton_class.ancestors
+          mixins_from_modules = {}
 
-          mixed_in_module = (ancestors_after - ancestors_before).first
+          Class.new do
+            # rubocop:disable Style/MethodMissingSuper, Style/MissingRespondToMissing
+            def method_missing(symbol, *args)
+            end
 
-          return "" if mixed_in_module.nil?
+            define_singleton_method(:include) do |mod|
+              before = singleton_class.ancestors
+              super(mod).tap do
+                mixins_from_modules[mod] = singleton_class.ancestors - before
+              end
+            end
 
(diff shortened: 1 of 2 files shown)


lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -309,17 +310,69 @@ def compile_mixins(constant)
...
+
+        sig { params(constant: Module).returns(String) }
+        def compile_mixes_in_class_methods(constant)
+          return "" if constant.is_a?(Class)
+
+          mixins_from_modules = {}
+
+          Class.new do
+            def method_missing(symbol, *args)
+              super
```
rafaelfranca, line 324: Isn't support going to raise an error?

paracycle: Indeed, I was mislead by the Rubocop rule, this is fixed.

paracycle: As for `method_missing`, the reason is what I noted in the PR description:

> ... defines `method_missing` both on the instance and class level so that we can ignore calls to some expected methods on the including class coming from some modules as they are being includes/extended. For example, `IdentityCache` just plain assumes that it will be able to call some AR methods in the `self.included` method. Implementing an empty `method_missing` allows those calls to happen in a no-op manner.

PR #49, opened 2020-04-23
Be *extras* careful

Morriar:
When used with the [extras](https://rubygems.org/gems/extras/versions/0.3.0) gem, `tapioca` crashes with:

```
/gems/extras-0.3.0/lib/extras/shell.rb:43:in `escape': undefined method `gsub' for #<Pathname:0x00007fb523a32840> (NoMethodError)
```

This actually hides two problems.

1. What happens is that `extras` aliases the `shellescape` method for [its own](https://github.com/envygeeks/extras/blob/master/lib/extras/shell.rb#L49) that doesn't support `Pathname`.

   This is fixed by this PR by only escaping the filename only and not all the other arguments we control.

2. Extras also [redefines the escaped value for the empty string](https://github.com/envygeeks/extras/blob/master/lib/extras/shell.rb#L27) which ends up running sorbet with no argument at all and produce an error.

   This PR fixes it by prepending the `-e` option to ensure Sorbet reads the empty string as a Ruby input.

See the included test for more details.

### DIFF
--- a/lib/tapioca/compilers/symbol_table/symbol_loader.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_loader.rb
@@ -33,7 +33,7 @@ module Tapioca
               file.write(Array(paths).join("\n"))
               file.flush
 
-              symbol_table_json_from("@#{file.path}")
+              symbol_table_json_from("@#{file.path.shellescape}")
             end, T.nilable(String))
 
             return Set.new if output.nil? || output.empty?
@@ -44,7 +44,7 @@ module Tapioca
 
           def ignored_symbols
             unless @ignored_symbols
-              output = symbol_table_json_from("''", table_type: "symbol-table-full-json")
+              output = symbol_table_json_from("-e ''", table_type: "symbol-table-full-json")
               json = JSON.parse(output)
               @ignored_symbols = SymbolTableParser.parse(json)
             end
@@ -61,7 +61,7 @@ module Tapioca
                 "--print=#{table_type}",
                 "--quiet",
                 input,
-              ].shelljoin,
+              ].join(' '),
               err: "/dev/null"
             ).read
           end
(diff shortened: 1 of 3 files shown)


rafaelfranca:
> 1. What happens is that `extras` aliases the `shellescape` method for [its own](https://github.com/envygeeks/extras/blob/master/lib/extras/shell.rb#L49) that doesn't support `Pathname`.

This looks like a thing we should be fixing in that gem, not on tapioca.

The second point makes sense.

Morriar:
@rafaelfranca 

> This looks like a thing we should be fixing in that gem, not on tapioca.

Good idea. I opened a PR here: https://github.com/envygeeks/extras/pull/2

The thing is, this gem will then need to be bumped in Jekyll-assets: https://rubygems.org/gems/jekyll-assets/versions/2.2.8

So in the mean time, I would like this to be merged so we can get `shopify-build` back on tapioca's master instead of a fix branch: https://github.com/Shopify/shopify-build/blob/master/Gemfile#L31

rafaelfranca:
Oh yeah, we can implement a workaround here, it is just that the gem is also wrong and should be fixed.

spec/tapioca/cli_spec.rb
```diff
@@ -57,29 +57,32 @@ def run(command, args = [], flags = {})
...
-      },
-      ["bundle", "install", "--quiet"]
-    ).read
+    Bundler.with_clean_env do
+      IO.popen(
+        {
+          "BUNDLE_GEMFILE" => (@repo_path / "Gemfile").to_s,
+        },
+        ["bundle", "install", "--quiet"]
+      ).read
```
paracycle, line 84: Can we not run this in `@repo_path` too then? That way we can drop the `BUNDLE_GEMFILE` from this call too.

Morriar: Good idea, done

lib/tapioca/compilers/symbol_table/symbol_loader.rb
```diff
@@ -61,7 +61,7 @@ def symbol_table_json_from(input, table_type: "symbol-table-json")
                 "--print=#{table_type}",
                 "--quiet",
                 input,
-              ].shelljoin,
+              ].join(' '),
```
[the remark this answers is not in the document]
Morriar: What about just escaping `SORBET`? I added a `sorbet_path` method that returns the escaped string version. Should do the trick.

PR #50, opened 2020-04-28
Recommend `bundle exec`

ghiculescu:
I dunno about you guys, but I always forget this.

### DIFF
--- a/lib/tapioca/config_builder.rb
+++ b/lib/tapioca/config_builder.rb
@@ -34,7 +34,7 @@ module Tapioca
         command = File.basename($PROGRAM_NAME)
         args = ARGV.join(" ")
 
-        "#{command} #{args}".strip
+        "bundle exec #{command} #{args}".strip
       end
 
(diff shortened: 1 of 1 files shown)


ghiculescu:
Oh, I didn't know that. By binstubs do you mean we should be encouraging something `bin/rails tapioca generate ...` ? @rafaelfranca

rafaelfranca:
I mean this https://bundler.io/man/bundle-binstubs.1.html. The `bin/rails` is a binstub but every gem you have in your project you should have a binstub.

ghiculescu:
wow, i just ran `bundle install --binstubs` and it generated a whole bunch. game changer.

ghiculescu:
Do you know if anyone is thinking about making that the default `bundle install` behaviour?

rafaelfranca:
I think it was at some point but it broke a few things. I "think" bundler 3 would make it the default again but "citation needed".

Morriar:
Should we also remove the `bundle exec` we use in the [README](https://github.com/Shopify/tapioca/blob/master/README.md#installation) then?

PR #51, opened 2020-04-29
Generate the list of all unresolved constants

Morriar:
This PR reproduces the behaviour of `srb rbi todo` inside Tapioca as I often end up using `srb` just for this feature.

With this, you just have to:

```
tapioca init
tapioca todo
```

It will generate the list of all unresolved constants under `sorbet/rbi/todo.rbi` and you get a project ready with Sorbet.

Like with `srb rbi todo`, all unresolved constants are generated as empty modules:

```ruby
# This file is autogenerated. Do not edit it by hand. Regenerate it with:
#   generate command

# typed: false

module ::UndefConstant; end
```


lib/tapioca/compilers/todos_compiler.rb
```diff
@@ -0,0 +1,34 @@
...
+      sig do
+        returns(String)
+      end
+      def compile
+        rbi = String.new
+        list_todos.each_line do |line|
+          next if line.include?("<") || line.include?("class_of")
+          rbi << "module #{line.strip.gsub('T.untyped::', '')}; end\n"
+        end
+        rbi
```
paracycle, line 19: Not that it matters a lot, but I am a fan of collecting and joining, whenever possible (no-side effect functional programming tendency). Also I think joining an array of strings would be slightly faster and more efficient than appending to a string. So I would suggest:
```suggestion
        list_todos.each_line.map do |line|
          next if line.include?("<") || line.include?("class_of")
          module_name = line.strip.gsub('T.untyped::', '')
          "module #{module_name}; end"
        end.compact.join("\n")
```

lib/tapioca/generator.rb
```diff
@@ -44,6 +44,37 @@ def build_gem_rbis(gem_names)
...
+        return
+      end
+
+      content = String.new
+      content << rbi_header(config.generate_command, "false")
+      content << rbi_string
+
+      outdir = File.dirname(todos_path)
+      FileUtils.mkdir_p(outdir)
+      File.write(config.todos_path, content)
```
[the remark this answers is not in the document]
Morriar: Good catch

lib/tapioca/compilers/todos_compiler.rb
Morriar: It's not entirely equivalent as the `join` strips the trailing `\n` hence the addition here: https://github.com/Shopify/tapioca/pull/51/commits/574529513d4bf3af5bd9437958d19e828694c298#diff-595c19bc567be3296e5404c802a7bc1fR67

PR #52, opened 2020-05-02
Auto-populate the post require file

Morriar:
Most of the friction when starting a new project with `tapioca` is that you have to figure out which gem is not loaded properly, find the corresponding require you should add to tapioca config and re-rerun `tapioca generate` again.

This PR proposes the command `tapioca require` in order to make this process easier.

When running `tapioca require`, `tapioca` will get all files listed in `sorbet/config` and grep them for `require 'X'` and copy all those requires to the `sorbet/tapioca/require.rb` file.

This file should then be reviewed by hand to remove all unnecessary (or unwanted) requires. For example, `require 'launch_nuclear_attack'` might not be a secure thing here.

In any case, I also altered the behavior of loading this file to catch the loading errors and give a bit more context to the user:

```
LoadError: cannot load such file -- something_wrong

Tapioca could not load all the files required by your application.
If you populated sorbet/tapioca/require.rb with tapioca require
you should probably review it and remove the faulty line.
```


rafaelfranca:
> This file should then be reviewed by hand to remove all unnecessary (or unwanted) requires.

How people will know which files to remove? I think they don't have the information to make that decision.

lib/tapioca/compilers/requires_compiler.rb
```diff
@@ -0,0 +1,101 @@
...
+          @sorbet_path = sorbet_path
+          @dirs = T.let([], T::Array[String])
+          @ignored = T.let([], T::Array[Regexp])
+          read_config if config?
+        end
+
+        sig { returns(T::Array[String]) }
+        def ruby_files
+          @dirs.map do |dir|
+            Dir.glob("#{dir}/**/*.rb", File::FNM_EXTGLOB).map do |file|
```
rafaelfranca, line 65: `flat_map` and return `[]` instead of `nil` if ignore and you don't need `compact` anymore.

Morriar:
New version based on experimentations on our projects.

What changed?
* Extracted SorbetConfig in its own module
* More tests for SorbetConfig
* More tests for the RequiresCompiler itself
* Addressed all comments from @rafaelfranca but one:

> How people will know which files to remove? I think they don't have the information to make that decision.

There are two things related to that:
1. if you know that one of your module do things when required (that can potentially destructive), you should remove that module
2. if you run `tapioca generate` and it crashes it will tell you which require to remove

I want to leave the message as like I explained in the PR header, things can get dangerous:
> For example, require 'launch_nuclear_attack' might not be a secure thing here.

rafaelfranca:
> 1. you know that one of your module do things when required (that can potentially destructive), you should remove that module

I'd not assume people understand the libraries they include in their apps, or even the files in the repository.

Do you have real example of those destructive files or things that should not be required? Maybe we can create a list of know issues and already remove them.

Morriar:
The "review & commit" is the standard message we output. For example when producing the RBIs here: https://github.com/Shopify/tapioca/blob/master/lib/tapioca/generator.rb#L44 

Not a real example but you can have something like that:

```ruby
module B
  def self.included(m)
    puts "malicious code"
  end
end
```

But in the other hand, the require already comes from your code so 🤷 .

lib/tapioca/compilers/requires_compiler.rb
```diff
@@ -0,0 +1,72 @@
...
+          else
+            [path.to_s]
+          end
+        end.sort.uniq
+      end
+
+      sig { params(file_path: String).returns(T::Enumerable[String]) }
+      def collect_requires(file_path)
+        File.read(file_path).lines.select do |line|
+          line =~ /^require ?\(?['"]/
```
paracycle, line 48: I think this is still a little too restrictive since it won't match `require`s that are indented like `  require "foo"`.

lib/tapioca/compilers/requires_compiler.rb
```diff
@@ -0,0 +1,72 @@
...
+            end
+          else
+            [path.to_s]
+          end
+        end.sort.uniq
+      end
+
+      sig { params(file_path: String).returns(T::Enumerable[String]) }
+      def collect_requires(file_path)
+        File.read(file_path).lines.select do |line|
```
paracycle, line 47: I wonder if it might make more sense to have a single regex and to `map` all lines to a (potential) require coming from that line. We can then compact to get rid of all the `nil`s. That would simplify the code by getting rid of `collect_required_name` and removing the `next unless name` in the loop using this.

That would also allow us to extract the regex to a constant so that it is more prominent what we are matching.

Also, it would allow us to generate a require file that is more uniform since would be in control of formatting the `require "..."` lines.

lib/tapioca/compilers/requires_compiler.rb
```diff
@@ -0,0 +1,72 @@
...
+          line =~ /^require ?\(?['"]/
+        end
+      end
+
+      sig { params(required_string: String).returns(T.nilable(MatchData)) }
+      def collect_required_name(required_string)
+        /['"]([^'"]+)['"]/.match(required_string)
```
paracycle, line 54: If you decide to keep this method, we can use the block arg of `match` to return a `T.nilable(String)` from this method:
```suggestion
        /['"]([^'"]+)['"]/.match(required_string) do |m| 
        	m[0]
        end
```
since the return value of `#match`, if supplied a block, is the return value of the block. And the block is only called if there was a match.

lib/tapioca/generator.rb
```diff
@@ -44,6 +44,38 @@ def build_gem_rbis(gem_names)
...
+        say("Nothing to do", :green)
+        return
+      end
+
+      # Clean all existing requires before regenerating the list so we update
+      # it with the new one found in the client code and remove the old ones.
+      File.delete(requires_path) if File.exist?(requires_path)
+
+      content = String.new
+      content << rbi_header(config.generate_command, "false")
```
paracycle, line 65: Ugh, this method name sticks out now that we are using it in a non-RBI setting. We need to change that at some point to something like `generated_file_header`. Feel free to do it as part of this PR.

lib/tapioca/generator.rb
```diff
@@ -44,6 +44,38 @@ def build_gem_rbis(gem_names)
...
+      content << rb_string
+
+      outdir = File.dirname(requires_path)
+      FileUtils.mkdir_p(outdir)
+      File.write(requires_path, content)
+
+      say("Done", :green)
+
+      say("All requires from this application have been written to #{name}.", [:green, :bold])
+      cmd = set_color('tapioca generate', :yellow, :bold)
```
paracycle, line 75:
```suggestion
      cmd = set_color("tapioca sync", :yellow, :bold)
```

lib/tapioca/generator.rb
```diff
@@ -113,11 +145,42 @@ def compiler
...
+        exit(1)
+      end
       say(" Done", :green)
       puts
     end
 
+    sig { params(file: String, error: LoadError).void }
+    def explain_failed_require(file, error)
+      say_error("\n\nLoadError: #{error}", :bold, :red)
+      say_error("\nTapioca could not load all the files required by your application.", :yellow)
```
paracycle, line 161: I wonder if we should be explicitly saying that these are "gem requires"? "All the files required by your application" sounds a bit vague.

lib/tapioca/compilers/requires_compiler.rb
Morriar: The goal with this heuristic is to avoid all the requires from inside the code, for example the conditional requires depending on platform, etc. as to minimize the amount of gems you need to remove manually because `tapioca` cannot load them.

I can go  and grab all of them and let the user remove them if you prefer.

paracycle: Hmm, not sure how I feel about this one. Conditional requires would still imply that we would need them for generating RBIs, and, in some cases, maybe even more so. There are also patterns like:
```ruby
begin
  require "foo"
rescue NameError
  require "foo_fallback"
end
```
that is used in some apps, so might be good to catch those too.

Do you have any examples of matching indented requires resulting in a lot of extra requires?

lib/tapioca/compilers/requires_compiler.rb
Morriar: I went with formatting at first, but not doing it was actually a better way to show the inconsistencies in the client coding style.

lib/tapioca/generator.rb
Morriar: I went with `Tapioca could not load all the gems required by your application.`

lib/tapioca/compilers/requires_compiler.rb
Morriar: Ok, I selected all `require`s whatever their indentation level.

lib/tapioca/compilers/requires_compiler.rb
Morriar: Since we now collect requires with possible indentation so I also needed to remove the indent. Keeping the original `"` or `'` was too much trouble so I went with the formatting after all.

PR #53, opened 2020-05-26
Smart properties extraction

katayounHillier:
Extract smart_properties RBI generator from shopify core.

Note: Things that are different between the expected generated RBI file in Tapioca with documentation in core :
- the documentation in generator has the wrong order of the generated sigs. 
-  `T.nilable` is missing in the documentation for some of the types. 
- `::` is missing in the document.


spec/tapioca/compilers/dsl/smart_properties_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+      RUBY
+
+      with_contents(content) do
+        parlour = Parlour::RbiGenerator.new(sort_namespaces: true)
+        subject.decorate(parlour.root, Post)
+        expect(parlour.rbi).to(eq("# typed: strong\n\n"))
+      end
+    end
+
+    it("generates RBI file if there are smart properties") do
```
paracycle, line 63: I would really like it if we could test each combination of how a `property` could be defined individually on a separate test. That would make it much easier to identify what went wrong when something is broken.

For example, `SmartProperties` has a lot of way of defining `accepts` and `required` etc, which we had to write a lot of code to work around. Specifically, we need to carefully test the cases where `accepts` could be an array of different types, where it could have a lambda as a value, etc etc.

KaanOzkan:
> 1. Regarding the incompatibility between the inline doc and the test output, did we update the doc to reflect the actual output?

We were going to update them once it's confirmed test output is accurate.

> 2. I think we need to add `require "smart_properties"` and `require "parlour"` to `sorbet/tapioca/require.rb` and run `bundle exec exe/tapioca sync` to generate the RBI file for the new `parlour` and `smart_properties` gem dependencies.

Did you mean a different file by `sorbet/tapioca/require.rb` ? It's not part of this repo.

> 3. As I noted inline, we need more test cases and we need them to be more targeted to one case.
Sounds good.

lib/tapioca/compilers/dsl/smart_properties.rb
```diff
@@ -0,0 +1,169 @@
...
+# frozen_string_literal: true
+
+require "parlour"
+
+begin
+  require "smart_properties"
+rescue LoadError
+  # means SmartProperties is not installed,
+  # so let's not even define the generator.
+  return
```
paracycle, line 11: @rafaelfranca This felt super hacky to do but it also seems like the best guard against `smart_properties` not being installed in the target application. If we can't load the gem, then not even defining this generator class seems to be the easiest way to guard it.

However, I can't be sure if there is a better way or if this is frowned upon. Can you comment?

rafaelfranca: I think this good way to solve this problem.

PR #58, opened 2020-06-02
Extract actionmailer

KaanOzkan:
resolves https://github.com/shopify/sorbet-issues/issues/185


spec/tapioca/compilers/dsl/action_mailer_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+      expected = <<~RUBY
+        # typed: strong
+        class NotifierMailer
+        end
+      RUBY
+
+      expect(rbi_for(content)).to(eq(expected))
+    end
+
+    it("generates correct RBI file for subclass with methods") do
```
paracycle, line 65: This is testing a case where the source method does not have a signature. Can we also add a test for where the original method does have a signature?

spec/tapioca/compilers/dsl/action_mailer_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+        # typed: strong
+        class NotifierMailer
+          sig { params(customer_id: T.untyped).returns(::ActionMailer::MessageDelivery) }
+          def self.notify_customer(customer_id); end
+        end
+      RUBY
+
+      expect(rbi_for(content)).to(eq(expected))
+    end
+  end
```
paracycle, line 84: Can we also add a test case where an abstract `class ApplicationMailer < ActionMailer::Base` has been defined with some helper instance methods and the `NotifierMailer` is subclassing from it? Since the generator is using `action_methods` we would expect it to ignore the instance methods coming from the abstract base class. We have an example such base class in Core: `components/platform/app/mailers/application_mailer.rb`

spec/tapioca/compilers/dsl/action_mailer_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+      with_content(content) do
+        expect(subject.processable_constants).to(eq(Set.new([NotifierMailer])))
+      end
+    end
+    it("ignores abstract subclasses") do
+      content = <<~RUBY
+        class NotifierMailer < ActionMailer::Base
+        end
+
+        class SecondayMailer < ActionMailer::Base
```
paracycle, line 30: I guess you mean:
```suggestion
        class SecondaryMailer < ActionMailer::Base
```

spec/tapioca/compilers/dsl/action_mailer_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+      content = <<~RUBY
+        class NotifierMailer < ActionMailer::Base
+        end
+
+        class User
+        end
+      RUBY
+
+      with_content(content) do
+        expect(subject.processable_constants).to(eq(Set.new([NotifierMailer])))
```
paracycle, line 22: Can this also please ensure that subclasses of `NotifierMailer` are also considered properly?

PR #59, opened 2020-06-02
fix rspec failure reporting

KaanOzkan:
calling `expect` inside a child process prevents rspec from reporting any failures


lib/tapioca/compilers/dsl/smart_properties.rb
```diff
@@ -140,7 +140,7 @@ def type_for(property)
             "T.untyped"
           elsif accepter == Array
             "T::Array[T.untyped]"
-          elsif BOOLEANS.any?(accepter)
+          elsif BOOLEANS.include?(accepter)
```
KaanOzkan, line 143: This is an error that was caught by this change

PR #60, opened 2020-06-03
Extract `ActionController::Helpers`

KaanOzkan:
resolves https://github.com/Shopify/sorbet-issues/issues/186


lib/tapioca/compilers/dsl/action_mailer.rb
```diff
@@ -29,7 +29,7 @@ module Dsl
       #
       # ~~~rbi
       # # notifier_mailer.rbi
-      # # typed: true
+      # # typed: strong
```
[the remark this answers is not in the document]
KaanOzkan: I was just keeping it consistent with what we expect in the tests. Generators are producing `#typed: strong`

If the pipeline is taking care of generating `typed: true` documentation can just say `typed: true` then

spec/tapioca/compilers/dsl/action_controller_helpers_spec.rb
```diff
@@ -0,0 +1,238 @@
...
+        class UserController
+          sig { returns(UserController::HelperProxy) }
+          def helpers; end
+        end
+
+        module UserController::HelperMethods
+          sig { params(user: T.untyped).returns(T.untyped) }
+          def greet(user); end
+
+          sig { params(user_id: Integer).returns(<VOID>) }
```
paracycle, line 186: Bah, we need to fix that. 😞 Looking into it

paracycle: Fixed

lib/tapioca/compilers/dsl/action_controller_helpers.rb
```diff
@@ -0,0 +1,129 @@
...
+      # end
+      #
+      # module UserController::HelperMethods
+      #    include MyHelper
+      #
+      #    sig { params(user: T.untyped).returns(T.untyped) }
+      #    def age(user); end
+      #
+      #    sig { params(args: T.untyped, blk: T.untyped).returns(T.untyped) }
+      #    def current_user_name(*args, &blk); end
```
paracycle, line 63:
```suggestion
      #    sig { returns(T.untyped) }
      #    def current_user_name; end
```

KaanOzkan: My bad forgot about it. Good catch.

PR #61, opened 2020-06-04
Add Sorbet signature and types generation

paracycle:
## Motivation
Gem source methods that have Sorbet signatures has been a problem and an opportunity for `tapioca` for a while now. 

The problem was that `tapioca` would get the source method parameters wrong, due to the wrapping by Sorbet. The opportunity was being able to generate RBI files with signatures when the methods were already declaring their signatures in source.

## Implementation

This PR solves the problem and addresses the opportunity by explicitly using the signatures in gem source.

In order to do this, the PR does 4 main things:

1. It checks for and processes Sorbet signatures associated with a method. This ensures that we do reflection properly on the actual method definition and not on the wrapper.
2. It generates a `sig { }` block based on the `Signature` defined for the method. This ends up being a little non-trivial due to the small ways in which the processed information not being 100% revertible. For example, `type_parameters` info is lost and needs to be recreated by scanning the `T.type_parameter` usages in `params` and `returns`. Similarly, `.void` ends up being serialized as `.returns(<VOID>)` unless explicitly handled.
3. It handles classes that are subclasses of `T::Struct` and generates `prop`/`const` definitions for them instead of the runtime visible accessor methods. This ensures that Sorbet does not complain about missing parameters in an initializer that it is expecting.
4. It tries to figure out if types need to be marked as `abstract!`, `interface!`, `final!` or `sealed!`, so that the generated code does not raise any errors. For example, if we don't mark classes that have `abstract` methods as `abstract!` then Sorbet would raise an error.

## Known Limitations

**Note:** What is still not handled yet are generic types, which we can address in a later PR. We would need to identify `type_member` and `type_template` constants and generate the equivalent method calls instead of naked constant definitions to handle that, which should be ~easy to do.

## Conclusion

The resulting output seems to be very robust and can generate equivalent RBIs to what some gems export from their RBI folder.


lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -464,7 +467,46 @@ def compile_method(symbol_name, constant, method)
...
+          [
+            signature_str,
+            indented("def #{method_name}#{parameters}; end"),
+          ].compact.join("\n")
+        end
+
+        TYPE_PARAMETER_MATCHER = /T\.type_parameter\(([[:word:]]+)\)/
+
+        sig { params(signature: T.untyped).returns(String) }
+        def compile_signature(signature)
```
Morriar, line 522: This too could be extracted so we can reuse it from the generators or would it conflict with the way we use Parlour in the generators?

sorbet/rbi/shims/sorbet.rbi
```diff
@@ -0,0 +1,15 @@
+module T::Private
```
Morriar, line 1: Don't we have the runtime mock for this?

spec/tapioca/compilers/_symbol_table_compiler_spec.rb
```diff
@@ -1703,5 +1703,120 @@ def foo(*_, &_); end
         RUBY
       )
     end
+
+    it("compiles signatures and structs in source files") do
+      expect(
+        compile(template(<<~RUBY))
+          class Foo
+            extend(T::Sig)
+
```
Morriar, line 1712: Might be good to also test signatures on attributes to show the current behavior?

spec/tapioca/compilers/_symbol_table_compiler_spec.rb
```diff
@@ -1703,5 +1703,120 @@ def foo(*_, &_); end
...
+            def initialize(*args, &blk); end
+
+            sig { abstract.void }
+            def do_it; end
+          end
+
+          class Foo
+            sig { params(a: Integer, b: String).returns(Integer) }
+            def bar(a, b:); end
+            sig { type_parameters(:U).params(a: T.type_parameter(:U)).returns(T.type_parameter(:U)) }
```
Morriar, line 1796: It would be easier to read if we added a blank line between definitions when there is a signature.

lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -433,6 +433,9 @@ def compile_method(symbol_name, constant, method)
           return unless method.owner == constant
           return if symbol_ignored?(symbol_name) && !method_in_gem?(method)
 
+          signature = T::Private::Methods.signature_for_method(method)
```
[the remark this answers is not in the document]
paracycle: Yes, once all the generators are in and the pipeline is working, we should do another pass over this to extract some reusable components out.

spec/tapioca/compilers/_symbol_table_compiler_spec.rb
paracycle: Do you mean `attr_reader`/`attr_writer` kind of thing? If so, I can add those.

sorbet/rbi/shims/sorbet.rbi
paracycle: `T::Private` exists in the exported RBIs but the types inside like `Final` or `Sealed` or `Abstract::Data` are not, so we need these for the static checker to not complain.

lib/tapioca/compilers/symbol_table/symbol_generator.rb
paracycle: Indeed, this can be extracted. Btw, a similar logic is also being used inside `sorbet-pry` as well, maybe it is time to extract a `sorbet-toolkit` gem.

lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -464,7 +509,51 @@ def compile_method(symbol_name, constant, method)
...
+            ".override"
+          when "overridable_override"
+            ".overridable.override"
+          when "overridable"
+            ".overridable"
+          else
+            ""
+          end
+
+          signature_body = +""
```
KaanOzkan, line 547:
```suggestion
          signature_body = ""
```

Idk if I'm missing something. Seems to be a new local variable.

spec/tapioca/compilers/_symbol_table_compiler_spec.rb
Morriar: ^ Yes

lib/tapioca/compilers/symbol_table/symbol_generator.rb
Morriar: Good idea, we could also extract the metrics parser and the config parser in such a gem.

spec/tapioca/compilers/_symbol_table_compiler_spec.rb
paracycle: Ugh, this is proving to be non-trivial since we are building strings just-in-time. I think I will leave it like this for now and revisit later.

spec/tapioca/compilers/_symbol_table_compiler_spec.rb
paracycle: Added

PR #62, opened 2020-06-04
Report test failures correctly

KaanOzkan:
continuation of https://github.com/Shopify/tapioca/pull/59 where we were having problems with rspec reporting the test failures. This PR removes the `expect` call out of the child process.

In order to more easily extract output from the child processes they are parsed as YAML and then outputted as an array of strings using `YAML#load`

### DIFF
--- a/spec/tapioca/compilers/dsl/action_controller_helpers_spec.rb
+++ b/spec/tapioca/compilers/dsl/action_controller_helpers_spec.rb
@@ -6,6 +6,12 @@ require "tapioca/compilers/dsl/action_controller_helpers"
 
 RSpec.describe(Tapioca::Compilers::Dsl::ActionControllerHelpers) do
   describe("#initialize") do
+    def rbi_for(content)
+      with_content(content) do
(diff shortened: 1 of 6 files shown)


spec/tapioca/compilers/dsl/action_controller_helpers_spec.rb
```diff
@@ -6,6 +6,12 @@
 
 RSpec.describe(Tapioca::Compilers::Dsl::ActionControllerHelpers) do
   describe("#initialize") do
+    def rbi_for(content)
```
paracycle, line 9: `rbi_for`? maybe `constants_from`?

KaanOzkan: copy past development tricked me

paracycle: 😂

paracycle: obviously the same rename applies to all instances, i just didn't comment on each.

PR #64, opened 2020-06-05
Extract ActiveSupport::CurrentAttributes generator

KaanOzkan:
This PR aims to extract our RBI generator for AS::CurrentAttributes DSL's from shopify core into tapioca.

- Added tests to the existing generator
- Modified the generator to maintain signatures in the source code (WIP)

resolves https://github.com/Shopify/sorbet-issues/issues/184


lib/tapioca/compilers/dsl/active_support_current_attributes.rb
```diff
@@ -0,0 +1,120 @@
+# typed: true
+# frozen_string_literal: true
+
+require "parlour"
+
```
KaanOzkan, line 5: I didn't have to `require "active_support"` here and therefore did not add a `rescue` block.

Is it possible that loading this file on systems without `ActiveSupport` will cause an error?

lib/tapioca/compilers/dsl/active_support_current_attributes.rb
```diff
@@ -0,0 +1,120 @@
...
+        sig { params(constant: T.class_of(::ActiveSupport::CurrentAttributes)).returns(T::Array[Symbol]) }
+        def instance_methods_for(constant)
+          constant.instance_methods(false)
+        end
+
+        sig { params(klass: Parlour::RbiGenerator::Namespace, method: String, class_method: T::Boolean).void }
+        def generate_method(klass, method, class_method:)
+          if method.end_with?("=")
+            parameter = Parlour::RbiGenerator::Parameter.new(T.must(method[0...-1]), type: "T.untyped")
```
KaanOzkan, line 111: Added a `T.must` here since `Parameter#new` takes on a `String` and not a `T.nilable(String)`. Want to make sure this is the best solution.

lib/tapioca/compilers/dsl/active_support_current_attributes.rb
```diff
@@ -0,0 +1,120 @@
...
+      #   sig { params(user_id: Integer).void }
+      #   def authenticate(user_id)
+      #     # ...
+      #   end
+      # end
+      # ~~~rb
+      #
+      # this generator will produce an RBI file with the following content:
+      # ~~~rbi
+      # # typed: strong
```
KaanOzkan, line 35: I added the documentation myself and left this as `strong`. Do we prefer to teach usage of `true`?

lib/tapioca/compilers/dsl/active_support_current_attributes.rb
paracycle: Yes, it probably would. Let's add the explicit `require "active_support"` here and the `rescue` block for it.

lib/tapioca/compilers/dsl/active_support_current_attributes.rb
paracycle: Yes, let's continue using `typed: true` please.

lib/tapioca/compilers/dsl/active_support_current_attributes.rb
paracycle: Weird that `method[0...-1]` would be considered nilable. 🤔

paracycle: Oh, I guess it is because `"abc"[10...1] == nil`. Actually, we don't have to worry about the parameter name since it does not matter. Let's just supply it as `"value"`.

PR #66, opened 2020-06-06
Switch to Minitest

paracycle:
This PR switches the gem from using RSpec for its tests, to using Minitest.

### Motivation

We are trying to consolidate on using Minitest for our projects so this project's use of RSpec ended up being an outlier. Moreover, it proved really hard (or hacky) to provide true per-test isolation using the RSpec runner but it seems to be much easier to do the same using Minitest.

We really need a good working and seamless isolation framework for our tests since we keep polluting the global namespace with a lot of types and keep doing runtime reflection on them, so tests end up affecting each other when not ran in a forked child process.

### Implementation

I tried to the simplest implementation possible by using `minitest-spec`, `minitest-hooks` and `minitest-reporters` to get the same general spec based testing framework. I opted to _not_ use the expectation based assertions, however, since I find the `assert` one much easier to read.

Finally, `minitest-fork_executor` handles the running of each test in isolation in a forked process.


tapioca.gemspec
```diff
@@ -33,10 +33,5 @@ Gem::Specification.new do |spec|
   spec.add_dependency("parlour", ">= 2.1.0")
   spec.add_dependency("thor", ">= 0.19.2")
 
-  spec.add_development_dependency("bundler", "~> 1.17")
```
Morriar, line 36: Why do we need to move that out to the Gemfile?

paracycle: No special reason, but I don't see the point of dev-dependencies in a Bundler world, so I just moved all of them into the `Gemfile`. Basically, the `Gemfile` is only used during development (dev/test environments) so whatever it has are dev dependencies.

Rakefile
```diff
@@ -3,10 +3,21 @@
...
+  t.test_files = FileList['spec/**/*_spec.rb']
+end
+
+task(:spec) do
+  begin
+    Rake::Task[:test].execute
+  rescue RuntimeError
+    exit(1)
+  end
 end
```
paracycle, line 21: I added this block for a weird reason. By default, if the Rake test task fails, Rake ends up printing an error message about the task having exited with a non-zero exit status and also prints a backtrace, which I found confusing and unnecessarily verbose.

So, I have a task here that executes the other task, catches the runtime error and exits with a non-zero status to signal failure. That prevents the Rake error message/backtrace from being printed in the console.

tapioca.gemspec
KaanOzkan: It confused me because we still have a `development, test` group defined.

paracycle: True, we probably don't need those either. TBH, I don't think anyone is asking Bundler to load a `test` group either in dev nor in CI anyway, so that might be something we can remove actually.

spec/spec_helper.rb
```diff
@@ -2,101 +2,21 @@
...
-    dir = Pathname.new(path)
-    # Create a "lib" folder
-    Dir.mkdir(dir.join("lib"))
-
-    contents.each do |file, content|
-      # Add our contents into their files in lib folder
-      File.write(dir.join("lib/#{file}"), content)
-    end
-
-    result = run_in_child do
```
KaanOzkan, line 60: Love how clean the helpers are now.

paracycle:
> Personally I'd like to have support for running a single test file automatically without proving the `TEST=path_to_test_file`.
> 
> Similarly running a single test in a file using the line number.

Yes, both of these are the bad parts of the Minitest runner. It seems we have work going on in the company to bring over these improvements to the Minitest runner too.

Until then, I guess it will have to be `TEST=path_to_test_file -n /a_string_that_exists_in_the_test_name/` to run individual tests.

PR #67, opened 2020-06-08
Extract ActiveRecordColumns

KaanOzkan:
This PR aims to extract our RBI generator for ActiveRecord DSL's from shopify core into tapioca. Specifically the DSL due to columns.

- Added sqlite3 in order to mock a database connection during the test
- Added tests to the existing generator

resolves Shopify/sorbet-issues#188


spec/tapioca/compilers/dsl/active_record_columns_spec.rb
```diff
@@ -0,0 +1,336 @@
...
+    end
+
+    def indented(str, indent)
+      str.lines.map! do |line|
+        next line if line.chomp.empty?
+        " " * indent + line
+      end.join
+    end
+
+    it("generates RBI file for class without custom attributes with StrongTypeGeneration") do
```
KaanOzkan, line 68: I didn't test this case "without" StrongTypeGeneration. 

Tested "classes with custom attributes with/without StrongTypeGeneration" down below. I thought it would add another long test without providing much value but I can add it.

lib/tapioca/compilers/dsl/active_record_columns.rb
```diff
@@ -0,0 +1,347 @@
...
+            column_name: String
+          ).returns([String, String])
+        end
+        def type_for(constant, column_name)
+          column_type = constant.attribute_types[column_name]
+
+          if Object.const_defined?(:StrongTypeGeneration) &&
+              !(constant.singleton_class < Object.const_get(:StrongTypeGeneration))
+            return ["T.untyped", "T.untyped"]
+          end
```
KaanOzkan, line 308: I couldn't decide on the best way to break this conditional into 2 lines. Rubocop didn't auto fix it for me.

spec/tapioca/compilers/dsl/active_record_columns_spec.rb
```diff
@@ -0,0 +1,336 @@
...
+
+      expected = indented(<<~RUBY, 2)
+        sig { returns([T.untyped, T.untyped]) }
+        def saved_change_to_body; end
+
+        sig { returns(T::Boolean) }
+        def saved_change_to_body?; end
+      RUBY
+      expect(output).to(include(expected))
+    end
```
KaanOzkan, line 334: Multiple `expect` calls were used here because the output is not continuous. This is what I could come up with but open to suggestions.

spec/tapioca/compilers/dsl/active_record_columns_spec.rb
```diff
@@ -0,0 +1,336 @@
...
+        parlour.rbi
+      end
+    end
+
+    def indented(str, indent)
+      str.lines.map! do |line|
+        next line if line.chomp.empty?
+        " " * indent + line
+      end.join
+    end
```
KaanOzkan, line 66: Copied from another test file. I think it's better to include it in `spec_helper.rb`?

spec/tapioca/compilers/dsl/active_record_columns_spec.rb
```diff
@@ -0,0 +1,336 @@
...
+          end
+        RUBY
+
+        "schema.rb" => <<~RUBY,
+          ActiveRecord::Migration.suppress_messages do
+            ActiveRecord::Schema.define do
+              create_table :posts do |t|
+              end
+            end
+          end
```
KaanOzkan, line 85: I decided to use `ActiveRecord::Migration.suppress_messages` so that during test runs we don't get `create_table...` output on console.

lib/tapioca/compilers/dsl/active_record_columns.rb
```diff
@@ -0,0 +1,347 @@
...
+            when ActiveRecord::AttributeMethods::TimeZoneConversion::TimeZoneConverter
+              "::ActiveSupport::TimeWithZone"
+            else
+              "T.untyped"
+            end
+
+          column = constant.columns_hash[column_name]
+          setter_type = getter_type
+
+          if column&.null ||
```
paracycle, line 335: Not sure if we are testing these conditions, but it would be nice to ensure that a field that is marked as `null: true` and `null: false` creates different return types and that `created_at`/`updated_at` columns and primary key columns are always nilable.

lib/tapioca/compilers/dsl/active_record_columns.rb
```diff
@@ -0,0 +1,347 @@
...
+        def type_for(constant, column_name)
+          column_type = constant.attribute_types[column_name]
+
+          if Object.const_defined?(:StrongTypeGeneration) &&
+              !(constant.singleton_class < Object.const_get(:StrongTypeGeneration))
+            return ["T.untyped", "T.untyped"]
+          end
+
+          getter_type =
+            case column_type
```
paracycle, line 311: Can we test each of these cases please? It can be a single test with columns of each of these types and we can assert the inclusion of the `field=` method signatures, which would tell us if we are mapping the types correctly, or not.

lib/tapioca/compilers/dsl/active_record_columns.rb
paracycle: It would be nice to extract that to a private method like:
```ruby
def generate_strong_types?
  Object.const_defined?(:StrongTypeGeneration) &&
  	constant.singleton_class < Object.const_get(:StrongTypeGeneration)
end
```

lib/tapioca/compilers/dsl/active_record_columns.rb
```diff
@@ -0,0 +1,347 @@
...
+          return unless constant.table_exists?
+
+          module_name = "#{constant}::GeneratedAttributeMethods"
+          root.create_module(module_name) do |mod|
+            constant.columns_hash.each_key do |column_name|
+              column_name = column_name.to_s
+              add_methods_for_attribute(mod, constant, column_name)
+            end
+
+            constant.attribute_aliases.each do |new_name, old_name|
```
paracycle, line 110: We need to test aliased attributes as well. We also need to test aliased attributes when the aliased attribute ends up conflicting with one of the generated attribute methods. For example:

```ruby
alias_attribute :nickname, :name
alias_attribute :name?, :name # yes, people do stuff like this
```

That's the point of the `old_method_names`/`new_method_names` shenanigans below.

lib/tapioca/compilers/dsl/active_record_columns.rb
KaanOzkan: It maybe how I'm running the tests but we don't actually generate any methods for `created_at/updated_at`

paracycle: I don't see us generating those columns, though.

paracycle: `t.timestamps`

KaanOzkan: I thought `created_at/updated_at` came automatically with every AR object regardless of the schema. I guess it's more so that `t.timestamps` is inserted automatically when an object is created.

spec/tapioca/compilers/dsl/active_record_columns_spec.rb
```diff
@@ -324,6 +324,481 @@ def saved_change_to_body; end
...
+        RUBY
+      }
+
+      expected = <<~RUBY
+        module Post::GeneratedAttributeMethods
+          sig { returns(T.untyped) }
+          def author; end
+
+          sig { params(value: T.untyped).returns(T.untyped) }
```
KaanOzkan, line 643: Methods generated for an alias don't specify the type.

lib/tapioca/compilers/dsl/active_record_columns.rb
alse: Hey @KaanOzkan I was wondering why primary_keys are set as nilable. Other columns with null: false are not nilable, so I was expecting the same for primary_key columns

paracycle: @alse The situation is more complicated, and, IMO, we've made a mistake with how we generate other column types, not primary keys.

In general an Active Record model would have all of its fields nilable:
```ruby
class User < ActiveRecord::Base
  # user has "id" and "name" fields
end

user = User.new
# => #<User:0x00007fefed98b090 id: nil, name: nil>
```

This is because AR models actually have a dual existence. Any model that is not persisted has no guarantees to the types of any of its fields, they can all be nil. This is most acute for the primary key field value which is almost never supplied as a parameter to the `.new` call. Only persisted models actually get the benefit of all the validations and the not-null annotations etc.

For a similar reason, we'd changed all AR associations to be nilable, we might actually do the same for columns as well, since that is actually the more correct way to type an AR model's column attributes under all conditions.

alse: That makes sense, thanks for thanks for the background @paracycle !

PR #70, opened 2020-06-11
Extract ActiveRecordAssociations DSL Generator

KaanOzkan:
This PR aims to extract our RBI generator for ActiveRecord DSL's from shopify core into tapioca. Specifically the DSL for associations.

Added tests to the existing generator
resolves Shopify/sorbet-issues#189

### DIFF
--- a/lib/tapioca/compilers/dsl/active_record_associations.rb
+++ b/lib/tapioca/compilers/dsl/active_record_associations.rb
@@ -1,0 +1,284 @@
+# typed: true
+# frozen_string_literal: true
+
+require "parlour"
+
+begin
+  require "active_record"
+rescue LoadError
+  return
+end
+
+module Tapioca
+  module Compilers
+    module Dsl
+      # `Tapioca::Compilers::Dsl::ActiveRecordAssociations` refines RBI files for subclasses of `ActiveRecord::Base`
+      # (see https://api.rubyonrails.org/classes/ActiveRecord/Base.html). This generator is only
+      # responsible for defining the methods that would be created for the association that
+      # are defined in the Active Record model.
+      #
+      # For example, with the following model class:
+      #
+      # ~~~rb
+      # class Post < ActiveRecord::Base
+      #   belongs_to :category
+      #   has_many :comments
+      #   has_one :author, class_name: "User"
+      # end
+      # ~~~
+      #
+      # this generator will produce the following methods in the RBI file
+      # `post.rbi`:
+      #
+      # ~~~rbi
+      # # post.rbi
+      # # typed: true
+      #
+      # class Post
+      #   include Post::GeneratedAssociationMethods
+      # end
+      #
+      # module Post::GeneratedAssociationMethods
+      #   sig { returns(T.nilable(::User)) }
+      #   def author; end
+      #
+      #   sig { params(value: T.nilable(::User)).void }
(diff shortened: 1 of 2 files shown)


KaanOzkan:
Comparing the generated methods with this rails documentation
https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html
For collection associations we are only generating these and not all specified. 
```
others
others=
other_ids
other_ids=
```

spec/tapioca/compilers/dsl/active_record_associations_spec.rb
```diff
@@ -0,0 +1,317 @@
...
+        end
+
+        class Current
+        end
+      RUBY
+
+      assert_equal(constants_from(content), ["Post"])
+    end
+
+    it("rejects abstract ActiveRecord subclasses") do
```
paracycle, line 38: Can we also add a case where `Post < Something < ActiveRecord::Base`?

lib/tapioca/compilers/dsl/active_record_associations.rb
```diff
@@ -0,0 +1,285 @@
...
+            constant: T.class_of(ActiveRecord::Base),
+            reflection: ReflectionType
+          ).returns(String)
+        end
+        def relation_type_for(constant, reflection)
+          "ActiveRecord::Associations::CollectionProxy" if !constant.table_exists? ||
+                                                            polymorphic_association?(reflection)
+
+          # Change to: "::#{reflection.klass.name}::ActiveRecord_Associations_CollectionProxy"
+          "::ActiveRecord::Associations::CollectionProxy[#{reflection.klass.name}]"
```
paracycle, line 267: Hmm, this is a problem, since there is no actual module `ActiveRecord::Associations::CollectionProxy` which has a generic parameter. That is just an RBI shim file we have in core that defines that module and we don't have that for any other class. I guess we can merge this as is and then fix this when we do the refactor to replace it with `Model::ActiveRecord::Associations::CollectionProxy`.

spec/tapioca/compilers/dsl/active_record_associations_spec.rb
```diff
@@ -0,0 +1,317 @@
...
+
+          sig { returns(T.nilable(T.untyped)) }
+          def reload_category; end
+        end
+      RUBY
+
+      assert_equal(rbi_for(content), expected)
+    end
+
+    it("generates RBI file for polymorphic belongs_to single association") do
```
paracycle, line 119: Can we also have a test that would test the:
```ruby
if reflection.through_reflection?
  polymorphic_association?(reflection.source_reflection)
```
part as well?

spec/tapioca/compilers/dsl/active_record_associations_spec.rb
KaanOzkan: I added it to this test case down below. I assumed that's what you were suggesting by commenting on this test

spec/tapioca/compilers/dsl/active_record_associations_spec.rb
KaanOzkan: Defined a table within this test which now goes through that path https://github.com/Shopify/tapioca/pull/70/files#diff-08e12da6ea17b18ee3475e651faf0189R231

PR #75, opened 2020-06-12
Update RBIs

KaanOzkan:
gem RBIs were outdated. Ran `tapioca sync`


sorbet/rbi/gems/minitest@5.14.1.rbi
```diff
@@ -1,8 +1,15 @@
 # This file is autogenerated. Do not edit it by hand. Regenerate it with:
-#   tapioca sync
+#   tapioca generate minitest
```
KaanOzkan, line 2: I needed to force tapioca to regenerate the RBI because when this RBI was first generated it `minitest-hooks` wasn't used in tapioca. `minitest-hooks` is bringing new modules from `minitest` that RBI didn't account for.

sorbet/rbi/gems/ansi@1.5.0.rbi
```diff
@@ -0,0 +1,281 @@
+# This file is autogenerated. Do not edit it by hand. Regenerate it with:
+#   tapioca generate ansi
```
KaanOzkan, line 2: After adding `require ansi/code` into sorbet/tapioca/require.rb I didn't want to remove all RBIs, run `bundle exec tapioca sync` and only commit the ansi RBI.

Instead I regenerated ansi only. That's why it wasn't generated by `sync`

sorbet/rbi/gems/minitest@5.14.1.rbi
paracycle: TBH, I think we need a script to encapsulate how this is being done and use that as the generation command. This is good for now.

KaanOzkan: I created an issue https://github.com/Shopify/sorbet-issues/issues/198

PR #76, opened 2020-06-15
Add DSL compilation pipeline runner

paracycle:
Resolves: Shopify/sorbet-issues#195

This PR adds a `dsl` command to `tapioca` which performs the following steps:

1. Assumes it is running inside a Rails application and loads the current Rails environment.
2. Optionally, eager loads the whole Rails application, unless it is supplied a list of constant names to generate DSLs for.
3. Loads all the generators inside `tapioca` as well as all the generators of the form `sorbet/tapioca/generators/**/*.rb` inside the target application.
4. Loops over all the constants to generate RBIs for and run each through the DSL generation pipeline.
5. Output an RBI file per constant in the `sorbet/rbi/dsl` folder (configurable using `outdir` flag).

This is working well inside of Core and should be working out-of-the-box in other Rails applications as well.


lib/tapioca/compilers/dsl_compiler.rb
```diff
@@ -0,0 +1,106 @@
...
+      private
+
+      sig { params(requested_generators: T::Array[String]).returns(Proc) }
+      def generator_filter(requested_generators)
+        return proc { true } if requested_generators.empty?
+
+        generators = requested_generators.map(&:downcase)
+
+        proc do |klass|
+          generator = klass.name&.sub(/^Tapioca::Compilers::Dsl::/, '')&.downcase
```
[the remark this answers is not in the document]
Morriar: (see following comment regarding how we create instances)

lib/tapioca/compilers/dsl_compiler.rb
```diff
@@ -0,0 +1,106 @@
...
+          generators.include?(generator)
+        end
+      end
+
+      sig { params(requested_generators: T::Array[String]).returns(T::Enumerable[Dsl::Base]) }
+      def gather_generators(requested_generators)
+        generator_filter = generator_filter(requested_generators)
+
+        Dsl::Base.descendants.select(&generator_filter).map(&:new)
```
Morriar, line 66: Only instanciating requested generators by their name will make it harder later if you want to introduce dependencies between them (ie. `GenA`, needs `GenB` to run before it). 

I think it would be better to instantiate all of them, then in a loop ask them if they need to run (and later ask them which other generators need to run before them). We could also imagine passing specific options to each generators and let them parse what they need if we want to alter their behavior from the `cli`.

lib/tapioca/compilers/dsl_compiler.rb
```diff
@@ -0,0 +1,106 @@
...
+
+        resolve_conflicts(parlour)
+
+        parlour.rbi("true").strip
+      end
+
+      sig { params(parlour: Parlour::RbiGenerator).void }
+      def resolve_conflicts(parlour)
+        Parlour::ConflictResolver.new.resolve_conflicts(parlour.root) do |msg, candidates|
+          $stderr.puts "=== Error ==="
```
Morriar, line 95: `say_error` could be used here too to add some colors.

lib/tapioca/config.rb
```diff
@@ -21,10 +21,11 @@ def outpath
 
     private_class_method :new
 
-    CONFIG_FILE_PATH = "sorbet/tapioca/config.yml"
-    SORBET_CONFIG = "sorbet/config"
+    TAPIOCA_PATH = T.let("sorbet/tapioca", String)
```
Morriar, line 24: What do you think about introducing a `SORBET_PATH` const in using it in all other paths?

lib/tapioca/generator.rb
```diff
@@ -108,6 +108,52 @@ def build_todos
       say("Please review changes and commit them.", [:green, :bold])
     end
 
+    sig { params(requested_constants: String).void }
+    def build_dsl(*requested_constants)
+      base_directory = Pathname.new("#{Config::DEFAULT_RBIDIR}/dsl")
```
Morriar, line 113: We could use a `DEFAULT_DSLDIR`

lib/tapioca/generator.rb
```diff
@@ -108,6 +108,52 @@ def build_todos
...
+      generation_command_template = config.generate_command
+
+      loader = Loader.new(Gemfile.new)
```
Morriar, line 116: I wonder if this could be a pipeline "step" that would be required by all other generators for DSL.

lib/tapioca/generator.rb
```diff
@@ -108,6 +108,52 @@ def build_todos
...
+        require File.expand_path(generator)
+      end
+
+      Dir["#{__dir__}/compilers/dsl/*.rb"].each do |generator|
+        require File.expand_path(generator)
+      end
+
+      compiler = Compilers::DslCompiler.new(
+        requested_constants: requested_constants.map(&:constantize),
+        requested_generators: ENV.fetch("GENERATORS", "").split(",").map(&:strip)
```
Morriar, line 132: Why do we rely on environment var instead of using a command option?

lib/tapioca/generator.rb
```diff
@@ -108,6 +108,52 @@ def build_todos
...
+
+      compiler.run do |constant, contents|
+        next if contents.nil?
+
+        command = format(generation_command_template, constant.name)
+        constant_name = Module.instance_method(:name).bind(constant).call
+        filename = base_directory.join(constant_name.underscore + ".rbi")
+
+        contents = <<~RBI
+          # DO NOT EDIT MANUALLY
```
Morriar, line 143: We're decoupling this in each new method generating a different kind of RBI. I think we can move it to the compiler itself.

lib/tapioca/loader.rb
```diff
@@ -24,6 +24,21 @@ def load_bundle(initialize_file, require_file)
...
+    sig { params(environment_load: T::Boolean, eager_load: T::Boolean).void }
+    def load_rails(environment_load: false, eager_load: false)
+      return unless File.exist?("config/application.rb")
+
+      safe_require("rails")
+
+      silence_deprecations
+
+      safe_require("rails/generators/test_case")
+      safe_require("./config/application")
```
rafaelfranca, line 36:
```suggestion
      safe_require("./config/environment")
```

if environment_load instead of `application`.

paracycle:
> I think we're missing the opportunity for some refactoring here.
> 

I agree, but I would like to do that work as a future improvement.

> I'm not against merging this PR has it is but this is work we should be doing at some point.

Definitely! The current design is not flexible enough and will cause problems in the future. However, the DSL generator works in the current form and most of the initial code duplication has been dealt with.

lib/tapioca/compilers/dsl_compiler.rb
```diff
@@ -0,0 +1,125 @@
...
+          candidates.each_with_index do |candidate, index|
+            error.puts "  #{index}. #{candidate.describe}"
+          end
+          report_error(error.to_s)
+        end
+      end
+
+      sig { params(error: String).returns(T.noreturn) }
+      def report_error(error)
+        handler = error_handler
```
Morriar, line 115: Instead of checking if `handler` is `nil`, let's initialize a default handler that print to `stderr` if none is provided?

lib/tapioca/compilers/dsl_compiler.rb
```diff
@@ -0,0 +1,125 @@
...
+        if handler
+          handler.call(error)
+        else
+          $stderr.puts(error)
```
KaanOzkan, line 119: Nitpick: It doesn't feel nice to call `stderr` directly when we have `say_error` but I guess we don't want to include it in this class for one call?

paracycle: Yes, this is supposed to be the ultimate fallback when no error handler was registered but there was still an error in processing. That is not supposed to happen since the only consumer of this class is the `Generator` and the handler is supplied there anyway, but I just wanted to be safer in the implementation.

PR #77, opened 2020-06-15
Extract ActiveRecord::Enum RBI generator into Tapioca

katayounHillier:
Resolves: Shopify/sorbet-issues#196

### testing 
  - dev test Test=spec/tapioca/compilers/dsl/active_record_enum_spec.rb
  - dev style 
  -  dev tc

### DIFF
--- a/lib/tapioca/compilers/dsl/active_record_enum.rb
+++ b/lib/tapioca/compilers/dsl/active_record_enum.rb
@@ -5,9 +5,8 @@ require "parlour"
 
 begin
   require "active_record"
-  require "action_controller"
 rescue LoadError
-  # means ActiveRecord::Enum is not installed,
+  # means ActiveRecord is not installed,
   # so let's not even define the generator.
   return
 end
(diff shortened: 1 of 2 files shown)


lib/tapioca/compilers/dsl/active_record_enum.rb
```diff
@@ -0,0 +1,112 @@
...
+        sig { override.returns(T::Enumerable[Module]) }
+        def gather_constants
+          ::ActiveRecord::Base.descendants.reject(&:abstract_class?)
+        end
+
+        private
+
+        sig { params(enum_map: T::Hash[T.untyped, T.untyped]).returns(String) }
+        def type_for_enum(enum_map)
+          value_type = enum_map.values.map { |v| v.class.name }.uniq
```
paracycle, line 88: I am not seeing this code-path being exercised properly. In order to test this, we need to supply enums like:
```ruby
enum foo: { a: 10, b: "text", c: true, d: [1,2,3] }
```
which should be generating a return type like `T::Hash[T.any(String, Symbol), T.any(Integer, String, TrueType, Array)]` or something like it.

Can we please add a test case for something similar?

paracycle: Also, it would be great if the same test had two enums on a single AR class, one like in my previous comment and another one with all the value types the same like:
```ruby
enum bar: { a: "a", b: "b", c: "c", d: "d" }
```
which should return `T::Hash[T.any(String, Symbol), String]`

(Of course, all of this with more meaningful names please 😄 )

katayounHillier:
I will rebase soon !

spec/tapioca/compilers/dsl/active_record_enum_spec.rb
```diff
@@ -87,6 +84,82 @@ def archived?; end
       assert_equal(rbi_for(content), expected)
     end
 
+    it("generates RBI file for classes with an enum attribute with hash values") do
+      content = <<~RUBY
+        class Conversation < ActiveRecord::Base
+          enum status: { active: 0, archived: 1 }
```
paracycle, line 90: Can we change the values here to be `String`s instead. It would be good to test the case where we discover a single non-`Integer` type.

PR #80, opened 2020-06-16
Extract UrlHelpers DSL generator

KaanOzkan:
This PR aims to extract our RBI generator for UrlHelpers DSL's from shopify core into tapioca. 

- Refactors #gather_constants to filter appropriate constants. Thanks to Ufuk.
- Refactors #decorate to use those constants and reduce hardcoding of built-in Rails classes.
- Adds tests
- Resolves Shopify/sorbet-issues#197


lib/tapioca/compilers/dsl/url_helpers.rb
```diff
@@ -0,0 +1,82 @@
...
+  return
+end
+
+module Tapioca
+  module Compilers
+    module Dsl
+      class UrlHelpers < Base
+        extend T::Sig
+
+        sig { override.params(root: Parlour::RbiGenerator::Namespace, constant: T.class_of(Object)).void }
```
KaanOzkan, line 18: I figured it was better to use `T.class_of(Object)` than `T.untyped`.

lib/tapioca/compilers/dsl/url_helpers.rb
```diff
@@ -0,0 +1,82 @@
...
+              mod.create_extend("GeneratedUrlHelpersModule") if
+                constant.singleton_class.ancestors.include?(GeneratedUrlHelpersModule)
+              mod.create_extend("GeneratedPathHelpersModule") if
+                constant.singleton_class.ancestors.include?(GeneratedPathHelpersModule)
+            end
+          end
+        end
+
+        sig { override.returns(T::Enumerable[T.untyped]) }
+        def gather_constants
```
KaanOzkan, line 65: I had to use `T.unsafe` in this method because sorbet thought `mod` was of type `BasicObject`. Can we manually specify the type of `mod`?

lib/tapioca/compilers/dsl/url_helpers.rb
paracycle:
```suggestion
        sig { override.params(root: Parlour::RbiGenerator::Namespace, constant: T.class_of(Module)).void }
```
would be better, since they are all modules.

lib/tapioca/compilers/dsl/url_helpers.rb
```diff
@@ -0,0 +1,82 @@
...
+              mod.create_include("ActionDispatch::Routing::PolymorphicRoutes")
+
+              path_helper_methods.each do |method|
+                mod.create_method(
+                  method.to_s,
+                  parameters: [Parlour::RbiGenerator::Parameter.new("*args", type: "T.untyped")],
+                  return_type: "String"
+                )
+              end
+            end
```
paracycle, line 36: Can we extract this to a method? That way we can move the `path_helper_methods` inside that method and add a guard clause to `return if path_helper_methods.empty?` so that we never even generate a file for this if there are no methods.

lib/tapioca/compilers/dsl/url_helpers.rb
```diff
@@ -0,0 +1,82 @@
...
+              mod.create_include("ActionDispatch::Routing::PolymorphicRoutes")
+
+              url_helper_methods.each do |method|
+                mod.create_method(
+                  method.to_s,
+                  parameters: [Parlour::RbiGenerator::Parameter.new("*args", type: "T.untyped")],
+                  return_type: "String"
+                )
+              end
+            end
```
paracycle, line 49: Same method extraction as the comment above.

lib/tapioca/compilers/dsl/url_helpers.rb
```diff
@@ -0,0 +1,82 @@
...
+            root.path(constant) do |mod|
+              mod.create_include("GeneratedUrlHelpersModule") if
+                constant.ancestors.include?(GeneratedUrlHelpersModule)
+              mod.create_include("GeneratedPathHelpersModule") if
+                constant.ancestors.include?(GeneratedPathHelpersModule)
+              mod.create_extend("GeneratedUrlHelpersModule") if
+                constant.singleton_class.ancestors.include?(GeneratedUrlHelpersModule)
+              mod.create_extend("GeneratedPathHelpersModule") if
+                constant.singleton_class.ancestors.include?(GeneratedPathHelpersModule)
+            end
```
paracycle, line 60:
```suggestion
            root.path(constant) do |mod|
	            create_mixins_for(mod, constant, GeneratedUrlHelpersModule)
	            create_mixins_for(mod, constant, GeneratedPathHelpersModule)
            end
```
where the method is defined as:
```ruby
def create_mixins_for(mod, constant, helper_module)
  mod.create_include(helper_module.name) if constant.ancestors.include?(helper_module)
  mod.create_extend(helper_module.name) if constant.singleton_class.ancestors.include?(helper_module)  
end

File: lib/tapioca/compilers/dsl/url_helpers.rb (hunk 2 above)
Comment (paracycle) [reply lib/tapioca/compilers/dsl/url_helpers.rb:65]: You can do a `T.let` to set the type

File: lib/tapioca/compilers/dsl/url_helpers.rb [hunk 6]
```diff
@@ -0,0 +1,82 @@
...
+            (T.unsafe(mod).ancestors.include?(GeneratedUrlHelpersModule) &&
+             !T.unsafe(mod).try(:superclass)&.ancestors&.include?(GeneratedUrlHelpersModule)) ||
+            (T.unsafe(mod).singleton_class.ancestors.include?(GeneratedUrlHelpersModule) &&
+             !T.unsafe(mod).singleton_class.try(:superclass)&.ancestors&.include?(GeneratedPathHelpersModule)) ||
+            (T.unsafe(mod).ancestors.include?(GeneratedPathHelpersModule) &&
+             !T.unsafe(mod).try(:superclass)&.ancestors&.include?(GeneratedPathHelpersModule)) ||
+            (T.unsafe(mod).singleton_class.ancestors.include?(GeneratedPathHelpersModule) &&
+             !T.unsafe(mod).singleton_class.try(:superclass)&.ancestors&.include?(GeneratedPathHelpersModule))
+          end.select(&T.unsafe(:name)) << "ActionDispatch::IntegrationTest"
+        end
```
paracycle, line 78:
```suggestion
          constants = ObjectSpace.each_object(Module).select do |mod|
            mod = T.let(mod, T.class_of(Module))
            includes_helper?(mod, GeneratedUrlHelpersModule)
     		  || includes_helper?(mod, GeneratedPathHelpersModule)
     		  || includes_helper?(mod.singleton_class, GeneratedUrlHelpersModule)
     		  || includes_helper?(mod.singleton_class, GeneratedPathHelpersModule)
          end.select(&:name)
          constants << "ActionDispatch::IntegrationTest"
          constants
        end
```
would be much nicer and readable.

```ruby
def includes_helper?(mod, helper)
  superclass_ancestors = [] 
  superclass_ancestors = mod.superclass.ancestors if Class === mod
  (mod.ancestors - superclass_ancestors).include?(helper)
end
```

lib/tapioca/compilers/dsl/url_helpers.rb
```diff
@@ -0,0 +1,82 @@
...
+module Tapioca
+  module Compilers
+    module Dsl
+      class UrlHelpers < Base
+        extend T::Sig
+
+        sig { override.params(root: Parlour::RbiGenerator::Namespace, constant: T.class_of(Object)).void }
+        def decorate(root, constant)
+          named_routes = Rails.application.routes.named_routes
+          path_helper_methods = named_routes.path_helpers_module.instance_methods(false)
```
paracycle, line 21: This can be just:
```suggestion
          path_helper_methods = GeneratedPathHelpersModule.instance_methods(false)
```
now

lib/tapioca/compilers/dsl/url_helpers.rb
```diff
@@ -0,0 +1,82 @@
...
+          url_helper_methods = named_routes.url_helpers_module.instance_methods(false)
```
paracycle, line 22: Same as above:
```suggestion
          url_helper_methods = GeneratedUrlHelpersModule.instance_methods(false)
```

spec/tapioca/compilers/dsl/url_helpers_spec.rb
```diff
@@ -0,0 +1,293 @@
...
+
+    content = <<~RUBY
+      class Application < Rails::Application
+      end
+    RUBY
+
+    it("gathers constants that include path_helpers_module") do
+      content += <<~RUBY
+        class MyClass
+          include Rails.application.routes.named_routes.path_helpers_module
```
paracycle, line 30: Let's always work with `Rails.application.routes.url_helpers`, since that is the actual thing we are trying to address.

You can use `include/extend Rails.application.routes.url_helpers(true/false)` to test various combinations.

spec/tapioca/compilers/dsl/url_helpers_spec.rb
```diff
@@ -0,0 +1,293 @@
...
+        end
+      RUBY
+
+      assert_equal(expected, rbi_for(content, :GeneratedUrlHelpersModule))
+    end
+
+    it("generates RBI for constant that includes path_helpers_module") do
+      content += <<~RUBY
+        class MyClass
+          include Rails.application.routes.named_routes.path_helpers_module
```
paracycle, line 203: Again, let's just concentrate on the `Rails.application.route.url_helpers` combinations.

lib/tapioca/compilers/dsl/url_helpers.rb
```diff
@@ -0,0 +1,80 @@
...
+          if constant == GeneratedPathHelpersModule
+            generate_module_for(root, GeneratedPathHelpersModule, path_helper_methods)
+          elsif constant == GeneratedUrlHelpersModule
+            generate_module_for(root, GeneratedUrlHelpersModule, url_helper_methods)
+          else
+            root.path(constant) do |mod|
+              create_mixins_for(mod, constant, GeneratedUrlHelpersModule)
+              create_mixins_for(mod, constant, GeneratedPathHelpersModule)
+            end
+          end
```
paracycle, line 32: This would be much nicer as a `case/when` block:
```suggestion
          case constant
          when GeneratedPathHelpersModule, GeneratedUrlHelpersModule
            generate_module_for(root, constant)
          else
            root.path(constant) do |mod|
              create_mixins_for(mod, constant, GeneratedUrlHelpersModule)
              create_mixins_for(mod, constant, GeneratedPathHelpersModule)
            end
          end
        end
```

lib/tapioca/compilers/dsl/url_helpers.rb
```diff
@@ -0,0 +1,80 @@
...
+module Tapioca
+  module Compilers
+    module Dsl
+      class UrlHelpers < Base
+        extend T::Sig
+
+        sig { override.params(root: Parlour::RbiGenerator::Namespace, constant: T.class_of(Module)).void }
+        def decorate(root, constant)
+          path_helper_methods = GeneratedPathHelpersModule.instance_methods(false)
+          url_helper_methods = GeneratedUrlHelpersModule.instance_methods(false)
```
paracycle, line 21: You don't need these anymore.

lib/tapioca/compilers/dsl/url_helpers.rb
```diff
@@ -0,0 +1,80 @@
...
+
+            helper_methods.each do |method|
+              mod.create_method(
+                method.to_s,
+                parameters: [Parlour::RbiGenerator::Parameter.new("*args", type: "T.untyped")],
+                return_type: "String"
+              )
+            end
+          end
+        end
```
paracycle, line 65:
```suggestion
        sig { params(root: Parlour::RbiGenerator::Namespace, constant: T.class_of(Module)).void }
        def generate_module_for(root, constant)
          root.create_module(constant.name) do |mod|
            mod.create_include("ActionDispatch::Routing::UrlFor")
            mod.create_include("ActionDispatch::Routing::PolymorphicRoutes")

            constant.instance_methods(false).each do |method|
              mod.create_method(
                method.to_s,
                parameters: [Parlour::RbiGenerator::Parameter.new("*args", type: "T.untyped")],
                return_type: "String"
              )
            end
          end
        end
```

Issue #79, opened 2020-06-15
what does tapioca do that `sorbet rbi gems` doesn't?


will:
I'm sorry, but I've read through the readme, and it's still not clear to me. When I run `sorbet init` or `sorbet rbi gems` it does generate rbi files for the gems. What does tapioca do differently than that?

paracycle:
👋 Hello @will, thanks for commenting. Yours is indeed a legitimate question and is also an indictment on the state of our documentation/README. Sorry about that!

`srb rbi gems` does indeed generate RBI files for gems and historically both `tapioca` and the `srb` tool date back to the same in-house Stripe code snippet for generating RBIs for gems. So, there are a lot of overlaps.

However, we have taken a slightly different route in building `tapioca` than the Sorbet team has with `srb rbi gems` that we feel make a big difference. I think it would be best for me to itemize them:

1. With `tapioca`, you are in complete control over which gems get loaded and how. We provide both `prerequire` and `postrequire` files that can do `require` calls so that you have the option to load all the code that you need RBI generation for into memory. `tapioca` handles the rest. You will need this for any gem that is marked `require: false` in the `Gemfile` and/or for any non-default part of a gem that you are requiring in your codebase explicitly. 

	In contrast, [this hard-coded lookup table](https://github.com/sorbet/sorbet/blob/master/gems/sorbet/lib/gem_loader.rb#L10) is what `srb rbi gems` provides. We didn't feel like maintaining a set of such lookup tables was going to be maintainable or desirable, so went a different way.
2. `tapioca` tries really hard to load all of your gem dependencies regardless of `groups` or other conditions, so that we can generate RBI files for all of your dependencies regardless of which environment you are using for RBI generation. I am not sure if `srb` does anything similar, but the last time we checked, it wasn't.
3. `tapioca` also does not rely on a pre-defined set of constants to start discovering other constants from like [how `srb` does (again via the same lookup table)](https://github.com/sorbet/sorbet/blob/master/gems/sorbet/lib/gem_loader.rb#L10). Instead, `tapioca` internally uses Sorbet to parse all of the code inside a gem to discover the statically analyzable types that is defines, and starts walking over those to do a better job of spanning all of type definitions in the gem. `srb`, by design, might miss some, since it depends on the internally defined constants.
[...]

PR #81, opened 2020-06-17
Link wiki response in README.md

KaanOzkan:
Copied Ufuk's answer from this issue https://github.com/Shopify/tapioca/issues/79 into a wiki https://github.com/Shopify/tapioca/wiki/How-does-tapioca-compare-to-%22srb-rbi-gems%22-%3F

This PR creates a new header in the README.md to link to the wiki answer

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,9 @@ If you ever run into a case, where you add a gem or update the version of a gem 
 You can use the command `tapioca require` to auto-populate the `sorbet/tapioca/require.rb` file with all the requires found
 in your application. Once the file generated, you should review it, remove all unnecessary requires and commit it.
 
+## How does tapioca compare to "srb rbi gems" ?
+[Please see the detailed answer on our wiki](https://github.com/Shopify/tapioca/wiki/How-does-tapioca-compare-to-%22srb-rbi-gems%22-%3F)
+
 ## Installation
 
 Add this line to your application's `Gemfile`:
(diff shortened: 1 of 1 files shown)


Morriar:
I think this should come sooner in the documentation and maybe not as a link but as text directly.

Most of the people that come to the `tapioca` repo have already eared about `srb` and might have the same question: "why should I use `tapioca` instead?". I think we need to answer this before even explaining how to install it.

Maybe another Idea would be to extract a short bullet list from the wiki post and put it int the introduction of the tool with a link to the post for more details?

Morriar:
It might also be useful to add this explanation in our internal docs: https://development.shopify.io/engineering/keytech/reference/sorbet/other_projects/

KaanOzkan:
It makes sense to me to put it earlier in the readme but I think it's too large of an answer to put the text in directly. 

Also I don't know how effectively we can extract it into a short bullet list. Because in my opinion if it's going to come before `installation` it needs to be quarter of its current length. I don't want to lose information/accuracy because chances are people won't go to the wiki if a summary is on the readme. But if everyone thinks it's possible to explain the points in a concise way then let's try it.

Morriar:
> Also I don't know how effectively we can extract it into a short bullet list. Because in my opinion if it's going to come before `installation` it needs to be quarter of its current length.

Good point! What do you think about moving the current section as it is right now just before `## Installation`?

KaanOzkan:
> What do you think about moving the current section as it is right now just before ## Installation?

I think thats better. I wasn't aware of how common/important this question was.

paracycle:
I also agree that this is an important question in people's minds about what makes this different from `srb` tooling, so it would be great to have it more prominently in the `README`.

README.md
```diff
@@ -34,6 +34,9 @@ If you ever run into a case, where you add a gem or update the version of a gem
 You can use the command `tapioca require` to auto-populate the `sorbet/tapioca/require.rb` file with all the requires found
 in your application. Once the file generated, you should review it, remove all unnecessary requires and commit it.
 
+## How does tapioca compare to "srb rbi gems" ?
```
Morriar, line 37:
```suggestion
## How does tapioca compare to "srb rbi gems" ?

```

To be consistent with the style in the rest of the document.

PR #84, opened 2020-06-22
Add Rails Model scope RBI generator

katayounHillier:
resolves https://github.com/Shopify/sorbet-issues/issues/159

Done these points for the generator so far
 1. A Shop::GeneratedRelationMethods module
 2. Generate all the methods on the module.

- Test
  - dev test  
  - dev style

### DIFF
--- a/lib/tapioca/compilers/dsl/active_record_scope.rb
+++ b/lib/tapioca/compilers/dsl/active_record_scope.rb
@@ -53,7 +53,7 @@ module Tapioca
         end
         def decorate(root, constant)
           scope_method_names = constant.send(:generated_relation_methods).instance_methods(false)
-          return if scope_method_names.blank?
+          return if scope_method_names.empty?
 
           module_name = "#{constant}::GeneratedRelationMethods"
           root.create_module(module_name) do |mod|
--- a/lib/tapioca/compilers/dsl/active_record_scope.rb
+++ b/lib/tapioca/compilers/dsl/active_record_scope.rb
@@ -1,0 +1,92 @@
+# typed: true
+# frozen_string_literal: true
+
+require "parlour"
+
+begin
+  require "active_record"
+rescue LoadError
+  return
+end
+
+module Tapioca
+  module Compilers
+    module Dsl
+      # `Tapioca::Compilers::Dsl::ActiveRecordScope` decorates RBI files for subclasses of
+      # `ActiveRecord::Base` which declare `scope` fields
+      # (see https://api.rubyonrails.org).
+      #
+      # For example, with the following `ActiveRecord::Base` subclass:
+      #
+      # ~~~rb
+      # class Post < ApplicationRecord
+      # end
+      # ~~~
+      #
+      # this generator will produce the RBI file `post.rbi` with the following content:
+      #
+      # ~~~rbi
+      # # post.rbi
+      # # typed: true
+      # class Post
+      # end
+      # ~~~
+      class ActiveRecordScope < Base
+        extend T::Sig
+
+        sig do
+          override.params(
+            root: Parlour::RbiGenerator::Namespace,
+            constant: T.class_of(::ActiveRecord::Base)
+          ).void
+        end
+        def decorate(root, constant)
+          scopes = constant.send(:generated_relation_methods).instance_methods(false)
+          return if scopes.blank?
+          module_name = "#{constant}::GeneratedRelationMethods"
+          root.create_module(module_name) do |mod|
+            generate_instance_methods(constant, mod)
+          end
+
+          root.path(constant) do |k|
+            k.create_include(module_name)
+
+            # scopes.each do |scope, value|
+            #   create_method(k, scope.to_s, class_method: true, return_type: "T.untyped")
+            # end
+          end
+        end
+
+        sig { override.returns(T::Enumerable[Module]) }
+        def gather_constants
+          ::ActiveRecord::Base.descendants.reject(&:abstract_class?)
+        end
+
+        private
+
+        sig do
+          params(
+            constant: T.class_of(::ActiveRecord::Base),
+            klass: Parlour::RbiGenerator::Namespace,
+          ).void
+        end
+        def generate_instance_methods(constant, klass)
+          methods = constant.send(:generated_relation_methods).instance_methods(false)
+          methods.each do |method|
+            method = method.to_s
+            return_type = "T.untyped"
+
+            create_method(
+              klass,
+              method,
+              parameters: [
+                Parlour::RbiGenerator::Parameter.new("*args", type: return_type),
+              ],
+              return_type: return_type,
+            )
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/tapioca/compilers/dsl/active_record_scope.rb
+++ b/lib/tapioca/compilers/dsl/active_record_scope.rb
@@ -20,6 +20,8 @@ module Tapioca
       #
       # ~~~rb
       # class Post < ApplicationRecord
+      #   scope :public_kind, -> { where.not(kind: 'private') }
+      #   scope :private_kind, -> {where(kind: 'private')}
       # end
       # ~~~
       #
@@ -29,6 +31,14 @@ module Tapioca
       # # post.rbi
       # # typed: true
       # class Post
+      #   extend Post::GeneratedRelationMethods
+      # end
+      # module Post::GeneratedRelationMethods
+      #   sig { params(args: T.untyped, blk: T.untyped).returns(T.untyped) }
+      #   def private_kind(*args, &blk); end
+
+      #   sig { params(args: T.untyped, blk: T.untyped).returns(T.untyped) }
+      #   def public_kind(*args, &blk); end
       # end
       # ~~~
       class ActiveRecordScope < Base
@@ -43,17 +53,16 @@ module Tapioca
         def decorate(root, constant)
           scopes = constant.send(:generated_relation_methods).instance_methods(false)
           return if scopes.blank?
+
           module_name = "#{constant}::GeneratedRelationMethods"
           root.create_module(module_name) do |mod|
-            generate_instance_methods(constant, mod)
+            scopes.each do |scope_method|
+              generate_instance_methods(scope_method, mod)
+            end
           end
 
           root.path(constant) do |k|
-            k.create_include(module_name)
-
-            # scopes.each do |scope, value|
-            #   create_method(k, scope.to_s, class_method: true, return_type: "T.untyped")
-            # end
+            k.create_extend(module_name)
           end
         end
 
@@ -66,25 +75,23 @@ module Tapioca
 
         sig do
           params(
-            constant: T.class_of(::ActiveRecord::Base),
-            klass: Parlour::RbiGenerator::Namespace,
+            scope_method: String,
+            mod: Parlour::RbiGenerator::Namespace,
           ).void
         end
-        def generate_instance_methods(constant, klass)
-          methods = constant.send(:generated_relation_methods).instance_methods(false)
-          methods.each do |method|
-            method = method.to_s
-            return_type = "T.untyped"
+        def generate_instance_methods(scope_method, mod)
+          # This return type should actually be Model::ActiveRecord_Relation
+          return_type = "T.untyped"
 
-            create_method(
-              klass,
-              method,
-              parameters: [
-                Parlour::RbiGenerator::Parameter.new("*args", type: return_type),
-              ],
-              return_type: return_type,
-            )
-          end
+          create_method(
+            mod,
+            scope_method,
+            parameters: [
+              Parlour::RbiGenerator::Parameter.new("*args", type: "T.untyped"),
+              Parlour::RbiGenerator::Parameter.new("&blk", type: "T.untyped"),
+            ],
+            return_type: return_type,
+          )
         end
       end
     end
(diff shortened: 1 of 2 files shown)


lib/tapioca/compilers/dsl/active_record_scope.rb
```diff
@@ -0,0 +1,92 @@
...
+          params(
+            constant: T.class_of(::ActiveRecord::Base),
+            klass: Parlour::RbiGenerator::Namespace,
+          ).void
+        end
+        def generate_instance_methods(constant, klass)
+          methods = constant.send(:generated_relation_methods).instance_methods(false)
+          methods.each do |method|
+            method = method.to_s
+            return_type = "T.untyped"
```
paracycle, line 77: Can we add a comment here saying that this return type should actually be `Model::ActiveRecord_Relation`?

lib/tapioca/compilers/dsl/active_record_scope.rb
```diff
@@ -0,0 +1,92 @@
...
+
+            create_method(
+              klass,
+              method,
+              parameters: [
+                Parlour::RbiGenerator::Parameter.new("*args", type: return_type),
```
paracycle, line 83:
```suggestion
                Parlour::RbiGenerator::Parameter.new("*args", type: "T.untyped"),
```
since the `return_type` and the arg type don't always have match.

Also, we need to supply a `&blk` parameter here as well.

lib/tapioca/compilers/dsl/active_record_scope.rb
```diff
@@ -0,0 +1,92 @@
...
+        private
+
+        sig do
+          params(
+            constant: T.class_of(::ActiveRecord::Base),
+            klass: Parlour::RbiGenerator::Namespace,
+          ).void
+        end
+        def generate_instance_methods(constant, klass)
+          methods = constant.send(:generated_relation_methods).instance_methods(false)
```
paracycle, line 74: We can get rid of this lookup if we supply the methods that we already looked up outside.

lib/tapioca/compilers/dsl/active_record_scope.rb
```diff
@@ -0,0 +1,92 @@
...
+        sig do
+          override.params(
+            root: Parlour::RbiGenerator::Namespace,
+            constant: T.class_of(::ActiveRecord::Base)
+          ).void
+        end
+        def decorate(root, constant)
+          scopes = constant.send(:generated_relation_methods).instance_methods(false)
+          return if scopes.blank?
+          module_name = "#{constant}::GeneratedRelationMethods"
```
paracycle, line 46:
```suggestion

          module_name = "#{constant}::GeneratedRelationMethods"
```

lib/tapioca/compilers/dsl/active_record_scope.rb
```diff
@@ -0,0 +1,92 @@
...
+          root.create_module(module_name) do |mod|
+            generate_instance_methods(constant, mod)
```
paracycle, line 48:
```suggestion
            generate_instance_methods(constant, mod, scope_methods)
```

lib/tapioca/compilers/dsl/active_record_scope.rb
```diff
@@ -0,0 +1,92 @@
...
+        extend T::Sig
+
+        sig do
+          override.params(
+            root: Parlour::RbiGenerator::Namespace,
+            constant: T.class_of(::ActiveRecord::Base)
+          ).void
+        end
+        def decorate(root, constant)
+          scopes = constant.send(:generated_relation_methods).instance_methods(false)
```
paracycle, line 44: Let's call this `scope_methods` instead.

lib/tapioca/compilers/dsl/active_record_scope.rb
```diff
@@ -0,0 +1,92 @@
...
+          return if scopes.blank?
+          module_name = "#{constant}::GeneratedRelationMethods"
+          root.create_module(module_name) do |mod|
+            generate_instance_methods(constant, mod)
+          end
+
+          root.path(constant) do |k|
+            k.create_include(module_name)
```
paracycle, line 52:
```suggestion
            k.create_extend(module_name)
```

lib/tapioca/compilers/dsl/active_record_scope.rb
```diff
@@ -0,0 +1,92 @@
...
+
+            # scopes.each do |scope, value|
+            #   create_method(k, scope.to_s, class_method: true, return_type: "T.untyped")
+            # end
```
paracycle, line 56:
```suggestion
```

We don't need any of these.

lib/tapioca/compilers/dsl/active_record_scope.rb
```diff
@@ -0,0 +1,92 @@
...
+
+        private
+
+        sig do
+          params(
+            constant: T.class_of(::ActiveRecord::Base),
+            klass: Parlour::RbiGenerator::Namespace,
+          ).void
+        end
+        def generate_instance_methods(constant, klass)
```
paracycle, line 73: Let's rename `klass` to `module` here, since we know we are adding methods to a module.

spec/tapioca/compilers/dsl/active_record_scope_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+        parlour = Parlour::RbiGenerator.new(sort_namespaces: true)
+        subject.decorate(parlour.root, Post)
+        parlour.rbi
+      end
+    end
+
+    it("generates RBI file for ActiveRecord classes with a scope field") do
+      content = <<~RUBY
+        class Post < ActiveRecord::Base
+          #extend Post::GeneratedRelationMethods
```
paracycle, line 55: We don't want these `extend`s in the source file. We want to get the `extend` line in the RBI output.

spec/tapioca/compilers/dsl/active_record_scope_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+          scope :public_kind, -> { where.not(kind: 'private') }
+        end
+
+      RUBY
+
+      expected = <<~RUBY
+        # typed: strong
+        class Post
+          include Post::GeneratedRelationMethods
```
paracycle, line 64: These should be:
```suggestion
          extend Post::GeneratedRelationMethods
```

spec/tapioca/compilers/dsl/active_record_scope_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+  describe("#decorate") do
+    def rbi_for(content)
+      with_content(content) do
+        parlour = Parlour::RbiGenerator.new(sort_namespaces: true)
+        subject.decorate(parlour.root, Post)
+        parlour.rbi
+      end
+    end
+
+    it("generates RBI file for ActiveRecord classes with a scope field") do
```
paracycle, line 52: Can we also add a case where we show that we generate an empty RBI file if the model has no defined scopes?

lib/tapioca/compilers/dsl/active_record_scope.rb
```diff
@@ -0,0 +1,92 @@
...
+      # end
+      # ~~~
+      #
+      # this generator will produce the RBI file `post.rbi` with the following content:
+      #
+      # ~~~rbi
+      # # post.rbi
+      # # typed: true
+      # class Post
+      # end
```
paracycle, line 32: Don't forget to complete this comment example!

spec/tapioca/compilers/dsl/active_record_scope_spec.rb
katayounHillier: There will be no RBI file. In `our RBI generator` we are making sure not to generate any RBI file if there is no scope defined  with this line of code `return if scopes.blank?`

lib/tapioca/compilers/dsl/active_record_scope.rb
katayounHillier: `module` is a key word, it causes issue. I can replace `klass` with `mod`, it wont cause any issue.

lib/tapioca/compilers/dsl/active_record_scope.rb
katayounHillier: I dont have any such variable as `scope_methods`, and Why do we need to pass `scope_methods` ?

lib/tapioca/compilers/dsl/active_record_scope.rb
paracycle: Yes, I meant to say `mod`.

spec/tapioca/compilers/dsl/active_record_scope_spec.rb
paracycle: That means the type won't be generated but there will be an empty RBI file as a result of running the generator.

It is the pipeline that prevents empty RBI files from not being written to disk, not the generator.

But either way, what we are trying to test is the case when the generator runs over a file that does not define any scopes.

lib/tapioca/compilers/dsl/active_record_scope.rb
katayounHillier: I added the lookup in decorate and added `scope_method`  and deleted `constant` since there wont be no need for it.

lib/tapioca/compilers/dsl/active_record_scope.rb
katayounHillier: I did iterate through scopes before calling the `generate_instance_methods`, therefore only sending the scope_method name as parameter.

lib/tapioca/compilers/dsl/active_record_scope.rb
```diff
@@ -0,0 +1,99 @@
...
+
+        private
+
+        sig do
+          params(
+            scope_method: String,
+            mod: Parlour::RbiGenerator::Namespace,
+          ).void
+        end
+        def generate_instance_methods(scope_method, mod)
```
paracycle, line 82: In this case, `generate_scope_method` would be a much better name.

lib/tapioca/compilers/dsl/active_record_scope.rb
paracycle: Thanks, that change makes sense. However, this variable is still not "scopes", it is just a list of method names that corresponds to defined scopes. So a better name would be `scope_method_names`.

spec/tapioca/compilers/dsl/active_record_scope_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+      RUBY
+
+      expected = <<~RUBY
+        # typed: strong
+
+      RUBY
+
+      assert_equal(expected, rbi_for(content))
+    end
+    it("generates RBI file for ActiveRecord classes with a scope field") do
```
paracycle, line 66:
```suggestion

    it("generates RBI file for ActiveRecord classes with a scope field") do
```

lib/tapioca/compilers/dsl/active_record_scope.rb
```diff
@@ -0,0 +1,99 @@
...
+      #
+      # this generator will produce the RBI file `post.rbi` with the following content:
+      #
+      # ~~~rbi
+      # # post.rbi
+      # # typed: true
+      # class Post
+      #   extend Post::GeneratedRelationMethods
+      # end
+      # module Post::GeneratedRelationMethods
```
paracycle, line 36:
```suggestion
      #
      # module Post::GeneratedRelationMethods
```

lib/tapioca/compilers/dsl/active_record_scope.rb
```diff
@@ -0,0 +1,99 @@
...
+      # `Tapioca::Compilers::Dsl::ActiveRecordScope` decorates RBI files for subclasses of
+      # `ActiveRecord::Base` which declare `scope` fields
+      # (see https://api.rubyonrails.org).
+      #
+      # For example, with the following `ActiveRecord::Base` subclass:
+      #
+      # ~~~rb
+      # class Post < ApplicationRecord
+      #   scope :public_kind, -> { where.not(kind: 'private') }
+      #   scope :private_kind, -> {where(kind: 'private')}
```
paracycle, line 24:
```suggestion
      #   scope :private_kind, -> { where(kind: 'private') }
```

spec/tapioca/compilers/dsl/active_record_scope_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+      RUBY
+
+      assert_equal(expected, rbi_for(content))
+    end
+
+    it("generates RBI file for ActiveRecord classes with multiple scope fields") do
+      content = <<~RUBY
+        class Post < ActiveRecord::Base
+          scope :public_kind, -> { where.not(kind: 'private') }
+          scope :private_kind, -> {where(kind: 'private')}
```
paracycle, line 93:
```suggestion
          scope :private_kind, -> { where(kind: 'private') }
```

lib/tapioca/compilers/dsl/active_record_scope.rb
```diff
@@ -0,0 +1,100 @@
...
+
+        sig do
+          override.params(
+            root: Parlour::RbiGenerator::Namespace,
+            constant: T.class_of(::ActiveRecord::Base)
+          ).void
+        end
+        def decorate(root, constant)
+          scope_method_names = constant.send(:generated_relation_methods).instance_methods(false)
+          return if scope_method_names.blank?
```
paracycle, line 56:
```suggestion
          return if scope_method_names.empty?
```

`Array#blank?` is an ActiveSupport extension and is not a part of Ruby. We already have `Array#empty?` in Ruby, and we are just interested in testing for the array being empty, so might as well use that.

PR #85, opened 2020-06-22
Make signature generation safer

paracycle:
There have been cases where signature generation was erroring out due to unloadable constants in some gem. Since signature generation is not an essential part of the operation, we can easily silence the exceptions we are getting during the process and act like there were no signatures.


paracycle:
> Can we add a test with that?

Thanks for that suggestion, it actually caught an edge-case. ❤️

PR #86, opened 2020-06-23
Custom type signature support for AR columns

KaanOzkan:
This PR adds support to understand custom types through `ActiveRecord::Type::Value` method signatures. We look at the method signatures of `deserialize, cast and serialize` methods and use their return types or argument types in the generated signatures.

resolves https://github.com/Shopify/sorbet-issues/issues/193

### DIFF
--- a/lib/tapioca/compilers/dsl/active_record_columns.rb
+++ b/lib/tapioca/compilers/dsl/active_record_columns.rb
@@ -361,8 +361,8 @@ module Tapioca
           signature = T::Private::Methods.signature_for_method(column_type.instance_method(method))
           return unless signature
 
-          signature.return_type.to_s if signature.return_type.to_s != "<VOID>" &&
-              signature.return_type.to_s != "<NOT-TYPED>"
+          return_type = signature.return_type.to_s
+          return_type if return_type != "<VOID>" && return_type != "<NOT-TYPED>"
         end
 
         def lookup_arg_type_of_method(column_type, method)
--- a/lib/tapioca/compilers/dsl/active_record_columns.rb
+++ b/lib/tapioca/compilers/dsl/active_record_columns.rb
@@ -348,14 +348,16 @@ module Tapioca
               !(constant.singleton_class < Object.const_get(:StrongTypeGeneration))
         end
 
-        def handle_unknown_type(constant)
-          if constant.method_defined?(:serialize)
-            method = constant.instance_method(:serialize)
-            signature = T::Private::Methods.signature_for_method(method)
+        def handle_unknown_type(column_type)
+          if column_type.ancestors.include?(ActiveModel::Type::Value)
+            signature = T::Private::Methods.signature_for_method(column_type.instance_method(:deserialize))
+            return signature.return_type.to_s if signature
 
-            if signature
-              return  signature.arg_types.first.last.to_s # Assuming it's correct for arg type to be getter and setter type
-            end
+            signature = T::Private::Methods.signature_for_method(column_type.instance_method(:cast))
+            return signature.return_type.to_s if signature
+
+            signature = T::Private::Methods.signature_for_method(column_type.instance_method(:serialize))
+            return signature.arg_types.first.last.to_s if signature
           end
 
           "T.untyped"
--- a/lib/tapioca/compilers/dsl/active_record_columns.rb
+++ b/lib/tapioca/compilers/dsl/active_record_columns.rb
@@ -349,18 +349,22 @@ module Tapioca
         end
 
         def handle_unknown_type(column_type)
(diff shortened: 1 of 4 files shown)


lib/tapioca/compilers/dsl/active_record_columns.rb
```diff
@@ -347,6 +347,21 @@ def do_not_generate_strong_types?(constant)
           Object.const_defined?(:StrongTypeGeneration) &&
               !(constant.singleton_class < Object.const_get(:StrongTypeGeneration))
         end
+
+        def handle_unknown_type(column_type)
+          if column_type.ancestors.include?(ActiveModel::Type::Value)
+            signature = T::Private::Methods.signature_for_method(column_type.instance_method(:deserialize))
+            return signature.return_type.to_s if signature
```
paracycle, line 354: We need to be a little bit more careful with return types since they can be `T::Private::Types::Void` or `T::Private::Types::NotTyped` both of which are internal Sorbet types.

It might also be nice to extract two methods from here so that this method can just be:
```ruby
def handle_unknown_type(column_type)
  return "T.unyped" unless column_type < ActiveModel::Type::Value

  lookup_return_type_of_method(column_type, :deserialize) ||
    lookup_return_type_of_method(column_type, :cast) ||
    lookup_arg_type_of_method(column_type, :serialize) ||
    "T.untyped"
end
```

lib/tapioca/compilers/dsl/active_record_columns.rb
```diff
@@ -347,6 +347,25 @@ def do_not_generate_strong_types?(constant)
           Object.const_defined?(:StrongTypeGeneration) &&
               !(constant.singleton_class < Object.const_get(:StrongTypeGeneration))
         end
+
+        def handle_unknown_type(column_type)
+          return "T.untyped" unless column_type.ancestors.include?(ActiveModel::Type::Value)
```
paracycle, line 352: Can we change this to be:
```suggestion
          return "T.untyped" unless column_type < ActiveModel::Type::Value
```

lib/tapioca/compilers/dsl/active_record_columns.rb
```diff
@@ -347,6 +347,25 @@ def do_not_generate_strong_types?(constant)
...
+
+          lookup_return_type_of_method(column_type, :deserialize) ||
+            lookup_return_type_of_method(column_type, :cast) ||
+            lookup_arg_type_of_method(column_type, :serialize) ||
+            "T.untyped"
+        end
+
+        def lookup_return_type_of_method(column_type, method)
+          signature = T::Private::Methods.signature_for_method(column_type.instance_method(method))
+          signature.return_type.to_s if signature && !signature.return_type.to_s.include?("T::Private::Types")
```
paracycle, line 362: This might not be the best way to test for `Void` and `NotTyped`. Can we test for them explicitly please? If you don't want to test by doing type comparison, you can also test the `.to_s` output being `<VOID>` or `<NOT-TYPED>`.

lib/tapioca/compilers/dsl/active_record_columns.rb
```diff
@@ -347,6 +347,28 @@ def do_not_generate_strong_types?(constant)
...
+            "T.untyped"
+        end
+
+        def lookup_return_type_of_method(column_type, method)
+          signature = T::Private::Methods.signature_for_method(column_type.instance_method(method))
+          return unless signature
+
+          signature.return_type.to_s if signature.return_type.to_s != "<VOID>" &&
+              signature.return_type.to_s != "<NOT-TYPED>"
+        end
```
KaanOzkan, line 366: This is how I refactored  it, thought it was the easiest to read option. Let me know 👍

lib/tapioca/compilers/dsl/active_record_columns.rb
```diff
@@ -359,7 +359,10 @@ def handle_unknown_type(column_type)
 
         def lookup_return_type_of_method(column_type, method)
           signature = T::Private::Methods.signature_for_method(column_type.instance_method(method))
-          signature.return_type.to_s if signature && !signature.return_type.to_s.include?("T::Private::Types")
+          return unless signature
+
+          signature.return_type.to_s if signature.return_type.to_s != "<VOID>" &&
+              signature.return_type.to_s != "<NOT-TYPED>"
```
paracycle, line 365: This is fine but we keep doing the `.to_s` over and over again, which is not cool. I will suggest:
```suggestion
          return_type = signature.return_type.to_s
          return_type if return_type != "<VOID>" && return_type != "<NOT-TYPED>"
```

paracycle: Basically, the `.to_s` call could potentially be very expensive so best to not repeat it.

PR #87, opened 2020-06-30
Various fixes based on testing on Core

paracycle:
This PR makes a few fixes to ensure everything works properly on Core. I've also taken the opportunity to increase strictness of DSL compilers to `strict`.

It is best to review commit-by-commit

### DIFF
--- a/lib/tapioca/compilers/dsl/smart_properties.rb
+++ b/lib/tapioca/compilers/dsl/smart_properties.rb
@@ -1,4 +1,4 @@
-# typed: true
+# typed: strict
 # frozen_string_literal: true
 
 require "parlour"
@@ -127,7 +127,10 @@ module Tapioca
           klass.create_method(property.reader.to_s, return_type: type) if block.call(property.reader.to_s)
         end
 
-        BOOLEANS = [[true, false], [false, true]].freeze
+        BOOLEANS = T.let([
+          [true, false],
+          [false, true],
+        ].freeze, T::Array[[T::Boolean, T::Boolean]])
 
         sig { params(property: ::SmartProperties::Property).returns(String) }
         def type_for(property)
@@ -159,6 +162,7 @@ module Tapioca
           type
         end
 
+        sig { params(type: Module).returns(String) }
         def name_of(type)
           name = Module.instance_method(:name).bind(type).call
           name.start_with?("::") ? name : "::#{name}"
--- a/lib/tapioca/compilers/dsl/url_helpers.rb
+++ b/lib/tapioca/compilers/dsl/url_helpers.rb
@@ -36,7 +36,7 @@ module Tapioca
 
           constants = ObjectSpace.each_object(Module).select do |mod|
             mod = T.cast(mod, T.class_of(Module))
-            next unless mod.name
+            next unless Module.instance_method(:name).bind(mod).call
 
             includes_helper?(mod, GeneratedUrlHelpersModule) ||
               includes_helper?(mod, GeneratedPathHelpersModule) ||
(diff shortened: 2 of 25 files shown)


lib/tapioca/compilers/dsl/url_helpers.rb
```diff
@@ -36,7 +36,7 @@ def gather_constants
 
           constants = ObjectSpace.each_object(Module).select do |mod|
             mod = T.cast(mod, T.class_of(Module))
-            next unless mod.name
+            next unless Module.instance_method(:name).bind(mod).call
```
katayounHillier, line 39: What sort of issues we are fixing with this change?

lib/tapioca/compilers/dsl/smart_properties.rb
```diff
@@ -127,7 +127,10 @@ def generate_methods_for_property(klass, property, &block)
           klass.create_method(property.reader.to_s, return_type: type) if block.call(property.reader.to_s)
         end
 
-        BOOLEANS = [[true, false], [false, true]].freeze
+        BOOLEANS = T.let([
+          [true, false],
+          [false, true]
+        ].freeze, T::Array[[T::Boolean, T::Boolean]])
```
katayounHillier, line 133: is this change just for making it more secure? or was there an actual issue that this change is fixing

lib/tapioca/compilers/dsl/url_helpers.rb
```diff
@@ -65,11 +65,13 @@ def generate_module_for(root, constant)
           end
         end
 
+        sig { params(mod: Parlour::RbiGenerator::Namespace, constant: T.class_of(Module), helper_module: Module).void }
         def create_mixins_for(mod, constant, helper_module)
-          mod.create_include(helper_module.name) if constant.ancestors.include?(helper_module)
-          mod.create_extend(helper_module.name) if constant.singleton_class.ancestors.include?(helper_module)
+          mod.create_include(T.must(helper_module.name)) if constant.ancestors.include?(helper_module)
```
katayounHillier, line 70: What does `T.must` do here?

lib/tapioca/compilers/dsl/url_helpers.rb
paracycle: If you have a type that redefines the `name` class method, then the value returned from it ends up being useless. For example, if I define a class `Foo`, I would expect `Foo.name == "Foo"` to be true. However, if I defined `Foo` as:
```ruby
class Foo
  def self.name
    nil
  end
end
```
then `Foo.name` would return `nil`, which is not what we want. We want to get the name of the class.

So the solution is to call the `name` instance method of `Module` and call that method on the object that we are working with. So in the above case `Module.instance_method(:name).bind(Foo).call` would correctly return `"Foo"`.

That is what this change is doing.

lib/tapioca/compilers/dsl/smart_properties.rb
paracycle: No, this is necessary for `typed: strict`. At that strictness level, all constants must have their types declared using `T.let`.

lib/tapioca/compilers/dsl/url_helpers.rb
paracycle: In general, the `name` method for a random `Module` could return `nil` if the module in question does not have a name. For example:
```ruby
a = Class.new
a.name # => nil
```
So the return type of the `.name` method is `T.nilable(String)`. But `create_include` is expecting a `String` which should not be `nil`. `T.must` tells Sorbet that we expect the return value of `helper_module.name` to **not** be `nil`. So Sorbet stops complaining. 

By the way, if code reaches this method, we **do** know that the module will definitely have a name, so this is safe to do.

PR #88, opened 2020-06-30
Improve method generation

paracycle:
## Motivation

Currently, `tapioca` is not generating `protected` or `private` class methods of modules/classes. However, that ends up being a problem when a user type is inheriting from or mixing in a class/module and tries to use its protected/private methods. In those cases, Sorbet would complain about missing methods since `tapioca` would not have generated the methods used. For example:
```ruby
# gem file: base.rb
class Base
  class << self
    private
   
    def activate!
      # do some activation
	end
  end
end


# user file: sub.rb
class Sub < Base
  activate!
end
```
would cause a `Method activate! does not exist on T.class_of(Sub)` error.

## Solution

The solution is to also generate the protected and private class methods too. However, we can't do that and still continue to generate methods using the visibility sections, since they don't apply to class methods defined using `self.`.

Thus, I also had to change the output to use `class << self; end` blocks when generating class methods. That way, the `protected`, `private` visibility declarations work properly for class methods.

## Tests

All the tests are updated to comply with the new output format. Moreover, one test file that exercises public/private/protected method generation is expanded to include such class methods too.

### DIFF
--- a/lib/tapioca/compilers/symbol_table/symbol_generator.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_generator.rb
@@ -409,7 +409,7 @@ module Tapioca
           )
 
           instance_methods = compile_directly_owned_methods(name, constant)
-          singleton_methods = compile_directly_owned_methods(name, singleton_class_of(constant), [:public])
+          singleton_methods = compile_directly_owned_methods(name, singleton_class_of(constant))
 
           return if symbol_ignored?(name) && instance_methods.empty? && singleton_methods.empty?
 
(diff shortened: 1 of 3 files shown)


lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -422,24 +422,44 @@ def compile_methods(name, constant)
 
         sig { params(module_name: String, mod: Module, for_visibility: T::Array[Symbol]).returns(String) }
```
KaanOzkan, line 423: Do we want to keep `for_visibility` as an argument for the future? It's not configured anymore.

paracycle: Yes, I thought it might come in handy, so didn't remove it.

PR #89, opened 2020-07-07
Use Spoom's config parser for Sorbet

Morriar:
I extracted the config parser for Sorbet inside [spoom](https://rubygems.org/gems/spoom) that I intend to use as a toolbox for Sorbet related things.

This PR, removes the parser we used and the related tests to use Spoom's one instead.


lib/tapioca/compilers/requires_compiler.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 # typed: strict
 
-require_relative '../sorbet_config_parser'
+require 'spoom'
```
paracycle, line 4: You probably don't need this since `Spoom` is loaded by default.

Morriar: Not require it gives me a "uninitialized constant `::Spoom`", should it not do it? Maybe something is wrong with my gem, I can investigate if it is supposed to be working.

paracycle: It seems you **do** need it after all. My bad, sorry for the confusion.

Morriar:
> It seems you do need it after all.

👍 

I rebased this on master and solved the conflicts in the `Gemfile.lock`.

Issue #65, opened 2020-06-05
Provide RBI generation for Protobuf


Morriar:
Sorbet recently [dropped support for Protobuf](https://github.com/sorbet/sorbet/commit/f20d4ccac5c893c48cf3407b6f8af65aac9ff3ab) and recommend using https://github.com/coinbase/protoc-gen-rbi instead which is a Go tool.

I think we can support this feature in Tapioca.

KaanOzkan:
Looks like the commit message on https://github.com/Shopify/shopify/commit/9eedbbd8179c79b95684be5e7929f990cd80e5dd was wrongly interpreted by Github as it "closes" this issue. Reopening this.

paracycle:
I looked into this a little bit and it looks like this won't be so complicated to pull off. Here are my findings:

1. Protocol Buffers tooling converts `.proto` files to `.rb` files, which are then checked into our repo as the file: `lib/shopify/protobuf/cart.rb`.
2. The types that we want to generate RBIs for in here are the ones declared at the end of the file under the `Shopify::Protobuf` namespace.
3. Each of those types includes a module named `Google::Protobuf::MessageExts`. So we can grab all these types using: `ObjectSpace.each_object(Class).select { |c| c < Google::Protobuf::MessageExts && !c.singleton_class? }`
4. Now, each of these types have a class method named `descriptor` which returns an instance of `Google::Protobuf::Descriptor` which is an enumerable.
5. When enumerated, each item is an instance of `Google::Protobuf::FieldDescriptor`.
6. Each of those objects has methods for getting the:
  - Name of the field using the `name` method
  - Type of the field using the `type` method (and using the `subtype` method if the type is complex)

This should be enough to generate methods that exist at runtime on these Protocol Buffers types.

paracycle:
An example session:
```ruby
pry> ObjectSpace.each_object(Class).select { |c| c < Google::Protobuf::MessageExts && !c.singleton_class? }.fifth.tap { |c| puts c.name }.descriptor.each { |f| puts "#{f.name}: #{f.type} #{f.subtype.try(:enummodule) || f.subtype.try(:msgclass)}" }
Shopify::Protobuf::DiscountApplication
key: string
title: string
description: string
value: message Google::Protobuf::DoubleValue
created_at: message Google::Protobuf::Timestamp
type: string
value_type: enum Shopify::Protobuf::VALUE_TYPE
[...]

PR #90, opened 2020-07-13
Add RBI generator and test for Protobuf

katayounHillier:
## Overview 
This PR :
   - Resolves: issue Shopify/tapioca#65
   - It generates RBI file for Protobuf. You can find [protobuf's documentation on the Google Developers site.](https://developers.google.com/protocol-buffers/)
You can also refer to [this for more info](https://github.com/protocolbuffers/protobuf/blob/master/ruby/ext/google/protobuf_c/message.c#L260-L377)
   - It adds tests for Protobuf RBI generator.

## Testing:
  - dev test 
  - dev style for robocup

I will add documentation to the RBI generator later.


lib/tapioca/compilers/dsl/protobuf.rb
```diff
@@ -0,0 +1,95 @@
...
+
+        sig do
+          override.params(
+            root: Parlour::RbiGenerator::Namespace,
+            constant: T.class_of(Google::Protobuf::MessageExts)
+          ).void
+        end
+        def decorate(root, constant)
+          descriptor = T.let(T.unsafe(constant).descriptor, Google::Protobuf::Descriptor)
+          return if descriptor.to_a.empty?
```
KaanOzkan, line 25: Nitpick: Wanted to make sure that `to_a` is a good way to check for emptiness. There might be an alternative that isn't as costly to allocate a (potentially) large array?

lib/tapioca/compilers/dsl/protobuf.rb
```diff
@@ -0,0 +1,95 @@
...
+        private
+
+        sig do
+          params(
+            descriptor: Google::Protobuf::FieldDescriptor
+          ).returns(String)
+        end
+        def type_of(descriptor)
+          case descriptor.type
+
```
KaanOzkan, line 49: extra new line

lib/tapioca/compilers/dsl/protobuf.rb
katayounHillier: Very good point Kaan! 
 `constant.descriptor` doesn't accept `empty?`, so this is one suggestion that we could come up with.
any other suggestion is appreciated ❤️

KaanOzkan: I found `any?` method locally or `none?`. I didn't go through the source just inspected its method at runtime so take those with a grain of salt 😀

spec/tapioca/compilers/dsl/protobuf_spec.rb
```diff
@@ -0,0 +1,216 @@
...
+          sig { returns(Cart::MYVALUETYPE) }
+          def value_type; end
+
+          sig { params(value: Cart::MYVALUETYPE).returns(Cart::MYVALUETYPE) }
+          def value_type=(value); end
+        end
+      RUBY
+
+      assert_equal(expected, rbi_for(content))
+    end
```
paracycle, line 214: Can we add a test that exercises all the different protobuf types that we map to Ruby types? One `Cart` message that has `int32_field`, `int64_field`, `uint32_field`, etc etc, enumerating all the types that we map would be enough. And you can just match the generated setter methods to assert that we figure out the correct Ruby types for them.

lib/tapioca/compilers/dsl/protobuf.rb
paracycle: Yeah, we can use `any?` here to mean not-`empty?` since we don't expect the contents to have falsy values. However, be careful that in general `any?` is not strictly the same an not-`empty?` since:
```ruby
[false, nil].any? #=> false
```

PR #94, opened 2020-07-22
Add documentation for strict type support

katayounHillier:
This PR resolves [issue #194](https://github.com/Shopify/sorbet-issues/issues/194)
Add a description on `tapioca` support for strict type and the usage of `StongTypeGeneration` module.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,8 +34,6 @@ If you ever run into a case, where you add a gem or update the version of a gem 
 You can use the command `tapioca require` to auto-populate the `sorbet/tapioca/require.rb` file with all the requires found
 in your application. Once the file generated, you should review it, remove all unnecessary requires and commit it.
 
-`tapioca` supports stronger/stricter type generation if class extends `StrongTypeGeneration` module. `StrongTypeGeneration` module does not allow the object to be initialize with a bad state by checking each non_nilable attribute to ensure they are not nil. `tapioca`  ActiveRecordColumns RBI generator allows for more strict type signature if the object class extend `StrongTypeGeneration` by generating sig with exact defined type instead of T.untyped.
-
 ## How does tapioca compare to "srb rbi gems" ?
 
 [Please see the detailed answer on our wiki](https://github.com/Shopify/tapioca/wiki/How-does-tapioca-compare-to-%22srb-rbi-gems%22-%3F)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -34,6 +34,8 @@ If you ever run into a case, where you add a gem or update the version of a gem
 You can use the command `tapioca require` to auto-populate the `sorbet/tapioca/require.rb` file with all the requires found
 in your application. Once the file generated, you should review it, remove all unnecessary requires and commit it.
 
+`tapioca` supports stronger/stricter type generation if class extends `StrongTypeGenerataion` module. `StrongTypeGenerataion` module does not allow the object to be initialize with a bad state by checking each non_nilable attribute to ensure they are not nil. `tapioca`  ActiveRecordColumns RBI generator allows for more strict type signature if the object class extend `StrongTypeGenerataion` by generating sig with exact defined type instead of T.untyped.
```
KaanOzkan, line 37: `StrongTypeGeneration` has a typo in all 3 usages.

KaanOzkan: I would prefer this to be lower if it's going to be in the README as it's a tinier detail compared to the main information at this part of the document. If a wiki page won't be visible enough it'd make sense to insert it towards the end such as after `Flags` section?

katayounHillier: like your suggestion, thanks Kaan!  should I add a header too? what should I use for header? is `StrongTypeGeneration Support` ok?

KaanOzkan: Yes we'd need a header as it doesn't really fit any headers right now. 
`Strong typing option for ActiveRecord column methods` is informative enough I think.

KaanOzkan:
I think we need to make this functionality more explicit:


Tapioca gives you the option to generate stricter type signatures for your ActiveRecord column types. By default, methods generated for columns that are defined in the schema have signatures of `T.untyped`. However, if the object extends a module with name `StrongTypeGeneration`, tapioca will generate stricter signatures that follow closely with the types defined in the schema. Expectation is that the `StrongTypeGeneration` module you define in your application won't allow objects to be initialized with "bad state" by checking all the attributes that are not nilable to ensure they are not nil.

README.md
```diff
@@ -103,6 +103,10 @@ This will generate the file `sorbet/rbi/todo.rbi` defining all unresolved consta
 - `--generate-command [command]`: The command to run to regenerate RBI files (used in header comment of the RBI files), defaults to the current command.
 - `--typed-overrides [gem:level]`: Overrides typed sigils for generated gem RBIs for gem `gem` to level `level` (`level` can be one of `ignore`, `false`, `true`, `strict`, or `strong`, see [the Sorbet docs](https://sorbet.org/docs/static#file-level-granularity-strictness-levels) for more details).
 
+### Strong typing option for ActiveRecord Column methods
```
KaanOzkan, line 106: nitpick: I think "Column" should be lowercase

PR #96, opened 2020-07-23
Avoid generating circular argument references like `_ = _`

Morriar:
With Ruby 2.7, a method definition containing a circular argument reference like this `def foo(a = a)` will create a parsing error. This change is reflected in Sorbet since https://github.com/sorbet/sorbet/pull/3283.

This means that for a gem containing this:

```ruby
class MyClass
  def foo(_ = nil); end
end
```

Tapioca will output the following RBI:

```ruby
class MyClass
  def foo(_ = _); end
end
```

Which once parsed by Sorbet will return a `Circular argument reference '_'`.

To avoid this, this pull-request replaces all the default values generated by `tapioca` as `T.unsafe(nil)` as we do in the stdlib RBIs.

cc @Edouard-chin


sorbet/rbi/gems/actioncable@5.2.4.3.rbi
```diff
@@ -1,6 +1,6 @@
 # DO NOT EDIT MANUALLY
 # This is an autogenerated file for types exported from the `actioncable` gem.
-# Please instead update this file by running `tapioca generate`.
+# Please instead update this file by running `tapioca sync`.
```
Morriar, line 3: I though the right command was `sync` and not `generate`? I can revert those changes if it should be `generate`.

Morriar:
PS: It's easier to read this PR commit by commit as the update of the RBIs creates a lot of noise.

lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -512,13 +512,13 @@ def compile_method(symbol_name, constant, method)
             when :req
               name
             when :opt
-              "#{name} = _"
+              "#{name} = T.unsafe(nil)"
```
Morriar, line 515: This will only work for gems RBIs.

Do we need to apply the same patch to DSLs?

katayounHillier: I assume if we are generating "_" for DSLs , It needs to be changed too so sorbet doesn't complain.

Morriar: I added a commit to fix it too.

Issue #29, opened 2020-01-09
Trim trailing blank lines in generated files


paracycle:
Tapioca ends up creating trailing empty lines, especially when it outputs an empty RBI file. It would be desirable to trim those so that the output is neater and more predictable.

Morriar:
>  when it outputs an empty RBI file

Should it produce empty files?

katayounHillier:
I haven't seen any empty file .

Morriar:
Here's an example of an empty file generated for `websocket-extensions@0.1.4.rbi`: https://github.com/Shopify/shopify/blob/master/sorbet/rbi/gems/websocket-extensions%400.1.4.rbi

I count 48 empty RBIs in Core.

katayounHillier:
We need to delete them. They seem to be  the  left over from year ago. I don't think the latest `tapioca` generate them again. should we delete all the empty file?

Morriar:
Are you sure?

Can you try the following on Core for example:
```
$ rm -rf sorbet/rbi/gems
$ bundle exec tapioca sync
$ cat sorbet/rbi/gems/websocket-extensions@*.rbi
```

The file is empty but still generated.

katayounHillier:
You are correct @Morriar.

katayounHillier:
but this file `websocket_extension@0.1.5.rbi` is generated on `tapioca` under sorbet/rbi/gems and it is not empty. 👀

katayounHillier:
I used `pry` to  debug, I got: 
`uninitialized constant` which is an indicator of missing `include`.
by adding `include "library-name" ` this issue is solved inside tapioca.
I need to do more testing in `shopify/core` to make sure this fix will resolve the issue in core too.

PR #97, opened 2020-07-29
Prevent generating empty RBI file

katayounHillier:
####  Background info
`Bundle exec tapioca generate` or `bundle exec tapioca sync` might generate empty RBI files for some gem or 
  gem's dependencies. 
 [issue #29](https://app.zenhub.com/workspaces/types-at-shopify-5e29d16af2821c45408b8c75/issues/shopify/tapioca/29)

#### What is this PR intending to do
  This PR preventing tapioca from generating empty RBI files. If there is no symbol finds for gem, no RBI file is 
  generated.
  This PR is not solving the root cause of empty RBI issue #29. It only prevents `tapioca` from generating empty `rbi` files. The issue #29 remains open till the root cause is fixed. 

#### Testing & Style
  modified existing tests. did not add any new test. open for suggestions.
  - dev test
  - dev style
  - bin/test


lib/tapioca/generator.rb
```diff
@@ -420,7 +420,7 @@ def compile_gem_rbi(gem)
       compiler = Compilers::SymbolTableCompiler.new
       gem_name = set_color(gem.name, :yellow, :bold)
       say("Compiling #{gem_name}, this may take a few seconds... ")
-
+      return if compiler.compile(gem) == "\n"
```
KaanOzkan, line 423: This means the gem will be compiled twice in the case where we don't return early. We can use the return value of the `compile(gem)` call instead of the second call down below.

katayounHillier: You are absolutely correct @KaanOzkan.  I am going to create a local variable to keep this value.

katayounHillier:
@KaanOzkan, you are correct. Yes, we still need to find the root cause on why some rbi files are blank and empty.

spec/tapioca/cli_spec.rb
```diff
@@ -636,10 +624,9 @@ class Foo::Secret
       output = execute("sync")
 
       assert_includes(output, "-- Removing: #{outdir}/outdated@5.0.0.rbi\n")
-      refute_includes(output, "++ Adding:")
       refute_includes(output, "-> Moving:")
 
-      assert_includes(output, <<~OUTPUT)
+      refute_includes(output, <<~OUTPUT)
```
Morriar, line 629: Why not delete it like the others?

lib/tapioca/generator.rb
```diff
@@ -420,7 +420,8 @@ def compile_gem_rbi(gem)
       compiler = Compilers::SymbolTableCompiler.new
       gem_name = set_color(gem.name, :yellow, :bold)
       say("Compiling #{gem_name}, this may take a few seconds... ")
-
+      compiled_gem_content = compiler.compile(gem)
+      return if compiled_gem_content == "\n"
```
paracycle, line 424: An early return here is problematic, since we will never do the cleanup that we are trying to do at the end of this method (which is fixable by moving the cleanup into an `ensure` section). However, worse than that, we are never signalling what happened with the RBI generation. Normally, we would be printing `Done`, but in this case, we should probably be printing `Skipped` or something.

So I don't think this very early return makes sense.

spec/tapioca/cli_spec.rb
```diff
@@ -583,19 +583,13 @@ class Foo::Secret
       output = execute("sync")
 
       refute_includes(output, "-- Removing:")
-      refute_includes(output, "++ Adding:")
```
paracycle, line 586: I don't think these should change at all.

lib/tapioca/generator.rb
paracycle: Btw, it would be better to check that the content is empty by doing `compiled_gem_content.strip.empty?` instead of checking for equality to some predefined string format.

spec/tapioca/cli_spec.rb
katayounHillier: I will revert all the changes back :)

PR #98, opened 2020-07-29
Remove method shorthand syntax

KaanOzkan:
I didn't explicitly test this solves the error as that work was already done

### DIFF
--- a/lib/tapioca/compilers/symbol_table/symbol_generator.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_generator.rb
@@ -29,7 +29,7 @@ module Tapioca
         def generate
           symbols
             .sort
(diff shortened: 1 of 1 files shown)


Morriar:
Do we need this patch if Sorbet get fixed?

KaanOzkan:
No we shouldn't. But you know more about your particular fix.

I don't know how slow the approval and release a new version process is for `sorbet` so based on Ufuk's suggestion opened this PR thinking it might be faster.

KaanOzkan:
Alex's fix was merged to Sorbet already (https://github.com/sorbet/sorbet/pull/3337) which should fix the error but as Ufuk mentioned for the sake of readability I'll merge this.

PR #99, opened 2020-08-04
Skip generating empty file

katayounHillier:
### Background info
`bundle exec tapioca generate` or `bundle exec tapioca sync` might generate empty RBI files for some gem or
gem's dependencies.

This PR resolves issue #29 and issue [#99](https://github.com/Shopify/tapioca/pull/99)

### What is this PR intending to do
This PR adds message to RBI file. If there is no symbol finds for gem, empty RBI file is
generated with added message.

This PR is not solving the root cause of empty RBI issue #29. It only highlights empty RBI files generated by tapioca. The issue #29 remains open till the root cause is fixed.

### Testing & Style

Still need to add test case for newly added output. 

`dev test`
`dev style`
`bin/test`


katayounHillier:
@paracycle as we discussed, I intentionally didn't fix the tests that are failing for visibility.

lib/tapioca/generator.rb
```diff
@@ -415,27 +415,42 @@ def rbi_header(command, reason: nil, strictness: nil)
...
 
+    sig { returns(String) }
+    def empty_rbi_comment
+      statement = <<~CONTENT
+        # THIS IS AN EMPTY RBI FILE.
+        # see https://github.com/Shopify/tapioca/blob/master/README.md
+      CONTENT
+
+      statement
+    end
```
paracycle, line 426: This does not need to be a method, let's make this a constant.

lib/tapioca/generator.rb
```diff
@@ -415,27 +415,42 @@ def rbi_header(command, reason: nil, strictness: nil)
...
+      rbi_body_content = compiler.compile(gem)
       content = String.new
       content << rbi_header(
         config.generate_command,
         reason: "types exported from the `#{gem.name}` gem",
         strictness: strictness
       )
-      content << compiler.compile(gem)
-
+      content << rbi_body_content
```
paracycle, line 442: This will again put multiple blank lines inside the file, which is what we are trying to avoid in the first place. Let's move this to the `else` block below.

lib/tapioca/generator.rb
```diff
@@ -415,27 +415,42 @@ def rbi_header(command, reason: nil, strictness: nil)
...
       FileUtils.mkdir_p(config.outdir)
       filename = config.outpath / gem.rbi_file_name
-      File.write(filename.to_s, content)
 
-      say("Done", :green)
+      if rbi_body_content.strip.empty?
+        content << empty_rbi_comment
+        output = "Done (Empty_rbi_output)"
```
paracycle, line 448:
```suggestion
        say("Done (empty output)", :yellow)
```

lib/tapioca/generator.rb
```diff
@@ -415,27 +415,42 @@ def rbi_header(command, reason: nil, strictness: nil)
...
+      else
+        output = "Done"
```
paracycle, line 450:
```suggestion
        say("Done", :green)
```

lib/tapioca/generator.rb
```diff
@@ -415,27 +415,42 @@ def rbi_header(command, reason: nil, strictness: nil)
...
+      end
+      File.write(filename.to_s, content)
+      say(output, :green)
```
paracycle, line 453:
```suggestion
```

lib/tapioca/generator.rb
```diff
@@ -422,20 +427,25 @@ def compile_gem_rbi(gem)
...
-      content << compiler.compile(gem)
 
       FileUtils.mkdir_p(config.outdir)
       filename = config.outpath / gem.rbi_file_name
-      File.write(filename.to_s, content)
 
-      say("Done", :green)
+      if rbi_body_content.strip.empty?
+        content << EMPTY_RBI_COMMENT
+        say("Done (EMPTY output)", :yellow)
```
paracycle, line 443: We don't have to shout 😄 
```suggestion
        say("Done (empty output)", :yellow)
```

spec/tapioca/cli_spec.rb
```diff
@@ -602,6 +602,17 @@ class Foo::Secret
...
 
+    it 'generate output  Done (empty output) after generating an empty RBI file' do
+      execute("generate", ["foo"])
+
+      output = execute("sync")
+
+      assert_includes(output, "++ Adding: #{outdir}/ast@2.4.0.rbi\n")
+      assert_includes(output, <<~OUTPUT)
+        Compiling ast, this may take a few seconds...   Done (empty output)
+      OUTPUT
```
paracycle, line 613: Can we also assert the contents of the `ast@2.4.0.rbi` file?

spec/tapioca/cli_spec.rb
```diff
@@ -602,6 +602,34 @@ class Foo::Secret
...
+    end
+
+    it 'Validate the content of a generated empty RBI file' do
+      execute("generate", ["foo"])
+      execute("sync")
+
+      assert_equal(<<~CONTENTS.chomp, File.read("#{outdir}/ast@2.4.1.rbi"))
```
paracycle, line 620: Actually, I feel like this is just one more assertion for the single scenario outlined above, so we should just merge the two tests into one. Basically, we are setting up a single scenario where we know there is an empty RBI that is being generated and we are trying to assert that all the expectations as a result of that happening are behaving as expected.

Thus, there is no need to setup the scenario again in a separate test.

README.md
```diff
@@ -8,6 +8,8 @@ As yet, no gem exports type information in a consumable format and it would be a
 
 When you run `tapioca sync` in a project, `tapioca` loads all the gems that are in your dependency list from the Gemfile into memory. It then performs runtime introspection on the loaded types to understand their structure and generates an appropriate RBI file for each gem with a versioned filename.
 
+## What to do to not get incomplete or empty RBI file for a gem: update the require.rb
```
paracycle, line 11:
```suggestion
## Manual gem requires
```

lib/tapioca/generator.rb
```diff
@@ -162,6 +162,11 @@ def sync_rbis_with_gemfile
 
     private
 
+    EMPTY_RBI_COMMENT = <<~CONTENT
+      # THIS IS AN EMPTY RBI FILE.
+      # see https://github.com/Shopify/tapioca/blob/master/README.md
```
paracycle, line 167:
```suggestion
      # see https://github.com/Shopify/tapioca/blob/master/README.md#manual-gem-requires
```

spec/tapioca/cli_spec.rb
```diff
@@ -602,6 +602,34 @@ class Foo::Secret
...
+        # DO NOT EDIT MANUALLY
+        # This is an autogenerated file for types exported from the `ast` gem.
+        # Please instead update this file by running `generate command`.
+
+        # typed: true
+
+        # THIS IS AN EMPTY RBI FILE.
+        # see https://github.com/Shopify/tapioca/blob/master/README.md
```
paracycle, line 628:
```suggestion
        # see https://github.com/Shopify/tapioca/blob/master/README.md#manual-gem-requires
```

spec/tapioca/cli_spec.rb
KaanOzkan: Wouldn't tests like these would have to be updated when `ast` releases a version `2.4.2`?

Previously this file never relied on real gems. It only uses `bar`, `foo`, `baz` . I think we can achieve the same functionality without checking for "existence of" or "contents of" `ast@2.4.1.rbi`

katayounHillier: I am going to add a fake gem and use that one. :)

PR #100, opened 2020-08-13
A few fixes for custom column type mapping

paracycle:
As I was testing the generators on Core, I ran into two problems with the custom column type mapping:

1. The column type definitions are passed as object instances and not classes. We were not handling them correctly.
2. If columns are mapped to generic types, we can't do the mapping correctly since type parameter information is lost.

This PR fixes these problems.

### DIFF
--- a/lib/tapioca/compilers/dsl/active_record_columns.rb
+++ b/lib/tapioca/compilers/dsl/active_record_columns.rb
@@ -349,9 +349,9 @@ module Tapioca
               !(constant.singleton_class < Object.const_get(:StrongTypeGeneration))
         end
 
-        sig { params(column_type: Module).returns(String) }
+        sig { params(column_type: Object).returns(String) }
         def handle_unknown_type(column_type)
-          return "T.untyped" unless column_type < ActiveModel::Type::Value
+          return "T.untyped" unless ActiveModel::Type::Value === column_type
 
           lookup_return_type_of_method(column_type, :deserialize) ||
             lookup_return_type_of_method(column_type, :cast) ||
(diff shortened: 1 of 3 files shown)


lib/tapioca/compilers/dsl/active_record_columns.rb
```diff
@@ -349,29 +349,37 @@ def do_not_generate_strong_types?(constant)
...
+        sig { params(column_type: ActiveModel::Type::Value, method: Symbol).returns(T.nilable(String)) }
         def lookup_return_type_of_method(column_type, method)
-          signature = T::Private::Methods.signature_for_method(column_type.instance_method(method))
+          signature = T::Private::Methods.signature_for_method(column_type.method(method))
           return unless signature
 
-          return_type = signature.return_type.to_s
-          return_type if return_type != "<VOID>" && return_type != "<NOT-TYPED>"
+          return_type = signature.return_type
+          return if T::Types::Simple === return_type && return_type.raw_type.singleton_class < T::Generic
```
KaanOzkan, line 368: I get ignoring the generic type, can you explain what `T::Types::Simple` is and has to be true to ignore Generics?

KaanOzkan: I'm also curious if we should use `===` instead of `<` to be consistent with the earlier change

paracycle: The value returned from `.return_type` is an instance of a subclass of `T::Types::Base` and the concrete subclass that wraps simple types is `T::Types::Simple`. There are other subclasses that represent union types (i.e `T.any`), intersection types, typed enumerables, etc.

In this case, we are interested in the return type representing a simple type which wraps a `raw_value` that has done an `extend T::Generic`.

paracycle: Made the change to `===`. Thanks, it was very good suggestion.

PR #103, opened 2020-08-18
Start generating tapioca RBI

paracycle:
We now have some runtime components in the form of RBI generators, so we need to start generating an RBI file for tapioca itself.

### DIFF
--- a/lib/tapioca/compilers/dsl/base.rb
+++ b/lib/tapioca/compilers/dsl/base.rb
@@ -1,6 +1,8 @@
 # typed: strict
 # frozen_string_literal: true
 
+require "parlour"
+
 module Tapioca
   module Compilers
     module Dsl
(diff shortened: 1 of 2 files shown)


lib/tapioca/compilers/dsl/base.rb
```diff
@@ -1,6 +1,8 @@
 # typed: strict
 # frozen_string_literal: true
 
+require "parlour"
```
paracycle, line 4: Up to now, we relied on the generators that subclass from `Base` to require `parlour` but actually, this file has a direct dependency on `parlour` types as they are in the signatures. So we really should have had this require here all along.

paracycle:
> I'm curious why tapioca needed a `gemfile.rb` in the first place

That file encapsulates the various bundler operations and provides a higher-level abstraction for some Gem spec properties.

PR #104, opened 2020-08-18
Add revision for git gems

paracycle:
Git gems usually do not change version numbers but they might still have changes if not fixed to a specific SHA. Thus, it makes more sense to track those gems' versions as the gem version plus the git revision they are at as installed on the system.

### DIFF
--- a/lib/tapioca/gemfile.rb
+++ b/lib/tapioca/gemfile.rb
@@ -28,7 +28,7 @@ module Tapioca
     sig { returns(T::Array[Gem]) }
     def dependencies
       @dependencies ||= begin
-        specs = definition.specs.to_a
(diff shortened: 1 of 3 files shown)


paracycle:
> I couldn't figure out what `locked_gems` represents but I guess it doesn't introduce a regression since tests are 🍏

It is basically the set of gems as they are represented in `Gemfile.lock`: https://github.com/rubygems/bundler/blob/master/lib/bundler/definition.rb#L83

We need to use the `locked_gems` since on Core with Nix, the normal gemspecs are taken over by Bundix and all the sources claim that they are installed locally or from rubygems. So, if we just rely on the gemspecs from the `Gemfile` then we can't get to add the revisions like this.

PR #106, opened 2020-08-19
Generate proper url helper RBIs for non-discoverable modules

paracycle:
`ActionDispatch::IntegrationTest` and `ActionView::Helpers` bring url helpers with them but don't necessarily include the url helper modules themselves (they either delegate to the actual modules or mix them into includers dynamically). Thus, we cannot discover that these modules include any url helpers.

However, we do want these modules to act like they include url helper modules, so we have to add them to the gathered constants list manually and also make sure that we generate the module includes for them unconditionally.

So far, we had been adding `ActionDispatch::IntegrationTest` to the gathered constants list manually, but we hadn't been checking if the appropriate module includes were being generated for it (surprise: they weren't). So, this commit adds tests to make sure that we generate includes for those modules specifically.

### DIFF
--- a/lib/tapioca/compilers/dsl/url_helpers.rb
+++ b/lib/tapioca/compilers/dsl/url_helpers.rb
@@ -6,6 +6,7 @@ require "parlour"
 begin
   require "rails"
   require "action_controller"
+  require "action_view"
 rescue LoadError
   return
 end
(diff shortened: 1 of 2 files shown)


paracycle:
> is there any generated RBI file in `core` that has been updated because of this change?

No, I am still working to integrate what we have into Core. These are the fixes coming from there.

PR #107, opened 2020-08-19
Improve app eager loading

paracycle:
## Motivation

We haven't been eager loading Zeitwerk dependencies so far, which was preventing us from discovering some modules that we could generate RBIs for. For example, `ActiveTax` uses Zeitwerk to do its autoloading and has a `FrozenRecord` subclass `ActiveTax::TaxAuthority`. We were not generating an RBI for that class since it was not eager loaded while we were eager loading the Rails app.

## Implementation

It turns out, Rails app finishes its eager loading [in an initializer](https://github.com/rails/rails/blob/7c275525a8f5d055a2e325950b477742d2d2398c/railties/lib/rails/application/finisher.rb#L127-L136) which does a little bit more than just eager loading all the Rails autoloaders. That includes eager loading all of the Zeitwerk dependencies as well. Thus, we need to replicate this logic in our eager loading of the Rails application as well.

## Tests

Made the `baz` gem autoloaded via Zeitwerk and added a `Role` class that has a smart property. Thus, we expect the DSL generation to eager load the `Role` class and generate an RBI for it.

I made sure that this test breaks if we disable eager loading of Zeitwerk dependencies.


paracycle:
> Link in the description is broken ...

Thanks, fixed it.

PR #108, opened 2020-08-19
Fix parameter names for attr writers with signature

paracycle:
## Motivation

Normally `attr_writer` generated methods do not have a name for the single parameter that the method accepts. However, if the `attr_writer` has a signature, then the parameter name should be the name of the attribute.

We have been defaulting the parameter name to `_` if the parameter name didn't exist. But, as outlined above, if the writer method has a signature, that ends up meaning that the parameter name in the signature does not match the one in the method definition.

## Implementation

This commit fixes that by trying to detect writer methods with a signature and naming the parameter properly in those cases.

## Tests

I fixed the test we have for signature generation to ensure that we generate the correct thing.


paracycle:
> My guess is if there's a default value it's not required?

That's exactly it. If there is a default parameter, then you don't need to supply a value for that parameter, thus it is optional.

PR #109, opened 2020-08-20
Modify sigil order in the `config/tapioca/require.rb` to comply with our own Ruby styleguide

Edouard-chin:
The sigil order don't comply with our Rubocop rule. Since this project and the Rubocop-sorbet one are both owned by Shopify, I'd like to have some consistency as otherwise running `tapioca generate rbis` would fail ruby lint out of the box.

Ref https://github.com/Shopify/rubocop-sorbet/blob/7cc2c9fbfd1f4fb4f2e052726e5b745e0235cbae/lib/rubocop/cop/sorbet/sigils/enforce_sigil_order.rb

### DIFF
--- a/lib/tapioca/cli.rb
+++ b/lib/tapioca/cli.rb
@@ -44,8 +44,8 @@ module Tapioca
       end
       create_file(Config::DEFAULT_POSTREQUIRE, skip: true) do
         <<~CONTENT
-          # frozen_string_literal: true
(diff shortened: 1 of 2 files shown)


paracycle (approved):
Great catch! Thanks for fixing this.

Edouard-chin:
Thanks @KaanOzkan !

PR #112, opened 2020-08-25
ActiveRecord changes return nilable types

Morriar:
All the ActiveRecord methods used to access the previous change state of a column can actually return `nil`.

For example:

```ruby
class Foo < ApplicationRecord; end

foo = Foo.new(name: "foo")
foo.name_previous_change # => nil
```

This pull-request uses `T.nilable` as return type for all the impacted methods.


paracycle (approved):
Good catch, thanks!

PR #118, opened 2020-09-01
Fix for mixes_in_class_method

### LINKED ISSUE #111: Fix mixes_in_class_methods discovery
The current implementation does not always produce the correct output. 

For example, instead of:
```ruby
mixes_in_class_methods(::ActiveModel::Validations::ClassMethods)
```
we produce:
```ruby
mixes_in_class_methods(::ActiveModel::Validations::HelperMethods)
```
in `module ActiveModel::Validations` which breaks type-checking.

We need to fix this and come up with a better way of implementing this feature.

katayounHillier:
Resolves #111 

## Motivation

Our solution prior to this PR dynamically discovered list of all module extends that happen as a result of including the module in question. Moreover, it also assumed that we could use the first item in that list to be the module we use for `mixes_in_class_methods`. This was a good assumption to make in cases where we had no more information about what would be the correct `mixes_in_class_methods` call.

However, for the case in #111, the first element in that list ended up being the `ActiveModel::Validations::HelperMethods` module. But, we know that `ActiveModel::Validations` is an `ActiveSupport::Concern` module, and will thus always use `ActiveModel::Validations::ClassMethods` as the primary class methods to be mixed in to base classes. Thus, our assumption ended up conflicting with what we know about the specific `ActiveSupport::Concern` case.

## Implementation

In order to handle this specific case, we added the extra handling for when a module extends `ActiveSupport::Concern`, in which case  `ClassMethods` would always be the correct `mixes_in_class_methods`.

## Testing

Added a new test case to cover the case where the constant extends a module named `"ActiveSupport::Concern"` which mixes in a module named `ClassMethods`. This is tested on a module definition that mimics what `ActiveModel::Validations` is doing, to ensure we treat that case correctly.


lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -384,7 +384,9 @@ def method_missing(symbol, *args)
             end.join("\n")
 
           mixed_in_module = dynamic_extends.find do |mod|
-            mod != constant && public_module?(mod)
+            mod != constant && qualified_name_of(mod) == "::ActiveSupport::Concern" ||
+            (singleton_class_of(constant).ancestors.include?(ActiveSupport::Concern) &&
+             mod == constant.const_get(:ClassMethods))
```
paracycle, line 389: We should not be trying to find a module in the extends that matches the criteria we are looking for. That is not the aim of this exercise.

We have two cases:

1. If the constant extends from a class **named** `"ActiveSupport::Concern"` and if the subconstantant `:ClassMethods` is a `Module`, then we should make `mixed_in_module` to be that module. Nothing more to do here and this is what we used to do before.
2. If the above case is not true, then this is a custom implementation of mixing in class methods, then the code above should try to do its best guess finding as before. Nothing more to do.

sorbet/rbi/gems/activemodel@5.2.4.3.rbi
```diff
@@ -1,6 +1,6 @@
 # DO NOT EDIT MANUALLY
 # This is an autogenerated file for types exported from the `activemodel` gem.
-# Please instead update this file by running `tapioca sync`.
+# Please instead update this file by running `tapioca generate activemodel`.
```
paracycle, line 3: Let's keep this `tapioca sync` please

spec/tapioca/compilers/symbol_table_compiler_spec.rb
```diff
@@ -1382,7 +1383,7 @@ class BasicObjectTest::B < ::BasicObject
       assert_equal(output, source)
     end
 
-    it("adds mixes_in_class_methods to modules that extend base classes") do
+    it("adds mixes_in_class_methods to modules that extend ActiveSupport::Concern classes") do
```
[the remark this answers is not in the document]
paracycle: We had a test before that constructed an `ActiveSupport::Concern` from scratch and tested this with that implementation. After all, we are only matching things **by name**, so we don't need to pull the actual `ActiveSupport::Concern` here.

lib/tapioca/compilers/symbol_table/symbol_generator.rb
katayounHillier: I didn't know we need to cover the else part, thanks Ufuk for explaining, here is what I understood from above: 
```
          extend  = singleton_class_of(constant).ancestors
          if extend.include?(ActiveSupport::Concern)
            mixed_in_module = dynamic_extends.find do |mod|
                mod == constant.const_get(:ClassMethods)
            end
          else
            mixed_in_module = dynamic_extends.find do |mod|
              mod != constant && public_module?(mod)
            end
          end
```
is this the correct translation of what you explained above? 
thnx!

sorbet/rbi/gems/activemodel@5.2.4.3.rbi
katayounHillier: I guess because I used `bundle exec tapioca generate` command, this text has been modified, I didn't change it manually. I guess I need to use `sync` command  instead of `generate` command ?

paracycle: you can manually update this line, but generally it is better to delete the RBI file and run `tapioca sync` instead.

lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -383,8 +383,18 @@ def method_missing(symbol, *args)
...
-          mixed_in_module = dynamic_extends.find do |mod|
-            mod != constant && public_module?(mod)
+          ancestors = singleton_class_of(constant).ancestors
+          extends_as_concern = ancestors.any? do |mod|
+            qualified_name_of(mod) == "::ActiveSupport::Concern"
+          end
+          class_methods_module = resolve_constant("#{name_of(constant)}::ClassMethods")
+
+          mixed_in_module = if extends_as_concern && Module === class_methods_module
+            constant.const_get(:ClassMethods)
```
paracycle, line 393: We already have the `class_methods_module`, why `const_get` again?
```suggestion
            class_methods_module
```

paracycle (commented):
The implementation looks good. However, we need to add tests to make sure the new code path works properly. Also, we need a test case that recreates the problem with `ActiveModel::Validations` (without using the actual implementation) to ensure that we don't regress.

KaanOzkan (approved):
lgtm

PR #135, opened 2020-09-23
Add a `--version` option

Morriar:
Port of https://github.com/Shopify/spoom/pull/12.

### DIFF
--- a/lib/tapioca/cli.rb
+++ b/lib/tapioca/cli.rb
@@ -96,7 +96,7 @@ module Tapioca
 
     desc "--version, -v", "show version"
     def __print_version
-      puts "Tapioca v#{Spoom::VERSION}"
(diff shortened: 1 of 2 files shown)


paracycle (approved):
❤️

lib/tapioca/cli.rb
```diff
@@ -92,6 +94,11 @@ def sync
       end
     end
 
+    desc "--version, -v", "show version"
+    def __print_version
+      puts "Tapioca v#{Spoom::VERSION}"
```
rafaelfranca, line 99: Should this be?

```suggestion
      puts "Tapioca v#{Tapioca::VERSION}"
```

Morriar: Copy pasta fail 🤦

rafaelfranca (approved):

PR #138, opened 2020-09-28
Push for increased typing strictness

paracycle:
## Motivation

The tests and a few other things were ignored from typing for a few reasons. However, there is value in pushing for increased typing, even in tests, since it prevents us from making silly errors. Thus, this PR makes a few changes to push for `typed: true` as a baseline (with a few exceptions).

```
$ bundle exec spoom tc metrics
No errors! Great job.
Sigils:
  files: 161
  ignore: 1 (0%)
  false: 1 (0%)
  true: 86 (53%)
  strict: 70 (43%)
  strong: 3 (1%)

Methods:
  methods: 33269
  signatures: 923 (2%)

Sends:
  sends: 7275
  typed: 5748 (79%)
```

## Implementation

- The implementation heavily relies on the built-in `Minitest` rewriter in Sorbet, which rewrites `describe` and `it` blocks into method definitions for proper testing. However, that requires the test definitions to be wrapped inside a test class, which is added for all tests.
- Now that we were adding a test class for each test, it also made it possible to extract some common functionality into a test superclass. This has allowed us to get rid of some repetitiveness in tests and centralize the late `require` for the class under test and declaring the `subject` block.
- The `rbi_for` method is refactored to accept a constant _name_ to be passed in instead of it expecting a hard-coded constant, which is only found inside the test snippets and fails type-checking.
- Add `rubocop-sorbet` and enforce `typed: true` as the baseline sigil. Plus, fix a couple of rubocop-sorbet cop errors.

## Tests

No extra tests added.

### DIFF
--- a/lib/tapioca/core_ext/class.rb
+++ b/lib/tapioca/core_ext/class.rb
@@ -19,8 +19,10 @@ class Class
   #   C.descendants # => [B, A, D]
   sig { returns(T::Array[Class]) }
   def descendants
-    T.cast(ObjectSpace.each_object(singleton_class).reject do |k|
+    result = ObjectSpace.each_object(singleton_class).reject do |k|
       T.cast(k, Module).singleton_class? || k == self
-    end, T::Array[Class])
+    end
+
+    T.cast(result, T::Array[Class])
   end
 end
--- a/lib/tapioca/core_ext/class.rb
+++ b/lib/tapioca/core_ext/class.rb
@@ -1,7 +1,9 @@
-# typed: false
+# typed: strict
 # frozen_string_literal: true
 
 class Class
+  extend T::Sig
+
   # Returns an array with all classes that are < than its receiver.
   #
   #   class C; end
@@ -15,9 +17,10 @@ class Class
   #
   #   class D < C; end
   #   C.descendants # => [B, A, D]
+  sig { returns(T::Array[Class]) }
   def descendants
-    ObjectSpace.each_object(singleton_class).reject do |k|
-      k.singleton_class? || k == self
-    end
+    T.cast(ObjectSpace.each_object(singleton_class).reject do |k|
+      T.cast(k, Module).singleton_class? || k == self
+    end, T::Array[Class])
   end
 end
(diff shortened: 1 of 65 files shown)


spec/spec_helper.rb
```diff
@@ -16,15 +16,87 @@
...
-    def indented(str, indent)
-      str.lines.map! do |line|
-        next line if line.chomp.empty?
-        " " * indent + line
-      end.join
+  end
+end
+
+class DslSpec < Minitest::Spec
+  before(:all) do
```
Morriar, line 28: Can we add a comment about what is happening here please?

katayounHillier (approved):

lib/tapioca/core_ext/class.rb
```diff
@@ -15,9 +17,10 @@ class Class
...
   #   class D < C; end
   #   C.descendants # => [B, A, D]
+  sig { returns(T::Array[Class]) }
   def descendants
-    ObjectSpace.each_object(singleton_class).reject do |k|
-      k.singleton_class? || k == self
-    end
+    T.cast(ObjectSpace.each_object(singleton_class).reject do |k|
+      T.cast(k, Module).singleton_class? || k == self
+    end, T::Array[Class])
```
Morriar, line 24: Wdyt about using a local variable so the cast is not split over 3 lines?

```suggestion
    res = ObjectSpace.each_object(singleton_class).reject do |k|
      T.cast(k, Module).singleton_class? || k == self
    end
    T.cast(res, T::Array[Class])
```

sorbet/config
```diff
@@ -1,4 +1,3 @@
 --dir
 .
 --ignore=/vendor
---ignore=/spec
```
Morriar, line 4: 🎉

Morriar (approved):
Very nice!

KaanOzkan (commented):
I created a new issue to make similar changes in core https://github.com/Shopify/sorbet-issues/issues/241

KaanOzkan (approved):
lgtm

spec/spec_helper.rb
paracycle: Done!

PR #148, opened 2020-10-20
Add support for T::Struct default values in gem RBI generation

Morriar:
In its current state, Tapioca does not produce default values for `T::Struct` in gems RBI.

So for the following code in a gem:

```rb
# gem.rb
class Foo < T::Struct
 const :a, String, default: "a"
end
```

Tapioca will generate the following RBI:

```rbi
# gem.rbi
class Foo < T::Struct
 const :a, String
end
```

Which will make valid uses of the `Foo` struct in clients wrongly invalid:

```rb
# client.rb

Foo.new # error: missing parameter `:a`
```

This PR allows Tapioca to produce default values when needed by the original implementation:

```rbi
# gem.rbi
class Foo < T::Struct
 const :a, String, default: T.unsafe(nil)
end
```

We use `T.unsafe(nil)` because the `default` field in the struct can contains anything from a primitive data type to a complex statement but since we explore the gem through meta-programming we only see the interpreted value. This is similar behaviour to how we treat arguments with default values.

### DIFF
--- a/lib/tapioca/compilers/symbol_table/symbol_generator.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_generator.rb
@@ -209,7 +209,11 @@ module Tapioca
             method = "const" if prop.fetch(:immutable, false)
             type = prop.fetch(:type_object, "T.untyped")
 
-            indented("#{method} :#{name}, #{type}")
+            if prop.key?(:default)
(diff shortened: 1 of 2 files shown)


KaanOzkan (approved):
Context on slack thread made sense to me, lgtm.

paracycle (approved):

PR #153, opened 2020-10-28
Gather IdentityCache::WithoutPrimaryIndex classes

KaanOzkan:
IdentityCache has a valid use case where classes only include `IdentityCache::WithoutPrimaryIndex` https://github.com/Shopify/identity_cache/blob/master/lib/identity_cache/without_primary_index.rb

We should support this in the generator as well

### DIFF
--- a/lib/tapioca/compilers/dsl/active_record_identity_cache.rb
+++ b/lib/tapioca/compilers/dsl/active_record_identity_cache.rb
@@ -108,7 +108,7 @@ module Tapioca
         sig { override.returns(T::Enumerable[Module]) }
         def gather_constants
           ::ActiveRecord::Base.descendants.select do |klass|
-            klass < IdentityCache || klass < IdentityCache::WithoutPrimaryIndex
+            klass < IdentityCache::WithoutPrimaryIndex
           end
         end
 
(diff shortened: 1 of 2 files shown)


rafaelfranca (approved):

lib/tapioca/compilers/dsl/active_record_identity_cache.rb
```diff
@@ -108,7 +108,7 @@ def decorate(root, constant)
         sig { override.returns(T::Enumerable[Module]) }
         def gather_constants
           ::ActiveRecord::Base.descendants.select do |klass|
-            klass < IdentityCache
+            klass < IdentityCache || klass < IdentityCache::WithoutPrimaryIndex
```
paracycle, line 111: I think the first always includes the second, so it should be enough to just check for the second.

KaanOzkan: Oh yeah it is enough to only check for the second. The `<` is not ideal to me then. I thought it would effectively be "recursive" when checking for a subclass.

paracycle: `<` does not mean subclass, though. It looks to see if the given module is in the list of ancestors. There is nothing recursive that is needed, every class that pulls in `IdentityCache` also pulls in `IdentityCache::WithoutPrimaryIndex` into its list of ancestors, as well.

KaanOzkan: Yes, I can't figure out what I was thinking when I wrote that comment now 😄

paracycle (approved):

PR #156, opened 2020-11-03
Better generics support

paracycle:
### Motivation

Since Tapioca uses runtime reflection to discover types, it hasn't been able to discover and generate generic types and type variables up to now. The problem was that Sorbet runtime erases all generic types at runtime and keeps no record of them. Similarly, type variables are declared in a way that the generated type variable value is stored in a subconstant of the type, which makes it hard to understand which subconstants represent type variables and which ones are real (value or module) constants. Moreover, type variables don't store all of their original data, which makes it impossible to generate the correct code for them, even if we could resolve them.

### Implementation

The implementation patches some Sorbet classes, most notably `T::Generic`, to intercept calls to generic type instantiation and type variable definitions. Tapioca stores the data from those calls in a `GenericTypeRegistry` which can then be used for looking up the original call details when we are trying to do code generation.

For the `[]` method on `T::Generic`, which is responsible for instantiating a generic type, we build the name of the instantiated concrete type and clone the current class/module so that we can return a type that is the same as the current type but is a different instance and has a different `name` method. We cache those cloned instances by their name, so that we don't keep instantiating a type every single time it is referenced. Thus `[Foo[Integer], Foo[Integer], Foo[Integer], Foo[String]]` will only result in 2 clones (1 for `Foo[Integer]` and another for `Foo[String]`) and 2 hash lookups (for the two `Foo[Integer]`s).

Once the above is done, when we query the name of the wrapped type on a `T::Types::Simple` it should give us the expected name for a `T::Generic` instance.

For type variables, we do something similar. We again intercept calls to `type_member` and `type_template`, pass all the arguments and the `T::Types::TypeVariable` instance generated by the actual `type_member`/`type_template` call on `T::Generic` (i.e. the result of `super`) to `GenericTypeRegistry`. The type registry, in turn, builds a `TypeVariableSerializer` with that data and stores that in a lookup up table keyed by the constant that defines the type variable and then the `T::Types::TypeVariable` instance generated by the normal `T::Generic` operation.

With that setup, during generation time, we can query each type against the type registry and check to see if there are any registered type variables for that type. This is a simple has lookup and is super cheap if the type is not generic.

If the type is generic, though, we build a `subconstant_name => subconstant_value` lookup and use the `type_variable => type_variable_serializer` lookup from the registry to resolve a `subconstant_name =>  type_variable_serializer` table. The key idea here is that for a generic type `Foo` that has a `Elem = type_member(:in)` declaration, the value of `Foo::Elem` will be the result of the `type_member` call. But that is the `type_variable` that we key our lookup table with, so we can relate `Foo::Elem` to its type serializer, which has all the data to generate `type_member(:in)`

### Tests
Added tests to verify that generic type generation works and that we can generate definitions for type variables.


spec/spec_helper.rb
```diff
@@ -1,7 +1,7 @@
 # typed: strict
 # frozen_string_literal: true
 
-require "tapioca"
+require "tapioca/all"
```
Morriar, line 4: I don't understand what we're doing here.

Since `tapioca` loads `tapioca/all`, can't we keep `require 'tapioca'`? 

What's the point of this commit?

paracycle: We've been doing the exports from the gem incorrectly up to now, and this commit is fixing it.

All the compilers and generators, etc are internal types of Tapioca and should not be loaded for clients that do a simple `gem "tapioca"` in their `Gemfile`. Those clients will end up requiring Tapioca and will get all kinds of compilers/generators loaded.

This didn't use to be a problem until now, since Tapioca was not modifying global state (much). Now that we are patching `T::Types::Simple` and `T::Types::Simple`, we really should not allow clients to load those classes unless they really need to.

The name `tapioca/all` for that optional require might not be the best. Maybe it should be `tapioca/internal` or something, I am open to ideas.

Basically:
- `gem 'tapioca'` loads `tapioca` which only loads the classes needed to write DSL generators.
- Running `tapioca` loads `cli.rb` which loads `tapioca/all` which loads everything that is needed to do a full Tapioca run.
- Similarly, spec helper loads `tapioca/all` so that test can find all the constants they expect.

Morriar: Makes sense, thanks for the explanation.

Yeah, if we want to discourage people to load this it should not be called `all`. `internal` is scary, I like it 👍

paracycle: I moved this into its own PR: https://github.com/Shopify/tapioca/pull/261

lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -255,6 +260,44 @@ def compile_subconstants(name, constant)
           "\n" + output.join("\n\n")
         end
 
+        sig { params(value: T.untyped).returns(String) }
```
Morriar: `sig { params(value: Class).returns(String) }` ?

lib/tapioca/sorbet_ext/generic_name_patch.rb
```diff
@@ -13,29 +13,35 @@ def [](*types)
...
 
       def type_member(variance = :invariant, fixed: nil, lower: T.untyped, upper: BasicObject)
         super.tap do |type_member|
           type_member.singleton_class.instance_exec do
-            define_method(:fixed) { fixed }
-            define_method(:lower) { lower unless lower == T.untyped }
-            define_method(:upper) { upper unless upper == BasicObject }
+            T.unsafe(self).define_method(:fixed) { fixed }
+            T.unsafe(self).define_method(:lower) { lower unless lower == T.untyped }
+            T.unsafe(self).define_method(:upper) { upper unless upper == BasicObject }
```
Morriar: Is `unless upper == BasicObject` to avoid printing the default values for those bounds in the generated RBI?

lib/tapioca/sorbet_ext/generic_name_patch.rb
```diff
@@ -13,29 +13,35 @@ def [](*types)
...
           end
+          T::Private::Abstract::Data.set_default(self, "type_variables", []) << type_template
         end
       end
+
+      def __type_variables
```
Morriar: So it's "internal" but not `private`?

lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -260,37 +260,40 @@ def compile_subconstants(name, constant)
...
-            output = type_constants_with_index.sort_by(&:last).map do |constant_name, value|
-              if T::Types::TypeMember === value
-                indented("#{constant_name} = type_member(#{compile_type_variable_args(value)})")
-              elsif T::Types::TypeTemplate === value
-                indented("#{constant_name} = type_template(#{compile_type_variable_args(value)})")
+            return unless type_variables && Array === type_variables
+
+            # For each subconstant defined under this constant
+            # we resolve it to a subconstant object and look up its index the
+            # value of it in the type variable list that we have collected.
```
Morriar: Seems like something is missing here: `look up its index the value of it`

lib/tapioca/sorbet_ext/generic_name_patch.rb
```diff
@@ -4,20 +4,52 @@
...
 
 module T
+  module Types
+    class CustomTypeVariable < T::Types::TypeVariable
+      def initialize(type_variable_type, type_variable, fixed, lower, upper)
+        parts = []
+        parts << ":#{type_variable.variance}" unless type_variable.variance == :invariant
+        parts << "fixed: #{fixed}" if fixed
+        parts << "lower: #{lower}" unless lower == T.untyped
+        parts << "upper: #{upper}" unless upper == BasicObject
```
Morriar: Aren't we returning `nil` if `upper == BasicObject` in  `TypeStoragePatch`?

Morriar: Ah it was moved around in this commit 👍

lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -480,29 +480,19 @@ def compile_methods(name, constant)
...
 
-          return if symbol_ignored?(name) && instance_methods.empty? && singleton_methods.empty?
+          return if symbol_ignored?(name) && !instance_methods && !singleton_methods
 
           [
             initialize_method || "",
             instance_methods,
             singleton_methods,
-          ].select { |b| b.strip != "" }.join("\n\n")
+          ].select { |b| b && b.strip != "" }.join("\n\n")
```
Morriar, line 511: `!b.strip.empty?` ?

lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -195,11 +195,11 @@ def compile_body(name, constant)
               compile_props(constant),
               compile_enums(constant),
               methods,
-            ].select { |b| b != "" }.join("\n\n")
+            ].select { |b| b && b != "" }.join("\n\n")
```
Morriar, line 198: `!b.empty?`?

lib/tapioca/generic_type_registry.rb
```diff
@@ -126,7 +129,7 @@ def lookup_or_initialize_type_variables(constant)
         @type_variables[object_id_of(constant)] ||= {}
       end
 
-      sig { params(constant: Module).returns(T.nilable(String)) }
+      sig { params(constant: Module).returns(T.nilable(String)).checked(:never) }
```
Morriar, line 132: Why `:never` checked?

Morriar (commented):
Looks amazing, left a couple of questions as to better understand what's going on.

lib/tapioca/sorbet_ext/generic_name_patch.rb
paracycle: This was refactored later so that we don't have to store anything on the constant itself. That's why `GenericTypeRegistry` was created.

lib/tapioca/compilers/symbol_table/symbol_generator.rb
paracycle: These comments were refactored heavily in later commits, so this was probably fixed.

lib/tapioca/sorbet_ext/generic_name_patch.rb
paracycle: Yes, `BasicObject` is the default value of `upper` and `T.untyped` is the default value of `lower`, so, I am not generating a `lower:`/`upper:` definition if those are the values to keep the noise levels down in the generated RBI file.

lib/tapioca/compilers/dsl/active_record_columns.rb
```diff
@@ -380,12 +379,7 @@ def lookup_return_type_of_method(column_type, method)
         sig { params(column_type: ActiveModel::Type::Value, method: Symbol).returns(T.nilable(String)) }
         def lookup_arg_type_of_method(column_type, method)
           signature = T::Private::Methods.signature_for_method(column_type.method(method))
-          return unless signature
-
-          arg_type = signature.arg_types.first.last
-          return if T::Types::Simple === arg_type && T::Generic === arg_type.raw_type
-
-          arg_type.to_s
+          signature.arg_types.first.last.to_s if signature
```
Morriar, line 382: Do you mind using intermediate variable with meaningful names so we can understand what we're getting from the `first` then `last`?

spec/tapioca/compilers/symbol_table_compiler_spec.rb
```diff
@@ -2193,6 +2201,27 @@ def do_it
...
           extend(T::Sig)
           extend(T::Helpers)
+          extend(T::Generic)
+
+          A = type_template(:in)
+          B = type_template(:out)
+          C = type_template
+
+          D = type_member(fixed: Integer)
+          E = type_member(fixed: Integer, upper: Numeric)
```
Morriar, line 2211: Can we add a test where the bound is itself generic?

Like here: https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20Foo%0A%20%20extend%20T%3A%3ASig%0A%20%20extend%20T%3A%3AGeneric%0A%0A%20%20E%20%3D%20type_member(upper%3A%20T%3A%3AArray%5BInteger%5D)%0A%0A%20%20sig%20%7Bparams(x%3A%20Integer).void%7D%0A%20%20def%20foo(x)%0A%20%20%20%20puts(x%20%2B%201)%0A%20%20end%0Aend

Morriar: Maybe also with a type modifier: https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20Foo%0A%20%20extend%20T%3A%3ASig%0A%20%20extend%20T%3A%3AGeneric%0A%0A%20%20E%20%3D%20type_member(upper%3A%20T.nilable(Integer))%0Aend

ttonelli:
I can't vouch for the implementation but I can confirm that this solves the problems in `flow-runtime` and `flow`, both of which use generic definitions from `maestro` and `shopify-adt`. Thanks!

spec/tapioca/compilers/symbol_table_compiler_spec.rb
paracycle: Yeah, good call.

lib/tapioca/compilers/symbol_table/symbol_generator.rb
paracycle: This is outdated

lib/tapioca/generic_type_registry.rb
paracycle: I think those were left over from something else, will fix.

lib/tapioca/sorbet_ext/generic_name_patch.rb
```diff
@@ -0,0 +1,56 @@
+# typed: false
+# frozen_string_literal: true
+
+require 'tapioca/sorbet_ext/name_patch'
+
+module T
```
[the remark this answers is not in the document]
paracycle: I improved documentation throughout.

lib/tapioca/generic_type_registry.rb
```diff
@@ -0,0 +1,178 @@
+# typed: strict
+# frozen_string_literal: true
+
+module Tapioca
+  # This class responsible for storing and looking up information related to generic types.
```
Morriar, line 5:
```suggestion
  # This class is responsible for storing and looking up information related to generic types.
```

lib/tapioca/generic_type_registry.rb
```diff
@@ -0,0 +1,178 @@
...
+  # The class stores 2 different kinds of data, in two separate lookup tables:
+  #   1. a lookup of generic type instances by name: `@generic_instances`
+  #   2. a lookup of type variable serializer by constant and type variable
+  #      instance: `@type_variables`
+  #
+  # By storing the above data, we can cheaply query each constant against this registry
+  # to see if it declares any generic type variables. This becomes a simple lookup in the
+  # `@type_variables` hash table with the given constant.
+  #
+  # If there are no entries, then we can cheaply know that we can skip generic type
```
Morriar, line 16:
```suggestion
  # If there is no entry, then we can cheaply know that we can skip generic type
```

lib/tapioca/generic_type_registry.rb
```diff
@@ -0,0 +1,178 @@
...
+
+    @type_variables = T.let(
+      {},
+      T::Hash[Integer, T::Hash[Integer, TypeVariableSerializer]]
+    )
+
+    class << self
+      extend T::Sig
+
+      # This method is responsible building the name of the instantiated concrete type
```
Morriar, line 66:
```suggestion
      # This method is responsible for building the name of the instantiated concrete type
```

lib/tapioca/generic_type_registry.rb
```diff
@@ -0,0 +1,178 @@
...
+      # and cloning the given constannt so that we can return a type that is the same
```
Morriar, line 67:
```suggestion
      # and cloning the given constant so that we can return a type that is the same
```

lib/tapioca/generic_type_registry.rb
```diff
@@ -0,0 +1,178 @@
...
+      # as the current type but is a different instance and has a different name method.
+      #
+      # We cache those cloned instances by their name in `@generic_instances`, so that
+      # we don't keep instantiating a new type every single time it is referenced.
+      # For example, `[Foo[Integer], Foo[Integer], Foo[Integer], Foo[String]]` will only
+      # result in 2 clones (1 for `Foo[Integer]` and another for `Foo[String]`) and
```
Morriar, line 73: Shouldn't it also create a type for the tuple: `[Foo[Integer], Foo[Integer], Foo[Integer], Foo[String]]`?

lib/tapioca/sorbet_ext/generic_name_patch.rb
```diff
@@ -0,0 +1,54 @@
+# typed: true
+# frozen_string_literal: true
+
+require "tapioca/sorbet_ext/name_patch"
+
+module T
+  module Generic
+    # This module intercepts calls to generic type instantiation and type variable definitions.
```
Morriar, line 8:
```suggestion
    # This module intercepts calls to generic type instantiations and type variable definitions.
```

lib/tapioca/sorbet_ext/generic_name_patch.rb
```diff
@@ -0,0 +1,54 @@
...
+    # Tapioca stores the data from those calls in a `GenericTypeRegistry` which can then be used
+    # for looking up the original call details when we are trying to do code generation.
```
Morriar, line 10:
```suggestion
    # to look up the original call details when we are trying to do code generation.
```

Morriar (commented):
The documentation helps a lot!

lib/tapioca/generic_type_registry.rb
paracycle: No, since the tuple is not generic.

cjhutchi (approved):
Went through this with @paracycle and team. Makes significantly more sense to me now and _somewhat_ understand what's going on (🤣). The group walkthrough was super useful in finding improvements, so I don't really have anything to add at this time. 

Really great work!

lib/tapioca/sorbet_ext/generic_name_patch.rb
```diff
@@ -0,0 +1,54 @@
...
+  module Generic
+    # This module intercepts calls to generic type instantiations and type variable definitions.
+    # Tapioca stores the data from those calls in a `GenericTypeRegistry` which can then be used
+    # to look up the original call details when we are trying to do code generation.
+    #
+    # We are interested in the data of the `[]`, `type_member` and `type_template` calls which
+    # are all needed to generate good generic information at runtime.
+    module TypeStoragePatch
+      def [](*types)
+        Tapioca::GenericTypeRegistry.register_type(self, types)
```
RyanBrushett, line 16: Maybe a comment here that `[]` from upstream simply returns `self`? Alternatively: use `super`

RyanBrushett (approved):
I very much appreciate your PR description and comments in the code. Thanks for this.

lib/tapioca/sorbet_ext/generic_name_patch.rb
paracycle: Yeah, good call!

paracycle: Done

Morriar (approved):

PR #159, opened 2020-11-04
Fix Sorbet ignore matching

paracycle:
### Motivation

I pushed a change to cache bundler installed gems on `master` which ended up breaking tests. The change installs gems into the `vendor/bundle` folder. Even though `/vendor` is an `--ignore` flag in `sorbet/config`, our tests were somehow still being affected by the existence of the `vendor` folder.

I investigated the cause to Sorbet ignore matching not exactly working exactly like a regular expression match. Instead, it tries to match whole path segments unless the ignore path starts with a `/` character in which case it tries to match the prefix of relative paths.

### Implementation

The new implementation first tries to get the relative path of each found file to the working path. All matches are done against this relative path. For each ignore string, if it starts with a `/`, the first segment of the relative path. Otherwise, we expect it to match a whole part of the relative path.

### Tests

I haven't added any extra tests, but all the existing ones pass.

### DIFF
--- a/lib/tapioca/compilers/requires_compiler.rb
+++ b/lib/tapioca/compilers/requires_compiler.rb
@@ -34,7 +34,8 @@ module Tapioca
           path = (Pathname.new(@sorbet_path) / "../.." / path).cleanpath
           if path.directory?
             Dir.glob("#{path}/**/*.rb", File::FNM_EXTGLOB).reject do |file|
-              file_ignored_by_sorbet?(config, file)
+              relative_file_path = Pathname.new(file).relative_path_from(path)
+              file_ignored_by_sorbet?(config, relative_file_path)
             end
           else
             [path.to_s]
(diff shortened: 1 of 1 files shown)


lib/tapioca/compilers/requires_compiler.rb
```diff
@@ -49,13 +50,40 @@ def collect_requires(file_path)
...
+          file_path_part_sequences = [file_path_part_sequences.first].to_enum if ignore.start_with?("/")
+
+          # we need to match whole segments
+          file_path_part_sequences.include?(ignore_parts)
         end
       end
 
+      sig { params(path: Pathname).returns(T::Array[String]) }
+      def path_parts(path)
+        T.unsafe(path).descend.map { |part| part.basename.to_s }
```
rafaelfranca, line 84: Why the unsafe here. Isn’t descend in the Pathname class?

rafaelfranca (approved):

lib/tapioca/compilers/requires_compiler.rb
paracycle: Sorry I forgot to note that. It is because of this: https://github.com/sorbet/sorbet/pull/3617

paracycle: Basically, the block argument should be optional but isn't

lib/tapioca/compilers/requires_compiler.rb
```diff
@@ -49,13 +50,40 @@ def collect_requires(file_path)
         end.compact
       end
 
-      sig { params(config: Spoom::Sorbet::Config, file: String).returns(T::Boolean) }
-      def file_ignored_by_sorbet?(config, file)
-        config.ignore.any? do |path|
-          Regexp.new(Regexp.escape(path)) =~ file
+      sig { params(config: Spoom::Sorbet::Config, file_path: Pathname).returns(T::Boolean) }
+      def file_ignored_by_sorbet?(config, file_path)
```
Morriar, line 54: Should we move this service to `Spoom`? Seems useful.

Morriar (approved):

lib/tapioca/compilers/requires_compiler.rb
paracycle: Yeah, I was thinking the same thing. Let's do it in a future refactor, and we can add tests around it as well.

PR #165, opened 2020-11-19
Make inclusion order deterministic

Morriar:
While bumping `spoom` the test suite returned errors because of differences in the inclusion order: https://github.com/Shopify/tapioca/runs/1424355030.

This PR makes the order of include/prepend/append deterministic sorting them by their qualified name so it doesn't happen again.

While the order may have an importance regarding behaviour, since the RBIs are only read by `sorbet-static`, this has no impact on the runtime execution.

### DIFF
--- a/lib/tapioca/compilers/symbol_table/symbol_generator.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_generator.rb
@@ -329,7 +329,7 @@ module Tapioca
               # properly for method resolution, so we generate an
               # include statement instead
               indented("include(#{qualified_name_of(mod)})")
-            end
+            end.sort
 
           includes = include
             .reverse
@@ -337,7 +337,7 @@ module Tapioca
             .select(&method(:public_module?))
             .map do |mod|
               indented("include(#{qualified_name_of(mod)})")
-            end
+            end.sort
 
           extends = extend
             .reverse
@@ -345,7 +345,7 @@ module Tapioca
             .select(&method(:public_module?))
             .map do |mod|
               indented("extend(#{qualified_name_of(mod)})")
-            end
+            end.sort
 
           (prepends + includes + extends).join("\n")
         end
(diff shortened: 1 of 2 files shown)


KaanOzkan (approved):
I'm curious as to how a spoom change caused this.

paracycle (changes requested):
This is not correct, I am afraid, since the order of included/extended modules matter and we should not be changing the order in any way.

There could be huge differences between inclusion orders like so:
```ruby
module A
  def foo; 1; end
end

module B
  def foo; 2; end
end

class C
  include A
  include B
end

class D
  include B
  include A
end

C.new.foo #=> 2
D.new.foo #=> 1
```

Coincidentally, that is why we have the `reverse` in the chain here.

Morriar:
Actually it also changes signature resolution: https://sorbet.run/#%23%20typed%3A%20true%0A%0Amodule%20A%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7B%20returns(String)%20%7D%0A%20%20def%20foo%3B%20%22%22%3B%20end%0Aend%0A%0Amodule%20B%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7B%20returns(Integer)%20%7D%0A%20%20def%20foo%3B%202%3B%20end%0Aend%0A%0Aclass%20C%0A%20%20include%20A%0A%20%20include%20B%0Aend%0A%0Aclass%20D%0A%20%20include%20B%0A%20%20include%20A%0Aend%0A%0AT.reveal_type(C.new.foo)%0AT.reveal_type(D.new.foo)

paracycle:
Yes, since Sorbet respects the same inclusion order. I had an error before where I hadn't included the reversal and Sorbet was matching `find` calls on AR objects to `Enumerable#find` instead of `ActiveRecord::Finders#find`.

Morriar:
> I'm curious as to how a spoom change caused this.

My guess is that since I changed the module hierarchy in `spoom` and added JSON support for coverage, JSON is now loaded earlier and monkey patches `String` before `Colorize`.

PR #169, opened 2020-11-26
Generate nested definitions for synthetic modules

paracycle:
### Motivation

Since Sorbet has started checking for accessing private constants and erroring out on referencing a private constant off of its parent, our generated DSL RBI files have started raising errors.

If the declaration of `Bar` was:
```ruby
module Foo
  class Bar
    # some definition
  end
  private_constant :Bar
end
```

then it is illegal to reference `Bar` as `Foo::Bar`. But we had been generating things like:
```ruby
module Foo
  class Bar
    include Foo::Bar::GeneratedMethods
  end
end

module Foo::Bar::GeneratedMethods
end
```

which raises the following error on recent Sorbet versions:
```
Non-private reference to private constant Foo::Bar referenced https://srb.help/5061
      3 |     include Foo::Bar::GeneratedMethods
                      ^^^^^^^^
```

### Implementation

The fix is to generate everything in a nested way, which actually ended up making the code simpler and easier to read as well.

The idea is to generate nested modules on the generated class/module namespace instead of the `root` namespace, which results in the output like below for the example above:
```ruby
module Foo
  class Bar
    include GeneratedMethods

	module GeneratedMethods
	end
  end
end
```

### Tests

No extra tests added, but current ones are updated with new expectations.

### DIFF
--- a/lib/tapioca/compilers/dsl/action_controller_helpers.rb
+++ b/lib/tapioca/compilers/dsl/action_controller_helpers.rb
@@ -120,7 +120,7 @@ module Tapioca
         def gather_includes(mod)
           mod.ancestors
             .reject { |ancestor| ancestor.is_a?(Class) || ancestor == mod || ancestor.name.nil? }
-            .map { |ancestor| T.must(ancestor.name) }
+            .map { |ancestor| "::#{ancestor.name}" }
             .reverse
         end
       end
--- a/lib/tapioca/compilers/dsl/action_controller_helpers.rb
+++ b/lib/tapioca/compilers/dsl/action_controller_helpers.rb
@@ -49,22 +49,22 @@ module Tapioca
       # # user_controller.rbi
       # # typed: strong
       # class UserController
-      #   sig { returns(UserController::HelperProxy) }
+      #   module HelperMethods
+      #     include MyHelper
+      #
+      #     sig { params(user: T.untyped).returns(T.untyped) }
+      #     def age(user); end
+      #
+      #     sig { returns(T.untyped) }
+      #     def current_user_name; end
+      #   end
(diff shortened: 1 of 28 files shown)


lib/tapioca/compilers/dsl/action_controller_helpers.rb
```diff
@@ -50,22 +50,22 @@ module Dsl
       # # user_controller.rbi
       # # typed: strong
       # class UserController
-      #   sig { returns(UserController::HelperProxy) }
-      #   def helpers; end
-      # end
+      #   module HelperMethods
+      #      include MyHelper
```
Morriar, line 54: Indent seems incorrect

lib/tapioca/compilers/dsl/active_support_current_attributes.rb
```diff
@@ -113,12 +113,9 @@ def instance_methods_for(constant)
 
         sig { params(klass: Parlour::RbiGenerator::Namespace, method: String, class_method: T::Boolean).void }
         def generate_method(klass, method, class_method:)
-          if method.end_with?("=")
-            parameter = Parlour::RbiGenerator::Parameter.new("value", type: "T.untyped")
-            klass.create_method(method, class_method: class_method, parameters: [parameter], return_type: "T.untyped")
-          else
-            klass.create_method(method, class_method: class_method, return_type: "T.untyped")
-          end
+          parameters = []
```
Morriar, line 116: It's a matter of preference but the previous form with a `if .. else` was stating the intent better as the condition was showed upfront. Now the condition is hidden at the end of an already long line.

Morriar (approved):
Nice!

PR #173, opened 2020-12-01
Use `polymorphic?` instead of `constructable?`

Morriar:
We've got reports of `tapioca dsl` failing with:

```
undefined method constructable?' for #<ActiveRecord::Reflection::BelongsToReflection> (NoMethodError)`
```

This seems to be related to an API change on Rails master: https://github.com/rails/rails/commit/371b4800266cbf1c659a9068f58412cb7a88e244

This PR uses `polymorphic?` instead of the now removed `constructable?` method.


rafaelfranca (approved):

paracycle (approved):
Awesome, thanks for finding and linking to the root cause.

PR #178, opened 2020-12-02
Make Rubocop TargetRubyVersion match gemspec

dduugg:
### Motivation

Rubocop should target the minimum ruby version required by the gem.

### Implementation

Updated `.rubocop.yml` to match the minimum required by `tapioca.gemspec`.

### Tests

None, but this can be enforced by enabling `Gemspec/RequiredRubyVersion`. I recommend doing so in https://github.com/Shopify/ruby-style-guide

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,7 +7,7 @@ require:
  - rubocop-sorbet
 
 AllCops:
-  TargetRubyVersion: 2.3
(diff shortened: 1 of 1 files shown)


KaanOzkan (approved):

Morriar (approved):
Thanks!

PR #184, opened 2020-12-10
Fix ActiveRecordColumns spec wrt StrongTypeGeneration

paracycle:
### Motivation

Contrary to what the documentation of `ActiveRecordColumns` generator says, we actually generate strong types for models if `StrongTypeGeneration` is not defined in the user application as a constant.

We only generate weak types if a constant named `StrongTypeGeneration` is defined in the user application and the current model does not extend from it.

Because of this disconnect, we were never really testing cases where we didn't have `StrongTypeGeneration` defined. However, that is usually our default state, so most test cases should have really be testing that.

### Implementation

Added separate test cases that test default behaviour and behaviour when `StrongTypeGeneration` is defined.

### Tests

Tests are updated to reflect the true state of affairs.

### DIFF
--- a/lib/tapioca/compilers/dsl/active_record_columns.rb
+++ b/lib/tapioca/compilers/dsl/active_record_columns.rb
@@ -17,19 +17,7 @@ module Tapioca
       # This generator is only responsible for defining the attribute methods that would be
       # created for the columns that are defined in the Active Record model.
       #
-      # **Note:** This generator, by default, generates weak signatures for column methods and treats each
-      # column to be `T.untyped`. This is done on purpose to ensure that the nilability of Active Record
-      # columns do not make it hard for existing code to adopt gradual typing. It is possible, however, to
-      # generate stricter type signatures for your Active Record column types. If your Active Record model extends
-      # a module with name `StrongTypeGeneration`, this generator will generate stricter signatures that follow
-      # closely with the types defined in the schema.
-      #
-      # The `StrongTypeGeneration` module you define in your application should add an `after_initialize` callback
-      # to the model and ensure that all the non-nilable attributes of the model are actually initialized with non-`nil`
-      # values.
-      #
       # For example, with the following model class:
-      #
       # ~~~rb
       # class Post < ActiveRecord::Base
       # end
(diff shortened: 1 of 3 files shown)


KaanOzkan (approved):
Do we not want to mention defining and using `StrongTypeGeneration` in the generator docs?

```
This generator, by default, generates *strong* signatures for column methods and ...
```

paracycle:
> Do we not want to mention defining and using `StrongTypeGeneration` in the generator docs?
> 
> ```
> This generator, by default, generates *strong* signatures for column methods and ...
> ```

Not really, since the default behaviour of the generator is to generate what users expect anyway. We will probably, optionally, export an implementation of `StrongTypeGeneration` from Tapioca so that people can use it to opt-in to strong type generation on a per-model basis.

Talking about something that does not change behaviour if absent and only exists in our codebase right now has no value for end users.

lib/tapioca/compilers/dsl/active_record_columns.rb
```diff
@@ -90,7 +78,7 @@ module Dsl
       #     sig { params(value: ::String).returns(::String) }
       #     def title=(value); end
       #
-      #     sig { params(args: T.untyped).returns(T::Boolean) }
+      #     sig { returns(T::Boolean) }
```
Morriar, line 81: Late to the battle but is this one correct?

Should it be `title?; end`?

manual/generator_activerecordcolumns.md
```diff
@@ -78,7 +66,7 @@ class Post
     sig { params(value: ::String).returns(::String) }
     def title=(value); end
 
-    sig { params(args: T.untyped).returns(T::Boolean) }
+    sig { returns(T::Boolean) }
```
Morriar, line 69: Ditto

Morriar:
^ @paracycle

lib/tapioca/compilers/dsl/active_record_columns.rb
paracycle: Thank you. Fixed in https://github.com/Shopify/tapioca/commit/16bb1be8947e60ca08b8eba6b929e695e18cd7f8

manual/generator_activerecordcolumns.md
paracycle: Same as above.

PR #185, opened 2020-12-10
Add `Money` column support for `ActiveRecordColumns` generator

paracycle:
### Motivation

The current `ActiveRecordColumns` generator matches on known AR Value type classes to map to basic types and then tries to match unknown types but looking up the signature of various `ActiveModel::Type` methods defined by the custom type.

However, the `shopify-money` gem does not have any signatures for those methods, so we misgenerate the column type as `::BigDecimal` instead of `::Money`.

### Implementation

- Added a development dependency on `shopify-money` gem
- Added an optional match for `MoneyColumn::ActiveRecordType` in `type_for` lookup

### Tests

- Expanded the different column type check test to test for `money_column` as well. 
- Added a negative test to ensure that we fallback gracefully if `MoneyColumn` constant is not defined.


KaanOzkan (approved):

Morriar (commented):
Nice 🎉

PR #186, opened 2020-12-11
Refactor test helpers

paracycle:
### Motivation

The motivation is to separate test file creation operations from the test mechanism so that
test helpers can depend on less context.

### Implementation

In the new setup, tests are responsible for declaring Ruby files they want as part of their setup using `add_ruby_file` which creates and requires the file by default. To add non-Ruby files, tests can call `add_content_file` which creates the file under the given path in the temp directory.

All test operations operate on files in the temp directory, by default, and the content helper removes the temp directory during teardown.

This allows for a more declarative test style and we don't have to keep passing around hashes of loosely structured data into test helper methods.

### Tests

Existing tests are refactored.

### DIFF
--- a/spec/tapioca/compilers/dsl/active_record_columns_spec.rb
+++ b/spec/tapioca/compilers/dsl/active_record_columns_spec.rb
@@ -7,11 +7,11 @@ class Tapioca::Compilers::Dsl::ActiveRecordColumnsSpec < DslSpec
   describe("Tapioca::Compilers::Dsl::ActiveRecordColumns") do
     describe("#initialize") do
       it("gathers no constants if there are no ActiveRecord subclasses") do
-        assert_empty(constants_from(""))
+        assert_empty(gathered_constants)
       end
 
       it("gathers only ActiveRecord subclasses") do
-        content = <<~RUBY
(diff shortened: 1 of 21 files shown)


spec/tapioca/compilers/dsl/active_record_columns_spec.rb
```diff
@@ -292,31 +286,29 @@ def money_column=(value); end
...
+            end
+
+            # Make `MoneyColumn` disappear artifically
+            Object.send(:remove_const, :MoneyColumn)
+          RUBY
 
-          output = rbi_for(:Post, files)
+          output = rbi_for(:Post)
 
           expected = indented(<<~RUBY, 4)
```
KaanOzkan, line 313:
```suggestion
          expected = indented(<<~RBI, 4)
```
For consistency

KaanOzkan (approved):
Thank you for this 😊

Morriar (approved):
Nice 👍

spec/tapioca/compilers/dsl/active_record_columns_spec.rb
paracycle: Good catch! I found one more instance of it.

PR #190, opened 2020-12-16
Move test helper modules into lib

paracycle:
### Motivation

This allows consumers of the library to use the same test helpers we use internally when they want to write tests for their generator classes.

### Implementation

This is based largely on the previous work introducing better content helper and an isolation helper. This PR moves those modules to the `lib` folder and provides a `Tapioca::TestHelper` module with a `Tapioca::TestHelper::Dsl` submodule so that consumer code can also `require "tapioca/test_helper` to get started with testing their own DSL generators.

The way this is supposed to work is as follows:

```ruby
require "tapioca/test_helper"

class MyDslGeneratorTest < Minitest::Test
  include Tapioca::TestHelper::Dsl

  def setup
    super
    self.generator_class = MyDslGenerator
  end

  def test_our_generator_finds_the_right_constants
    add_ruby_file("foo.rb", <<~RUBY)
      # Some Ruby code
    RUBY

    assert_equal([:Post, :Comment], gathered_constants)
  end

  def test_our_generator_generates_proper_rbi_files
    add_ruby_file("foo.rb", <<~RUBY)
      # Some Ruby code
    RUBY

    expected = <<~RBI
      # typed: strong
      # Some RBI output
    RBI

    assert_equal(expected, rbi_for("Post"))
  end
end
```

The included test helper provides the `gathered_constants` and `rbi_for` methods for assertions. It also provides the ability to define fixture files for processing and supplies the isolation for each test case to run independently so that they don't keep redefining the same constants.

### Tests

No new tests.


elsom25:
🙏 🙇

spec/tapioca/compilers/dsl/active_record_typed_store_spec.rb
```diff
@@ -4,386 +4,388 @@
...
-    end
+        assert_equal(rbi_for(:Post), expected)
+      end
 
-    it("generates methods with Time type for attributes with time type") do
-      add_ruby_file("post.rb", <<~RUBY)
-        class Post < ActiveRecord::Base
-          typed_store :metadata do |s|
-            s.time(:review_time)
+      it("generates methods with DteTime type for attributes with datetime type") do
```
KaanOzkan, line 258: Looks like pre existing typo here with `DteTime`

KaanOzkan (approved):
I wish github/git had better diff support for indentation 😭

spec/dsl_spec.rb
```diff
@@ -3,82 +3,44 @@
...
+      # Set the class under test (generator class) as the class instance
+      T.unsafe(self).generator_class = Object.const_get(generator_class_name)
+    end
+  end
+
+  class << self
+    extend T::Sig
+
+    sig { params(camel_cased_word: String).returns(String) }
+    private def underscore(camel_cased_word)
```
Morriar, line 36:
```suggestion
    private def snake_case_word(camel_cased_word)
```

Morriar (approved):
Nice cleaning!

spec/dsl_spec.rb
paracycle: This method was directly copied from `String#underscore` in Active Support: http://api.rubyonrails.org/classes/String.html#method-i-underscore So, I kept the name the same.

PR #192, opened 2020-12-17
Stop assuming that constants will have sensible hashes

### LINKED ISSUE #191: Declaring a class method called "hash" leads to an unhelpful error
We had a class declaration like this:

```
class FooTest
  class << self
    def hash
      @hash ||= {}
    end
  end
end
```

When we ran Tapioca on it, we saw this:

```
  Compiling foo_project, this may take a few seconds... bundler: failed to load command: tapioca (2.6.6/bin/tapioca)
TypeError: no implicit conversion of Hash into Integer
  ruby/2.6.0/set.rb:254:in `include?'
```

We found it comes from here: https://github.com/Shopify/tapioca/blob/3937c0a658e000f6881bec9827e219992af13f1f/lib/tapioca/compilers/symbol_table/symbol_generator.rb#L311

It is true that `Set.new.include?(FooTest)` blows up like that.

I was going to create a PR, but thought I'd check how you'd like to handle it first. Thanks!

paracycle:
### Motivation

Fixes #191

If a constant overrides the `hash` method at the constant level in an incompatible way by returning a value that is not an `Integer`, then `Set` stops working for that constant.

However, we were using a `Set` for filtering out ignorable ancestors when discovering mixins, which was failing with a cryptic error message if it ever came across a constant that did such a thing.

### Implementation

~The solution is to use array difference rather than sets and enumeration, which actually makes the code simpler in the process.~

The new implementation uses a `Set` of `object_id`s to compare the inherited ancestors against the ancestors of the `constant` or the singleton class of the `constant` to find interesting ancestors. This allows us to not depend on `==`, `eql?`, `equal?` or `hash` for comparing ancestors.

### Tests

Added an expanded version of the reported failure case as a test and a test that makes sure that overridden `==`, `eql?` and `equal?` methods don't affect us.

### DIFF
--- a/lib/tapioca/compilers/symbol_table/symbol_generator.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_generator.rb
@@ -294,19 +294,8 @@ module Tapioca
         def compile_mixins(constant)
           singleton_class = singleton_class_of(constant)
 
-          inherited_ancestors_ids = Set.new(
-            inherited_ancestors_of(constant).map { |mod| object_id_of(mod) }
-          )
-          interesting_ancestors = constant.ancestors.reject do |mod|
-            inherited_ancestors_ids.include?(object_id_of(mod))
-          end
-
-          inherited_singleton_class_ancestors_ids = Set.new(
-            inherited_ancestors_of(singleton_class).map { |mod| object_id_of(mod) }
-          )
-          interesting_singleton_class_ancestors = singleton_class.ancestors.reject do |mod|
-            inherited_singleton_class_ancestors_ids.include?(object_id_of(mod))
-          end
+          interesting_ancestors = interesting_ancestors_of(constant)
+          interesting_singleton_class_ancestors = interesting_ancestors_of(singleton_class)
 
           prepend = interesting_ancestors.take_while { |c| !are_equal?(constant, c) }
           include = interesting_ancestors.drop(prepend.size + 1)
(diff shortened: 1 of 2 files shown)


lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -307,8 +306,7 @@ def compile_mixins(constant)
               Module.ancestors
             end
 
-          interesting_ancestors =
-            constant.ancestors.reject { |mod| ignorable_ancestors.include?(mod) }
+          interesting_ancestors = constant.ancestors - ignorable_ancestors
```
Morriar, line 309: Isn't `Array#-` also using `hash`?

From the docs:

```
Array Difference

Returns a new array that is a copy of the original array, removing any items that also appear in other_ary. The order is preserved from the original array.

It compares elements using their hash and eql? methods for efficiency.
```

paracycle: It turns out that 2.4 uses `hash` but anything above uses `eql?` I guess. I need to find a different solution if we want to keep supporting 2.4.

paracycle: Ok, I changed the implementation. Can you review again please?

KaanOzkan (approved):
We could also check for `Integer` and use the current `Set` solution in that case. Avoid this unlikely "edge case" with an extra O(n) check?

Also nitpick: moving `inherited_singleton_class_ancestors` assignment after `interesting_ancestors` assignment would make it easier to read, at least for me.

paracycle:
> We could also check for `Integer` and use the current `Set` solution in that case. Avoid this unlikely "edge case" with an extra O(n) check?

However, we would need to do that check for all of the ancestors, since any of them could have this weird behaviour. I actually have another solution in mind, maybe I'll try that instead.

> Also nitpick: moving `inherited_singleton_class_ancestors` assignment after `interesting_ancestors` assignment would make it easier to read, at least for me.

Yeah, good call. Actually I noticed that we use a `Set` in that code path as well. So I should rework that too.

paracycle:
@KaanOzkan I streamlined the implementation again so that it uses a `Set` of `object_id`s (based on a suggestion by @Morriar)

KaanOzkan (approved):
I wouldn't be against adding some comments to this functionality. Such as information about what "interesting" means. It took me a quite a while to understand the `compile_mixins` method and that the change is functionally equivalent but maybe that's just me. 

So many `ancestors` 😄

paracycle:
> I wouldn't be against adding some comments to this functionality. 

Fair, we should do that.

> Such as information about what "interesting" means. 

Actually, it is a poor name, it should be something that's more like `non_inherited_ancestors`. We are basically removing all the ancestors that already exist on the `superclass` (and its singleton class), so that we are only left with the ones that are relevant to this constant.
 
> It took me a quite a while to understand the `compile_mixins` method and that the change is functionally equivalent but maybe that's just me.

This is also fair. There are some simplifications we can do around it.

> So many `ancestors` 😄

😂

PR #193, opened 2020-12-19
Add DSL compiler for Sidekiq::Worker

jeffcarbs:
### Motivation

[Sidekiq](https://github.com/mperham/sidekiq) is a gem for background processing. You use it by defining "worker" classes that `include Sidekiq::Worker` and define a `perform` method:
```ruby
class NotifierWorker
  include Sidekiq::Worker
  def perform(customer_id)
    # ...
  end
end
```

You never actually call the `perform` instance method directly, but instead you call one of three class methods:
- `def self.perform_async(*args)`
- `def self.perform_at(datetime_or_time, *args)`
- `def self.perform_in(seconds, *args)`

This compiler would generate:
```ruby
# notifier_worker.rbi
# typed: true
class NotifierWorker
  sig { params(customer_id: T.untyped).returns(String) }
  def self.perform_async(customer_id); end

  sig { params(interval: T.any(DateTime, Time), customer_id: T.untyped).returns(String) }
  def self.perform_at(interval, customer_id); end

  sig { params(interval: Numeric, customer_id: T.untyped).returns(String) }
  def self.perform_in(interval, customer_id); end
end
```

This is similar to ActionMailer where you define instance methods but only ever interact with it by calling class methods.

### Implementation

Based off of the compiler for `ActionMailer`. Commits are broken down logically so commit-by-commit review might be best.

### Tests

Added tests 🙂


lib/tapioca/compilers/dsl/sidekiq_worker.rb
```diff
@@ -0,0 +1,83 @@
...
+
+            # `perform_at` and is just an alias for `perform_in` so both methods technically
+            # accept a date, time, or numeric but we're typing them differently so they
+            # semantically make sense.
+            perform_at_params = [
+              Parlour::RbiGenerator::Parameter.new('interval', type: 'T.any(DateTime, Time)'),
+              *perform_async_params,
+            ]
+            perform_in_params = [
+              Parlour::RbiGenerator::Parameter.new('interval', type: 'Numeric'),
```
jeffcarbs, line 65: These are both named `interval` because it's what they're called in sidekiq source: https://github.com/mperham/sidekiq/blob/3b5ae30c4e5e9e760268243ab5c14664a2f8d236/lib/sidekiq/worker.rb#L167-L179

paracycle (approved):
Thanks @jeffcarbs, this looks great. Apologies for the lack of documentation on how to write a good generator, but I've been putting this off since things are still in flux wrt the final API we want.

Great job discovering how to do it, though! 😍

Issue #177, opened 2020-12-02
[Feature request] Support ORM Sequel


nvh0412:
Hi! tapioca. Thank you for creating an amazing gem, it helps me resolve many problems which I have to deal manually with `sorbet gem`.

We are using ActiveRecord and Sequel gems, both of them inside our codebases. After I tried using Tapioca for Sequel project, just realized that we didn't have any dsl compiler for Sequel, only ActiveRecord at the moment.

Is there any chance that I can request a DSL compiler for Sequel? please let me know if I can contribute to this project too. I'm willing to do it, and again thanks for this great gem.

paracycle:
👋 Hello, thank you so much for using Tapioca. We are super glad that you are having a good time.

We would also love a Sequel DSL compiler for Tapioca but I am afraid we don't use Sequel at Shopify, so none of us have any experience with the gem to be able to build one.

Would you be interested in building one in a PR if we supported you along the way? I think that would be best outcome.

PR #196, opened 2020-12-27
Feat/sequel compilers

nvh0412:
### Motivation
Resolved https://github.com/Shopify/tapioca/issues/177

### Implementation
Based on our current ActiveRecord Implementation, we implement the same way for Sequel Association and Column.


lib/tapioca/generator.rb
```diff
@@ -223,13 +223,15 @@ def say_error(message = "", *color)
 
     sig { params(eager_load: T::Boolean).void }
     def load_application(eager_load:)
-      say("Loading Rails application... ")
+      say("Loading Rails application and Non-Rails application")
```
Morriar, line 226:
```suggestion
      say("Loading application...")
```

lib/tapioca/compilers/dsl/sequel_columns.rb
```diff
@@ -0,0 +1,213 @@
...
+        end
+
+        sig do
+          params(
+            constant: T.class_of(Sequel::Model),
+            column_name: String
+          ).returns([String, String])
+        end
+        def type_for(constant, column_name)
+          return ["T.untyped", "T.untyped"] if do_not_generate_strong_types?(constant)
```
Morriar, line 169: Do you also need the `StrongTypeGeneration` pattern for sequel or is it to mirror what we do in the Active Record generators?

We might want the reverse choice, something like `UnsafeTypeGeneration` as to encourage safe typing by default.

Morriar (changes requested):
I'm not familiar with sequel but the changes look good 👍 

Can you add some tests for your compilers, like it's done for [Active Record](https://github.com/Shopify/tapioca/blob/master/spec/tapioca/compilers/dsl/active_record_columns_spec.rb) please?

Thanks for you pull-request!

rafaelfranca:
I don't think we should maintain Sequel compilers in tapioca. Our compilers system make it possible to people to write their own compilers, so I think a compiler to Sequel should be in its own gem, maintained by people that uses Sequel.

nvh0412:
It makes sense, @rafaelfranca. I realized that Sequel compilers should be maintained by people that use Sequel instead of putting it into this Gem when I was implementing this stuff. Let me close this PR. Thank you, @Morriar too

Issue #201, opened 2021-01-11
Tapioca rendering incomplete RBI for httpclient gem


RyanBrushett:
When running tapioca sync or generate on a project that includes the httpclient gem, the resulting RBI file has a syntax error:

```
...

WebAgent::CookieManager = 
```

I'll include the entire generated RBI in an expandable block at the end.

I've tracked  the problem down to this call to [resolve constant](https://github.com/Shopify/tapioca/blob/master/lib/tapioca/compilers/symbol_table/symbol_generator.rb#L795) where it seems that we're failing to find `HTTPClient::CookieManager`. Thus, I think the resulting RBI should look like this:

```
WebAgent::CookieManager = HTTPClient::CookieManager
```

<details>
  <summary>Full RBI</summary>

```
# DO NOT EDIT MANUALLY
# This is an autogenerated file for types exported from the `httpclient` gem.
# Please instead update this file by running `tapioca sync`.

# typed: true

module HTTP
  extend(::HTTP::Chainable)

  class << self
    def [](headers); end
  end
end

HTTP::CHARSET_RE = T.let(T.unsafe(nil), Regexp)

class HTTP::ContentType < ::Struct
  def charset; end
  def charset=(_); end
  def mime_type; end
  def mime_type=(_); end

  class << self
    def [](*_arg0); end
    def inspect; end
    def members; end
    def new(*_arg0); end
    def parse(str); end

    private

    def charset(str); end
    def mime_type(str); end
  end
end

HTTP::MIME_TYPE_RE = T.let(T.unsafe(nil), Regexp)

class HTTP::Message
  def initialize; end

  def body; end
  def body=(body); end
  def body_encoding; end
  def code; end
  def content; end
  def content_type; end
  def content_type=(content_type); end
  def contenttype; end
  def contenttype=(content_type); end
  def cookies; end
  def dump(dev = T.unsafe(nil)); end
  def header; end
  def headers; end
  def http_body; end
  def http_body=(body); end
  def http_header; end
  def http_header=(_arg0); end
  def http_version; end
  def http_version=(http_version); end
  def ok?; end
  def peer_cert; end
  def peer_cert=(_arg0); end
  def previous; end
  def previous=(_arg0); end
  def reason; end
  def reason=(reason); end
  def redirect?; end
  def see_other?; end
  def status; end
  def status=(status); end
  def status_code; end
  def version; end
  def version=(version); end

  class << self
    def create_query_part_str(query); end
    def escape(str); end
[...]

PR #202, opened 2021-01-12
Fix RBI compile for constant aliases

RyanBrushett:
### Motivation

Tapioca is compiling RBI files that contain syntax errors in some cases where a class ends up overwriting its own superclass by aliasing that superclass to itself.

### Implementation

We can fix this by having `resolve_constant` take an optional `inherit` parameter which defaults to `false`. In the codepath where this comes up, we're comparing the name of the constant we see with the name of the constant returned by calling `Object.const_get` with that name. It seems like sort of a sanity check. Like, checking that this constant does in fact exist and we can find it, I guess? So we don't mind passing `inherit: true` to const_get here _I don't think_. Correct me if I'm wrong.

@paracycle was wondering if we even needed this `are_equal?` check in this case at all, and I think he was going to look into that.

### Tests

If you take away the optional `inherit` flag OR set it to `false` in the `name_of` method, the test will fail as the RBI generated will look as follows:

```ruby
        HTTPClient = WebMockHTTPClient

        class WebAgent
        end

        WebAgent::CookieManager =  # This is the problem

        class WebMockHTTPClient
        end
```

Note that I think this bug was hidden for a while simply because having that empty assignment doesn't make a typecheck fail unless it's the last line of the RBI file! So I got lucky to find it. I think this actually affects a lot more than just httpclient but that's where I noticed it.

Also: Thanks @paracycle and @KaanOzkan for helping me figure this out 💖

Resolves https://github.com/Shopify/tapioca/issues/201

### DIFF
--- a/lib/tapioca/compilers/symbol_table/symbol_generator.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_generator.rb
@@ -74,9 +74,9 @@ module Tapioca
           compile(symbol, constant)
         end
 
-        sig { params(symbol: String).returns(BasicObject).checked(:never) }
-        def resolve_constant(symbol)
-          Object.const_get(symbol, false)
+        sig { params(symbol: String, inherit: T::Boolean).returns(BasicObject).checked(:never) }
+        def resolve_constant(symbol, inherit: false)
+          Object.const_get(symbol, inherit)
         rescue NameError, LoadError, RuntimeError, ArgumentError, TypeError
           nil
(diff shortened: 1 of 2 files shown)


paracycle (approved):
Looks great!

KaanOzkan (approved):
:shipit: 

Thanks for the detailed description. Feel free to ping `@Shopify/sorbet` in the future for visibility from the whole team 😊

PR #203, opened 2021-01-13
Fix Ruby 3.0 compatibility

casperisfine:
The good old name argument change. `options` has to be passed as kw args.


paracycle (commented):
You beat me to it. However, it is a problem for us to keep testing Ruby 2.4 while still being Ruby 3.0 compliant due to the Rails version upgrade. 

I think we need to split the Gemfile into two parts with Rails 5.2 on 2.4 and Rails 6.1 on 2.5+. I can work on this PR to do that, if you want.

casperisfine:
> it is a problem for us to keep testing Ruby 2.4 while still being Ruby 3.0 compliant due to the Rails version upgrade.

Yes. But at this point shouldn't we just drop 2.4? It's quite old already.

casperisfine:
> I can continue on this PR to do all that and resolve the test errors.

🆗

paracycle:
This was superseded by #352 and #358

PR #204, opened 2021-01-13
Fix Ruby 3.0 compat

casperisfine:
This is the fix from https://github.com/Shopify/tapioca/pull/203, but without all the changes to try to make the suite pass on 3.0.

THere is a pile of dependency issues to run CI on 3.0:

  - Bundler is too old
  - Rails etc is too old
  - google-protobuf won't install on 3.0 right now
  - probably other stuffs I haven't figured yet.

So I'm thinking in the meantime we can simply merge the fix without testing it on CI.


paracycle (approved):
Yeah, this is much better. Thank you!

PR #205, opened 2021-01-13
Fix bug in compile_alias when constant names are mismatched

RyanBrushett:
### Motivation

Fixing an RBI typo I found. When compiling, there is a chance the constant name that we want to use for an alias doesn't match because that name has itself been overwritten with something.

```ruby
class MyClient
end

module MyModule
  OriginalClient = ::MyClient
end

class MockClient < MyClient
end

MyClient = MockClient
```

In this case, we do a `name_of` on `OriginalClient` which is pointing to `MyClient` but `resolve_constant` results in a `MockClient` because of the last line there. So `are_equal?(constant, resolve_constant(name, inherit: true))` is false and we end up with an empty assignment like `MyModule::OriginalClient =`

### Implementation

~Remove the equality check... It seems to work just fine when running it on my project also.~

~**I don't actually know if this is OK**. So I defer to y'all. @paracycle what was the reason for this equality check in the first place?~

Removing the equality check was actually not OK.
Instead, if `name` turns up blank, we'll set it to an anonymous class or module with `"#{constant.class}.new"`

### Tests

I did add tests! ~If you add the equality check back in that I removed, you'll see the test fail and OriginalClient won't be assigned to anything at all.~
If you remove the lines I've added to symbol_generator.rb, you'll see the test start to fail 👍

### DIFF
--- a/lib/tapioca/compilers/symbol_table/symbol_generator.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_generator.rb
@@ -122,8 +122,9 @@ module Tapioca
         def compile_alias(name, constant)
           return if symbol_ignored?(name)
 
-          constant_name = name_of(constant)
-          target = constant_name.nil? ? "#{constant.class}.new" : constant_name
+          target = name_of(constant)
+          # If target has no name, let's make it an anonymous class or module with `Class.new` or `Module.new`
+          target = "#{constant.class}.new" unless target
 
           add_to_alias_namespace(name)
 
--- a/lib/tapioca/compilers/symbol_table/symbol_generator.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_generator.rb
@@ -792,7 +792,6 @@ module Tapioca
           return name if name
           name = raw_name_of(constant)
           return if name.nil?
-          return unless are_equal?(constant, resolve_constant(name, inherit: true))
           name = "Struct" if name =~ /^(::)?Struct::[^:]+$/
           name
         end
--- a/lib/tapioca/compilers/symbol_table/symbol_generator.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_generator.rb
@@ -123,11 +123,13 @@ module Tapioca
           return if symbol_ignored?(name)
 
           constant_name = name_of(constant)
+          target = constant_name.nil? ? "#{constant.class}.new" : constant_name
+
           add_to_alias_namespace(name)
 
           return if IGNORED_SYMBOLS.include?(name)
 
-          indented("#{name} = #{constant_name}")
+          indented("#{name} = #{target}")
         end
 
         sig do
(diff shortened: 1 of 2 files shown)


spec/tapioca/compilers/symbol_table_compiler_spec.rb
```diff
@@ -531,6 +531,37 @@ class WebMockHTTPClient
...
+      output = template(<<~RBI)
+        class MockClient
+        end
+
+        MyClient = MockClient
+
+        module MyModule
+        end
+
+        MyModule::OriginalClient = MyClient
```
paracycle, line 559: I don't think this is what we want to generate, though. Let us call the first `MyClass` class `MyClass (1)` and the second one `MyClass (2)` to disambiguate them. At runtime after the `bar.rb` file is loaded `MyModule::OriginalClient` would be aliasing `MyClass (1)` and `MyClass (2)` would really be an alias to `MockClient`.

But the RBI we generate is saying that `MyModule::OriginalClient` is an alias to `MyClient (2)` which is an alias to `MockClient`. This does not match the runtime behaviour.

I think that's why we had the `are_equal?` check in the first place.

For this case, I'd prefer to not generate `MyModule::OriginalClient` at all, or to generate it as `MyModule::OriginalClient = Class.new` or `MyModule::OriginalClient = Module.new`, whichever is appropriate.

RyanBrushett: Ah interesting OK. I'll go back to the drawing board 👍

RyanBrushett:
When you get a chance, let me know if this is a bit closer to what you had in mind @paracycle.

spec/tapioca/compilers/symbol_table_compiler_spec.rb
```diff
@@ -531,6 +531,52 @@ class WebMockHTTPClient
       assert_equal(output, compile)
     end
 
+    it("compiles a constant which is aliased to a constant that has been overwritten as a placeholder") do
+      add_ruby_file("bar.rb", <<~RUBY)
```
paracycle, line 535: Super nitpicky, but can we give this "file" a better name? Maybe:
```suggestion
      add_ruby_file("overwritten_class_module_references.rb", <<~RUBY)
```
or something

lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -123,11 +123,13 @@ def compile_alias(name, constant)
           return if symbol_ignored?(name)
 
           constant_name = name_of(constant)
+          target = constant_name.nil? ? "#{constant.class}.new" : constant_name
```
paracycle, line 126:
```suggestion
          target = name_of(constant)
          # If the target has no name, let's make it an anonymous 
          # class or module with `Class.new` or `Module.new`
          target = "#{constant.class}.new" unless target
```
maybe?

paracycle (approved):
This looks much much better. Let's test this on the project you were getting errors in before merging. I want to make sure that we are not introducing more problems over and beyond this test case (even though other tests seem to be passing fine).

RyanBrushett:
Thanks for the feedback!
I tested with my project and it looks good. Generated RBI free of any dangling assignments with no regressions/changes in the RBI files that had already generated correctly 🎉

paracycle (approved):
👌

PR #210, opened 2021-01-22
Use the collection proxies

kddeisz:
### Motivation

This ensures that the associations generated will receive the generated scopes as well.

### Implementation

Putting a `ActiveRecord_Associations_CollectionProxy` class onto the generated classes. I wasn't sure if you wanted this to be a different DSL, which I could see a case for as well. Just let me know.

### DIFF
--- a/lib/tapioca/compilers/dsl/active_record_associations.rb
+++ b/lib/tapioca/compilers/dsl/active_record_associations.rb
@@ -70,7 +70,7 @@ module Tapioca
       #     sig { params(ids: T::Array[T.untyped]).returns(T::Array[T.untyped]) }
       #     def comment_ids=(ids); end
       #
-      #     sig { returns(::Comment::ActiveRecord_Associations_CollectionProxy) }
+      #     sig { returns(::Comment::PrivateCollectionProxy) }
       #     def comments; end
       #
       #     sig { params(value: T::Enumerable[::Comment]).void }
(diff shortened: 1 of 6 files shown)


rafaelfranca (commented):
Those classes are private of the framework, so while I think it is fine to put them in rbis, since the visibility of them is not checked, it will fail if people use in their signatures.

The name of those classes are by design ugly because we don't intend people to be using then directly in their applications.

I fear that exposing them in the rbi will tell people that they can use in their signatures and in their code.

Even if we document in the rbis that people should not use them I'm not sure if it will be enough.

kddeisz:
The RBIs don't necessarily need to match the class name of the actual constant since its a private const and we're not using it directly anyway. How about we just name it something different so folks don't get the wrong impression?

kddeisz:
GeneratedCollectionProxy?

paracycle:
@kddeisz Thanks for the PR. Apart from @rafaelfranca's concerns about the naming, we did actually intend this to be a separate generator that was just responsible for generating stuff related to AR relations, which include `Model::ActiveRecord_Relation`, `Model::ActiveRecord_AssociationRelation` as well as `Model::ActiveRecord_Associations_CollectionProxy`. All of those need to be covered along with the differences in the types of parameters they expect and the results they return.

At the same time, this is actually a good idea that we can maybe start exposing that interface with less work by just exposing the `GeneratedRelationMethods` from them and also implementing some of the methods that we are already exposing from the `ActiveRecord::Associations::CollectionProxy` generic hack that we currently have in Core that this code still depends on.

kddeisz:
Closing in favor of https://github.com/Shopify/tapioca/tree/uk-add-ar-relations

PR #214, opened 2021-01-25
Add support for AR nested attributes for associations

paracycle:
### Motivation

Add support for [Active Record Nested Attributes](https://api.rubyonrails.org/v6.1.0/classes/ActiveRecord/NestedAttributes/ClassMethods.html)

### Implementation

Each nested attribute definition [adds an entry in the `nested_attributes_options` attribute of the AR model class](https://github.com/rails/rails/blob/914caca2d31bd753f47f9168f2a375921d9e91cc/activerecord/lib/active_record/nested_attributes.rb#L343-L345). Thus, enumerating the keys of the `nested_attributes_options` hash gives us all the attribute values for which `accepts_nested_attributes_for` was called. For each of these association names, the name of the generated setter method can [easily be generated](https://github.com/rails/rails/blob/914caca2d31bd753f47f9168f2a375921d9e91cc/activerecord/lib/active_record/nested_attributes.rb#L370).

### Tests

Expanded current tests to also include `accepts_nested_attributes_for` calls for the associations.

### DIFF
--- a/lib/tapioca/compilers/dsl/active_record_associations.rb
+++ b/lib/tapioca/compilers/dsl/active_record_associations.rb
@@ -103,13 +103,8 @@ module Tapioca
             module_name = "GeneratedAssociationMethods"
 
             model.create_module(module_name) do |mod|
-              constant.reflections.each do |association_name, reflection|
-                if reflection.collection?
-                  populate_collection_assoc_getter_setter(mod, constant, association_name, reflection)
-                else
-                  populate_single_assoc_getter_setter(mod, constant, association_name, reflection)
-                end
-              end
+              populate_nested_attribute_writers(mod, constant)
+              populate_associations(mod, constant)
             end
 
             model.create_include(module_name)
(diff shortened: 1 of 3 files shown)


lib/tapioca/compilers/dsl/active_record_associations.rb
```diff
@@ -44,6 +46,9 @@ module Dsl
       #     sig { params(value: T.nilable(::User)).void }
       #     def author=(value); end
       #
+      #     sig { params(attributes: T.untyped).returns(T.untyped) }
+      #     def author_attributes=(attributes); end
```
Morriar, line 50: The doc says:

> The attribute writer is named after the association, which means that in the following example, two new methods are added to your model: `author_attributes=(attributes)` and `pages_attributes=(attributes)`

Shouldn't we see two methods being added to the model? I only see `=` methods in this example.

Morriar (approved):

lib/tapioca/compilers/dsl/active_record_associations.rb
paracycle: `accepts_nested_attributes_for :author, :pages` adds `author_attributes=` and `pages_attributes=` (note that two names are passed to `accepts_nested_attributes_for`: `:author` and `:pages`). One writer method for each association that has a nested attribute.

In this example case, we have 3 writers added since we called the method with associations.

Morriar: Urgh I don't know how to read anymore...

PR #221, opened 2021-02-08
Support for modules with multiple helpers

ttonelli:
### Motivation

The code to generate module helpers currently assumes that a module can have a single helper. For example, it's either an `interface!` or `abstract!` or `sealed!`, etc. But Sorbet allows a module to have multiple. For example, algebraic data types often use `interface!` and `sealed!` together.

### Implementation

We modify the `compile_module_helpers` method to assemble all possible helpers instead of picking the first that matches.

### Tests

I updated the test for signatures and structs to check for an algebraic data type example.

### DIFF
--- a/spec/tapioca/compilers/symbol_table_compiler_spec.rb
+++ b/spec/tapioca/compilers/symbol_table_compiler_spec.rb
@@ -2094,6 +2094,7 @@ class Tapioca::Compilers::SymbolTableCompilerSpec < Minitest::HooksSpec
         module Adt
           extend(T::Sig)
           extend(T::Helpers)
+
           interface!
           sealed!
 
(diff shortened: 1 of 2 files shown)


paracycle (approved):
Great catch. Thank you!

spec/tapioca/compilers/symbol_table_compiler_spec.rb
```diff
@@ -2090,7 +2090,32 @@ def something
         end
       RUBY
 
+      add_ruby_file("adt.rb", <<~RUBY)
+        module Adt
+          extend(T::Sig)
+          extend(T::Helpers)
+          interface!
```
Morriar, line 2097: Can we add a new line before so the helpers stand out?

Morriar (commented):
It's almost incredible that things were working 😂

spec/tapioca/compilers/symbol_table_compiler_spec.rb
ttonelli: Yep, pushed the fix.

Morriar (approved):
Thanks for indulging me! ❤️

PR #222, opened 2021-02-08
Adjust void return for props

ttonelli:
### Motivation

A `prop` can be a `Proc`; if the return type of the `Proc` is `void`, the current code generates `returns(<VOID>)`, which is incorrect.

### Implementation

I reused the approach used elsewhere in the same file, which simply `gsub`s the string from `.returns(<VOID>)` to `.void`.

### Tests

Included a `prop` with the issue in the test cases.


paracycle (approved):
👌

Morriar (approved):

PR #223, opened 2021-02-09
A small hack to support generics

ttonelli:
### Motivation

Tapioca currently does not generate type arguments to generic types, which breaks sorbet type checking.

### Implementation

Sorbet runtime does type erasure, so the type argument information is not available from their APIs. So I monkey patched `T::Generic`'s `[]` method to intercept references to generic types. Then I created a wrapper called `GenericReference` that encapsulates the generic type (under a `T::Types::Simple` as usual) together with the type arguments. This is then used to generate signatures that contain the arguments.

In order for this to work properly, however, tapioca needs to also generate sorbet signatures like `T::Generic`. These were explicitly removed from generation and I don't have context why. In this PR I brought them back, but that is debatable (maybe adding an allowlist or only allowing `T::Generic` is enough?).

### Tests

I haven't added tests yet (in fact I broke a bunch of tests). I wanted to discuss the approach before adding and fixing tests.

### DIFF
--- a/lib/tapioca/compilers/symbol_table/symbol_generator.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_generator.rb
@@ -3,6 +3,41 @@
 
 require 'pathname'
 
+module T
+  module Private
+    module Types
+      #
+      # A reference to a generic type, keeping track of the type parameters for generation.
+      #
+      class GenericReference < T::Types::Base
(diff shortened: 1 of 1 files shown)


ttonelli:
Closing this experiment as there's now a proper implementation in https://github.com/Shopify/tapioca/pull/156

PR #224, opened 2021-02-11
Fixing dynamic mixin includes

ttonelli:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,6 +13,7 @@ end
 gem("yard", "~> 0.9.25")
 gem("pry-byebug")
 gem("minitest")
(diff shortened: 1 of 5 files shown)


paracycle (commented):
Thiago, this is a good idea and one I've been meaning to tackle soon, but it needs a Sorbet version check before doing this, since we try to support much older versions of Sorbet as well.

Basically, the ability to declare multiple `mixes_in_class_modules` or giving it multiple arguments is a feature that we added to Sorbet at a particular version and it would break type-checking for any version of Sorbet before that.

paracycle:
Superseded by #363

PR #226, opened 2021-02-14
Fix bad link in generators.md

connorshea:
Needs an underscore in the filename for sidekiq worker docs.


Morriar:
I think the table of content is generated.

It looks like a problem with how we create the filename: https://github.com/Shopify/tapioca/blob/aad17fa218b84aa762a94821d52175d96b7c6a5b/tasks/generator_documentation.rake#L93

paracycle (approved):
Thanks for this PR. The actual fix was to add the correctly named file to the repo, instead of the incorrectly named one. We should have caught this on CI when doing synchronization checks for docs.

I've taken the liberty to make the correct change on this PR and I have also implemented a better check for the synchronization assertion that should take untracked files into consideration.

Morriar (approved):
Good catch! Thanks to both of you ❤️

PR #230, opened 2021-02-17
Change with_clean_env to with_unbundled_env

cjhutchi:
### Motivation

Removes the deprecation warnings for `.with_clean_env`.

### Implementation

Use the recommended change as provided by `Bundler`. Switch from `.with_clean_env` to `.with_unbundled_env`.

### Tests

Only changed within the `cli_spec.rb` file.

### DIFF
--- a/sorbet/rbi/shims/bundler.rbi
+++ b/sorbet/rbi/shims/bundler.rbi
@@ -3785,7 +3785,7 @@ module Bundler
 
   def self.which(executable); end
 
-  def self.with_clean_env(); end
+  def self.with_unbundled_env(); end
 
   def self.with_original_env(); end
 end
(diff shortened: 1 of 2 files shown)


KaanOzkan (approved):

paracycle (commented):
Does that mean we are dropping support for Bundler versions that don't have this? If so, we should bump the lower bound on the `bundler` gemspec dependency.

Morriar (approved):

cjhutchi:
> Does that mean we are dropping support for Bundler versions that don't have this? If so, we should bump the lower bound on the `bundler` gemspec dependency.

@paracycle This only effects a couple of tests from the test suite. Shouldn't have any overall effect on `tapioca` as a whole. And looks like `#with_unbundled_env` has been present since `Bundler 2.1.x` so the CI pipeline should be fine as well since we're testing against `2.4.x+`

paracycle:
> This only effects a couple of tests from the test suite.

Right! Sorry, I missed the fact that this was just a test change.

paracycle (approved):

PR #231, opened 2021-02-17
Add check command

cjhutchi:
### Motivation

We need an easy way to check whether or not the RBI files are out-of-date after updates are made to files, Gems are bumped, etc. This adds a `check` command to `tapioca` that will re-generate the DSL (excluding Gems) to diff against to determine whether or not `tapioca dsl` needs to be run. 

### Implementation

Create a `tmpdir` using [`Dir.mktmpdir`](https://ruby-doc.org/stdlib-2.7.2/libdoc/tmpdir/rdoc/Dir.html), generate the DSL in the temp directory, and diff against that to determine if RBIs are out of sync.

The command returns with an exit status of `0` if no changes are detected, and `1` if changes are detected. This will allow for it to be used in CI pipelines.

### Tests

Adds two additional tests that ensure that the command returns the correct exit code as well as messaging to advise on correct next steps.

### DIFF
--- a/lib/tapioca/cli.rb
+++ b/lib/tapioca/cli.rb
@@ -74,9 +74,17 @@ module Tapioca
       aliases: ["--gen", "-g"],
       banner: "generator [generator ...]",
       desc: "Only run supplied DSL generators"
+    option :verify,
+      type: :boolean,
+      default: false,
+      desc: "Verifies correctness of RBIs"
     def dsl(*constants)
       Tapioca.silence_warnings do
-        generator.build_dsl(constants)
+        if options[:verify]
+          generator.verify_dsl(constants)
+        else
+          generator.build_dsl(constants)
+        end
       end
     end
 
--- a/lib/tapioca/cli.rb
+++ b/lib/tapioca/cli.rb
@@ -81,7 +81,7 @@ module Tapioca
     def dsl(*constants)
       Tapioca.silence_warnings do
         if options[:verify]
-          generator.verify_dsl(constants)
+          generator.build_dsl(constants, should_verify: true)
         else
           generator.build_dsl(constants)
         end
--- a/lib/tapioca/cli.rb
+++ b/lib/tapioca/cli.rb
@@ -77,14 +77,10 @@ module Tapioca
     option :verify,
       type: :boolean,
       default: false,
-      desc: "Verifies correctness of RBIs"
+      desc: "Verifies RBIs are up-to-date"
     def dsl(*constants)
       Tapioca.silence_warnings do
-        if options[:verify]
-          generator.build_dsl(constants, should_verify: true)
-        else
-          generator.build_dsl(constants)
-        end
+        generator.build_dsl(constants, should_verify: options[:verify])
       end
     end
 
--- a/lib/tapioca/config_builder.rb
+++ b/lib/tapioca/config_builder.rb
@@ -44,7 +44,7 @@ module Tapioca
       def default_command
         command = File.basename($PROGRAM_NAME)
         # Hack to avoid flags ending up in the header of the RBIs
-        args = ARGV.reject { |arg| arg.include?("--") }.join(" ")
+        args = ARGV.grep_v(/^-/).join(" ")
 
         "#{command} #{args}".strip
       end
--- a/lib/tapioca/config_builder.rb
+++ b/lib/tapioca/config_builder.rb
@@ -43,7 +43,8 @@ module Tapioca
       sig { returns(String) }
       def default_command
         command = File.basename($PROGRAM_NAME)
-        args = ARGV.join(" ")
+        # Hack to avoid flags ending up in the header of the RBIs
+        args = ARGV.reject { |arg| arg.include?("--") }.join(" ")
 
         "#{command} #{args}".strip
       end
--- a/lib/tapioca/generator.rb
+++ b/lib/tapioca/generator.rb
@@ -148,27 +148,13 @@ module Tapioca
         filename = compile_dsl_rbi(constant, contents, outpath: Pathname.new(outpath))
         rbi_files_to_purge.delete(filename) if filename
       end
-
-      if rbi_files_to_purge.any? && !should_verify
-        say("")
-        say("Removing stale RBI files...")
-
-        rbi_files_to_purge.sort.each do |filename|
-          remove(filename)
-        end
-      end
-
       say("")
+
       if should_verify
-        if (error = verify_dsl_rbi(tmp_dir: Pathname.new(outpath)))
-          say("RBI files are out-of-date, please run `#{config.generate_command}` to update.")
-          say("Reason: ", [:red])
-          say(error)
-          exit(1)
-        else
-          say("Nothing to do, all RBIs are up-to-date.")
-        end
+        perform_dsl_verification(outpath)
       else
+        purge_stale_dsl_rbi_files(rbi_files_to_purge)
+
         say("Done", :green)
 
         say("All operations performed in working directory.", [:green, :bold])
@@ -307,10 +293,10 @@ module Tapioca
       constant_map.values
     end
 
-    sig { params(requested_constants: T::Array[String]).returns(T::Set[Pathname]) }
-    def existing_rbi_filenames(requested_constants)
+    sig { params(requested_constants: T::Array[String], path: Pathname).returns(T::Set[Pathname]) }
+    def existing_rbi_filenames(requested_constants, path: config.outpath)
       filenames = if requested_constants.empty?
-        Pathname.glob(config.outpath / "**/*.rbi")
+        Pathname.glob(path / "**/*.rbi")
       else
         requested_constants.map do |constant_name|
           dsl_rbi_filename(constant_name)
@@ -546,8 +532,8 @@ module Tapioca
 
     sig { params(tmp_dir: Pathname).returns(T.nilable(String)) }
     def verify_dsl_rbi(tmp_dir:)
-      existing_rbis = get_file_list(dir: config.outpath).sort
-      new_rbis = get_file_list(dir: tmp_dir).sort
+      existing_rbis = existing_rbi_filenames([]).sort
+      new_rbis = existing_rbi_filenames([], path: tmp_dir).grep_v(/gem|shim/).sort
 
       return "New file(s) introduced." if existing_rbis.length != new_rbis.length
 
@@ -566,9 +552,30 @@ module Tapioca
       nil
     end
 
-    sig { params(dir: Pathname).returns(T::Array[Pathname]) }
-    def get_file_list(dir:)
-      Dir.glob("#{dir}/**/*.rbi").reject { |e| e =~ /gems|shims/ }.map { |f| Pathname.new(f) }
+    sig { params(dir: String).void }
+    def perform_dsl_verification(dir)
+      if (error = verify_dsl_rbi(tmp_dir: Pathname.new(dir)))
+        say("RBI files are out-of-date, please run `#{config.generate_command}` to update.")
+        say("Reason: ", [:red])
+        say(error)
+        exit(1)
+      else
+        say("Nothing to do, all RBIs are up-to-date.")
+      end
+    ensure
+      FileUtils.remove_entry(dir)
+    end
+
+    sig { params(files: T::Set[Pathname]).void }
+    def purge_stale_dsl_rbi_files(files)
+      if files.any?
+        say("Removing stale RBI files...")
+
+        files.sort.each do |filename|
+          remove(filename)
+        end
+        say("")
+      end
     end
   end
 end
--- a/lib/tapioca/generator.rb
+++ b/lib/tapioca/generator.rb
@@ -149,7 +149,7 @@ module Tapioca
         rbi_files_to_purge.delete(filename) if filename
       end
 
-      if !rbi_files_to_purge.empty? && !should_verify
+      if rbi_files_to_purge.any? && !should_verify
         say("")
         say("Removing stale RBI files...")
 
@@ -568,7 +568,7 @@ module Tapioca
 
     sig { params(dir: Pathname).returns(T::Array[Pathname]) }
     def get_file_list(dir:)
-      Dir.glob("#{dir}/**/*.rbi").reject { |e| e =~ /gems/ }.map { |f| Pathname.new(f) }
+      Dir.glob("#{dir}/**/*.rbi").reject { |e| e =~ /gems|shims/ }.map { |f| Pathname.new(f) }
     end
   end
 end
--- a/lib/tapioca/generator.rb
+++ b/lib/tapioca/generator.rb
@@ -160,7 +160,7 @@ module Tapioca
 
       say("")
       if should_verify
-        if error = verify_dsl_rbi(tmp_dir: Pathname.new(outpath))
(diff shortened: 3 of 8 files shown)


spec/tapioca/cli_spec.rb
```diff
@@ -569,6 +575,107 @@ def body=(body); end
...
+
+        it 'does nothing if no RBIs have changed' do
+          output = execute("dsl", "--verify")
+
+          assert_includes(output, <<~OUTPUT)
+            Nothing to do, all RBIs are up-to-date.
+          OUTPUT
+        end
+
+        it ' returns the correct exit code' do
```
KaanOzkan, line 593: I think these `returns correct exit code` tests be combined with the tests just above them, I'd personally prefer it since it makes the file shorter and tad bit easier to read, but it's your call. No strong opinions.

spec/tapioca/cli_spec.rb
```diff
@@ -4,6 +4,7 @@
 require "spec_helper"
 require "pathname"
 require "shellwords"
+require "pry"
```
KaanOzkan, line 7: debug statement

lib/tapioca/generator.rb
```diff
@@ -522,5 +547,35 @@ def compile_dsl_rbi(constant, contents)
...
+
+      raise(OutOfSyncError.new, "New file(s) introduced.") if current_count != new_count
+
+      desynced_files = []
+
+      (0..current_count - 1).each do |i|
+        desynced_files << test_rbis[i] unless FileUtils.identical?(existing_rbis[i], test_rbis[i])
+      end
+
+      if desynced_files.count > 0
```
KaanOzkan, line 567: I'd prefer `!empty?` or `any?`

lib/tapioca/generator.rb
```diff
@@ -149,10 +161,23 @@ def build_dsl(requested_constants)
       end
 
       say("")
-      say("Done", :green)
+      if should_verify
+        begin
+          verify_dsl_rbi(tmp_dir: Pathname.new(outpath))
+        rescue OutOfSyncError => e
```
KaanOzkan, line 167: I don't love using exceptions for control flow. Also like other languages I'm sure using exceptions in Ruby are slower than its alternatives.

It's only thrown within `verify_dsl_rbi` so I think it can be replaced with something simple such as return value if we don't want to overcomplicate.

lib/tapioca/generator.rb
```diff
@@ -522,5 +547,35 @@ def compile_dsl_rbi(constant, contents)
...
+
+      desynced_files = []
+
+      (0..current_count - 1).each do |i|
+        desynced_files << test_rbis[i] unless FileUtils.identical?(existing_rbis[i], test_rbis[i])
+      end
+
+      if desynced_files.count > 0
+        filenames = desynced_files.map(&:to_s).each do |file|
+          file.sub!(tmp_dir.to_s, "sorbet/rbi/dsl")
```
vinistock, line 569: NIT: instead of mapping to strings and then iterating again, can we use a single map and invoke `to_s` on `file` inside the loop?

lib/tapioca/generator.rb
```diff
@@ -522,5 +547,35 @@ def compile_dsl_rbi(constant, contents)
...
       filename
     end
+
+    sig { params(tmp_dir: Pathname).void }
+    def verify_dsl_rbi(tmp_dir:)
+      existing_rbis = get_file_list(dir: config.outpath).sort
+      test_rbis = get_file_list(dir: tmp_dir).sort
+
+      current_count = existing_rbis.count
+      new_count = test_rbis.count
```
vinistock, line 557: NIT: `length` instead of `count`, since it's faster.

spec/tapioca/cli_spec.rb
cjhutchi: Well.. TIL that `pry` does something with `$?` because when using that initially I was getting deprecation errors advising to use `$CHILD_STATUS`. Then after removing this line I get failing tests... turns out `$CHILD_STATUS` exists on `pry` so had to go back to `$?`...

spec/tapioca/cli_spec.rb
```diff
@@ -569,6 +572,95 @@ def body=(body); end
...
+        end
+
+        it 'advises of new file(s) and returns exit_status 1' do
+          output = execute("dsl", "--verify")
+
+          assert_includes(output, <<~OUTPUT)
+            RBI files are out-of-date, please run `generate command` to update.
+            Reason: New file(s) introduced.
+          OUTPUT
+          assert_includes($?.to_s, "exit 1") # rubocop:disable Style/SpecialGlobalVars
```
cjhutchi, line 618: Want to make a special note about this. I was intending to use `Process.last_status` which wasn't available in Ruby 2.4. So I opted to use this with the `rubocop` exception as it's just for a test to work around the `IO.popen(...)`

lib/tapioca/cli.rb
```diff
@@ -74,9 +74,17 @@ def todo
       aliases: ["--gen", "-g"],
       banner: "generator [generator ...]",
       desc: "Only run supplied DSL generators"
+    option :verify,
+      type: :boolean,
+      default: false,
+      desc: "Verifies correctness of RBIs"
```
Morriar, line 80:
```suggestion
      desc: "Verifies RBIs are up-to-date"
```

lib/tapioca/cli.rb
```diff
@@ -74,9 +74,17 @@ def todo
...
     def dsl(*constants)
       Tapioca.silence_warnings do
-        generator.build_dsl(constants)
+        if options[:verify]
+          generator.build_dsl(constants, should_verify: true)
+        else
+          generator.build_dsl(constants)
+        end
```
Morriar, line 87:
```suggestion
        generator.build_dsl(constants, should_verify: options[:verify])
```

lib/tapioca/generator.rb
```diff
@@ -522,5 +543,32 @@ def compile_dsl_rbi(constant, contents)
 
       filename
     end
+
+    sig { params(tmp_dir: Pathname).returns(T.nilable(String)) }
```
Morriar, line 547: Could be an array of errors for you can return all of them, not just the first one.

lib/tapioca/generator.rb
```diff
@@ -522,5 +543,32 @@ def compile_dsl_rbi(constant, contents)
...
+
+        return "File(s) updated:\n  - #{filenames}"
+      end
+
+      nil
+    end
+
+    sig { params(dir: Pathname).returns(T::Array[Pathname]) }
+    def get_file_list(dir:)
+      Dir.glob("#{dir}/**/*.rbi").reject { |e| e =~ /gems/ }.map { |f| Pathname.new(f) }
```
Morriar, line 571: Aren't you also getting the `shims/` like this?

Why not directly taking the `**/dsl/*.rbi` files?

spec/tapioca/cli_spec.rb
```diff
@@ -569,6 +572,95 @@ def body=(body); end
...
+          execute("dsl")
+        end
+
+        it 'does nothing and returns exit_status 0' do
+          output = execute("dsl", "--verify")
+
+          assert_includes(output, <<~OUTPUT)
+            Nothing to do, all RBIs are up-to-date.
+          OUTPUT
+          assert_includes($?.to_s, "exit 0") # rubocop:disable Style/SpecialGlobalVars
```
Morriar, line 588: I guess making `execute` return the result code would be a lot of changes?

spec/tapioca/cli_spec.rb
```diff
@@ -569,6 +572,95 @@ def body=(body); end
...
+            class Image
+              include(SmartProperties)
+
+              property :title, accepts: String
+            end
+          RUBY
+        end
+
+        after do
+          File.delete(repo_path / "lib" / "image.rb") if File.exist?(repo_path / "lib" / "image.rb")
```
Morriar, line 608:
```suggestion
          FileUtils.rm_f(repo_path / "lib" / "image.rb")
```

lib/tapioca/generator.rb
```diff
@@ -116,16 +116,26 @@ def build_todos
...
 
-    sig { params(requested_constants: T::Array[String]).void }
-    def build_dsl(requested_constants)
+    sig do
+      params(
+        requested_constants: T::Array[String],
+        should_verify: T::Boolean,
+      ).void
+    end
+    def build_dsl(requested_constants, should_verify: false)
```
RyanBrushett, line 125: I'm thinking a bit `out loud` but happy to pair on this or talk more about it:

It feels like this method is doing too much. Like maybe it shouldn't need to know whether or not the user is verifying the files but instead maybe it could take an `outpath`, build things happily and put them wherever it's requested, and some other pieces of code that call it could be concerned with the act of verification (or otherwise), communicating with the user, etc. ya know?

lib/tapioca/generator.rb
cjhutchi: I actually build the error message within the method. So we return `New file(s) added.` or a list of modified files.

lib/tapioca/generator.rb
cjhutchi: Good call, I don't believe the test suite includes any shim tests. But I updated to exclude that directory. Kept it relatively generic because we just dump the temporary RBIs into a tmp dir without the overall `rbi/dsl` structure.

spec/tapioca/cli_spec.rb
cjhutchi: Yeah, would require changing every single test. Previously, I was setting a global variable, but was causing the tests to be flakey. This seems to have resolved the flakiness and didn't require modifying each test.

lib/tapioca/generator.rb
cjhutchi: Yeah, I had those thoughts originally as well. After talking with @paracycle this was the chosen approach, and future refactors will take place to clean things up. Focus is getting the feature out in working order, than future improvements can be made to the code to tidy things up.

lib/tapioca/generator.rb
Morriar: I think this method does too much, it shouldn't be controlling the message.

What do you think about returning a struct like this instead?

```rb
class DslState < T::Struct
  new_files, T::Array[String]
  update_files, T::Array[String]

  def dirty?
    !new_files.empty? || !updated_files.empty?
  end
end
```

So the `verify_dsl_rbi` method builds the current state and the client decide what to do with the information?

spec/tapioca/cli_spec.rb
Morriar: Can we add a method `last_execute_status` to `CliSpec` and store the result during `execute` instead?

cjhutchi: That's what I was doing before. Due to the async nature of `IO.popen(..)` it wasn't getting set correctly each time and was causing flakiness.

lib/tapioca/generator.rb
cjhutchi: I'm not opposed to that at all, but I'm going to opt to leaving that to the future refactors @paracycle and I discussed.

lib/tapioca/config_builder.rb
```diff
@@ -43,7 +43,8 @@ def default_options(command)
       sig { returns(String) }
       def default_command
         command = File.basename($PROGRAM_NAME)
-        args = ARGV.join(" ")
+        # Hack to avoid flags ending up in the header of the RBIs
+        args = ARGV.reject { |arg| arg.include?("--") }.join(" ")
```
paracycle, line 47:
```suggestion
        args = ARGV.grep_v(/^-/).join(" ")
```
which would drop all arguments that start with `-`, which the short option aliases start with.

lib/tapioca/generator.rb
```diff
@@ -116,16 +116,26 @@ def build_todos
...
-
-      say("Compiling DSL RBI files...")
+      if should_verify
+        say("Checking for out-of-date RBIs...")
+      else
+        say("Compiling DSL RBI files...")
+      end
       say("")
 
+      outpath = should_verify ? Dir.mktmpdir : config.outpath
```
paracycle, line 136: Are we not cleaning up after ourselves by removing the tmp directory?

lib/tapioca/generator.rb
```diff
@@ -135,11 +145,11 @@ def build_dsl(requested_constants)
       )
 
       compiler.run do |constant, contents|
-        filename = compile_dsl_rbi(constant, contents)
+        filename = compile_dsl_rbi(constant, contents, outpath: Pathname.new(outpath))
         rbi_files_to_purge.delete(filename) if filename
       end
 
-      unless rbi_files_to_purge.empty?
+      if rbi_files_to_purge.any? && !should_verify
```
paracycle, line 152: I think this decision would be much cleaner if the actual actions were extracted:
```ruby
if should_verify
  perform_dsl_verification(outpath)
else
  purge_stale_dsl_rbi_files(rbi_files_to_purge)
end
```

lib/tapioca/generator.rb
```diff
@@ -522,5 +543,32 @@ def compile_dsl_rbi(constant, contents)
...
+
+        return "File(s) updated:\n  - #{filenames}"
+      end
+
+      nil
+    end
+
+    sig { params(dir: Pathname).returns(T::Array[Pathname]) }
+    def get_file_list(dir:)
+      Dir.glob("#{dir}/**/*.rbi").reject { |e| e =~ /gems|shims/ }.map { |f| Pathname.new(f) }
```
paracycle, line 571: We shouldn't need the `reject` here since the outpath should be just the `sorbet/rbi/dsl` folder for existing files and the temp folder for the new files. We don't expect any gem or shim RBI files in there, and we could potentially decide to add more RBI folders which would mean that we would need to exclude them in the future as well.

lib/tapioca/generator.rb
paracycle: Agreed that this can be better, but I think this whole file needs a refactor anyway, so I am fine with the current state.

spec/tapioca/cli_spec.rb
paracycle: I can make the bigger change for all the tests later.

RyanBrushett (approved):
🎩 seems to work neatly to me!
Let's cut an issue for the refactors somewhere IMO since there are a few worth tracking.
Possible future improvement: Quieter output that just writes the result and the reason like

```
RBI files are out-of-date, please run `tapioca dsl` to update.
Reason: New file(s) introduced.
```

lib/tapioca/config_builder.rb
cjhutchi: Smart. Was jumping the gun on this one. I'll add the change. 👍🏻

lib/tapioca/generator.rb
cjhutchi: Good point. I'll double check, previously when we were using the block syntax in a different method, it cleaned up automatically after the block executed. May need to manually do that. I'll double check and update if necessary.

lib/tapioca/generator.rb
cjhutchi: This was because I'm using the same method to grep the existing `./sorbet/` folder as well. Can probably clean this up.

paracycle (approved):

vinistock (approved):
🚀

KaanOzkan (approved):
Thanks Cory, nice job 🙂

Morriar (approved):

PR #236, opened 2021-02-23
Add ActiveRecord relations generator

paracycle:


spec/tapioca/compilers/dsl/active_record_relations_spec.rb
```diff
@@ -0,0 +1,602 @@
...
+
+          module GeneratedAssociationRelationMethods
+            sig { returns(PrivateAssociationRelation) }
+            def all; end
+
+            sig { params(args: T.untyped, blk: T.untyped).returns(PrivateAssociationRelation) }
+            def create_with(*args, &blk); end
+
+            sig { params(args: T.untyped, blk: T.untyped).returns(PrivateAssociationRelation) }
+            def create_with!(*args, &blk); end
```
rafaelfranca, line 190: Those methods with `!` should not be generated. They are private API of the framework and applications should not be calling.

spec/tapioca/compilers/dsl/active_record_relations_spec.rb
```diff
@@ -0,0 +1,602 @@
...
+            def where!(*args, &blk); end
+          end
+
+          class PrivateAssociationRelation < ::ActiveRecord::AssociationRelation
+            include CommonRelationMethods
+            include GeneratedAssociationRelationMethods
+            Elem = type_member(fixed: Post)
+
+            sig { returns(ActiveRecord::Associations::Association) }
+            def proxy_association; end
```
rafaelfranca, line 534: This is not public API.

spec/tapioca/compilers/dsl/active_record_relations_spec.rb
paracycle: Done. Except it seems `uniq!` is public API, so I included that, but it doesn't appear here since this tests against Rails 5.2

spec/tapioca/compilers/dsl/active_record_relations_spec.rb
paracycle: Removed.

manual/generator_activerecordrelations.md
```diff
@@ -0,0 +1,3 @@
+## ActiveRecordRelations
+
+[No description provided]
```
Morriar, line 3: 😢

Morriar (commented):
Looks very good!

jeffcarbs:
@paracycle - any plans to revisit this PR?

Also do you have thoughts on if/how those `Private…` classes could be exposed at runtime so they can be used in online sigs? Monkey-patching is obviously possible but that means tapioca now needs a runtime component. I could imagine just requiring the consumer to either:
- use `T::Sig::WithoutRuntime.sig`
- define it themselves, eg inside the model: `PrivateRelation = ActiveRecord_Relation`

lib/tapioca/compilers/dsl/active_record_relations.rb
```diff
@@ -0,0 +1,498 @@
...
+                    Parlour::RbiGenerator::Parameter.new(
+                      "unique_by:",
+                      type: "T.any(T::Array[Symbol], Symbol)",
+                      default: "nil"
+                    ),
+                  ],
+                  return_type: "ActiveRecord::Result"
+                )
+              when :proxy_association
+                # skip - private method
```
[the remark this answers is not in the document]
paracycle: For now, this is the only private method on that module, yeah, so that is the only one we explicitly skip (so that it doesn't look like an omission). If any more private methods are added in Rails, we won't match them anyway, since we opt-in to the methods we generate instead of opting out.

paracycle:
Ok, I've rebased and refactored the final version of this PR, but have not added documentation for it yet, which is what I will be doing next. For all other aspects, this should be good for a re-review.

lib/tapioca/compilers/dsl/active_record_relations.rb
```diff
@@ -0,0 +1,465 @@
...
+            T::Array[Symbol]
+          )
+
+          QUERY_METHODS = T.let(begin
+            # Grab all Query methods
+            query_methods = ActiveRecord::QueryMethods.instance_methods(false)
+            # Grab all Spawn methods
+            query_methods |= ActiveRecord::SpawnMethods.instance_methods(false)
+            # Remove the ones we know are private API
+            query_methods -= [:arel, :build_subquery, :construct_join_dependency, :extensions, :spawn]
```
Morriar, line 88: Should we move this a `PRIVATE_API_METHODS` constant?

paracycle: I was trying to keep all the constant definitions as "lists of methods we want" and that's why I didn't pull this out to a constant. But, if you think it makes sense, I can do it.

PR #243, opened 2021-03-02
Compile T::Enum with the `enums` block

jeffcarbs:
### Motivation

With sorbet's `T::Enum` the enum values themselves are actually instances of the class that inherits from `T::Enum`. Tapioca doesn't handle these specially which affects sorbet's ability to perform exhaustiveness checking. Here's a repro [on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0A%23%20Normal%20enum%20defintion%0Aclass%20Foo%20%3C%20T%3A%3AEnum%0A%20%20enums%20do%0A%20%20%20%20A%20%3D%20new%0A%20%20end%0Aend%0A%0A%23%20What%20tapoica%20generates%0Aclass%20TapiocaFoo%20%3C%20T%3A%3AEnum%0Aend%0ATapiocaFoo%3A%3AA%20%3D%20T.let(T.unsafe(nil)%2C%20TapiocaFoo)%0A%0A%23%20Test%20normal%0Asig%20%7B%20params(arg0%3A%20Foo).void%20%7D%0Adef%20test_normal(arg0)%0A%20%20case%20arg0%0A%20%20when%20Foo%3A%3AA%0A%20%20else%0A%20%20%20%20T.absurd(arg0)%0A%20%20end%0Aend%0A%0A%23%20Test%20tapioca%0Asig%20%7B%20params(arg0%3A%20TapiocaFoo).void%20%7D%0Adef%20test_tapioca(arg0)%0A%20%20case%20arg0%0A%20%20when%20TapiocaFoo%3A%3AA%0A%20%20else%0A%20%20%20%20T.absurd(arg0)%0A%20%20end%0Aend).

### Implementation

Specially handle `T::Enum` to define the enums using the `enums do ... end` block and ignore the enum values from the normal compilation.

### Tests

See updated tests.

### DIFF
--- a/lib/tapioca/compilers/symbol_table/symbol_generator.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_generator.rb
@@ -95,6 +95,7 @@ module Tapioca
           return if alias_namespaced?(name)
           return if seen?(name)
           return unless parent_declares_constant?(name)
+          return if T::Enum === constant # T::Enum instances are defined via `compile_enums`
 
           mark_seen(name)
           compile_constant(name, constant)
@@ -183,6 +184,7 @@ module Tapioca
               compile_mixins(constant),
               compile_mixes_in_class_methods(constant),
               compile_props(constant),
+              compile_enums(constant),
               methods,
             ].select { |b| b != "" }.join("\n\n")
           end
@@ -215,6 +217,25 @@ module Tapioca
               indented("#{method} :#{name}, #{type}")
             end
           end.join("\n")
+        end
+
+        sig { params(constant: Module).returns(String) }
+        def compile_enums(constant)
+          return "" unless T::Enum > constant
+
+          # T::Enum doesn't currently give a nice way to get the actual enum name so you
+          # need to parse it from the string representation which is in the form `#<SomeEnum::A>`
+          enums = T.cast(constant, T::Enum).values.map do |value|
+            value.to_s.match(/#<([\w:]+)>/)[1].split('::').last
+          end
+
+          content = [
+            indented('enums do'),
+            *enums.map { |e| indented("  #{e} = new") }.join("\n"),
+            indented('end'),
+          ]
+
+          content.join("\n")
         end
 
         sig { params(name: String, constant: Module).returns(T.nilable(String)) }
--- a/lib/tapioca/compilers/symbol_table/symbol_generator.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_generator.rb
@@ -221,12 +221,10 @@ module Tapioca
 
         sig { params(constant: Module).returns(String) }
         def compile_enums(constant)
-          return "" unless T::Enum > constant
+          return "" unless constant < T::Enum
 
-          # T::Enum doesn't currently give a nice way to get the actual enum name so you
-          # need to parse it from the string representation which is in the form `#<SomeEnum::A>`
-          enums = T.cast(constant, T::Enum).values.map do |value|
-            value.to_s.match(/#<([\w:]+)>/)[1].split('::').last
+          enums = T.cast(constant, T::Enum).values.map do |enum_type|
+            enum_type.instance_variable_get(:@const_name).to_s
           end
 
           content = [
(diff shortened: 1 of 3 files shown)


lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -217,6 +219,25 @@ def compile_props(constant)
...
         end
 
+        sig { params(constant: Module).returns(String) }
+        def compile_enums(constant)
+          return "" unless T::Enum > constant
+
+          # T::Enum doesn't currently give a nice way to get the actual enum name so you
+          # need to parse it from the string representation which is in the form `#<SomeEnum::A>`
+          enums = T.cast(constant, T::Enum).values.map do |value|
+            value.to_s.match(/#<([\w:]+)>/)[1].split('::').last
```
paracycle, line 229: I think we can do:
```suggestion
            value.instance_variable_get(:@const_name).to_s
```
as well.

lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -217,6 +219,25 @@ def compile_props(constant)
           end.join("\n")
         end
 
+        sig { params(constant: Module).returns(String) }
+        def compile_enums(constant)
+          return "" unless T::Enum > constant
+
+          # T::Enum doesn't currently give a nice way to get the actual enum name so you
+          # need to parse it from the string representation which is in the form `#<SomeEnum::A>`
+          enums = T.cast(constant, T::Enum).values.map do |value|
```
paracycle, line 228: Can we rename `value` to `enum_type` to make it more obvious that we are dealing with types and not instances?
```suggestion
          enums = T.cast(constant, T::Enum).values.map do |enum_type|
```

lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -217,6 +219,25 @@ def compile_props(constant)
           end.join("\n")
         end
 
+        sig { params(constant: Module).returns(String) }
+        def compile_enums(constant)
+          return "" unless T::Enum > constant
```
paracycle, line 224: Small nitpick, but can we swap the order of this comparison please? I find it much easier to read `constant < T::Enum` as "`constant` inherits from `T::Enum`" personally.

```suggestion
          return "" unless constant < T::Enum
```

paracycle (commented):
Looks great, thank you. Just a couple of comments.

lib/tapioca/compilers/symbol_table/symbol_generator.rb
jeffcarbs: I was originally thinking that but wasn't sure how you'd feel about reaching into a private instance method. I think it's much better than the regex tbh 👍

paracycle: Yeah, I'd rather that we explicitly reach into the functionality instead of relying on a name serialization side-effect to grab this. It would make it easier to troubleshoot if the functionality changes down the line.

jeffcarbs:
@paracycle - Thanks for the suggestions. Just committed them so this should be good for another look.

paracycle (approved):
Yeap, this looks great!

paracycle:
@jeffcarbs You had signed the CLA before, right? I think it might be our CLA bot acting up. Do you mind amending your last commit and force-pushing to trigger it again?

jeffcarbs:
Amending and force-pushing worked 👍

lib/tapioca/compilers/symbol_table/symbol_generator.rb
paracycle: It turns out that this advice was pretty bad, the `XPath` constant from the `capybara` gem messes with the `<` operator 🤦 :
```
pry> XPath < T::Enum
=> #<XPath::Expression:0x0000559c53120378
 @arguments=
  [:<, #<XPath::Expression:0x0000559c531203c8 @arguments=[], @expression=:this_node>, T::Enum],
 @expression=:binary_operator>
pry> T::Enum > XPath
=> nil
pry> ancestors_of(XPath)
=> [XPath, XPath::DSL]
```

I'll revert the comparison order back again to fix this and will do the same for other locations that might be doing similar checks too.

PR #244, opened 2021-03-02
Purge stale DSL RBI files during `dsl` operation

paracycle:
### Motivation

Do a better job of keeping DSL RBI files in sync by removing stale RBI files during generation.

### Implementation

We collect the set of existing DSL RBI files in the filesystem before we start any DSL RBI generation. We then remove items from the set as RBI files are generated. Whatever we are left with in that set should be RBI files that are no longer needed.

The only complication for this strategy is the fact that the user can supply a list of requested constant names, in which case, we need to build the initial list as the filenames of the RBIs that those constant names would have generated.

### Tests

See included tests.

### DIFF
--- a/lib/tapioca/generator.rb
+++ b/lib/tapioca/generator.rb
@@ -121,6 +121,8 @@ module Tapioca
       load_application(eager_load: requested_constants.empty?)
       load_dsl_generators
 
+      rbi_files_to_purge = existing_rbi_filenames(requested_constants)
+
       say("Compiling DSL RBI files...")
       say("")
 
@@ -133,7 +135,17 @@ module Tapioca
       )
 
       compiler.run do |constant, contents|
-        compile_dsl_rbi(constant, contents)
+        filename = compile_dsl_rbi(constant, contents)
+        rbi_files_to_purge.delete(filename) if filename
+      end
+
+      unless rbi_files_to_purge.empty?
+        say("")
+        say("Removing stale RBI files...")
+
+        rbi_files_to_purge.sort.each do |filename|
+          remove(filename)
+        end
       end
 
       say("")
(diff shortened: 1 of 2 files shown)


lib/tapioca/generator.rb
```diff
@@ -253,13 +265,38 @@ def load_dsl_generators
...
         rescue NameError
-          nil
+          [name, nil]
+        end
+      end.to_h
+
+      unprocessable_constants = constant_map.select { |_, v| v.nil? }
+      unless unprocessable_constants.empty?
+        unprocessable_constants.each do |name, _|
+          say("Error: Cannot find constant '#{name}'", :red)
```
vinistock, line 279: Question: in which scenario would the file exist but the constant be missing? Is this in the case of renaming a class and forgetting to remove/change the corresponding RBI?

Morriar (approved):
Nice, this will be useful!

lib/tapioca/generator.rb
Morriar: I believe this one is about the constant you pass to Tapioca: `tapioca dsl MyConstant`

paracycle: Yes, if a constant that previously existed was removed or renamed, then the RBI could have been left dangling. For the full run, that will be cleaned up at the end of the run. I also wanted to make sure that the RBI is cleaned up even if someone supplied the old constant name explicitly.

paracycle: Btw, most people don't remove RBI files of constants that are removed or renamed, so this does happen a lot in our codebase.

vinistock (approved):

lib/tapioca/generator.rb
```diff
@@ -316,6 +358,7 @@ def add(filename)
 
     sig { params(filename: Pathname).void }
     def remove(filename)
+      return unless filename.exist?
```
KaanOzkan, line 361: This is a nice touch in case user passes in bad input ie. not a fully qualified name 👍

KaanOzkan (approved):

PR #246, opened 2021-03-04
Allow for custom sorbet executables via TPC_SORBET_EXE env var

### LINKED ISSUE #245: Support alternative Sorbet install locations
Currently, the path of the `sorbet-static` binary is hardcoded in `lib/tapioca/compilers/sorbet.rb`. However, I have a [custom build of Sorbet](https://github.com/sorbet/sorbet/issues/3852#issuecomment-765865954) that I use because Sorbet doesn't yet support my M1 mac. However, when I try to run `tapioca sync` or other tapioca commands, only empty rbi files are generated because tapioca attempts to use a version of Sorbet that crashes on my system.

When running `srb` commands, there's an escape hatch for this issue: [environment variable `SRB_SORBET_EXE`](https://github.com/sorbet/sorbet/blob/47af81c2b70269349fd9f38da3b6ff7f4f593cb3/gems/sorbet/README.md#environment-variables), which allows one to define a custom path to `sorbet-static`. Would it be possible for tapioca to support this variable or a similar env var? There are also other escape hatch approaches (e.g., an argument) if we don't want to use environment variables.

I'm happy to make a PR for this change if it's a feature you'd like to add!

nathunsmitty:
### Motivation

Fixes #245. Allows users to specify an alternative installation path for `sorbet-static` via `TPC_SORBET_EXE`, similar to the `SRB_SORBET_EXE` used by `srb`.

### Implementation

The path to `sorbet-static` was hardcoded in `lib/tapioca/compilers/sorbet.rb`. I added a conditional to check if `TPC_SORBET_EXE` is set, and if so use that value instead. If `TPC_SORBET_EXE` is not set, behavior should be unchanged.

I wasn't sure where/if I should add documentation to let users know `TPC_SORBET_EXE` exists. Open to suggestions!

### Tests

Unit tests added in `spec/tapioca/compilers/sorbet_spec.rb`. I also ran my fork of tapioca on my system with my custom sorbet executable, to confirm it works. Not sure if it makes sense to add any integration or e2e tests for this change.

### DIFF
--- a/lib/tapioca/compilers/sorbet.rb
+++ b/lib/tapioca/compilers/sorbet.rb
@@ -8,6 +8,7 @@ module Tapioca
   module Compilers
     module Sorbet
       SORBET = Pathname.new(Gem::Specification.find_by_name("sorbet-static").full_gem_path) / "libexec" / "sorbet"
+      EXE_PATH_ENV_VAR = "TAPIOCA_SORBET_EXE"
 
       class << self
         extend(T::Sig)
@@ -26,7 +27,7 @@ module Tapioca
 
         sig { returns(String) }
         def sorbet_path
-          sorbet_path = ENV.fetch("TAPIOCA_SORBET_EXE", SORBET)
+          sorbet_path = ENV.fetch(EXE_PATH_ENV_VAR, SORBET)
           sorbet_path = SORBET if sorbet_path.empty?
           sorbet_path.to_s.shellescape
         end
--- a/lib/tapioca/compilers/sorbet.rb
+++ b/lib/tapioca/compilers/sorbet.rb
@@ -26,7 +26,12 @@ module Tapioca
 
         sig { returns(String) }
         def sorbet_path
-          SORBET.to_s.shellescape
+          custom_sorbet_path = ENV["TPC_SORBET_EXE"]
+          if custom_sorbet_path.nil? || custom_sorbet_path.empty?
+            SORBET.to_s.shellescape
+          else
+            custom_sorbet_path.shellescape
+          end
         end
       end
     end
--- a/lib/tapioca/compilers/sorbet.rb
+++ b/lib/tapioca/compilers/sorbet.rb
@@ -26,12 +26,9 @@ module Tapioca
 
         sig { returns(String) }
         def sorbet_path
-          custom_sorbet_path = ENV["TPC_SORBET_EXE"]
-          if custom_sorbet_path.nil? || custom_sorbet_path.empty?
-            SORBET.to_s.shellescape
-          else
-            custom_sorbet_path.shellescape
-          end
+          sorbet_path = ENV.fetch("TAPIOCA_SORBET_EXE", SORBET)
+          sorbet_path = SORBET if sorbet_path.empty?
+          sorbet_path.to_s.shellescape
         end
       end
     end
--- a/spec/tapioca/compilers/sorbet_spec.rb
+++ b/spec/tapioca/compilers/sorbet_spec.rb
@@ -26,4 +26,11 @@ class Tapioca::Compilers::SorbetSpec < Minitest::Spec
       assert_equal(Tapioca::Compilers::Sorbet.sorbet_path, default_path)
     end
   end
+
+  it("returns the default sorbet path if TAPIOCA_SORBET_EXE is not set") do
+    default_path = Tapioca::Compilers::Sorbet::SORBET.to_s.shellescape
+    with_custom_sorbet_exe_path(nil) do
+      assert_equal(Tapioca::Compilers::Sorbet.sorbet_path, default_path)
+    end
+  end
 end
--- a/spec/tapioca/compilers/sorbet_spec.rb
+++ b/spec/tapioca/compilers/sorbet_spec.rb
@@ -1,0 +1,29 @@
+# typed: strict
+# frozen_string_literal: true
+
+require "spec_helper"
+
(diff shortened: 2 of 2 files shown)


lib/tapioca/compilers/sorbet.rb
```diff
@@ -26,7 +26,12 @@ def run(*args)
 
         sig { returns(String) }
         def sorbet_path
-          SORBET.to_s.shellescape
+          custom_sorbet_path = ENV["TPC_SORBET_EXE"]
+          if custom_sorbet_path.nil? || custom_sorbet_path.empty?
+            SORBET.to_s.shellescape
+          else
+            custom_sorbet_path.shellescape
+          end
```
paracycle, line 34:
```suggestion
          sorbet_path = ENV.fetch("TAPIOCA_SORBET_EXE", SORBET)
          sorbet_path = SORBET if sorbet_path.empty?
          sorbet_path.to_s.shellescape
```
would be much more succinct.

spec/tapioca/compilers/sorbet_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+class Tapioca::Compilers::SorbetSpec < Minitest::Spec
+  before do
+    @temp_env_value = T.let(nil, T.nilable(String))
+    @temp_env_value = ENV["TPC_SORBET_EXE"]
+  end
+
+  after do
+    ENV["TPC_SORBET_EXE"] = @temp_env_value
+  end
```
paracycle, line 14: I'd rather use a helper method instead of changing test state:
```ruby
def with_custom_sorbet_exe_path(path, &block)
  sorbet_exe_env_value = ENV["TAPIOCA_SORBET_EXE"]
  begin
    ENV["TAPIOCA_SORBET_EXE"] = path
    block.call(path)
  ensure
    ENV["TAPIOCA_SORBET_EXE"] = sorbet_exe_env_value
  end
end
```
and then use it in tests like:
```ruby
with_custom_sorbet_exe_path("custom_value") do |custom_path|
  assert_equal(Tapioca::Compilers::Sorbet.sorbet_path, custom_path)
end
```
This way, the test is way more readable, you don't need any cleanup steps and reduce the number of local variables you have to use in tests.

spec/tapioca/compilers/sorbet_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  it("returns the value of TPC_SORBET_EXE if set") do
+    custom_path = 'bin/custom-sorbet-static'
+    ENV["TPC_SORBET_EXE"] = custom_path
+    sorbet_path = Tapioca::Compilers::Sorbet.sorbet_path
+    assert_equal(sorbet_path, custom_path)
+  end
+
+  it("returns the default sorbet path if TPC_SORBET_EXE is not set") do
```
paracycle, line 23: This is not exactly testing that the value is not set. It is testing the case that the value is set but it is set to an empty string. Can you add another test case for when the value of the env var is `nil` to test for env var not being set?

spec/tapioca/compilers/sorbet_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    ENV["TPC_SORBET_EXE"] = ''
+    gem_path = Gem::Specification.find_by_name("sorbet-static").full_gem_path
+    default_path = File.join(gem_path, "libexec", "sorbet")
```
paracycle, line 26: You should be able to get this more easily by:
```suggestion
    default_path = Tapioca::Compilers::Sorbet::SORBET
```

paracycle (commented):
Thank you for taking the time to put this together. I have a few comments for improvement that I've noted inline.

Also, seeing `TPC_` repeated, I decided that it doesn't look nice. Explicitly writing out `TAPIOCA_` for the env var prefix will be much better.

spec/tapioca/compilers/sorbet_spec.rb
nathunsmitty: Much easier! Slight modification, I made this `Tapioca::Compilers::Sorbet::SORBET.to_s.shellescape`, since `SORBET` is a `Pathname` and not a string.

spec/tapioca/compilers/sorbet_spec.rb
nathunsmitty: This is great, thanks for the suggestion!

lib/tapioca/compilers/sorbet.rb
```diff
@@ -26,7 +26,9 @@ def run(*args)
 
         sig { returns(String) }
         def sorbet_path
-          SORBET.to_s.shellescape
+          sorbet_path = ENV.fetch("TAPIOCA_SORBET_EXE", SORBET)
```
paracycle, line 29: Apologies for seeing this late, but I think we can do one more small refactor by pulling this env var name out to a constant on this class. We can refer to the env var in tests by using the constant as well. 

That way, if we ever decide to change the name or something, we would only need to do it in one place.

nathunsmitty: @paracycle done! Apologies for finishing this on a weekend, I'm in no rush on this change; you can re-review on Monday. Thank you for the thorough code review – I learned a couple of new things about Ruby from you! 🙂

paracycle (approved):
Awesome, thank you!

lib/tapioca/compilers/sorbet.rb
paracycle: Amazing! Great to hear that and thanks again for the contribution.

PR #247, opened 2021-03-04
Generate better Struct definitions

paracycle:

### DIFF
--- a/lib/tapioca/compilers/symbol_table/symbol_generator.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_generator.rb
@@ -155,7 +155,11 @@ module Tapioca
 
           header =
             if constant.is_a?(Class)
-              indented("class #{name}#{compile_superclass(constant)}")
+              if constant < Struct && struct_definition?(constant)
+                indented("#{name} = #{compile_struct_definition(constant)} do")
+              else
+                indented("class #{name}#{compile_superclass(constant)}")
+              end
             else
               indented("module #{name}")
             end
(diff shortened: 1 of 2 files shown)


lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -169,8 +169,15 @@ def compile_module(tree, name, constant)
 
           scope =
             if constant.is_a?(Class)
-              superclass = compile_superclass(constant)
-              RBI::Class.new(name, superclass_name: superclass)
+              if constant < Struct && struct_definition?(constant)
+                members = constant.members.map(&:to_s)
```
Morriar, line 173: Should we make `RBI::Struct` take symbols for its `members`?

spec/tapioca/rbi/model_spec.rb
```diff
@@ -97,6 +100,9 @@ class ModelSpec < Minitest::HooksSpec
           cls << singleton_class
           assert_equal("::Foo::Bar::<self>", singleton_class.to_s)
 
+          struct = RBI::Struct.new("Foo", members: ["foo", "bar"], keyword_init: true)
```
Morriar, line 103: We might want to add special treatment for it in [Index](https://github.com/Shopify/tapioca/blob/master/lib/tapioca/rbi/index.rb#L43) so we can also index the members as if they were real methods. So we can detect conflicting redefinitions / duplicates.

Morriar (approved):
> I have NOT, however, added a parser for this syntax

I'll take care of it once this is merged 👍

vinistock (approved):

Issue #237, opened 2021-02-24
sync: LoadError: cannot load such file -- RMagick


jaredbeck:
Trying to install tapioca for the first time, `sync` was unable to compile `carrierwave`. 

My app does not use RMagick. RMagick is not a runtime dependency of CarrierWave.

```
bundle exec tapioca -v
Tapioca v0.4.14
bundle exec tapioca sync
...
  Compiling carrierwave, this may take a few seconds... bundler: failed to load command: tapioca (/Users/jared/.rbenv/versions/2.6.6/bin/tapioca)
LoadError: cannot load such file -- RMagick
..../ruby/gems/2.6.0/gems/activesupport-6.1.3/lib/active_support/dependencies.rb:332:in `require'
..../ruby/gems/2.6.0/gems/activesupport-6.1.3/lib/active_support/dependencies.rb:332:in `block in require'
..../ruby/gems/2.6.0/gems/activesupport-6.1.3/lib/active_support/dependencies.rb:299:in `load_dependency'
..../ruby/gems/2.6.0/gems/activesupport-6.1.3/lib/active_support/dependencies.rb:332:in `require'
..../ruby/gems/2.6.0/gems/carrierwave-2.1.1/lib/carrierwave/processing/rmagick.rb:65:in `rescue in block in <module:RMagick>'
..../ruby/gems/2.6.0/gems/carrierwave-2.1.1/lib/carrierwave/processing/rmagick.rb:62:in `block in <module:RMagick>'
..../ruby/gems/2.6.0/gems/activesupport-6.1.3/lib/active_support/concern.rb:136:in `class_eval'
..../ruby/gems/2.6.0/gems/activesupport-6.1.3/lib/active_support/concern.rb:136:in `append_features'
..../ruby/gems/2.6.0/gems/tapioca-0.4.14/lib/tapioca/compilers/symbol_table/symbol_generator.rb:349:in `include'
..../ruby/gems/2.6.0/gems/tapioca-0.4.14/lib/tapioca/compilers/symbol_table/symbol_generator.rb:349:in `block (2 levels) in compile_mixes_in_class_methods'
..../ruby/gems/2.6.0/gems/tapioca-0.4.14/lib/tapioca/compilers/symbol_table/symbol_generator.rb:359:in `compile_mixes_in_class_methods'
..../ruby/gems/2.6.0/gems/tapioca-0.4.14/lib/tapioca/compilers/symbol_table/symbol_generator.rb:184:in `block in compile_body'
[...]

Issue #238, opened 2021-02-25
Make errors from DSL generators more approachable


paracycle:
Currently, when an error is raised during the processing of a DSL generator, we fail with a printed stack trace and nothing else. That makes it really hard to debug the failure, since it is not even obvious which source file the generator is failing on.

We need to display better errors that include:
- which generator caused an error
- for which constant
- with what error message and stack trace

mutecipher:
Going to dump some ideas in here, as I'll be away on vacation next week, and want to have this information available if someone decides to pick up some work. 

- [ ] Centralized error handler for `tapioca`
- [ ] Update runs to no longer fail/report errors immediately. Instead, store them in the error handler, and provide errors/warnings after a run has completed
  - Complete a best effort run, get done what we can, but advise what has failed/errored/etc. 
  - We should be able to generate enough information to advise on potential fixes in the source, or what may require a shim
- [ ] Separate concerns between business logic and application UI
  - [ ]  Break up `generators.rb` into separately clearly defined generator classes
    - Invert dependencies and pass in CLI to deal with reporting?
    - Leave reporting to CLI only, and only gather errors/warnings in the generator classes
      - Use the [Mediator pattern](https://refactoring.guru/design-patterns/mediator)(?)
 
This should get us into a good state where we can provide better context surrounding errors encountered during a run while providing _best-effort_ completion and helpful messaging to solve source related issues.

mutecipher:
Closing this after #584 and #578.

Will add/open smaller for focused issues for additional error messages in the future rather than maintaining this issue. 👍🏻

PR #252, opened 2021-03-09
Mixed in class methods detection should be best effort

paracycle:
### Motivation

Fixes: #237

We try to do something sneaky to detect modules that are mixed into a class through inclusion of another module, where we try to infer that information from the change in the ancestor list of a test class after including the said module to that class.

However, that operation is risky since a `self.included` method on the target module could potentially run any Ruby code and it turns out some code in the wild even try to require optional gems when the concern is being included.

### Implementation

Since this detection is best effort, it should be fine to catch all kinds of `Exception`s during `include` and bail out of the operation without further side-effects.

### Tests

Added a test to verify that we can still generate module definitions even if `self.included` fails.

### DIFF
--- a/lib/tapioca/compilers/symbol_table/symbol_generator.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_generator.rb
@@ -364,9 +364,13 @@ module Tapioca
             end
 
             define_singleton_method(:include) do |mod|
-              before = singleton_class.ancestors
-              super(mod).tap do
-                mixins_from_modules[mod] = singleton_class.ancestors - before
+              begin
+                before = singleton_class.ancestors
+                super(mod).tap do
+                  mixins_from_modules[mod] = singleton_class.ancestors - before
+                end
+              rescue Exception # rubocop:disable Lint/RescueException
+                # this is a best effort, bail if we can't perform this
               end
             end
 
--- a/spec/tapioca/compilers/symbol_table_compiler_spec.rb
+++ b/spec/tapioca/compilers/symbol_table_compiler_spec.rb
@@ -1662,6 +1662,59 @@ class Tapioca::Compilers::SymbolTableCompilerSpec < Minitest::HooksSpec
       assert_equal(output, compile)
     end
 
+    it("safely bails out of generating mixes_in_class_methods for modules that do weird things") do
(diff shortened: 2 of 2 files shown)


KaanOzkan (approved):
Catching makes sense to me 👍

Morriar (approved):
Not for this pull-request but I wonder if we could monkey patch `mixes_in_class_method` so it register the list of receiver and mixes modules while we load the app?

paracycle:
> Not for this pull-request but I wonder if we could monkey patch `mixes_in_class_method` so it register the list of receiver and mixes modules while we load the app?

That's not quite what we are doing here, though. It would have been simpler if that was the case, indeed.

What we are trying to detect here is given a random module `Foo`, does it have any side-effects from the `include Foo` operation, like extending the target class with other modules etc. Basically we are trying to detect what `mixes_in_class_methods` stanza we should "generate" for `Foo`, not how to detect the `mixes_in_class_methods` call in `Foo` (since we don't expect random gem code to have any such usage).

lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -364,9 +364,13 @@ def method_missing(symbol, *args)
...
-              before = singleton_class.ancestors
-              super(mod).tap do
-                mixins_from_modules[mod] = singleton_class.ancestors - before
+              begin
+                before = singleton_class.ancestors
+                super(mod).tap do
+                  mixins_from_modules[mod] = singleton_class.ancestors - before
+                end
+              rescue Exception # rubocop:disable Lint/RescueException
+                # this is a best effort, bail if we can't perform this
```
jaredbeck, line 373: Might be nice to `warn` here, otherwise this "silent failure" could make future debugging difficult. @paracycle

paracycle: Agreed in principle, but I would like to address the few places where we unconditionally and silently rescue errors similar to this into a more uniform framework. The output is already pretty noisy and I would not like to make it more noisy by surfacing these errors as is.

I had created the #238 issue to handle that in the DSL generation part of the codebase, but will most certainly extend the same thing to other parts of the codebase too. It will probably be a file-based logging operation with an indication in the UI that there might have been some errors encountered during the operation, so best to inspect the file.

PR #253, opened 2021-03-10
Properly handle types that override `<`

paracycle:
### Motivation

The `T::Enum` detection code was giving us a false-positive match for the [`XPath` constant from the `xpath` gem that defines a custom `<` class method](https://github.com/teamcapybara/xpath/blob/9860a6b20b31be97f6bd23a42ff6573c35f910c3/lib/xpath/dsl.rb#L122). Since `XPath` does not also have a `values` method, we were getting a hard failure.

### Implementation

Since `constant < X` actually means `constant.<(X)`, when we try to run `constant < T::Enum`, if `constant` has a custom `<` (class) method defined, then the check would not mean what we want to detect, i.e. inheritance.
    
On the other hand, we are pretty much guaranteed `T::Enum > constant` will work since we don't expect `T::Enum` to have an overridden `>` method.

### Tests

Added a test to trigger this case explicitly to make sure we don't regress.

/cc @jeffcarbs

### DIFF
--- a/lib/tapioca/compilers/symbol_table/symbol_generator.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_generator.rb
@@ -221,7 +221,7 @@ module Tapioca
 
         sig { params(constant: Module).returns(String) }
         def compile_enums(constant)
-          return "" unless constant < T::Enum
+          return "" unless T::Enum > constant
 
           enums = T.cast(constant, T::Enum).values.map do |enum_type|
             enum_type.instance_variable_get(:@const_name).to_s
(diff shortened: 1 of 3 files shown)


lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -221,7 +221,7 @@ def compile_props(constant)
 
         sig { params(constant: Module).returns(String) }
         def compile_enums(constant)
-          return "" unless constant < T::Enum
+          return "" unless T::Enum > constant
```
Morriar, line 224: It seems we're just moving the problem around, can't we call directly the definition from `Object`?

paracycle: Yes, there is still some risk but much much less so since `T::Enum` is not a random type, but `constant` is. I'd rather not use the base method if we don't have to since it is much slower to do it.

vinistock (approved):

Morriar (approved):

Issue #254, opened 2021-03-11
Replace the header in `tapioca` so that it doesn't change so often


cjhutchi:
The header in RBI files that `tapioca` generates is dynamic, and therefore changes too often. We should make it static to avoid unnecessary changes to RBIs.

PR #259, opened 2021-03-15
Remove Config.generate_command and update tests

cjhutchi:
Closes: https://github.com/Shopify/tapioca/issues/254
### Motivation

Dynamic headers cause false positives and unnecessary changes when regenerating RBIs. Hard coded the commands for generation
and removed the `generate_command` from `Tapioca::Config` as it is no longer necessary.

### Implementation

Updated tests and implementations to use hardcoded `tapioca <command>`.

### Tests

No new tests added, but updated existing tests.


Morriar (commented):
Can we make it the default value of the option instead of removing the option?

cjhutchi:
> Can we make it the default value of the option instead of removing the option?

@Morriar There's some difficulty with that. Mainly because there is no default value for generating RBIs. Depending on what we generate different commands are used (i.e. Gems created with `tapioca sync`, and DSLs with `tapioca dsl`).

Also, feels kind of like a weird/non-standard flag to me. As you're basically telling the program to lie about what it's accomplished. 

I'm open to being convinced otherwise as to where it's useful.

Morriar:
> There's some difficulty with that.

Can't we reach the same result with something like:

```rb
    class_option :generate_command,
                  aliases: ["--cmd", "-c"],
                  banner: "command",
                  desc: "The command to run to regenerate RBI files",
                  default: "tapioca"
```

then:

```rb
say("RBI files are out-of-date, please run `#{config.command} dsl` to update.")
```

?

cjhutchi:
> > There's some difficulty with that.
> 
> Can't we reach the same result with something like:
> 
> ```ruby
>     class_option :generate_command,
>                   aliases: ["--cmd", "-c"],
>                   banner: "command",
>                   desc: "The command to run to regenerate RBI files",
>                   default: "tapioca"
> ```
> 
> then:
> 
> ```ruby
> say("RBI files are out-of-date, please run `#{config.command} dsl` to update.")
> ```
> 
> ?

That's more or less what we're trying to get rid of. Yes, you could specify the command that you want in the header, but that seems an obtuse way of accomplishing this task IMO.

Morriar:
> That's more or less what we're trying to get rid of. Yes, you could specify the command that you want in the header, but that seems an obtuse way of accomplishing this task IMO. In this case, we'd have to specify a `-C/--cmd` flag anytime we pass flags into `tapioca` in order to generate the correct header.

My understanding of the road map was to create a config file to store such option so we don't have to pass them each time but still have the option available.

Stating `tapioca X` in the header when we actually expect them to run `bundle tapioca X` or another command will come around to bite us with people committing files produced by the wrong version of `tapioca`.

cjhutchi:
> > That's more or less what we're trying to get rid of. Yes, you could specify the command that you want in the header, but that seems an obtuse way of accomplishing this task IMO. In this case, we'd have to specify a `-C/--cmd` flag anytime we pass flags into `tapioca` in order to generate the correct header.
> 
> My understanding of the road map was to create a config file to store such option so we don't have to pass them each time but still have the option available.
> 
> Stating `tapioca X` in the header when we actually expect them to run `bundle exec tapioca X` or another command will come around to bite us with people committing files produced by the wrong version of `tapioca`.

According to the Slack thread, Rafael suggested we always use `tapioca dsl` but I believe we've had discussions of a binstub but nothing concrete has come out of that yet.

lib/tapioca/cli.rb
```diff
@@ -19,10 +19,6 @@ class Cli < Thor
                   aliases: ["--out", "-o"],
                   banner: "directory",
                   desc: "The output directory for generated RBI files"
-    class_option :generate_command,
-                  aliases: ["--cmd", "-c"],
-                  banner: "command",
-                  desc: "The command to run to regenerate RBI files"
```
paracycle, line 25: I suggest that we still leave the option in place and show a warning to users if they are using the option. It would be better than making them fail weirdly on CI or something, in case they are using this.

lib/tapioca/generator.rb
```diff
@@ -510,14 +510,13 @@ def compile_gem_rbi(gem)
...
 
-      command = format(config.generate_command, constant.name)
       constant_name = Module.instance_method(:name).bind(constant).call
       rbi_name = constant_name.underscore + ".rbi"
       filename = outpath / rbi_name
 
       out = String.new
       out << rbi_header(
-        command,
+        "tapioca dsl",
```
paracycle, line 519: Let's make this:
```suggestion
        "tapioca dsl #{constant_name}",
```
since it should always be safe to run the DSL command for a single file and it will always be faster than a full run.

spec/tapioca/cli_spec.rb
```diff
@@ -363,7 +362,7 @@ module Foo::Undef2; end
       assert_equal(<<~CONTENTS.chomp, File.read("#{outdir}/post.rbi"))
         # DO NOT EDIT MANUALLY
         # This is an autogenerated file for dynamic methods in `Post`.
-        # Please instead update this file by running `generate command`.
+        # Please instead update this file by running `tapioca dsl`.
```
paracycle, line 365:
```suggestion
        # Please instead update this file by running `tapioca dsl Post`.
```

spec/tapioca/cli_spec.rb
```diff
@@ -448,7 +447,7 @@ def title=(title); end
       assert_equal(<<~CONTENTS.chomp, File.read("#{outdir}/baz/role.rbi"))
         # DO NOT EDIT MANUALLY
         # This is an autogenerated file for dynamic methods in `Baz::Role`.
-        # Please instead update this file by running `generate command`.
+        # Please instead update this file by running `tapioca dsl`.
```
paracycle, line 450:
```suggestion
        # Please instead update this file by running `tapioca dsl Baz::Role`.
```

spec/tapioca/cli_spec.rb
```diff
@@ -465,7 +464,7 @@ def title=(title); end
       assert_equal(<<~CONTENTS.chomp, File.read("#{outdir}/post.rbi"))
         # DO NOT EDIT MANUALLY
         # This is an autogenerated file for dynamic methods in `Post`.
-        # Please instead update this file by running `generate command`.
+        # Please instead update this file by running `tapioca dsl`.
```
paracycle, line 467:
```suggestion
        # Please instead update this file by running `tapioca dsl Post`.
```

spec/tapioca/cli_spec.rb
```diff
@@ -480,7 +479,7 @@ def title=(title); end
       assert_equal(<<~CONTENTS.chomp, File.read("#{outdir}/namespace/comment.rbi"))
         # DO NOT EDIT MANUALLY
         # This is an autogenerated file for dynamic methods in `Namespace::Comment`.
-        # Please instead update this file by running `generate command`.
+        # Please instead update this file by running `tapioca dsl`.
```
paracycle, line 482:
```suggestion
        # Please instead update this file by running `tapioca dsl Namespace::Comment`.
```

paracycle (commented):
We would also like to remove the default for `generate_command` in `config_builder.rb` and the `default_command` method in the same file.

lib/tapioca/cli.rb
paracycle: In other words, deprecate now with removed functionality and then remove the option in a later version.

paracycle:
> Stating `tapioca X` in the header when we actually expect them to run `bundle exec tapioca X` or another command will come around to bite us with people committing files produced by the wrong version of `tapioca`.

We run the same risk with people doing `rails xxx` as well. Hopefully, people are using our tools either with the binstubs or via `bundle exec`.

One thing we can do, however, is to make Tapioca create the binstub in `tapioca init` and then to set these commands as `bin/tapioca xxx` unconditionally. I think that would address @Morriar's concern.

paracycle:
The binstub creation code should be something like:
```ruby
installer = Bundler::Installer.new(Bundler.root, Bundler.definition)
spec = Bundler.definition.specs.find {|s| s.name == "tapioca" }
installer.generate_bundler_executable_stubs(spec, { force: true })
```

lib/tapioca/generator.rb
```diff
@@ -63,7 +63,7 @@ def build_requires
 
       content = String.new
       content << rbi_header(
-        config.generate_command,
+        "tapioca require",
```
Morriar, line 66: It should be `bin/tapioca require` since we install the binstub right?

paracycle: Yeah, we should have all commands as `bin/tapioca xxx` now, since we know, for a properly initialized app, it should always work.

lib/tapioca/cli.rb
```diff
@@ -22,7 +22,7 @@ class Cli < Thor
     class_option :generate_command,
                   aliases: ["--cmd", "-c"],
                   banner: "command",
-                  desc: "The command to run to regenerate RBI files"
+                  desc: "[DEPRECATED] The command to run to regenerate RBI files"
```
paracycle, line 25: This will be visible if the user runs `tapioca help` or similar. For proper deprecation, it would be great if all the methods that accept this option could show a warning that the option is deprecated and will be removed if the option is supplied.

spec/tapioca/cli_spec.rb
```diff
@@ -142,6 +141,8 @@ def execute(command, args = [], flags = {})
 
         # Add your extra requires here
       CONTENTS
+
+      assert_path_exists(repo_path / "bin/tapioca")
```
paracycle, line 145: 👍

paracycle (commented):
Like @Morriar pointed out, we need to `s|tapioca |bin/tapioca |g` or something in file headers and commands that we show.

Also, we need the `config_builder.rb` changes.

lib/tapioca/config_builder.rb
```diff
@@ -10,6 +10,9 @@ class << self
 
       sig { params(command: Symbol, options: T::Hash[String, T.untyped]).returns(Config) }
       def from_options(command, options)
+        puts(<<~MSG) if options.include?("generate_command")
+          DEPRECATION: The `-c` and `--cmd` flags will be removed in a future release.
+        MSG
         Config.from_hash(
           merge_options(default_options(command), config_options, options)
         )
```
paracycle, line 18: Small improvement might be to make sure that we don't get that option set by any config file either:
```suggestion
        merged_options = merge_options(default_options(command), config_options, options)

        puts(<<~MSG) if merged_options.include?("generate_command")
          DEPRECATION: The `-c` and `--cmd` flags will be removed in a future release.
        MSG
        
        Config.from_hash(merged_options)
```

lib/tapioca/config_builder.rb
```diff
@@ -66,7 +60,7 @@ def merge_options(*options)
     DEFAULT_OPTIONS = T.let({
       "postrequire" => Config::DEFAULT_POSTREQUIRE,
       "outdir" => nil,
-      "generate_command" => default_command,
+      "generate_command" => Config::DEFAULT_COMMAND,
```
paracycle, line 63: And, based on the message above, we wouldn't want to set any defaults for this command either.

paracycle (approved):

lib/tapioca/config_builder.rb
```diff
@@ -10,9 +10,13 @@ class << self
 
       sig { params(command: Symbol, options: T::Hash[String, T.untyped]).returns(Config) }
       def from_options(command, options)
-        Config.from_hash(
-          merge_options(default_options(command), config_options, options)
-        )
+        merged_options = merge_options(default_options(command), config_options, options)
+
+        puts(<<~MSG) if options.include?("generate_command")
```
paracycle, line 15: This line now needs to be:
```suggestion
        puts(<<~MSG) if merged_options.include?("generate_command")
```

vinistock (approved):

PR #261, opened 2021-03-16
Stop exporting internals from the gem

paracycle:
### Motivation

Tapioca should have never really been exporting its internal types from the gem. All the compilers, etc are for the CLI to operate properly and should not be exported by default.

### Implementation

Moved all the internal requires to `tapioca/internal.rb`. Also, moved the `sorbet-runtime` initialization and configuration to `tapioca/cli.rb` since that is the application which should be configuring the runtime.

Moving the Sorbet runtime configuration to CLI entrypoint also meant that we are running runtime typechecking during test runs now. That surfaced a few type errors that are fixed in this PR as well.

### Tests

No extra added tests.

### DIFF
--- a/lib/tapioca/compilers/symbol_table/symbol_generator.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_generator.rb
@@ -223,7 +223,7 @@ module Tapioca
         def compile_enums(constant)
           return "" unless T::Enum > constant
 
-          enums = T.cast(constant, T::Enum).values.map do |enum_type|
+          enums = T.unsafe(constant).values.map do |enum_type|
             enum_type.instance_variable_get(:@const_name).to_s
           end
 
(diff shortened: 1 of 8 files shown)


Morriar (approved):

lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -859,7 +859,7 @@ def signature_of(method)
           nil
         end
 
-        sig { params(constant: Module).returns(String) }
+        sig { params(constant: T::Types::Base).returns(String) }
         def type_of(constant)
```
Morriar, line 863: Actually I wonder if we should move this method as a service of `T::Types::Base` since we monkey patch it now?

paracycle: We don't yet monkey-patch it, but yes, that's a good idea. I'll do that once I've merged the first PR that starts the monkey-patching business.

PR #267, opened 2021-03-22
Improve binstub generation with output

paracycle:
### Motivation

We haven't been showing any output for binstub generation.

### Implementation

Add output that `bin/tapioca` was created or force created depending the existence of `bin/tapioca`

### Tests

Updated tests to check for the `bin/tapioca` output. Also changed test to run `bundle exec tapioca` and not `bundle exec bin/tapioca` since (a) we cannot rely on the binstub existing and (b) if running the binstub, it is meaningless to prepend a `bundle exec` to it.

### DIFF
--- a/lib/tapioca/cli/main.rb
+++ b/lib/tapioca/cli/main.rb
@@ -114,9 +114,15 @@ module Tapioca
       end
 
       def generate_binstub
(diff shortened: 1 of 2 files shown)


cjhutchi (approved):
Good catch

vinistock (approved):

PR #272, opened 2021-03-29
Make generic type creation safe for sealed classes

paracycle:
### Motivation

Sealed classes refuse to be subclassed in any file outside the original file they were declared in, but we use subclassing to represent concrete generic types at runtime. So for a generic sealed type, we need to jump through extra hoops to trick Sorbet runtime into thinking that what we are doing is safe.

### Implementation

Just some hacky state changing wrt the internals of how `sealed!` operates when doing the subclassing.

### Tests

Added an extra test that fails in the absence of the new code path.

### DIFF
--- a/lib/tapioca/generic_type_registry.rb
+++ b/lib/tapioca/generic_type_registry.rb
@@ -100,7 +100,7 @@ module Tapioca
           # the generic class `Foo[Bar]` is still a `Foo`. That is:
           # `Foo[Bar].new.is_a?(Foo)` should be true, which isn't the case
           # if we just clone the class. But subclassing works just fine.
-          Class.new(constant)
+          create_sealed_safe_subclass(constant)
         else
           # This can only be a module and it is fine to just clone modules
           # since they can't have instances and will not have `is_a?` relationships.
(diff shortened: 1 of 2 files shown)


cjhutchi (approved):
🙏🏻 You're a legend

vinistock (approved):

lib/tapioca/generic_type_registry.rb
```diff
@@ -150,6 +150,32 @@ def register_type_variable(constant, type_variable_type, type_variable, fixed, l
...
+        # We just return a subclass of the constant.
+        return Class.new(constant) unless T::Private::Sealed.sealed_module?(constant)
+
+        # Since sealed classes can normally not be subclassed, we need to trick
+        # sealed classes into thinking that the generic type we are
+        # creating by subclassing is actually safe for sealed types.
+        #
+        # Get the filename the sealed class was declared in
+        decl_file = constant.instance_variable_get(:@sorbet_sealed_module_decl_file)
+        begin
```
Morriar, line 165: Why do we need the `begin`/`ensure` blocks here? Are we expecting something to fail?

Morriar (approved):

lib/tapioca/generic_type_registry.rb
paracycle: The `ensure` is to set the state of `@sorbet_sealed_module_decl_file` back to its original value after we've changed it in the `begin` block.

paracycle: Since calling `sealed!` updates that value and before that we've removed it as well.

Morriar: If nothing is raising from what's inside the `begin`, do we need the `ensure`? Could this just be a plain sequence of instructions instead of a begin block?

paracycle: I just want to be sure that if anything *does* go wrong, the previous state will always be restored. Imagine that the `sealed!` call inside the block throws an error for some reason.

PR #274, opened 2021-03-30
Add a silent option for DSL generation

paracycle:
### Motivation

We print a lot of output in DSL generation which is not always helpful. Moreover, the DSL output in `--verify` mode is never helpful, just the final changed/added/removed message is good enough.

### Implementation

Added a `--silent` option for DSL generation and forced the `--verify` mode to activate the silent mode.

The silent mode only changes the output of files generated and nothing else for now.

### Tests

Added a test to ensure that `--silent` behaves as expected.

### DIFF
--- a/lib/tapioca/cli/main.rb
+++ b/lib/tapioca/cli/main.rb
@@ -66,9 +66,13 @@ module Tapioca
         type: :boolean,
         default: false,
         desc: "Verifies RBIs are up-to-date"
+      option :silent,
+        aliases: ["-s"],
+        type: :boolean,
+        desc: "Silences file creation output"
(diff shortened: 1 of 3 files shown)


Morriar (approved):

vinistock (approved):

KaanOzkan (approved):

paracycle:
I renamed the flag to `quiet` since silent sounded like it should make the whole output less verbose. Happy to hear your thoughts.

KaanOzkan (approved):
Good call.

vinistock (approved):
I agree with `quiet` 👍

PR #275, opened 2021-03-30
Print full command line for `tapioca dsl` in verify

paracycle:
### Motivation

I realized that we can keep a file name to constant name lookup table as we are doing DSL generation and then we can use that when reporting the diff in a verify run to report the exact command they can run to make RBI files complete.

### Implementation

- Maintain a filename to constant name lookup table as we are doing DSL generation.
- Change the diff output to be a filename to cause lookup.
- When reporting the error, build a list of the constant names from the files in the diff and print them at the end of the `bin/tapioca dsl` command.
- Print the diff by grouping the filename to cause lookup by the cause and printing all the files for a given cause in the output.

### Tests

Updates existing tests for changed expectations.

### DIFF
--- a/lib/tapioca/generator.rb
+++ b/lib/tapioca/generator.rb
@@ -134,7 +134,7 @@ module Tapioca
       end
       say("")
 
-      outpath = should_verify ? Dir.mktmpdir : config.outpath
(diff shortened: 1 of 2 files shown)


cjhutchi (approved):
🙏🏻 Huge improvement

Morriar (approved):

PR #276, opened 2021-04-01
Bump version to v0.4.20

cjhutchi:

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -8,7 +8,7 @@ GIT
 PATH
   remote: .
   specs:
-    tapioca (0.4.19)
+    tapioca (0.4.20)
       bundler (>= 1.17.3)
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -189,7 +189,7 @@ GEM
     smart_properties (1.15.0)
     sorbet (0.5.6357)
       sorbet-static (= 0.5.6357)
-    sorbet-runtime (0.5.6357)
+    sorbet-runtime (0.5.6361)
```
paracycle, line 192: Can we not comment this change? Not that it makes a difference for the release, but it would be great if the versions of runtime and static were in sync.

cjhutchi: Just saw that. Removing now

paracycle (approved):

PR #277, opened 2021-04-02
Extend protobuf support for map and repeated

kddeisz:
### Motivation

Was reviewing a PR that involved using protobuf objects directly and saw that they could not be properly typed.

### Implementation

Fields in protobuf can be marked as "repeated" or "map", which loosely correspond to arrays and hashes, respectively. This commit extends our support for protobuf by correctly marking the types of these fields. It also adds an initialize method to each of the protobuf classes so that creating them can be typed.

Basically, when you have `repeated` fields (https://developers.google.com/protocol-buffers/docs/reference/ruby-generated#repeated-fields), it looks like:

```ruby
repeated :examples, :string, 1
```

which corresponds to:

```ruby
sig { returns(Google::Protobuf::RepeatedField[String]) }
def examples; end

sig { params(value: Google::Protobuf::RepeatedField[String]).returns(Google::Protobuf::RepeatedField[String]) }
def examples=(value); end
```

and in the initialize method, it adds the following parameter:

```ruby
sig { params(examples: T.any(Google::Protobuf::RepeatedField[String], T::Array[String])).void }
def initialize(examples: Google::Protobuf::RepeatedField.new(:string)); end
```

This logic includes protobuf defaulting the value for the argument as well as the fact that you can pass an array for the value and it will automatically convert it into a `RepeatedField`.

Similarly, when you have `map` fields (https://developers.google.com/protocol-buffers/docs/reference/ruby-generated#map-fields), it looks like:

```ruby
map :examples, :string, :string, 1
```

which corresponds to:

```ruby
sig { returns(Google::Protobuf::Map[String, String]) }
def examples; end

sig { params(value: Google::Protobuf::Map[String, String]).returns(Google::Protobuf::Map[String, String]) }
def examples=(value); end
```

and in the initialize method, it adds the following parameter:

```ruby
sig { params(examples: T.any(Google::Protobuf::Map[String, String], T::Hash[String, String])).void }
def initialize(examples: Google::Protobuf::Map.new(:string, :string)); end
```

This commit includes making `Google::Protobuf::RepeatedField` and `Google::Protobuf::Map` generic so that it supports the same interface (and so we get the `Enumerable` typing for free). It also adds types to the `initialize` method so that we can write those in tests and they get typed for free. _Technically_ you can also use string keys for the `initialize` method, but I doubt it's being used a lot and it's cleaner this way because we can declare types for all of the keys.

### Tests

Yup, tests updated.

### DIFF
--- a/lib/tapioca/compilers/dsl/protobuf.rb
+++ b/lib/tapioca/compilers/dsl/protobuf.rb
@@ -62,32 +62,98 @@ module Tapioca
       # end
       # ~~~
       class Protobuf < Base
+        # Parlour doesn't support type members out of the box, so adding the
+        # ability to do that here. This should be upstreamed.
+        class TypeMember < Parlour::RbiGenerator::RbiObject
+          extend T::Sig
+
+          sig { params(other: Object).returns(T::Boolean) }
+          def ==(other)
+            TypeMember === other && name == other.name
+          end
+
+          sig do
+            override
+              .params(indent_level: Integer, options: Parlour::RbiGenerator::Options)
+              .returns(T::Array[String])
+          end
+          def generate_rbi(indent_level, options)
+            [options.indented(indent_level, "#{name} = type_member")]
+          end
+
+          sig do
+            override
+              .params(others: T::Array[Parlour::RbiGenerator::RbiObject])
+              .returns(T::Boolean)
+          end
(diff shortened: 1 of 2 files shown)


paracycle (approved):
Looks great! Thank you.

PR #291, opened 2021-04-20
Compile T.type_alias

jeffcarbs:
### Motivation

We're have a `T.type_alias` defined in an internal gem that we want to use inside a consuming application. Seems like it ought to come over in the RBI since the constant is available at runtime.

### Implementation

Currently tapioca ignores `T.type_alias` since it starts with `T::Private::Types::`. This PR updates the spot where that bailout happens and adds a special case to handle it.

### Tests

Extended one test and added another.


paracycle (approved):
Thanks for rebasing this. I've tested this change on our monolith and it works great.

PR #293, opened 2021-04-23
Stop generating more methods than necessary

paracycle:
### Motivation

We used to treat methods that didn't have a source location the same as methods that explicitly were not defined in the current gem. That resulted in Tapioca creating more method definitions than necessary.

We would only skip method generation for a method if the constant it was on was an ignored type (i.e. a built-in type), so that we wouldn't keep redefining methods for built-in types. However, for all other types, especially types that come from other gems, we would just keep on generating all the methods regardless of if they were defined by this gem or not.

Moreover, the source location check was happening at the wrong location, before unwrapping the method signature. Thus, many methods with signatures would not be generated when the previous problem was fixed, since our code would see them as being defined in Sorbet runtime.

### Implementation

The fix is to return a more fine-grained result from `method_in_gem?` which signals yes/no/don't-have-source-location. Based on that we can skip generating don't-have-source-location cases if they are for built-in types and totally ignore the methods that have a source location and are definitely not defined in the current gem.

Additionally, if we try to unwrap the method signature and we get an exception, that means the signature block raised an error. If we continue with the method as is, the source location checks would think the method definition does not belong in the gem (since the method is still wrapped), and we would thus skip the method generation. To avoid that, the `signature_for` method is not raising a custom exception to signal that exceptional case, so that we can at least continue generating "a" method definition.

### Tests

Updated existing tests.

### DIFF
--- a/lib/tapioca/compilers/symbol_table/symbol_generator.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_generator.rb
@@ -567,10 +567,25 @@ module Tapioca
         def compile_method(symbol_name, constant, method)
           return unless method
           return unless method.owner == constant
-          return if symbol_ignored?(symbol_name) && !method_in_gem?(method)
 
-          signature = signature_of(method)
-          method = T.let(signature.method, UnboundMethod) if signature
+          begin
+            signature = signature_of(method)
+            method = T.let(signature.method, UnboundMethod) if signature
+
+            case method_in_gem?(method)
+            when nil
+              # This means that this is a C-method. Thus, we want to
+              # skip it only if the constant is an ignored one, since
+              # that probably means that we've hit a C-method for a
+              # core type.
+              return if symbol_ignored?(symbol_name)
+            when false
+              # Do not process this method, if it is not defined by the current gem
+              return
+            end
+          rescue SignatureBlockError
+            signature = nil
+          end
 
           method_name = method.name.to_s
           return unless valid_method_name?(method_name)
@@ -747,12 +762,12 @@ module Tapioca
           " " * @indent + str
         end
 
-        sig { params(method: UnboundMethod).returns(T::Boolean) }
+        sig { params(method: UnboundMethod).returns(T.nilable(T::Boolean)) }
         def method_in_gem?(method)
-          source_location = method.source_location&.first
-          return false if source_location.nil?
+          source_location, _ = method.source_location
+          return nil unless source_location
 
-          gem.contains_path?(source_location)
+          source_location == "(eval)" || gem.contains_path?(source_location)
         end
 
         sig { params(constant: Module, strict: T::Boolean).returns(T::Boolean) }
@@ -935,11 +950,13 @@ module Tapioca
           Class.instance_method(:superclass).bind(constant).call
         end
 
+        SignatureBlockError = Class.new(StandardError)
+
         sig { params(method: T.any(UnboundMethod, Method)).returns(T.untyped) }
         def signature_of(method)
           T::Private::Methods.signature_for_method(method)
         rescue LoadError, StandardError
-          nil
+          raise SignatureBlockError
         end
 
         sig { params(constant: T::Types::Base).returns(String) }
--- a/spec/tapioca/compilers/symbol_table_compiler_spec.rb
+++ b/spec/tapioca/compilers/symbol_table_compiler_spec.rb
@@ -2306,16 +2306,10 @@ class Tapioca::Compilers::SymbolTableCompilerSpec < Minitest::HooksSpec
           prop :bar, String
           const :baz, T::Hash[String, T.untyped]
           prop :quux, T.untyped, default: T.unsafe(nil)
-
-          class << self
-            def inherited(s); end
-          end
         end
 
         class Baz
           abstract!
-
-          def initialize(*args, &blk); end
 
           sig { abstract.void }
           def do_it; end
@@ -2452,10 +2446,6 @@ class Tapioca::Compilers::SymbolTableCompilerSpec < Minitest::HooksSpec
           Elem = type_member
 
           const :foo, T.untyped
-
-          class << self
-            def inherited(s); end
-          end
         end
       RBI
 
@@ -2530,10 +2520,6 @@ class Tapioca::Compilers::SymbolTableCompilerSpec < Minitest::HooksSpec
           prop :l, T::Array[Foo], default: T.unsafe(nil)
           prop :m, T::Hash[Foo, Foo], default: T.unsafe(nil)
           prop :n, Foo, default: T.unsafe(nil)
-
-          class << self
-            def inherited(s); end
-          end
         end
       RBI
 


lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -747,12 +762,12 @@ def indented(str)
...
-        sig { params(method: UnboundMethod).returns(T::Boolean) }
+        sig { params(method: UnboundMethod).returns(T.nilable(T::Boolean)) }
         def method_in_gem?(method)
-          source_location = method.source_location&.first
-          return false if source_location.nil?
+          source_location, _ = method.source_location
+          return nil unless source_location
 
-          gem.contains_path?(source_location)
+          source_location == "(eval)" || gem.contains_path?(source_location)
```
vinistock, line 770: When is the source_location equal to (eval)?

vinistock (approved):

lib/tapioca/compilers/symbol_table/symbol_generator.rb
paracycle: Ah, forgot to mention that in the PR description, good catch. That happens when:
```
class Foo
  module_eval("def foo; end")
end

Foo.instance_method(:foo).source_location
# => ["(eval)", 1]
```

Morriar (approved):

egiurleo:
I've written a failing test for Ufuk's case above: https://github.com/Shopify/tapioca/blob/emily/method-gen/spec/tapioca/gem/pipeline_spec.rb#L584

amomchilov:
> Unlike mixin tracking, though, this is much harder to do, since we can't intercept def calls and analyze the backtrace. It won't be enough to intercept define_method calls (even though it will take us some of the way there), since methods can always be defined via class_eval/module_eval/eval using def.

Does hooking [`method_added`](https://rubyapi.org/3.3/o/module#method-i-method_added) solve that? @paracycle

paracycle:
> > Unlike mixin tracking, though, this is much harder to do, since we can't intercept def calls and analyze the backtrace. It won't be enough to intercept define_method calls (even though it will take us some of the way there), since methods can always be defined via class_eval/module_eval/eval using def.
> 
> Does hooking [`method_added`](https://rubyapi.org/3.3/o/module#method-i-method_added) solve that? @paracycle

That is exactly what @egiurleo's PR in the comment above is doing, but in general I am afraid it will be brittle since you need one constant to forget to call `super` from their `method_added` implementation and you never get your one called.

Gemfile
```diff
@@ -7,7 +7,7 @@ gemspec
 CURRENT_RAILS_VERSION = "7.1"
 rails_version = ENV.fetch("RAILS_VERSION", CURRENT_RAILS_VERSION)
 
-gem "minitest"
+gem "minitest", "< 5.25.0" # minitest 5.25.0+ is incompatible with minitest-hooks
```
KaanOzkan, line 10: I think this can be removed https://github.com/minitest/minitest/commit/6d838438ee53897df20b5aef3df35fc291018b9e

paracycle: Sure, I'll clean up this branch quite a lot, but I needed to make sure the tests weren't failing.

lib/tapioca/gem/pipeline.rb
```diff
@@ -126,35 +126,38 @@ def symbol_in_payload?(symbol_name)
...
 
         gem.contains_path?(source_file)
       end
 
-      sig { params(method: UnboundMethod).returns(T::Boolean) }
-      def method_in_gem?(method)
-        source_location = method.source_location&.first
-        return false if source_location.nil?
+      sig { params(method_name: Symbol, owner: Module).returns(T.any([String, Integer], NilClass, T::Boolean)) }
+      def method_definition_in_gem(method_name, owner)
```
amomchilov, line 140: The name is less-than-obvious, but I can't think of a better one. How about we add a doc comment? The return value is particularly tricky to grok

```suggestion
      # @returns - The source file and line number, if this method was defined in this pipeline's gem
                 - `nil` if the source location is unknown
                 - `true` if the method was defined via `eval` without a known file/line number
                 - `false` if this method wasn't defined by this gem
      sig { params(method_name: Symbol, owner: Module).returns(T.any([String, Integer], NilClass, T::Boolean)) }
      def method_definition_in_gem(method_name, owner)
```

lib/tapioca/runtime/reflection.rb
```diff
@@ -33,6 +33,10 @@ module Reflection
       UNDEFINED_CONSTANT = T.let(Module.new.freeze, Module)
 
       REQUIRED_FROM_LABELS = T.let(["<top (required)>", "<main>"].freeze, T::Array[String])
+      # this looks something like:
+      # "(eval at /path/to/file.rb:123)"
+      # and we are just interested in the "/path/to/file.rb" part
+      EVAL_SOURCE_FILE_PATTERN = T.let(/\(eval at (.+):\d+\)/, Regexp)
```
amomchilov, line 39: Can we anchor the start or end of this with `^` and/or `$`?

lib/tapioca/runtime/reflection.rb
```diff
@@ -177,17 +185,49 @@ def descendants_of(klass)
         T.unsafe(result)
       end
 
+      sig { params(constant_name: T.any(String, Symbol)).returns(T.nilable([String, Integer])) }
+      def const_source_location(constant_name)
+        return unless Object.respond_to?(:const_source_location)
+
+        file, line = Object.const_source_location(constant_name)
+
+        # Ruby 3.3 adds automatic definition of source location for evals if
```
amomchilov, line 194: GOOD

lib/tapioca/runtime/reflection.rb
```diff
@@ -177,17 +185,49 @@ def descendants_of(klass)
...
+        # `file` and `line` arguments are not provided. This results in the source
+        # file being something like `(eval at /path/to/file.rb:123)`. We try to parse
+        # this string to get the actual source file.
+        file = file&.sub(EVAL_SOURCE_FILE_PATTERN, "\\1")
```
amomchilov, line 198:
```suggestion
        file = file&.sub(EVAL_SOURCE_FILE_PATTERN, '\1')
```

amomchilov (approved):
Fantastic work!

lib/tapioca/gem/pipeline.rb
```diff
@@ -136,25 +136,34 @@ def constant_in_gem?(name)
...
+        source_file = source_file&.sub(EVAL_SOURCE_FILE_PATTERN, "\\1")
+
+        # If the source location of the constant isn't available or is "(eval)", all bets are off.
+        return true if source_file.nil? || source_file == "(eval)"
 
         gem.contains_path?(source_file)
       end
 
-      sig { params(method: UnboundMethod).returns(T::Boolean) }
+      sig { params(method: UnboundMethod).returns(T.nilable(T::Boolean)) }
```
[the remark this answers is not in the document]
paracycle: Went for the inheritance based solution that you proposed.

lib/tapioca/runtime/reflection.rb
```diff
@@ -184,14 +184,20 @@ def descendants_of(klass)
...
         resolved_loc = locations.find { |loc| REQUIRED_FROM_LABELS.include?(loc.label) }
-        return "" unless resolved_loc
+        return ["", 0] unless resolved_loc
 
-        resolved_loc.absolute_path || ""
+        # Find the location of the last frame in this file to get the most accurate line number.
+        resolved_loc = locations.find { |loc| loc.absolute_path == resolved_loc.absolute_path }
+        return ["", 0] unless resolved_loc
+
+        [resolved_loc.absolute_path || "", resolved_loc.lineno]
```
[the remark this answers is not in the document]
paracycle: This should be addressed with the new `SourceLocation` class that is introduced.

vinistock (approved):
This is great. Two questions:

1. Did we test this in Core? Does it pass type checking afterwards?
2. Do you know if there is a performance hit or memory increase for tracking the methods?

paracycle:
> This is great. Two questions:

Thanks for the review

> 
> 1. Did we test this in Core? Does it pass type checking afterwards?

I triggered the build pipeline for testing against Core now.

> 2. Do you know if there is a performance hit or memory increase for tracking the methods?

There will definitely be a performance and memory hit, but I have no idea how much. We do extra operations for every method definition and we store their locations, etc, so this is certainly not free.

paracycle:
Ok, tested this against Core. The build pipeline was giving a lot of errors. I debugged against Core and found the problem and pushed a fix on this branch.

I also did `time bin/tapioca gem --all` timing for both the version of Tapioca on Core and this branch:

The current Tapioca version on Core:
```
bin/tapioca gem --all  2803.99s user 89.11s system 380% cpu 12:40.88 total
```

This branch:
```
bin/tapioca gem --all  2826.87s user 93.45s system 392% cpu 12:23.26 total
```

So, it seems like we use about the same amount of time to generate all gem RBIs on Core.

I am not sure if the timing diff is real, but if this branch is indeed faster, then I assume that would be because we collect all method definitions once when loading all the gems, and then we always check against that lookup table for any method. That somehow might be faster than looking up the source location for each method as we come across it.

PR #303, opened 2021-05-05
Fix RBI generation for gems

Morriar:
### Motivation

My previous PR on RBI generation (#297) broke two things:

1. Grouping the include and extend in separate groups meant I changed the [order in which they were generated](https://github.com/Shopify/tapioca/commit/2ad440590e747403012cdbba7b49c2baa0aae072#diff-3e55697f4d9bb01a769a76f008853abea864fdb0c8de83956ee0303bfddc1764L1556-R1561) and includes were always printed before extends which is actually incorrect. So the first commit (2bb2a7e) groups them together to get the right order again.

2. Both params types and return types need to be sanitized for weird types such as `<VOID>`. It was done on the [whole signature string before](https://github.com/Shopify/tapioca/commit/d2107da489c440ae655226b33d8fee10a7051688#diff-6c17c60032864d0164ef50591234ad8238992fb3dad7d6ac33158f333ea61190L680-L684) and I forgot to do it on parameters types. The second commit (86b2982) fixes this and adds a related test.

With these fixes we can generate the RBIs for Shopify's core without any error! 🎉 

### Tests

See included tests.

### DIFF
--- a/spec/tapioca/rbi_spec.rb
+++ b/spec/tapioca/rbi_spec.rb
@@ -431,8 +431,8 @@ module Tapioca
           rbi << RBI::Class.new("S2")
           rbi << RBI::Method.new("m1")
           rbi << RBI::Method.new("m2", is_singleton: true)
-          rbi << RBI::Include.new("I")
           rbi << RBI::Extend.new("E")
+          rbi << RBI::Include.new("I")
           rbi << RBI::MixesInClassMethods.new("MICM")
           rbi << RBI::Helper.new("h")
           rbi << RBI::TStructConst.new("SC", "Type")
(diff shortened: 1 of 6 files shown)


spec/tapioca/rbi_spec.rb
```diff
@@ -431,8 +431,8 @@ def self.m2; end
           rbi << RBI::Class.new("S2")
           rbi << RBI::Method.new("m1")
           rbi << RBI::Method.new("m2", is_singleton: true)
-          rbi << RBI::Include.new("I")
```
Morriar, line 434: Just changed the order in a few tests to show that we keep it when compiling.

paracycle (approved):
Yeah, the `include`/`extend` ordering is tricky. My first implementation had also completely messed it up. :)

PR #307, opened 2021-05-10
Add helper method underscore

### LINKED ISSUE #286: Tapioca assumes ActiveSupport string core extensions are present
I have a plan ol' Ruby project (no Rails) that has Sidekiq workers. I'd love to use Tapiocal to generate the RBI files for the Sidekiq workers. However,  I'm running into an error because it's assumed `String` has the  ActiveSupport string extension method `underscore` .

For example, there are two uses of `underscore` in `lib/tapioca/generator.rb`.

```ruby
def dsl_rbi_filename(constant_name)
  config.outpath / "#{constant_name.underscore}.rbi"
end
```

```ruby
def compile_dsl_rbi(constant_name, contents, outpath: config.outpath, quiet: false)
  return if contents.nil?

  rbi_name = constant_name.underscore + ".rbi"

  ...
end
```

vinistock:
Closes #286

### Motivation

We can't depend on active_support being available, so I moved our spec implementation of `underscore` into a helper method so that we can use it in projects that do not include active_support.

I considered adding it as a core extension to string, but I'm not a fan of that approach as it could potentially lead to mistakes. For example, a project including Tapioca could use `underscore` in `String` during development and it would work fine, but then would break outside of a development environment, where Tapioca is not included. This is why I made it a class method.

### Implementation

Move the implementation of `underscore` into a helper class method inside `Generator`. Please, let me know if we have a better place for utilities such as this one.

### Tests

This is a refactor. Existing tests already cover it.

### DIFF
--- a/lib/tapioca/generator.rb
+++ b/lib/tapioca/generator.rb
@@ -3,6 +3,7 @@
 
 require 'pathname'
 require 'thor'
+require "tapioca/core_ext/string"
 
 module Tapioca
   class Generator < ::Thor::Shell::Color
@@ -285,7 +286,7 @@ module Tapioca
     def constantize(constant_names)
       constant_map = constant_names.map do |name|
         begin
-          [name, name.constantize]
+          [name, Object.const_get(name)]
         rescue NameError
           [name, nil]
         end
--- a/lib/tapioca/core_ext/string.rb
+++ b/lib/tapioca/core_ext/string.rb
@@ -1,0 +1,18 @@
+# typed: strict
+# frozen_string_literal: true
+
+class String
+  extend T::Sig
+
+  sig { returns(String) }
+  def underscore
+    return self unless /[A-Z-]|::/.match?(self)
+
+    word = to_s.gsub("::", "/")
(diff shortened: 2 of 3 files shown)


lib/tapioca/generator.rb
```diff
@@ -11,6 +11,18 @@ class Generator < ::Thor::Shell::Color
     sig { returns(Config) }
     attr_reader :config
 
+    sig { params(camel_cased_word: String).returns(String) }
+    def self.underscore(camel_cased_word)
```
paracycle, line 15: Can we instead move this to an actual `String` extension inside the `core_ext` folder? We'd done the same for `descendants`, so that we don't have to change the use sites.

paracycle (commented):
We will also need a similar treatment for `constantize` or we need to replace it with `const_get` instead.

lib/tapioca/generator.rb
vinistock: Ah, perfect! So, we won't be polluting the core classes. I'll make the change!

Morriar (approved):

paracycle (approved):

PR #308, opened 2021-05-11
Resolve security alerts

mutecipher:
### Motivation

Ran `bundle update rails` to resolve the following security advisories:
* CVE-2021-22885,
* CVE-2021-22880,
* GHSA-vr8q-g5c7-m54m, and
* GHSA-cfjv-5498-mph5.

### Implementation

Ran `bundle update rails` which updated related dependencies. Needed to go back and lock Nokogiri at v1.10.10 as it was bumped to v1.11.0 which no longer supports Ruby 2.4.

### Tests

No changes to tests, but current tests continue to pass.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -34,7 +34,7 @@ group(:development, :test) do
   # Fix version to 0.14.1 since it is the last version to support Ruby 2.4
   gem("shopify-money", "= 0.14.1", require: false)
   gem("sidekiq", "~>5.0", require: false) # Version 6 dropped support for Ruby 2.4
-  gem("nokogiri", "= 1.10.10", require: false) # Lock to last supported for Ruby 2.4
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -34,6 +34,7 @@ group(:development, :test) do
   # Fix version to 0.14.1 since it is the last version to support Ruby 2.4
   gem("shopify-money", "= 0.14.1", require: false)
   gem("sidekiq", "~>5.0", require: false) # Version 6 dropped support for Ruby 2.4
+  gem("nokogiri", "= 1.10.10", require: false) # Lock to last supported for Ruby 2.4
```
rafaelfranca, line 37:
```suggestion
  gem("nokogiri", "1.10.10", require: false) # Lock to last supported for Ruby 2.4
```

PR #309, opened 2021-05-11
Update minimum supported Ruby versions

mutecipher:
Signed-off-by: Cory Hutchison <cory.hutchison@shopify.com>

### Motivation

Ruby 2.4 support has ended as of Apr 5, 2020. As well, Ruby 2.5 support
has ended as of the end of March 2021. So this PR removes support for
those two versions and updates the related CI pipelines to test against
the currently supported versions:

2.6, 2.7 and 3.0

### Implementation

Update `tapioca.gemspec` to specify a minimum Ruby version of `>= 2.6` and update the GitHub workflow.

### Tests

No specific changes to tests required.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [ 2.4, 2.5, 2.6, 2.7 ]
+        ruby: [ 2.6, 2.7, 3.0 ]
     name: Test Ruby ${{ matrix.ruby }}
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


rafaelfranca (approved):

vinistock (approved):

rafaelfranca:
While I support removing old rubies it seems sorbet is still supporting 2.4. I think we need to make a decision based in the cost. Can you create a branch that keep support to 2.4 and 2.5 to see how hard it is to support both?

mutecipher:
Existing PR already in place (https://github.com/Shopify/tapioca/pull/203). Gonna piggy back off it. Also, going to split up some of the work into separate PRs.

PR #326, opened 2021-05-25
Fix ActiveJob generator

paracycle:
### Motivation

For Active Job classes, we were generating the signatures wrong in two ways:

1. `perform_later` returns an instance of the job class, or [`false` if the job couldn't be enqueued](https://github.com/rails/rails/blob/5aaaa1630ae9a71b3c3ecc4dc46074d678c08d67/activejob/lib/active_job/enqueuing.rb#L68).
2. `perform_now` returns the return value of `perform` immediately. It does not return an instance of the job class.

### Implementation

Make relevant updates.

### Tests

Updated existing tests.


kddeisz (approved):
Wow, TIL

vinistock (approved):

PR #329, opened 2021-05-25
Serialize generic types with unspecified type variables

paracycle:
### Motivation

The generic serialization code is serializing generic types with unspecified type variables incorrectly. Sorbet currently allows one to type `Foo.new` even if `Foo` is a generic type that expects type variables, and treats `Foo.new` as if it was `Foo[T.untyped].new` or similar.

However, Sorbet complains about `T.let(T.unsafe(nil), Foo)` usage, with the error message saying that `Foo` is a generic class without type arguments.

Thus, if the user code has a constant definition like `FOO = Foo.new` (where `Foo` is a generic class), then the generated RBI code for it, `FOO = T.let(T.unsafe(nil), Foo)` would end up raising a type checking error.

### Implementation

The fix is to treat generic types differently when grabbing their name. If their name does not already have the type variables, then we add `T.untyped` for each of the type variables it would expect and generate the type name like that.

### Tests

Added an extra test to check for this specific case.

### DIFF
--- a/lib/tapioca/compilers/symbol_table/symbol_generator.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_generator.rb
@@ -145,7 +145,15 @@ module Tapioca
         def compile_object(tree, name, value)
           return if symbol_ignored?(name)
           klass = class_of(value)
-          klass_name = name_of(klass)
+
+          klass_name = if klass == ObjectSpace::WeakMap
+            # WeakMap is an implicit generic with one type variable
+            "ObjectSpace::WeakMap[T.untyped]"
+          elsif T::Generic === klass
+            generic_name_of(klass)
+          else
+            name_of(klass)
+          end
 
           if klass_name == "T::Private::Types::TypeAlias"
             tree << RBI::Const.new(name, "T.type_alias { #{T.unsafe(value).aliased_type} }")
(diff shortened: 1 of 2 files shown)


lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -145,7 +145,15 @@ def compile_alias(tree, name, constant)
         def compile_object(tree, name, value)
           return if symbol_ignored?(name)
           klass = class_of(value)
-          klass_name = name_of(klass)
+
+          klass_name = if klass == ObjectSpace::WeakMap
+            # WeakMap is an implicit generic with one type variable
+            "ObjectSpace::WeakMap[T.untyped]"
```
Morriar, line 151: Why can't we treat this one like the other classes extending `T::Generic`? https://github.com/sorbet/sorbet/blob/master/rbi/stdlib/objspace.rbi#L170-L171

paracycle: As you point out, the fact that this is generic is encoded in the RBI file, but no artifact of that exists at runtime.

vinistock (approved):

lib/tapioca/compilers/symbol_table/symbol_generator.rb
Morriar: I wonder if such a list should exist in `sorbet-runtime` so it can "tag" known generic types?

paracycle: TBH, there aren't that many of them, and, IMO, `ObjectSpace::WeakMap` shouldn't be marked as a generic type anyway. I'll see what upstream thinks about making it non-generic.

Morriar: I think `sorbet-runtime` could leave some info in the instance about this so we can pick it up from `tapioca`

PR #330, opened 2021-05-25
Treat default gems properly

paracycle:
### Motivation

If the version of a default gem specified in the `Gemfile` is the version of that default gem that is bundled with the Ruby version, then Bundler just uses the default gem from the Ruby library directory.

But Bundler still acts like the gem is installed in where it would have been, which prevents Tapioca from generating a proper RBI for the gem and we end up generating an empty gem.

### Implementation

This PR makes sure that if a gem is marked as a default gem, we check against the files listed in its gemspec inside the Ruby library folder instead of looking inside what the gem install folder would have been.

### Tests

No added tests.

### DIFF
--- a/lib/tapioca/gemfile.rb
+++ b/lib/tapioca/gemfile.rb
@@ -108,8 +108,14 @@ module Tapioca
 
       sig { returns(T::Array[Pathname]) }
       def files
-        @spec.full_require_paths.flat_map do |path|
-          Pathname.glob((Pathname.new(path) / "**/*.rb").to_s)
+        if default_gem?
+          @spec.files.map do |file|
+            ruby_lib_dir.join(file)
+          end
+        else
+          @spec.full_require_paths.flat_map do |path|
+            Pathname.glob((Pathname.new(path) / "**/*.rb").to_s)
+          end
         end
       end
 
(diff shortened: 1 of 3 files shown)


lib/tapioca/gemfile.rb
```diff
@@ -125,11 +131,25 @@ def rbi_file_name
...
+        else
+          to_realpath(path).start_with?(full_gem_path) || has_parent_gemspec?(path)
+        end
       end
 
       private
 
+      sig { returns(T::Boolean) }
+      def default_gem?
+        @spec.respond_to?(:default_gem?) && @spec.default_gem?
```
vinistock, line 145: Is the `respond_to?` for older versions of Bundler? When would it not respond to `default_gem?`?

vinistock (approved):

lib/tapioca/gemfile.rb
paracycle: I think so. To be perfectly honest, I cargo-culted it from [some of the Bundler code](https://github.com/rubygems/rubygems/blob/96e5cff3df491c4d943186804c6d03b57fcb459b/bundler/lib/bundler/rubygems_integration.rb#L94) that was doing similar checks so I assumed it could be an older version Rubygems, maybe.

PR #331, opened 2021-05-25
Suppress call validation errors during CLI run

paracycle:
### Motivation

The way we modify generic types means we sometimes are not able to properly validate inline checks or call signatures. For that reason, Tapioca tries to run with most stuff turned off.

However, there is a call validation that happens inside `T::Props::Private::SetterFactory#build_setter_proc` which does not respect the `default_checked_level` value.

### Implementation

To suppress those errors from being raised at runtime in production, we need to explicitly suppress call validation errors by setting a no-op handler.

### Tests

No tests.

### DIFF
--- a/exe/tapioca
+++ b/exe/tapioca
@@ -5,6 +5,8 @@ require 'sorbet-runtime'
 
 begin
   T::Configuration.default_checked_level = :never
+  # Suppresses call validation errors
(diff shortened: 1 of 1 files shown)


ttonelli (commented):
thanks!

PR #334, opened 2021-05-27
Add attributes accessors to RBI model

Morriar:
### Motivation

Sorbet's RBIs support attributes definitions.

This PR adds them to the RBI model so we can actually generate them if needed (and later parse them when used in shims).

### Tests

See automated tests.

### DIFF
--- a/spec/tapioca/rbi/printer_spec.rb
+++ b/spec/tapioca/rbi/printer_spec.rb
@@ -58,6 +58,21 @@ module Tapioca
           RBI
         end
 
+        it("builds attributes") do
+          rbi = RBI::Tree.new
+          rbi << RBI::AttrReader.new(:m1)
+          rbi << RBI::AttrWriter.new(:m2, :m3, visibility: Visibility::Public)
+          rbi << RBI::AttrAccessor.new(:m4, visibility: Visibility::Private)
+          rbi << RBI::AttrReader.new(:m5, visibility: Visibility::Protected)
+
+          assert_equal(<<~RBI, rbi.string)
+            attr_reader :m1
+            attr_writer :m2, :m3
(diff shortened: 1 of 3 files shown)


KaanOzkan (approved):

vinistock (approved):

spec/tapioca/rbi/printer_spec.rb
```diff
@@ -58,6 +58,21 @@ class << self; end
...
+          rbi = RBI::Tree.new
+          rbi << RBI::AttrReader.new(:m1)
+          rbi << RBI::AttrWriter.new(:m2, :m3, visibility: Visibility::Public)
+          rbi << RBI::AttrAccessor.new(:m4, visibility: Visibility::Private)
+          rbi << RBI::AttrReader.new(:m5, visibility: Visibility::Protected)
+
+          assert_equal(<<~RBI, rbi.string)
+            attr_reader :m1
+            attr_writer :m2, :m3
+            private attr_accessor :m4
```
paracycle, line 71: Umm, [this will result in a type check error](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20Foo%0A%20%20private(attr_accessor(%3Am4))%0Aend%0A), though, since `attr_reader` returns `T.untyped` but `private` expects `T.any(String, Symbol)`. 

Also, I believe this usage will only be valid on Ruby 3.1: https://bugs.ruby-lang.org/issues/17314

The best way to generate private/protected attributes is in a `private` / `public` section.

Morriar: If I want to be able to display some error/warning about it, I need the node in the tree with its location.

PR #335, opened 2021-05-27
Add `checked` signature builder

Morriar:
### Motivation

A signature accept the sig builder `checked`.

This PR adds it to the RBI model so we can generate them (and later parse them).

### Tests

See automated tests.


KaanOzkan (approved):

vinistock (approved):

Morriar:
> I have a question about this: Why would anyone add a `checked` attribute to an RBI file? `checked` is about the runtime behaviour and RBI files are totally ignored at runtime.

If I want to be able to display some error/warning about it, I need the node in the tree with its location.

PR #337, opened 2021-05-27
Introduce RBI parsing

Morriar:
### Motivation

This PR introduce `RBI::Parser`, a way to parse a RBI string into `RBI::Node`s:

```rb
rbi = <<~RBI
  module Foo
    def foo; end
  end
RBI

tree = RBI::Parser.parse_string(rbi)
```

The `tree` can then be visited/manipulated/printed as any other `RBI::Node` hierarchy.

For example, `puts tree.string(print_locs: true)` will output:

```rb
# -:1:0-3:3
module Foo
  # -:2:2-2:14
  def foo; end
end
```

### Tests

See automated tests.


lib/tapioca/rbi/model.rb
```diff
@@ -468,6 +468,11 @@ def initialize(names = [], loc: nil, comments: [])
       def empty?
         names.empty?
       end
+
+      sig { params(name: String).void }
```
RyanBrushett, line 472: Should this return `T::Array[String]`?

RyanBrushett (approved):

KaanOzkan (approved):

PR #339, opened 2021-05-28
Drop Ruby 2.4 support

Morriar:
### Motivation

Support for Ruby 2.4 has ended [more than a year ago](https://www.ruby-lang.org/en/news/2020/04/05/support-of-ruby-2-4-has-ended/) we're not doing any good by keeping it alive.

Furthermore, it's now blocking active development on `tapioca`: 

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


Morriar:
#320 seems to be doing way more than removing support for 2.4?

It's open for 11 days, I believe we can merge this one then rebase the other on top of it. Let's keep the ball rolling.

PR #350, opened 2021-06-10
Allow to print a RBI node by its qualified name / to_s representation

Morriar:
### Motivation

Right now when you work with a RBI tree, if you print a node to know what you're dealing with, all you get is the default Ruby `to_s` implementation like `#<Tapioca::RBI::Class:0xXXXXXX>` or `#<Tapioca::RBI::Module:0xXXXXXX>` which is not super user-friendly (especially if you're debugging).

This PR adds various printing methods to nodes so when you do `puts my_rbi_node` you get `::Foo::Bar` instead of `#<Tapioca::RBI::Class:0xXXXXXX>`.

This is particularly useful if you want to use a node in a log message or error without having to visit it's content recursively with the tree printer.

### Implementation

1. Add `fully_qualified_name` for uniquely identifiable nodes such as `Module`, `Class`, `Const`, `Method`
2. Add `fully_qualifed_names` for nodes referring to multiple entities such as `attr_reader :a, :b`
3. Add `to_s` to all nodes

Depending on the nodes `fully_qualified_name(s)` and `to_s` behave differently:

```rb
mod = RBI::Module.new("Foo")
puts mod.fully_qualified_name # ::Foo::bar
puts mod.to_s # ::Foo::Bar (same as `fully_qualified_node`)

attr = RBI::AttrAccessor.new(:a1, :a2)
mod << attr
puts attr.fully_qualified_names # ::Foo#a1, ::Foo#a1=, ::Foo#a2, ::Foo#a2=
puts attr.to_s # ::Foo.attr_accessor(:a1, :a2)
```

### Tests

See included tests.

### DIFF
--- a/lib/tapioca/rbi/model.rb
+++ b/lib/tapioca/rbi/model.rb
@@ -28,6 +28,13 @@ module Tapioca
         tree.nodes.delete(self)
         self.parent_tree = nil
       end
+
+      sig { returns(T.nilable(Scope)) }
+      def parent_scope
+        parent = T.let(parent_tree, T.nilable(Tree))
+        parent = parent.parent_tree until parent.is_a?(Scope) || parent.nil?
+        parent
+      end
     end
 
     class Comment < Node
@@ -89,6 +96,9 @@ module Tapioca
       extend T::Helpers
 
       abstract!
+
+      sig { abstract.returns(String) }
+      def fully_qualified_name; end
     end
 
     class Module < Scope
@@ -101,6 +111,12 @@ module Tapioca
       def initialize(name, loc: nil, comments: [])
         super(loc: loc, comments: comments)
         @name = name
+      end
+
+      sig { override.returns(String) }
+      def fully_qualified_name
+        return name if name.start_with?("::")
+        "#{parent_scope&.fully_qualified_name}::#{name}"
       end
     end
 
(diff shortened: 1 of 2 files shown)


Morriar:
This will need [this](https://github.com/sorbet/sorbet/pull/4261) to be merged on Sorbet first

lib/tapioca/rbi/model.rb
```diff
@@ -28,6 +28,16 @@ def detach
...
       end
+
+      sig { returns(T.nilable(Scope)) }
+      def parent_scope
+        parent = T.let(parent_tree, T.nilable(Tree))
+        while parent
+          return parent if parent.is_a?(Scope)
+          parent = parent.parent_tree
+        end
+        nil
```
vinistock, line 39: Suggestion:
```ruby
parent = T.let(parent_tree, T.nilable(Tree))
parent = parent.parent_tree until parent.is_a?(Scope) || parent.nil?
parent
```

lib/tapioca/rbi/model.rb
```diff
@@ -186,11 +232,58 @@ def initialize(name, *names, visibility: Visibility::Public, sigs: [], loc: nil,
...
+      def fully_qualified_names; end
+    end
+
+    class AttrAccessor < Attr
+      extend T::Sig
+
+      sig { override.returns(T::Array[String]) }
+      def fully_qualified_names
+        parent_name = parent_scope&.fully_qualified_name
+        names.map { |name| ["#{parent_name}##{name}", "#{parent_name}##{name}="] }.flatten
```
vinistock, line 246: Instead of using `map` and then `flatten`, simply use `flat_map`.

```ruby
names.flat_map { |name| ["#{parent_name}##{name}", "#{parent_name}##{name}="] }
```

vinistock (approved):

PR #364, opened 2021-06-19
Support `tapioca sync --verify`

jeffcarbs:
### Motivation

`tapioca dsl` has a `--verify` flag that will "exit 1" if your DSL RBI is not up-to-date. This makes it easy to add a CI check to prevent stale RBI. We're starting to use `tapioca sync` and I found myself wanting something similar to ensure gem RBI is up-to-date.

### Implementation

Followed a similar pattern as verification for DSL.

### Tests

See added test cases

### DIFF
--- a/lib/tapioca/generator.rb
+++ b/lib/tapioca/generator.rb
@@ -600,7 +600,7 @@ module Tapioca
     def perform_dsl_verification(dir)
       diff = verify_dsl_rbi(tmp_dir: dir)
 
-      report_diff_and_exit_if_out_of_date(diff, 'dsl')
+      report_diff_and_exit_if_out_of_date(diff, "dsl")
     ensure
       FileUtils.remove_entry(dir)
     end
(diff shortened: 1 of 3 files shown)


lib/tapioca/generator.rb
```diff
@@ -624,5 +631,37 @@ def purge_stale_dsl_rbi_files(files)
...
+        say("Once it is complete, be sure to commit and push any changes", :green)
+
+        say("")
+
+        say("Reason:", [:red])
+        diff.group_by(&:last).sort.each do |cause, diff_for_cause|
+          say(build_error_for_files(cause, diff_for_cause.map(&:first)))
+        end
+
+        exit(1)
```
jeffcarbs, line 663: This is nearly identical to a block of code in the DSL, only difference is the suggested command (`sync` here, `dsl` in the DSL one). I considered generalizing to a separate method but decided against it to keep the diff smaller.

Morriar: Very similar to [`verify_dsl_rbi`](https://github.com/Shopify/tapioca/blob/b8e7ceeaead2df91b847051e4c8d8b35dbf842d4/lib/tapioca/generator.rb#L545) indeed. Can we extract it as a helper so it can be shared instead of duplicating it?

paracycle (approved):
Thank you, this looks great!

Morriar (approved):

PR #368, opened 2021-06-22
Add block builders to RBI trees and scopes

Morriar:
### Motivation

Make the creation of nested RBI trees and scopes easier.

Let's try to build this RBI string:

```rb
module Foo
  class Bar
    class Baz < Bar
      class << self; end
    end
  end
end
```

Before this PR, the only way was to save the reference of the created node to reuse it:

```rb
scope1 = RBI::Module.new("Foo")
scope2 = RBI::Class.new("Bar")
scope3 = RBI::Class.new("Baz", superclass_name: "Bar")
scope4 = RBI::SingletonClass.new

rbi = RBI::Tree.new
rbi << scope1
scope1 << scope2
scope2 << scope3
scope3 << scope4
```

With this PR, we can now use the block builder syntax:

```rb
rbi = RBI::Tree.new do |tree|
  tree << RBI::Module.new("Foo") do |mod|
    mod << RBI::Class.new("Bar") do |klass1|
      klass1 << RBI::Class.new("Baz", superclass_name: "Bar") do |klass2|
        klass2 << RBI::SingletonClass.new
      end
    end
  end
end
```

### Tests

See automated tests.

### DIFF
--- a/lib/tapioca/rbi/model.rb
+++ b/lib/tapioca/rbi/model.rb
@@ -72,10 +72,17 @@ module Tapioca
       sig { returns(T::Array[Node]) }
       attr_reader :nodes
 
-      sig { params(loc: T.nilable(Loc), comments: T::Array[Comment]).void }
-      def initialize(loc: nil, comments: [])
+      sig do
+        params(
+          loc: T.nilable(Loc),
+          comments: T::Array[Comment],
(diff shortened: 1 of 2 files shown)


vinistock (approved):
Nice

KaanOzkan (approved):

PR #369, opened 2021-06-23
Handle identity_cache aliases

### LINKED ISSUE #322: Identity cache attribute aliases are ignored by DSL generator
When adding a cache attribute with Identity cache, we ignore the alias and generate incorrect methods for the RBIs.

```ruby
class Post
  cache_attribute :author, by: :id
end
```

Should generate
```ruby
class Post
  def self.fetch_multi_author_by_id(id); end
end
```

But actually generates
```ruby
class Post
  def self.fetch_multi_by_id(id); end
end
```

vinistock:
Closes #322 

### Motivation

Identity cache provides two mechanisms for caching: `cache_index` and `cache_attribute`.

When invoking [cache_index](https://github.com/Shopify/identity_cache/blob/01e90ed9fffe960ff96e9e54eaa7a03fa47a4d85/lib/identity_cache/with_primary_index.rb#L55), it invokes `cache_attribute_by_alias` which will create the alias methods for `cache_index` that we are not currently generating. After that, `cache_index` generates other methods of its own, which are not created by `cache_attribute`.

When invoking [cache_attribute](https://github.com/Shopify/identity_cache/blob/771b0f98324274c8fd79bf2d2094ca9545f3af45/lib/identity_cache/configuration_dsl.rb#L124), it only invokes `cache_attribute_by_alias` and generates the methods with aliases. These [methods](https://github.com/Shopify/identity_cache/blob/771b0f98324274c8fd79bf2d2094ca9545f3af45/lib/identity_cache/cached/attribute_by_one.rb#L16) do not have the `include` argument.

The signatures are different (the alias methods do not have the `include` argument), but they all add entries to the cache variables that keep track of attributes.

### TL;DR

- Both `cache_index` and `cache_attribute` generate the alias methods which do not contain the `include` argument
- `cache_index` generates extra methods (with the `include` argument), which should not be added to the RBI for `cache_attribute`

The challenge is telling which method call create the cached attribute.

### Implementation

I added a new method to create the alias methods and added a few checks to the existing `create_fetch_by_methods`.

I couldn't think of an easy way of knowing if the cached attribute came from `cache_index` or from `cache_attribute`, so I added safe guards to not generate the methods in the RBIs if the `constant` does not actually respond to them. Please, let me know if you have a better suggestion on how to identify these.

### Tests

See included tests.

### DIFF
--- a/lib/tapioca/active_record_column_type_helper.rb
+++ b/lib/tapioca/active_record_column_type_helper.rb
@@ -1,21 +1,19 @@
 # typed: strict
 # frozen_string_literal: true
 
-module ColumnTypeHelper
+class ActiveRecordColumnTypeHelper
   extend T::Sig
 
-  private
-
-  sig do
-    params(
-      constant: T.class_of(ActiveRecord::Base),
-      column_name: String
-    ).returns([String, String])
+  sig { params(constant: T.class_of(ActiveRecord::Base)).void }
+  def initialize(constant)
+    @constant = constant
   end
-  def type_for(constant, column_name)
-    return ["T.untyped", "T.untyped"] if do_not_generate_strong_types?(constant)
 
-    column_type = constant.attribute_types[column_name]
+  sig { params(column_name: String).returns([String, String]) }
+  def type_for(column_name)
+    return ["T.untyped", "T.untyped"] if do_not_generate_strong_types?(@constant)
+
+    column_type = @constant.attribute_types[column_name]
 
     getter_type =
       case column_type
@@ -41,14 +39,14 @@ module ColumnTypeHelper
         handle_unknown_type(column_type)
       end
 
-    column = constant.columns_hash[column_name]
+    column = @constant.columns_hash[column_name]
     setter_type = getter_type
 
     if column&.null
       return ["T.nilable(#{getter_type})", "T.nilable(#{setter_type})"]
     end
 
-    if column_name == constant.primary_key ||
+    if column_name == @constant.primary_key ||
         column_name == "created_at" ||
         column_name == "updated_at"
       getter_type = "T.nilable(#{getter_type})"
@@ -56,6 +54,8 @@ module ColumnTypeHelper
 
     [getter_type, setter_type]
   end
+
+  private
 
   sig { params(constant: Module).returns(T::Boolean) }
   def do_not_generate_strong_types?(constant)
--- a/lib/tapioca/compilers/dsl/identity_cache.rb
+++ b/lib/tapioca/compilers/dsl/identity_cache.rb
@@ -166,11 +166,10 @@ module Tapioca
         def create_fetch_by_methods(field, klass, constant)
           field_length = field.key_fields.length
           fields_name = field.key_fields.join("_and_")
-          suffix = field.send(:fetch_method_suffix)
           is_cache_index = field.instance_variable_defined?(:@attribute_proc)
 
           # Both `cache_index` and `cache_attribute` generate aliased methods
-          create_aliased_fetch_by_methods(klass, constant, field.alias_name.to_s, field_length, suffix)
+          create_aliased_fetch_by_methods(field, klass, constant)
 
           # If the method used was `cache_index` a few extra methods are created
           if is_cache_index
@@ -219,16 +218,16 @@ module Tapioca
 
         sig do
           params(
+            field: T.untyped,
             klass: Parlour::RbiGenerator::Namespace,
             constant: T.class_of(::ActiveRecord::Base),
-            alias_name: String,
-            length: Integer,
-            suffix: String,
           ).void
         end
-        def create_aliased_fetch_by_methods(klass, constant, alias_name, length, suffix)
-          type, _ = type_for(constant, alias_name)
+        def create_aliased_fetch_by_methods(field, klass, constant)
+          type, _ = type_for(constant, field.alias_name.to_s)
           multi_type = type.delete_prefix("T.nilable(").delete_suffix(")")
+          length = field.key_fields.length
+          suffix = field.send(:fetch_method_suffix)
 
           if length == 1
             klass.create_method(
--- a/lib/tapioca/compilers/dsl/identity_cache.rb
+++ b/lib/tapioca/compilers/dsl/identity_cache.rb
@@ -2,6 +2,7 @@
 # frozen_string_literal: true
 
 require "parlour"
+require "tapioca/column_type_helper"
 
 begin
   require "rails/railtie"
@@ -64,6 +65,7 @@ module Tapioca
       # ~~~
       class IdentityCache < Base
         extend T::Sig
+        include ColumnTypeHelper
 
         COLLECTION_TYPE = T.let(
           ->(type) { "T::Array[::#{type}]" },
@@ -168,7 +170,7 @@ module Tapioca
           is_cache_index = field.instance_variable_defined?(:@attribute_proc)
 
           # Both `cache_index` and `cache_attribute` generate aliased methods
-          create_aliased_fetch_by_methods(klass, field_length, suffix)
+          create_aliased_fetch_by_methods(klass, constant, field.alias_name.to_s, field_length, suffix)
 
           # If the method used was `cache_index` a few extra methods are created
           if is_cache_index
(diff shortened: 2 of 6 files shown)


paracycle:
Hmm, I am not sure why we need to do extra work. When `cache_attribute_by_alias` is called from `cache_index` it is passed a proc, but when it is called from `cache_attribute` it is passed an attribute. We can use that to discriminate between the two types of methods that need to be generated.

lib/tapioca/compilers/dsl/identity_cache.rb
```diff
@@ -177,7 +178,63 @@ def create_fetch_by_methods(field, klass, constant)
...
+        end
+
+        sig do
+          params(
+            field: T.untyped,
+            klass: Parlour::RbiGenerator::Namespace,
+            constant: T.class_of(::ActiveRecord::Base),
+          ).void
+        end
+        def create_aliased_fetch_by_methods(field, klass, constant)
```
paracycle, line 222: We shouldn't really need a new method, right? Since the library just generates the same kind of attributes for both the `cache_index` and `cache_attribute` calls. We just weren't handling the aliases set in those attributes properly before, from what I understand.

vinistock: It was not just a matter of handling the alias name, we were ignoring all methods created by `cache_attribute_by_alias`.

`cache_index` generates extra methods [here](https://github.com/Shopify/identity_cache/blob/01e90ed9fffe960ff96e9e54eaa7a03fa47a4d85/lib/identity_cache/with_primary_index.rb#L62-L92) which are not generated by `cache_attribute`. These are  the ones we were already generating in the RBIs.

However, we were completely ignoring the methods generated by `cache_attribute_by_alias` [here](https://github.com/Shopify/identity_cache/blob/771b0f98324274c8fd79bf2d2094ca9545f3af45/lib/identity_cache/cached/attribute_by_one.rb#L16-L25).

Also, the extra methods created by `cache_index` have a different signature.

vinistock:
@paracycle where is the information of whether a proc or attribute was used to generate the method available inside the DSL generator?

paracycle:
> @paracycle where is the information of whether a proc or attribute was used to generate the method available inside the DSL generator?

https://github.com/Shopify/identity_cache/blob/771b0f98324274c8fd79bf2d2094ca9545f3af45/lib/identity_cache/cached/attribute.rb#L11-L15

paracycle:
For example, the following class definition:
```ruby
class Foo < ActiveRecord::Base
  include IdentityCache  
  cache_index :foo  
  cache_attribute :bar  
end  
```
generates `Foo.cache_indexes` as:
```ruby
[
#<IdentityCache::Cached::AttributeByOne:0x00007fc5b830c5e8
  @alias_name=:id,
  @attribute_proc=
   #<Proc:0x00007fc5b830c7a0 /Users/ufuk/.gem/ruby/3.0.1/gems/identity_cache-1.1.0/lib/identity_cache/with_primary_index.rb:56 (lambda)>,
  @key_field=:foo,
  @key_fields=[:foo],
  @model=Foo (call 'Foo.connection' to establish a connection),
  @unique=false>,
#<IdentityCache::Cached::AttributeByOne:0x00007fc5b831f5a8
  @alias_name=:bar,
  @attribute=:bar,
  @key_field=:id,
  @key_fields=[:id],
  @model=Foo (call 'Foo.connection' to establish a connection),
  @unique=true>
]
```

Note how the attribute for `:bar` has an `@attribute` ivar, but the one for `:foo` has an `@attribute_proc` one.

lib/tapioca/compilers/dsl/identity_cache.rb
paracycle: Ok, in that case, this method should generate only the methods that `cache_attribute_by_alias` would have generated and nothing more.

paracycle:
Actually for each entry in `cache_indexes` we can get them to tell us what method we should be generating:
```ruby
Foo.cache_indexes.map { |cached_attribute| "fetch_" + cached_attribute.send(:fetch_method_suffix) }
# =>
```
after I've also added a `cache_attribute :zot, by: [:zopf, :ding]` definition to `Foo`

paracycle:
And this gets you the original `cache_index` calls:
```ruby
Foo.cache_indexes.select { |cached_attribute| cached_attribute.instance_variable_defined?(:@attribute_proc) }
# => [#<IdentityCache::Cached::AttributeByOne:0x00007fc5b830c5e8
#  @alias_name=:id,
#  @attribute_proc=
#   #<Proc:0x00007fc5b830c7a0 /Users/ufuk/.gem/ruby/3.0.1/gems/identity_cache-1.1.0/lib/identity_cache/with_primary_index.rb:56 (lambda)>,
#  @key_field=:foo,
#  @key_fields=[:foo],
#  @model=Foo (call 'Foo.connection' to establish a connection),
#  @unique=false>]
```

vinistock:
I cleaned up the implementation and used the suggestion to determine whether we are coming from a `cache_index` or `cache_attribute`.

Question: for `cache_attribute`, is there a way of finding out the type of the field from within Tapioca if we previously generated the RBI for it?

For example, if we use `cache_attribute` to cache a column that is a `String`, is there a way of getting that type from within this generator?

Right now, the return types for alias methods are a bit limited. If their suffix starts with `id`, then it was generated alongside with `cache_index` and we know the return type is `Integer`. For any other attribute, we would need the information from the database column, so I currently put `T.untyped`.

vinistock:
I extracted the logic for determining the type of a column into a helper module that I'm now including on the two DSL generators. We can now get the types for the attributes 👍.

lib/tapioca/compilers/dsl/identity_cache.rb
```diff
@@ -164,46 +166,90 @@ def create_fetch_field_methods(field, klass, returns_collection:)
...
+              klass.create_method(
+                "fetch_multi_by_#{fields_name}",
+                class_method: true,
+                parameters: [
+                  Parlour::RbiGenerator::Parameter.new("index_values", type: "T.untyped"),
+                  Parlour::RbiGenerator::Parameter.new("includes:", default: "nil", type: "T.untyped"),
+                ],
+                return_type: COLLECTION_TYPE.call(constant)
+              )
+            end
```
paracycle, line 216: I think it would be great if this was encapsulated in a `create_index_fetch_by_methods` method.

lib/tapioca/compilers/dsl/identity_cache.rb
```diff
@@ -64,6 +65,7 @@ module Dsl
       # ~~~
       class IdentityCache < Base
         extend T::Sig
+        include ColumnTypeHelper
```
paracycle, line 68: Can we use composition instead of inheritance here? Seen as we are extracting this out to some reusable code, we might want some other DSL generators to use the same logic without necessarily including a module. So it would be great if this was something like:
```ruby
@column_type_helper = ActiveRecordColumnTypeHelper.new(constant)
#...
@column_type_helper.type_for(column_name)
```
and if it was exported from the gem along with the DSL generation `Base` class.

mojanjz (approved):

lib/tapioca/active_record_column_type_helper.rb
```diff
@@ -0,0 +1,98 @@
...
+
+    column = @constant.columns_hash[column_name]
+    setter_type = getter_type
+
+    if column&.null
+      return ["T.nilable(#{getter_type})", "T.nilable(#{setter_type})"]
+    end
+
+    if column_name == @constant.primary_key ||
+        column_name == "created_at" ||
```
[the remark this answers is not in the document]
paracycle: A record that is not persisted yet, will definitely have its `created_at` field as `nil`.

lib/tapioca/active_record_column_type_helper.rb
```diff
@@ -0,0 +1,98 @@
...
+  def do_not_generate_strong_types?(constant)
+    Object.const_defined?(:StrongTypeGeneration) &&
+        !(constant.singleton_class < Object.const_get(:StrongTypeGeneration))
+  end
+
+  sig { params(column_type: Object).returns(String) }
+  def handle_unknown_type(column_type)
+    return "T.untyped" unless ActiveModel::Type::Value === column_type
+
+    lookup_return_type_of_method(column_type, :deserialize) ||
```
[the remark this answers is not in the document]
vinistock: In this case, since we are looking for a specific method from an Active Model value, I think it's better to use the specific arguments.

Generally, hashes as arguments tend to be used for generic option parameters or a set of related parameters, but not for specific ones.

If we define the argument as a hash, any hash will be valid for that argument (even an empty hash), which tends to be too loose for most cases. In this particular method, having the column type and the method name is mandatory, so a hash or an array would be too broad.

PR #372, opened 2021-06-28
Add DSL generator for `config` gem

paracycle:
### Motivation

We have about 30 services that in some way use the `config` gem which is really hard to statically analyze, so adding a DSL RBI generator for it.

### Implementation

The `config` gem reads config files and/or env variables to generate a `Config::Options` instance with the correct methods, etc. This means that the constant that the gem defines, which is named `Settings` by default, is not a `Module` but is instead a value constant.

To be able to represent this properly, I've decided to create a `SettingsConfigOptions` class that inherits from `Config::Options` class and I define the config methods on that class. The `Settings` class is in turn declared as an instance of the `SettingsConfigOptions` class.

This way, we are able to represent the object very well and we don't have to do any modifications to the way DSL RBI generation works.

### Tests

Added tests for the new config generator.


lib/tapioca/compilers/dsl/config.rb
```diff
@@ -0,0 +1,104 @@
...
+      #   config.const_name = "AppSettings"
+      # end
+      # ```
+      # this generator will produce the following RBI file:
+      # ```rbi
+      # AppSettings = T.let(T.unsafe(nil), AppSettingsConfigOptions)
+      #
+      # class AppSettingsConfigOptions < ::Config::Options
+      #   sig { returns(T.untyped) }
+      #   def github; end
```
vinistock, line 41: How does the gem handle nested values (e.g.: `token`)? Could we return an object that has the nested keys as methods instead of `T.untyped`?

paracycle: I am returning `T.untyped` for exactly this reason. The way the library works is based on `OpenStruct` etc and is super dynamic, so it will be very messy to define a different type for each nested object to represent it properly.

We are not very interested in the typing here, hopefully people are not using these config values in very complex ways. We are more interested in the method calls being known.

lib/tapioca/compilers/dsl/config.rb
```diff
@@ -0,0 +1,104 @@
...
+            .gsub(/#{CONFIG_OPTIONS_SUFFIX}$/, "")
+          method_names = constant.instance_methods(false)
+
+          return if method_names.empty?
+
+          root.create_constant(constant_name, value: "T.let(T.unsafe(nil), #{constant.name})")
+
+          root.create_class(option_class_name, superclass: "::Config::Options") do |mod|
+            mod.create_extend("T::Generic")
+            mod.create_constant("Elem", value: "type_member(fixed: T.untyped)")
```
vinistock, line 67: Is `Elem` used anywhere?

vinistock (approved):

KaanOzkan (approved):
LGTM pending  ruby 2.6-2.7 version failures

lib/tapioca/compilers/dsl/config.rb
paracycle: Since `Config::Options` is enumerable, we need to redefine the generic type member in the subclass. That's why I have to declare `Elem` here, so it is only used if someone wants to use `Settings` as an enumerable, each entry will be `T.untyped`. Otherwise, it is not directly used.

PR #374, opened 2021-06-30
Do not print trailing spaces with empty comments

Morriar:
### Motivation

Avoid printing a trailing space on empty comments lines

Be the following sequence of comments:

```rb
tree = RBI::Tree.new(comments: [
  RBI::Comment.new("typed: true"),
  RBI::Comment.new(""),
  RBI::Comment.new("Some comment")
])
```

Before this PR, the rendering included a trailing space on the second line:

```rb
# typed: true
# <-- trailing space
# Some comment
```

With this PR, no trailing space is rendered:

```rb
# typed: true
#<-- no trailing space
# Some comment
```

This plays better with cops.

### Tests

See automated tests.

### DIFF
--- a/lib/tapioca/rbi/printer.rb
+++ b/lib/tapioca/rbi/printer.rb
@@ -118,7 +118,10 @@ module Tapioca
 
       sig { override.params(v: Printer).void }
       def accept_printer(v)
-        v.printl("# #{text.strip}")
+        text = self.text.strip
+        v.printt("#")
+        v.print(" #{text}") unless text.empty?
+        v.printn
       end
     end
 
--- a/spec/tapioca/rbi/printer_spec.rb
+++ b/spec/tapioca/rbi/printer_spec.rb
@@ -455,6 +455,22 @@ module Tapioca
           RBI
         end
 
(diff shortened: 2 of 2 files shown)


KaanOzkan (approved):

vinistock (approved):

lib/tapioca/rbi/printer.rb
```diff
@@ -118,7 +118,10 @@ class Comment
 
       sig { override.params(v: Printer).void }
       def accept_printer(v)
-        v.printl("# #{text.strip}")
+        text = self.text.strip
```
paracycle, line 121: Lol, I am so sure that you first wrote this as:
```ruby
text = text.strip
```
and you were wondering why it didn't work. This is such a gotcha part of the language.

Morriar: What do you mean?

paracycle: The fact that you have to use a `self.text` on the right hand side is so unintuitive to me. I, for one, have to keep thinking twice about that.

PR #375, opened 2021-06-30
Add RBI merge feature

Morriar:
### Motivation

Allow to programatically merge two RBI files.

Be two RBI trees:

```rb
rbi1 = RBI::Parser.parse_string(<<~RBI)
  class A
    include A
    attr_reader :a

    # Comment
    def m1; end
  end
RBI

rbi2 = RBI::Parser.parse_string(<<~RBI)
  class A
    extend B
    attr_accessor :b

    # Comment
    sig { void }
    def m1; end

    sig { params(a: Integer).void }
    def m2(a); end
  end
RBI
```

One can merge them using the `merge` method:

```rb
res = rbi1.merge(rbi2)

assert_equal(<<~RBI, res.string)
  class A
    include A
    extend B
    attr_reader :a
    attr_accessor :b

    sig { void }
    def m1; end

    sig { params(a: Integer).void }
    def m2(a); end
  end
RBI
```

This will automatically merge matching definitions and produce the resulting tree.

In case of conflicts, `merge` will create git-like merge conflicts asking for manual intervention from the end-user:

```rb
rbi1 = RBI::Parser.parse_string(<<~RBI)
  class A
    attr_reader :a
    def m1; end
  end
RBI

rbi2 = RBI::Parser.parse_string(<<~RBI)
  class A
    attr_accessor :a
    def m1(a); end
    def m2; end
  end
RBI

res = rbi1.merge(rbi2)

assert_equal(<<~RBI, res.string)
  class A
    <<<<<<< left
    attr_reader :a
    def m1; end
    =======
    attr_accessor :a
    def m1(a); end
    >>>>>>> right
    def m2; end
  end
RBI
```

Conflicts on scope definitions are also handled:

```rb
rbi1 = RBI::Parser.parse_string(<<~RBI)
  class A
    def m1; end
  end
RBI

rbi2 = RBI::Parser.parse_string(<<~RBI)
  class A < B
    def m2; end
  end
RBI

res = rbi1.merge(rbi2)

assert_equal(<<~RBI, res.string)
  <<<<<<< left
  class A
  =======
  class A < B
  >>>>>>> right
    def m1; end
    def m2; end
  end
RBI
```

### Implementation

The implementation relies on a visitor like other RBI rewriters plus the usage of `RBI::Index` (#373) to locate nodes related to the same definition.

Nodes content for matching definitions are merged.

In case of conflict we create a new `ConflictTree` node that we add in the tree. The node is then printed with the `<<< === >>>` markers. A slightly different node `ScopeConflict` is used for scope definitions conflicts as to put the conflict markers only around the scope header (and not around the whole module or class which would make for a poor visibility when applied to scopes with hundreds of properties).

### Tests

See automated tests.


lib/tapioca/rbi/model.rb
```diff
@@ -390,6 +396,12 @@ def initialize(name, loc: nil, comments: [])
       def to_s
         name
       end
+
+      sig { params(other: T.nilable(Object)).returns(T::Boolean) }
+      def ==(other)
+        return false unless other.instance_of?(Param)
+        name == T.cast(other, Param).name
```
KaanOzkan, line 414: Side note, couldn't `instance_of?` be supported in Sorbet's flow typing?

Morriar: There is already an issue for that and it's less trivial than what it seems https://github.com/sorbet/sorbet/issues/48

KaanOzkan (approved):

lib/tapioca/rbi/rewriters/merge_trees.rb
```diff
@@ -0,0 +1,589 @@
...
+            scope_conflict = ScopeConflict.new(left: left, right: right, left_name: @left_name, right_name: @right_name)
+            left.replace(scope_conflict)
+          end
+
+          sig { params(left: Node, right: Node).void }
+          def make_conflict_tree(left, right)
+            @conflicts << Conflict.new(left: left, right: right, left_name: @left_name, right_name: @right_name)
+            tree = left.parent_conflict_tree
+            unless tree
+              tree = ConflictTree.new(left_name: @left_name, right_name: @right_name)
```
paracycle, line 173: Is there way we can make the conflict resolution strategy configurable? For example, I would very much like us to use the bundler `/rbi` folder contents from gems, if they exist and then for Tapioca to merge its generated RBI onto that. However, I think I would like the Tapioca generated nodes to overwrite the bundled RBI nodes without needing any manual override. If the conflict resolution strategy was pluggable, I could easily create a `RightConflictTree` class that would always use the nodes from the right tree, and pass it in here.

Btw, I am not necessarily asking this to be configurable as part of this PR, just want to understand if we have plans on doing that.

paracycle (approved):

lib/tapioca/rbi/rewriters/merge_trees.rb
Morriar: Yes it's the plan.

Follow ups will add ways to keep/discard definitions depending on the source as well as a way to annotate definitions that should be kept 👍

PR #379, opened 2021-07-06
Bump sorbet to 0.5.6482

RyanBrushett:
### Motivation

Bump sorbet!

https://github.com/Shopify/tapioca/pull/377
https://github.com/Shopify/tapioca/pull/378


KaanOzkan (approved):
We could be doing less work and merge the dependabot PRs one after the other by the way. It's not important that version will be mismatched for a few minutes 🙂 .

PR #382, opened 2021-07-09
Support excluding specific DSL generators

jeffcarbs:
### Motivation

There are times where we've wanted to ignore certain DSL generators, either temporarily to support migration or permanently when it causes conflicts in our codebase (e.g. if we have our own generator we'd like to use). 

### Implementation

Add a `--exclude_generators` flag similar to `--exclude` for gems and `--generators` for specifically requesting certain generators.

### Tests

See added tests.

----

It's a small PR, but still may be best reviewed commit-by-commit.


spec/support/repo/Gemfile
```diff
@@ -14,8 +14,9 @@ gem("psych")
 # ensure that we don't fail.
 gem("extras")
 # Needed for DSL generation tests. The simplest
-# gem that we have a DSL generator for.
+# gems that we have a DSL generator for.
 gem("smart_properties")
+gem("sidekiq")
```
jeffcarbs, line 19: Added sidekiq so we have more than one type of DSL generator and can accurately test exclusion.

lib/tapioca/compilers/dsl_compiler.rb
```diff
@@ -54,21 +55,28 @@ def run(&blk)
...
 
-      sig { params(requested_generators: T::Array[String]).returns(T.proc.params(klass: Class).returns(T::Boolean)) }
-      def generator_filter(requested_generators)
-        return ->(_klass) { true } if requested_generators.empty?
-
-        generators = requested_generators.map(&:downcase)
+      sig do
+        params(requested_generators: T::Array[String],
+          excluded_generators: T::Array[String]).returns(T.proc.params(klass: Class).returns(T::Boolean))
+      end
```
Morriar, line 61:
```suggestion
      sig do
        params(
          requested_generators: T::Array[String],
          excluded_generators: T::Array[String]
        ).returns(T.proc.params(klass: Class).returns(T::Boolean))
      end
```

lib/tapioca/compilers/dsl_compiler.rb
```diff
@@ -54,21 +55,28 @@ def run(&blk)
...
         end
       end
 
-      sig { params(requested_generators: T::Array[String]).returns(T::Enumerable[Dsl::Base]) }
-      def gather_generators(requested_generators)
-        generator_filter = generator_filter(requested_generators)
+      sig do
+        params(requested_generators: T::Array[String],
+          excluded_generators: T::Array[String]).returns(T::Enumerable[Dsl::Base])
+      end
```
Morriar, line 77:
```suggestion
      sig do
        params(
          requested_generators: T::Array[String],
          excluded_generators: T::Array[String]
        ).returns(T::Enumerable[Dsl::Base])
      end
```

lib/tapioca/compilers/dsl_compiler.rb
```diff
@@ -21,12 +21,13 @@ class DslCompiler
...
           requested_generators: T::Array[String],
+          excluded_generators: T::Array[String],
           error_handler: T.nilable(T.proc.params(error: String).void)
         ).void
       end
-      def initialize(requested_constants:, requested_generators: [], error_handler: nil)
+      def initialize(requested_constants:, requested_generators: [], excluded_generators: [], error_handler: nil)
         @generators = T.let(
-          gather_generators(requested_generators),
+          gather_generators(requested_generators, excluded_generators),
```
Morriar, line 30: Could we produce an error if the requested/excluded generator names passed do not match any existing generator?

spec/tapioca/cli_spec.rb
```diff
@@ -608,6 +665,32 @@ def body=(body); end
         end
       end
 
+      describe("with excluded files") do
+        before do
+          execute("dsl")
+        end
+
+        it "advises of removed file(s) and returns exit_status 1" do
+          output = execute("dsl", "--verify", exclude_generators: "SidekiqWorker")
```
Morriar, line 674: Can we add a test showing the behaviour for `exclude_generators: "NotFound"` please?

Morriar (commented):
I like the feature. I suggested a few changes:

lib/tapioca/compilers/dsl_compiler.rb
jeffcarbs: Added, see last commit (`Error if requested/excluded generator does not exist`).

spec/tapioca/cli_spec.rb
jeffcarbs: Added, see last commit (`Error if requested/excluded generator does not exist`).

lib/tapioca/generator.rb
```diff
@@ -307,6 +308,26 @@ def constantize(constant_names)
       constant_map.values
     end
 
+    sig { params(generator_names: T::Array[String]).returns(T::Array[T.class_of(Compilers::Dsl::Base)]) }
+    def constantize_generators(generator_names)
+      generator_map = generator_names.map do |name|
+        [name, Object.const_get("Tapioca::Compilers::Dsl::#{name}")]
```
paracycle, line 314: This is probably a behavior change. 

Previously we were stripping `Tapioca::Compilers::Dsl::` prefix from the supplied generator name and matching it to the names of the generators that we have. That meant that if someone had a `Foo::Bar::BazGenerator` class defined, and they want to just run it, they would be able to do `--generators "Foo::Bar::BazGenerator"`. In this new way, that won't work since we unconditionally assuming the namespace.

We could fix that by looking up by the `Tapioca::Compilers::Dsl::` prefix first and if that fails, falling back to looking up the given name, and if that also work, falling back to treating it as `nil`.

What do you think about that?

paracycle (commented):
I am also concerned about passing these options via the command-line and would really like to see these flags also integrated into the `config.yml` file.

Regrettably, I had poor foresight when I put that config format together, since it plain assumes that the keys will be `sync`/`generate` keys. Fortunately, the config file and format is not documented, so we can break it if we have to.

Maybe we should move all the existing keys to be under a `gem` key and start adding the DSL flags under a top-level `dsl` key in another PR?

lib/tapioca/generator.rb
jeffcarbs: Ah you're right! I had thought about that use-case (supplying your own compiler) but for some reason thought that `gsub` code wouldn't have supported it, but you're right that it would've.

I was also thinking along the same lines of falling back to trying to find the passed generator in the global namespace. I'll update this code to do that.

jeffcarbs:
@paracycle - Updated to respect passing custom generators so this should be good for another review.

Re: your point about configuration, I think nesting would probably make sense. It would make it less intuitive to pass those options via the command line (e.g. do you support `gem.exclude` or something?) but that might be a feature. I feel like which generators to include/exclude or which gems to exclude should be a repo-wide decision and not something you can change on a one-off CLI run.

paracycle (approved):
Thanks, this looks great to me. Let's get this merged after CI is green.

jeffcarbs:
Cool, specs are ✅

Morriar (approved):

lib/tapioca/generator.rb
```diff
@@ -307,6 +308,32 @@ def constantize(constant_names)
...
+        generator_klass = ["Tapioca::Compilers::Dsl::#{name}", name].find do |potential_name|
+          break Object.const_get(potential_name)
+        rescue NameError
+          # Skip if we can't find generator by the potential name
+        end
+
+        [name, generator_klass]
+      end.to_h
+
+      unprocessable_generators = generator_map.select { |_, v| v.nil? }
```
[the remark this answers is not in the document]
jeffcarbs: It would have to be an `if` after that internal `find` since we run through that rescue multiple times and I think that would probably be less readable since you'd have to explicitly `return nil` in the map and then `compact` after.

A more readable middle-ground might be an `each_with_object` or just initializing those enumerables outside of the loop:
```ruby
generator_map = {}
unprocessable_generators = []

generator_names.each do |name|
  # Try to find built-in tapioca generator first, then globally defined generator.
  ["Tapioca::Compilers::Dsl::#{name}", name].each do |potential_name|
    generator_map[name] = Object.const_get(potential_name)
    break
  rescue NameError
    # Skip if we can't find generator by the potential name
  end

  unprocessable_generators << name unless generator_map[name]
end
```

I could also update the `constantize` method right above this to follow a similar pattern since that's what I was following.

Issue #346, opened 2021-06-02
tapioca dsl are not adding trailing new lines to auto-generated files.


andynguyenshopify:
It seems like autogenerated files in (sorbet/rbi/dsl) for dynamic methods which are created by running `bin/tapioca dsl` are missing new lines. We should end these files with a newline according to https://rubystyle.guide/#newline-eof

For instance running
`bin/tapioca dsl MyModule::MyClass`

Will generate this file
```ruby
module MyModule
  class MyClass
  end
end
```
Will generate this file with no trailing newline.
`sorbet/rbi/dsl/my_module/my_class.rbi`

PR #388, opened 2021-07-14
Use our internal RBI backend for DSL generators

Morriar:
### Motivation

Use our RBI generation framework instead of `Parlour`.

### Implementation

Added a `RBI::Builder` to make an interface compatible with our current generators and minimize the diffs.

### Tests

See automated tests.

This was also tested by regenerating all Core DSL RBIs with success. Typecheking is passing and diff is minimal.

One interesting side effect is the time required to generate all the RBIs:

* Before: 3m 31s
* After: 2m 23s

Fixes https://github.com/Shopify/tapioca/issues/346.


lib/tapioca/compilers/dsl/state_machines.rb
```diff
@@ -121,16 +120,19 @@ module Dsl
       class StateMachines < Base
         extend T::Sig
 
-        sig { override.params(root: Parlour::RbiGenerator::Namespace, constant: ::StateMachines::ClassMethods).void }
+        sig { override.params(root: RBI::Tree, constant: ::StateMachines::ClassMethods).void }
         def decorate(root, constant)
           return if constant.state_machines.empty?
 
-          root.path(constant) do |klass|
+          root.create_path(T.unsafe(constant)) do |klass|
```
Morriar, line 127: `Tree#create_path` expects a `Module` but we get a `StateMachines::ClassMethods` for argument `constant` which doesn't seem to be a module.

One notable problem here is that the redefinition of `decorate` does not follow Liskov substitution principle. We do some shady thing (https://github.com/Shopify/tapioca/blob/master/lib/tapioca/compilers/dsl/base.rb#L28-L36) to avoid detection but this is weird design.

We should refactor this part to offer a correct/safe interface.

spec/tapioca/cli_spec.rb
```diff
@@ -451,47 +452,50 @@ def title=(title); end
       assert_path_exists("#{outdir}/namespace/comment.rbi")
       refute_path_exists("#{outdir}/user.rbi")
 
-      assert_equal(<<~CONTENTS.chomp, File.read("#{outdir}/baz/role.rbi"))
+      assert_equal(<<~CONTENTS, File.read("#{outdir}/baz/role.rbi"))
```
Morriar, line 455: No need to `chomp` since this PR fixes https://github.com/Shopify/tapioca/issues/346.

spec/tapioca/compilers/dsl/active_record_enum_spec.rb
```diff
@@ -36,9 +36,13 @@ class Conversation < ActiveRecord::Base
 
       expected = <<~RBI
         # typed: strong
+
         class Conversation
           include EnumMethodsModule
 
+          sig { returns(T::Hash[T.any(String, Symbol), Integer]) }
```
Morriar, line 43: Like for RBIs, all methods are sorted before the subconstants

spec/tapioca/compilers/dsl/protobuf_spec.rb
```diff
@@ -52,16 +52,17 @@ class Tapioca::Compilers::Dsl::ProtobufSpec < DslSpec
 
       expected = <<~RBI
         # typed: strong
+
         class Cart
+          sig { params(customer_id: Integer, shop_id: Integer).void }
```
Morriar, line 57: Like in gems RBIs, `initialize` methods are always sorted first.

spec/tapioca/compilers/symbol_table_compiler_spec.rb
```diff
@@ -268,10 +268,10 @@ def to_bar; end
 
         class String
           include ::Comparable
-          include ::JSON::Ext::Generator::GeneratorMethods::String
```
Morriar, line 271: This order changes because we removed Parlour has a dependency.

spec/tapioca/compilers/dsl/url_helpers_spec.rb
```diff
@@ -158,9 +158,10 @@ class Application < Rails::Application
 
       expected = <<~RBI
         # typed: strong
+
         module GeneratedUrlHelpersModule
-          include ::ActionDispatch::Routing::PolymorphicRoutes
           include ::ActionDispatch::Routing::UrlFor
+          include ::ActionDispatch::Routing::PolymorphicRoutes
```
Morriar, line 164: This is now in the same order we generate them:
image

Not sure why Parlour was changing the order?

spec/tapioca/compilers/dsl/url_helpers_spec.rb
```diff
@@ -233,11 +236,10 @@ class Application < Rails::Application
...
         # typed: strong
-        module ActionDispatch
-          class IntegrationTest
-            include GeneratedPathHelpersModule
-            include GeneratedUrlHelpersModule
-          end
+
+        class ActionDispatch::IntegrationTest
+          include GeneratedUrlHelpersModule
+          include GeneratedPathHelpersModule
```
Morriar, line 242: Now in the same order we want them to be generated:

image

spec/tapioca/compilers/dsl/url_helpers_spec.rb
paracycle: Nice! Thankfully, the order didn't matter here.

spec/tapioca/cli_spec.rb
```diff
@@ -451,47 +452,50 @@ def title=(title); end
...
         # DO NOT EDIT MANUALLY
         # This is an autogenerated file for dynamic methods in `Job`.
         # Please instead update this file by running `bin/tapioca dsl Job`.
 
         # typed: true
+
         class Job
-          sig { params(foo: T.untyped, bar: T.untyped).returns(String) }
-          def self.perform_async(foo, bar); end
+          class << self
```
paracycle, line 479: Why aren't these methods generated as `self.foo` anymore? I thought DSL generation was not doing `class << self` for class methods.

paracycle (commented):
Looks great! I just have one question.

spec/tapioca/cli_spec.rb
Morriar: My bad, I added the `nest_singleton_methods` rewriter to the DSL output:  https://github.com/Shopify/tapioca/pull/388/files#diff-5f22751f990a0340da79d3ef08e0c2785b41d8516748fc7fc3c20851b5bc3c28R91 I'll remove it.

It is not used in the `dsl_spec` tho, hence why those are not rewritten in the DSL specs like Kaan noticed: https://github.com/Shopify/tapioca/pull/388/files#diff-64b06a679c19070519b7f3394c9ba6d1a8c5c8d44c4bf61a9f13dc832f9b9a46R80-R83

paracycle: Yeah, that's how I noticed the difference in behaviour. :)

paracycle (approved):

PR #390, opened 2021-07-15
Merge cli/main.rb into cli.rb

Morriar:
I spend my time opening `cli.rb` and being surprised that the file is empty before going to `cli/main.rb`.

If we're not going to have more files under `cli/`, it makes little sense to create this separation.

Let's merge them until we figure a better separation.


rafaelfranca (approved):

rafaelfranca:
I'd go futher and remove the `Main` constant. If there is only one, it is clear that it is main.

Morriar:
Also removed the `Cli` constant 👍

vinistock (approved):

PR #397, opened 2021-07-21
Ignore vendor folder by default

miry:
### Motivation

There is some practices to install gems in `vendor/bundle`.
During type checking or tools around sorbet like `spoom` this line presented in each projects that I work in.

### Implementation

Generate a sorbet config with ignore vendor directive

### DIFF
--- a/spec/tapioca/cli/init_spec.rb
+++ b/spec/tapioca/cli/init_spec.rb
@@ -20,6 +20,7 @@ module Tapioca
         assert_equal(<<~CONTENTS, File.read(repo_path / "sorbet/config"))
           --dir
           .
+          --ignore=vendor
         CONTENTS
         assert_path_exists(repo_path / "sorbet/tapioca/require.rb")
         assert_equal(<<~CONTENTS, File.read(repo_path / "sorbet/tapioca/require.rb"))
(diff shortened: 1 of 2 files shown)


spec/tapioca/cli/init_spec.rb
```diff
@@ -20,6 +20,7 @@ class InitSpec < CliSpec
         assert_equal(<<~CONTENTS, File.read(repo_path / "sorbet/config"))
           --dir
           .
+          --ignore=vendor
```
miry, line 23: This test duplicates the code 1:1 - feel free to suggest another way.

rafaelfranca (changes requested):
I don't think we should ignore the entire vendor folder, specially by default. `vendor` in Rails application is a place for code that belongs to the application but was copied from some vendor and could or could not be modified. `vendor/plugins` is a thing, and `vendor/assets` as well, although the later is not relevant for sorbet, but the latter is.

miry:
@rafaelfranca Good to know about other folders.

What do you think to ignore `vendor/bundle` only in this case?
If it does not also work, I am ok to close PR and keep updating config in my projects.

Morriar (changes requested):
I don't think Tapioca should do this by default. Type-checking (or not type-checking) a specific directory seems to be an end-user decision. I would also not generate this configuration if the project does not contain a `vendor/` directory.

Issue #281, opened 2021-04-07
Verbose / debug mode for "tapioca dsl"


bessey:
I am trying to debug this error in "tapioca dsl"

```
Loading DSL generator classes... Done
Compiling DSL RBI files...

bundler: failed to load command: tapioca (/usr/local/bundle/bin/tapioca)
Traceback (most recent call last):
         .... trimmed ...
	 7: from /usr/local/bundle/gems/tapioca-0.4.20/lib/tapioca/compilers/dsl/active_record_associations.rb:148:in `populate_associations'
	 6: from /usr/local/bundle/gems/tapioca-0.4.20/lib/tapioca/compilers/dsl/active_record_associations.rb:148:in `each'
	 5: from /usr/local/bundle/gems/tapioca-0.4.20/lib/tapioca/compilers/dsl/active_record_associations.rb:150:in `block in populate_associations'
	 4: from /usr/local/bundle/gems/tapioca-0.4.20/lib/tapioca/compilers/dsl/active_record_associations.rb:227:in `populate_collection_assoc_getter_setter'
	 3: from /usr/local/bundle/gems/tapioca-0.4.20/lib/tapioca/compilers/dsl/active_record_associations.rb:265:in `type_for'
	 2: from /usr/local/bundle/gems/tapioca-0.4.20/lib/tapioca/compilers/dsl/active_record_associations.rb:291:in `polymorphic_association?'
	 1: from /usr/local/bundle/gems/tapioca-0.4.20/lib/tapioca/compilers/dsl/active_record_associations.rb:291:in `polymorphic_association?'
/usr/local/bundle/gems/tapioca-0.4.20/lib/tapioca/compilers/dsl/active_record_associations.rb:290:in `polymorphic_association?': undefined method `through_reflection?' for nil:NilClass (NoMethodError)
bash-5.1# 
```

but since there is no output pertaining to a specific constant, my only option is to review alphabetically how far through all constants `tapioca dsl` got so far, and try to guess what constant its on now. It would be very helpful if there were a `--verbose` flag that output "About to process X::Y::Z` _before_ running, so that I could see on what constant its breaking. Without cracking open the gem / guesswork.

paracycle:
Agreed the error reporting is super cryptic here and I'll look into making it better. However, this looks very much like an error that was raised by one of our applications (reported [here](https://github.com/Shopify/tapioca/issues/233)) which ended up being because of association declaration without the expected association column existing. 

It was something like:
```ruby
[...]

PR #402, opened 2021-07-27
Adds a `verbose` flag to the `dsl` command

mutecipher:
Closes: https://github.com/Shopify/tapioca/issues/281

### Motivation

Adds a `--verbose` flag to the `dsl` command. The output will now show which constant we're processing at runtime, so that if errors were to arise, we can pin point where the problem occurred.

### Implementation

Added a `verbose` option to the `dsl` command in `cli.rb`. The information is then printed based on whether or not that flag is set to true.

### Tests

Added additional test case ensuring that output is as expected.

### DIFF
--- a/lib/tapioca/cli.rb
+++ b/lib/tapioca/cli.rb
@@ -79,9 +79,14 @@ module Tapioca
       aliases: ["-q"],
       type: :boolean,
       desc: "Supresses file creation output"
+    option :verbose,
+      aliases: ["-V"],
+      type: :boolean,
+      desc: "Verbose output for debugging purposes"
     def dsl(*constants)
       Tapioca.silence_warnings do
-        generator.build_dsl(constants, should_verify: options[:verify], quiet: options[:quiet])
+        generator.build_dsl(constants, should_verify: options[:verify], quiet: options[:quiet],
+verbose: options[:verbose])
       end
     end
 
--- a/lib/tapioca/cli.rb
+++ b/lib/tapioca/cli.rb
@@ -86,8 +86,12 @@ module Tapioca
       desc: "Supresses file creation output"
     def dsl(*constants)
       Tapioca.silence_warnings do
-        generator.build_dsl(constants, should_verify: options[:verify], quiet: options[:quiet],
-verbose: options[:verbose])
+        generator.build_dsl(
+          constants,
+          should_verify: options[:verify],
+          quiet: options[:quiet],
+          verbose: options[:verbose]
+        )
       end
     end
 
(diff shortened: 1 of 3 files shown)


mojanjz (approved):
looks great!

KaanOzkan (approved):

lib/tapioca/cli.rb
```diff
@@ -79,9 +79,14 @@ def todo
       aliases: ["-q"],
       type: :boolean,
       desc: "Supresses file creation output"
+    option :verbose,
```
Morriar, line 82: It might be interesting to make this a [`class_option`](https://github.com/Shopify/tapioca/blob/master/lib/tapioca/cli.rb#L36) and [pass it to the generator instance](https://github.com/Shopify/tapioca/blob/master/lib/tapioca/cli.rb#L152-L154) under construction so it can be reused in other commands.

lib/tapioca/generator.rb
```diff
@@ -147,6 +148,10 @@ def build_dsl(requested_constants, should_verify: false, quiet: false)
       compiler.run do |constant, contents|
         constant_name = Module.instance_method(:name).bind(constant).call
 
+        if verbose
+          say("Processing: ", [:yellow])
+          say(constant_name)
+        end
```
Morriar, line 154:
```suggestion
        end
        
```

vinistock:
Should we have a check that `verbose` is not used together with `quiet`? It might be overkill, but the two options are conflicting - at least conceptually speaking.

mutecipher:
> Should we have a check that `verbose` is not used together with `quiet`? It might be overkill, but the two options are conflicting - at least conceptually speaking.

@vinistock Good catch. Updated the default behaviour to be:

Individually: The options work as expected. (i.e. `--verbose` provides verbose output, and `--quiet` silences output)
Together: They negate one another and it just provides the default behaviour. (i.e. standard output)

lib/tapioca/cli.rb
```diff
@@ -81,7 +86,8 @@ def todo
       desc: "Supresses file creation output"
     def dsl(*constants)
       Tapioca.silence_warnings do
-        generator.build_dsl(constants, should_verify: options[:verify], quiet: options[:quiet])
+        generator.build_dsl(constants, should_verify: options[:verify], quiet: options[:quiet],
+verbose: options[:verbose])
```
rafaelfranca, line 90: Fix the indentation please

mutecipher: ✅

KaanOzkan (approved):

PR #405, opened 2021-07-28
Better sorting for RBI nodes

Morriar:
### Motivation

Fix an early return in the node sorting as well as a conflicting node rank.

### Implementation

Instead of returning early from the block, when two nodes are at a conflicting position in a tree, we preserve the original order in the non-sorted tree.

This means that mixins are always kept in the same order they were added to the tree even if moved around to satisfy the order against other node such as methods or constants.

### Tests

See the new test file.


lib/tapioca/rbi/rewriters/sort_nodes.rb
```diff
@@ -11,11 +11,11 @@ class SortNodes < Visitor
         def visit(node)
           return unless node.is_a?(Tree)
           visit_all(node.nodes)
+          original_order = original_nodes_order(node.nodes.to_a)
           node.nodes.sort! do |a, b|
-            return 0 if a.is_a?(Mixin) || b.is_a?(Mixin)
```
Morriar, line 15: This was returning early. Thanks @paracycle 👀

lib/tapioca/rbi/rewriters/sort_nodes.rb
```diff
@@ -25,29 +25,29 @@ def visit(node)
...
+          when TEnumBlock           then 60
           when Method
             if node.name == "initialize"
-              7
+              71
             elsif !node.is_singleton
-              8
+              72
             else
-              9
```
Morriar, line 41: This was using the same rank number as scopes and consts

lib/tapioca/rbi/rewriters/sort_nodes.rb
```diff
@@ -11,11 +11,11 @@ class SortNodes < Visitor
         def visit(node)
           return unless node.is_a?(Tree)
           visit_all(node.nodes)
+          original_order = original_nodes_order(node.nodes.to_a)
           node.nodes.sort! do |a, b|
-            return 0 if a.is_a?(Mixin) || b.is_a?(Mixin)
-
             res = node_rank(a) <=> node_rank(b)
             res = node_name(a) <=> node_name(b) if res == 0
+            res = (original_order[a] || 0) <=> (original_order[b] || 0) if res == 0
```
Morriar, line 18: Fall back on the original order if everything else fails

lib/tapioca/rbi/rewriters/sort_nodes.rb
```diff
@@ -11,11 +11,11 @@ class SortNodes < Visitor
         def visit(node)
           return unless node.is_a?(Tree)
           visit_all(node.nodes)
+          original_order = original_nodes_order(node.nodes.dup)
```
Morriar, line 14: Passing a copy so we always preserve the original order since the array will be sorted in place.

lib/tapioca/rbi/rewriters/sort_nodes.rb
```diff
@@ -25,29 +25,29 @@ def visit(node)
...
           case node
-          when Group                then kind_rank(node.kind)
-          when Include, Extend      then 0
-          when Helper               then 1
-          when TypeMember           then 2
-          when MixesInClassMethods  then 3
-          when TStructField         then 4
-          when TEnumBlock           then 5
+          when Group                then group_rank(node.kind)
+          when Include, Extend      then 10
```
Morriar, line 29: Using 10, 20, etc. so we can add some sub-ordering when needed (example with Method a few lines below)

lib/tapioca/rbi/rewriters/sort_nodes.rb
```diff
@@ -70,6 +70,15 @@ def node_name(node)
...
           end
         end
+
+        sig { params(nodes: T::Array[Node]).returns(T::Hash[Node, Integer]) }
+        def original_nodes_order(nodes)
+          order = {}
+          nodes.each_with_index do |node, i|
+            order[node] = i
+          end
+          order
```
paracycle, line 80: Nitpick:
```suggestion
          nodes.each_with_index.each_with_object do |node, index, result|
            result[node] = index
          end
```

might be more idiomatic

lib/tapioca/rbi/rewriters/sort_nodes.rb
```diff
@@ -11,11 +11,11 @@ class SortNodes < Visitor
         def visit(node)
           return unless node.is_a?(Tree)
           visit_all(node.nodes)
+          original_order = original_nodes_order(node.nodes.dup)
           node.nodes.sort! do |a, b|
-            return 0 if a.is_a?(Mixin) || b.is_a?(Mixin)
-
             res = node_rank(a) <=> node_rank(b)
             res = node_name(a) <=> node_name(b) if res == 0
+            res = (original_order[a] || 0) <=> (original_order[b] || 0) if res == 0
```
paracycle, line 18: I am not quite clear how we fallback to this case for mixins? Is it that their node names are always the same? What prevents mixin nodes from being sorted alphabetically?

Morriar: `node_name` always returns `nil` for nodes that shouldn't be sorted alphabetically: https://github.com/Shopify/tapioca/blob/master/lib/tapioca/rbi/rewriters/sort_nodes.rb#L69-L70

lib/tapioca/rbi/rewriters/sort_nodes.rb
Morriar: It would have been nice like this but:

```
irb(main):001:0> arr = ["a", "b", "c"]
=> ["a", "b", "c"]
irb(main):002:0> arr.each_with_index.each_with_object({}) { |x, i, res| res[x] = i }
(irb):2:in `block in <main>': undefined method `[]=' for nil:NilClass (NoMethodError)
	from (irb):2:in `each'
	from (irb):2:in `each_with_index'
	from (irb):2:in `each_with_object'
	from (irb):2:in `<main>'
	from /opt/rubies/3.0.1/lib/ruby/gems/3.0.0/gems/irb-1.3.5/exe/irb:11:in `<top (required)>'
	from /opt/rubies/3.0.1/bin/irb:23:in `load'
	from /opt/rubies/3.0.1/bin/irb:23:in `<main>'
```

My understanding is that we need to destructure the couple:

```rb
nodes.each_with_index.each_with_object({}) do |node_plus_index, result|
  node, index = *node_plus_index
     result[node] = index
  end
end
```

Do we still want this change?

paracycle (approved):

lib/tapioca/rbi/rewriters/sort_nodes.rb
paracycle: yeah, maybe we need `|(node, index), result|`?

paracycle: yes, that works:
```ruby
["a", "b", "c"].each_with_index.each_with_object({}) { |(x, i), res| res[x] = i }
# => {"a"=>0, "b"=>1, "c"=>2}
```

paracycle: There is also the much shorter:
```ruby
["a", "b", "c"].map.with_index.map.to_h
# => {"a"=>0, "b"=>1, "c"=>2}
```
:)

vinistock (approved):

lib/tapioca/rbi/rewriters/sort_nodes.rb
Morriar: I went with `original_order = node.nodes.map.with_index.to_h` and removed the now-useless method 👍

PR #414, opened 2021-08-04
Add `gem` command to replace `sync` and `generate` commands

paracycle:
### Motivation

With hindsight, the decision to have two separate commands to basically do the same gem RBI generation was a bad judgement call on my behalf. It is not super intuitive when people should use `generate` and when they should use `sync`. Moreover, the fact that `sync` does not accept gem names as parameters makes the cognitive load even higher.

Additionally, now that we have a DSL compilation command, it makes sense to have a Gem compilation command as well. Thus, the `gem` command.

### Implementation

The way the `gem` command works is:
- `tapioca gem`: works like how `tapioca sync` used to work. Syncs the version of the gems in the `Gemfile.lock` with the RBI files.
- `tapioca gem gem-name[...]`: works like how `tapioca generate gem-name[...]` used to work. Generates RBI files for the specified gems.
- `tapioca gem --all`: works like how `tapioca generate` used to work. (Re)generates RBI files for all gems in the `Gemfile.lock`.
- `tapioca gem --verify`: works like how `tapioca sync --verify` used to work. Verifies that the versions of the gems in the `Gemfile.lock` is synced to the RBI files.

I've added deprecation notices to `sync` and `generate` to make it obvious that those commands will be going away but their functionality stays intact for the time being. I added the notices to print both at the top of the output (to make it obvious to people who manually run those commands) and at the bottom of the output (to make it obvious to people who run the command as part of an automation and only look at the result of it).

### Tests

Added new tests for the unified command.


lib/tapioca/cli.rb
```diff
@@ -97,12 +142,24 @@ def dsl(*constants)
       banner: "gem:level [gem:level ...]",
       desc: "Overrides for typed sigils for generated gem RBIs"
     def generate(*gems)
+      deprecation_message = <<~MSG
+        DEPRECATION: The `generate` command will be removed in a future release.
+
+        Start using the `gem` command with gem names or the `--all` flag instead.
```
Morriar, line 148: We could even use the args to suggest the exact replacement:

```rb
"Use `tapioca gem #{args.join(" ")}` instead"
```

lib/tapioca/cli.rb
```diff
@@ -97,12 +142,24 @@ def dsl(*constants)
       banner: "gem:level [gem:level ...]",
       desc: "Overrides for typed sigils for generated gem RBIs"
     def generate(*gems)
+      deprecation_message = <<~MSG
```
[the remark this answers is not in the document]
paracycle: My plan is to release `0.5` with this inside and give people a chance to switch over to the new syntax. By `0.6` or something we can remove completely. We don't have to do a deprecation dance but it still feels better to do it if we can.

lib/tapioca/cli.rb
paracycle: This is a good idea, I'll implement this.

lib/tapioca/cli.rb
```diff
@@ -77,7 +77,52 @@ def dsl(*constants)
       end
     end
 
-    desc "generate [gem...]", "generate RBIs from gems"
+    desc "gem [gem...]", "generate RBIs from gems"
+    option :all,
```
vinistock, line 81: In my opinion, it would be more intuitive for the default gem command be to generate all and `sync` be an option. E.g.:
```shell
bin/tapioca gem # generates everything
bin/tapioca gem rails # generates specific gem
bin/tapioca gem --update # (or --sync) updates for all gems
bin/tapioca gem rails --update # (or --sync) updates for a specific gem
```
To my taste, this feels more consistent. In the way it's currently implemented, the gem command can either generate or sync, depending on whether we pass a gem name as an argument.

Thoughts?

paracycle: I've been pushing for the default action to be `sync`, since that works properly ~90% of the time and implies huge time-saving benefits. For example, on Core, if everyone were to regenerate the whole gem RBI set, everybody would be waiting for 5-6 mins while all random gem RBIs are generated.

That's why I've made `sync` the default action. If you give specific gem names, then those gems get generated and if you supply `--all` then all gems get generated. There is no case of syncing specific gems anyway, since `sync` is meant to bring things in line with Gemfile versions.

lib/tapioca/cli.rb
```diff
@@ -77,7 +77,52 @@ def dsl(*constants)
...
+      default: false,
+      desc: "Verifies RBIs are up-to-date"
+    def gem(*gems)
+      Tapioca.silence_warnings do
+        all = options[:all]
+        verify = options[:verify]
+
+        unless gems.empty?
+          raise MalformattedArgumentError, "Option '--all' must be provided without any other arguments" if all
+          raise MalformattedArgumentError, "Option '--verify' must be provided without any other arguments" if verify
```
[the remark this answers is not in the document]
paracycle: Done!

Morriar (approved):

PR #416, opened 2021-08-05
Add ActiveStorage getter and setter generator for has_one_attached and has_many_attached

### LINKED ISSUE #376: ActiveStorage dynamic proxy methods not surfaced
I'm actually not 100% sure if this issue belongs here, in sorbet-rails, or somewhere else. Please redirect me to wherever would have responsibility for this!

> [`ActiveStorage` has been designed to having you go through the `ActiveStorage::Attached::One` proxy that provides the dynamic proxy to the associations and factory methods](https://apidock.com/rails/ActiveStorage/Attached/Macros/has_one_attached)

However, generated RBIs do not include those proxy methods, which means operations like `model.file.**open**` raise a typechecking error, despite that method both existing and working. A workaround is to use `model.file**_blob**.open`, but this is not the documented API so I'm nervous using it - I imagine it's more likely to change in future.

I'd expect the RBI for `ActiveStorage::Attached::One` to include these methods so that typechecking can be used [when accessing the files as documented](https://edgeguides.rubyonrails.org/active_storage_overview.html#downloading-files).

mojanjz:
### Motivation
closes #376 

Based on ActiveStorage [documentation](https://edgeguides.rubyonrails.org/active_storage_overview.html#has-one-attached) on `has_one_attached` and `has_many_attached`, we are missing getter/setter methods for the attachments. 

Here is the source code for these getters and setters: [`has_one_attached`](https://github.com/rails/rails/blob/83217025a171593547d1268651b446d3533e2019/activestorage/lib/active_storage/attached/model.rb#L53) and [`has_many_attached`](https://github.com/rails/rails/blob/83217025a171593547d1268651b446d3533e2019/activestorage/lib/active_storage/attached/model.rb#L128). The other methods are picked up by `ActiveRecord` scope and association generators.

### Implementation
Added an `ActiveStorage` generator that adds the getter and setter methods for the attachments to the `ActiveRecord` model.

Added tests for ActiveStorage associations and scopes to `active_record_associations_spec` and `active_record_scope_spec`. I put the tests in a new `describe` block, but if there is a more acceptable way to do this let me know.

### Tests
Added tests for the generator


Morriar:
You'll have to run `bin/docs` and commit the results to update the doc pages 📝

sorbet/rbi/shims/activestorage.rbi
```diff
@@ -0,0 +1,9 @@
+# typed: strict
+
+# ActiveStorage dynamically adds a mixin and class attributes
+# so replicating that behaviour here.
+
+class ActiveRecord::Base
+  include(::ActiveStorage::Attached::Model)
+  class_attribute :attachment_reflections, instance_writer: false, default: {}
+end
```
Morriar, line 9: You're missing a newline at the end of this file

lib/tapioca/compilers/dsl/active_storage.rb
```diff
@@ -0,0 +1,42 @@
...
+        end
+
+        sig { override.params(root: RBI::Tree, constant: T.class_of(::ActiveRecord::Base)).void }
+        def decorate(root, constant)
+          return if constant.reflections.empty?
+
+          root.create_path(constant) do |scope|
+            constant.attachment_reflections.each do |field_name, _value|
+              scope << RBI::Method.new(field_name)
+              scope << RBI::Method.new("#{field_name}=")
```
Morriar, line 30: I think you're missing the parameter `attachable` for the setter: https://github.com/rails/rails/blob/83217025a171593547d1268651b446d3533e2019/activestorage/lib/active_storage/attached/model.rb#L138

lib/tapioca/compilers/dsl/active_storage.rb
```diff
@@ -0,0 +1,42 @@
...
+            ::ActiveStorage::Reflection::HasManyAttachedReflection)
+        end
+
+        sig { override.params(root: RBI::Tree, constant: T.class_of(::ActiveRecord::Base)).void }
+        def decorate(root, constant)
+          return if constant.reflections.empty?
+
+          root.create_path(constant) do |scope|
+            constant.attachment_reflections.each do |field_name, _value|
+              scope << RBI::Method.new(field_name)
```
Morriar, line 29: Maybe we could provide a signature?

```rb
sig { returns(T.any(ActiveStorage::Attached::One, ActiveStorage::Attached::Many)) }
def photo; end
```

Morriar (commented):
If we don't have them already, can you add tests to the related generators so we can ensure that the [`has_one`](https://github.com/rails/rails/blob/83217025a171593547d1268651b446d3533e2019/activestorage/lib/active_storage/attached/model.rb#L70), [`has_many`](https://github.com/rails/rails/blob/83217025a171593547d1268651b446d3533e2019/activestorage/lib/active_storage/attached/model.rb#L155) and [`scope`](https://github.com/rails/rails/blob/83217025a171593547d1268651b446d3533e2019/activestorage/lib/active_storage/attached/model.rb#L73) are generated properly please? 🙏

lib/tapioca/compilers/dsl/active_storage.rb
```diff
@@ -0,0 +1,66 @@
+# typed: strict
+# frozen_string_literal: true
+
+begin
+  require "active_record"
+  require "active_storage"
+  require "active_storage/attached"
```
paracycle, line 7: I think we need to require `active_storage/reflections` here since `attached` should be handled by the default require:
```suggestion
  require "active_record"
  require "active_storage"
  require "active_storage/reflection"
```
since we want to check that they are extending `ActiveStorage::Reflection::ActiveRecordExtensions::ClassMethods`.

lib/tapioca/compilers/dsl/active_storage.rb
```diff
@@ -0,0 +1,66 @@
...
+      # ~~~
+      class ActiveStorage < Base
+        extend T::Sig
+
+        sig { override.params(root: RBI::Tree, constant: T.class_of(::ActiveRecord::Base)).void }
+        def decorate(root, constant)
+          return if constant.reflections.empty?
+
+          root.create_path(constant) do |scope|
+            constant.attachment_reflections.each do |field_name, _value|
```
paracycle, line 47: The `value` parameter here should be used since it tells us if this is a one or many attachment:
```ruby
class User < ActiveRecord::Base
  has_one_attached :foo
  has_many_attached :bar
end

User.attachment_reflections
=> {"foo"=>
  #<ActiveStorage::Reflection::HasOneAttachedReflection:0x00007fb9d616d5b8
   @active_record=User (call 'User.connection' to establish a connection),
   @klass=nil,
   @name=:foo,
   @options={:dependent=>:purge_later, :service_name=>nil},
   @plural_name="foos",
   @scope=nil>,
 "bar"=>
  #<ActiveStorage::Reflection::HasManyAttachedReflection:0x00007fb9d2c2dc90
   @active_record=User (call 'User.connection' to establish a connection),
   @klass=nil,
   @name=:bar,
   @options={:dependent=>:purge_later, :service_name=>nil},
   @plural_name="bars",
   @scope=nil>}
```

Note how the class of the value tells us which kind it is.

lib/tapioca/compilers/dsl/active_storage.rb
```diff
@@ -0,0 +1,66 @@
...
+      #  def photo; end
+      #  def photo=; end
+      #  def blogs; end
+      #  def blogs=; end
+      # end
+      # ~~~
+      class ActiveStorage < Base
+        extend T::Sig
+
+        sig { override.params(root: RBI::Tree, constant: T.class_of(::ActiveRecord::Base)).void }
```
paracycle, line 42:
```suggestion
        sig { override.params(root: RBI::Tree, constant: ::ActiveStorage::Reflection::ActiveRecordExtensions::ClassMethods).void }
```

lib/tapioca/compilers/dsl/active_storage.rb
paracycle: Actually, I think it is better to use `reflect_on_all_attachments` here instead, since it is a method defined on `ActiveStorage::Reflection::ActiveRecordExtensions::ClassMethods`: http://api.rubyonrails.org/classes/ActiveStorage/Reflection/ActiveRecordExtensions/ClassMethods.html#method-i-reflect_on_all_attachments
```suggestion
            constant.reflect_on_all_attachments.each do |reflection|
```

lib/tapioca/compilers/dsl/active_storage.rb
```diff
@@ -0,0 +1,66 @@
...
+
+          root.create_path(constant) do |scope|
+            constant.attachment_reflections.each do |field_name, _value|
+              getter_sig = RBI::Sig.new(
+                return_type: "T.any(ActiveStorage::Attached::One, ActiveStorage::Atached::Many)"
+              )
+              scope << RBI::Method.new(field_name, sigs: [getter_sig])
+              setter_sig = RBI::Sig.new(params: [RBI::SigParam.new(field_name,
+                "T.any(ActiveStorage::Attached::One, ActiveStorage::Attached::Many")])
+              scope << RBI::Method.new("#{field_name}=", params: [RBI::Param.new("attachable")], sigs: [setter_sig])
```
paracycle, line 54:
```suggestion
              type = type_of(reflection)
              name = reflection.name.to_s
              
              getter_sig = RBI::Sig.new(return_type: type)
              setter_sig = RBI::Sig.new(params: [RBI::SigParam.new("attachable", type)])

              scope << RBI::Method.new(name, sigs: [getter_sig])
              scope << RBI::Method.new("#{name}=", params: [RBI::Param.new("attachable")], sigs: [setter_sig])
```
where `type_of` is a private method that returns `"ActiveStorage::Attached::One"` or `"ActiveStorage::Attached::Many"` depending on the type of the reflection.

lib/tapioca/compilers/dsl/active_storage.rb
```diff
@@ -0,0 +1,66 @@
...
+            end
+          end
+        end
+
+        sig { override.returns(T::Enumerable[Module]) }
+        def gather_constants
+          ActiveRecord::Base.descendants.reject(&:abstract_class?)
```
paracycle, line 61: Let's change this to only scan for classes that extend `ActiveStorage::Reflection::ActiveRecordExtensions::ClassMethods`
```suggestion
          ActiveRecord::Base.descendants
            .reject(&:abstract_class?)
            .grep(::ActiveStorage::Reflection::ActiveRecordExtensions::ClassMethods)
```

lib/tapioca/compilers/dsl/active_storage.rb
```diff
@@ -0,0 +1,66 @@
...
+      # this generator will produce the RBI file `post.rbi` with the following content:
+      #
+      # ~~~rbi
+      # # typed: true
+      # class Post
+      #  def photo; end
+      #  def photo=; end
+      #  def blogs; end
+      #  def blogs=; end
+      # end
```
paracycle, line 37: This is missing signatures and the `attachable` argument on setters.

sorbet/rbi/shims/activestorage.rbi
paracycle, line 9: Since this happens dynamically, we really should not encode this as if it is statically true. We really should not need this if we operate over the `ActiveStorage::Reflection::ActiveRecordExtensions::ClassMethods` type anyway.

paracycle (changes requested):

lib/tapioca/compilers/dsl/active_storage.rb
```diff
@@ -0,0 +1,73 @@
...
+        sig do
+          override.params(root: RBI::Tree,
+            constant: T.all(Module, ::ActiveStorage::Reflection::ActiveRecordExtensions::ClassMethods)).void
+        end
+        def decorate(root, constant)
+          return if constant.reflect_on_all_attachments.empty?
+
+          root.create_path(constant) do |scope|
+            constant.reflect_on_all_attachments.each do |reflection|
+              type = "T.untyped"
```
mojanjz, line 51: it seems like attachable can be any of the types mentioned [here](https://github.com/rails/rails/blob/83217025a171593547d1268651b446d3533e2019/activestorage/lib/active_storage/attached/one.rb#L20-L29)

spec/tapioca/compilers/dsl/active_record_associations_spec.rb
```diff
@@ -396,4 +396,138 @@ def commenters_attributes=(attributes); end
...
+        database: ":memory:"
+      )
+
+      T.bind(self, Tapioca::Compilers::Dsl::ActiveRecordAssociationsSpec)
+      add_ruby_file("application.rb", <<~RUBY)
+        ENV["DATABASE_URL"] = "sqlite3::memory:"
+
+        require "active_storage/engine"
+
+        class Dummy < Rails::Application
```
mojanjz, line 419: creating a dummy app to run all Rails initializer to load all ActiveStorage classes.

lib/tapioca/compilers/dsl/active_storage.rb
```diff
@@ -0,0 +1,73 @@
...
+              name = reflection.name.to_s
+
+              getter_sig = RBI::Sig.new(return_type: type)
+              setter_sig = RBI::Sig.new(params: [RBI::SigParam.new("attachable", type)])
+
+              scope << RBI::Method.new(name, sigs: [getter_sig])
+              scope << RBI::Method.new("#{name}=", params: [RBI::Param.new("attachable")],
```
paracycle, line 58: I think we prefer the utility methods instead of these low level ones:
```suggestion
			  scope.create_method(
			  	name, 
			  	return_type: type
			  ) 

        	  scope.create_method(
			  	"#{name}=", 
			  	parameters: [create_param("attachable", type: "T.untyped")],
			  	return_type: "T.untyped"
			  ) 
```

lib/tapioca/compilers/dsl/active_storage.rb
paracycle: Right, the type of the `attachable` parameter is `T.untyped`, but the return type of the getter is still the type that you were discovering previously. You should bring that back.

paracycle (changes requested):

lib/tapioca/compilers/dsl/active_storage.rb
```diff
@@ -0,0 +1,96 @@
...
+                return_type: "void"
+              )
+            end
+          end
+        end
+
+        sig do
+          params(reflection: T.any(::ActiveStorage::Reflection::HasOneAttachedReflection,
+            ::ActiveStorage::Reflection::HasManyAttachedReflection)).returns(String)
+        end
```
paracycle, line 77: We are not that interested in the concrete type here, we can just do:
```suggestion
        sig { params(reflection: Class).returns(String) }
```

Morriar (approved):
Amazing, thanks!

PR #418, opened 2021-08-09
Extract reflection methods into a module

paracycle:
### Motivation

We use reflection helper methods throughout the codebase, but they kept being replicated and we ended up having lots of copies of how to grab the actual name of a constant, etc. We need a more unified central way to do that.

### Implementation

This implementation extracts said reflection helper methods into a helper module that can be mixed into a class (done for `SymbolGenerator` and `Dsl::Base`) or can have the helper methods called on the module itself.

I've also unified some other patterns wrt reflection in the codebase as well.

### Tests

Added an extra test to validate that we get expected results for a lying class. Also some changes to existing tests.

### DIFF
--- a/lib/tapioca/compilers/dsl/base.rb
+++ b/lib/tapioca/compilers/dsl/base.rb
@@ -10,6 +10,8 @@ module Tapioca
         extend T::Sig
         extend T::Helpers
 
+        include Reflection
+
         abstract!
 
         sig { returns(T::Set[Module]) }
--- a/lib/tapioca/compilers/dsl/base.rb
+++ b/lib/tapioca/compilers/dsl/base.rb
@@ -43,6 +43,18 @@ module Tapioca
 
         private
 
+        sig { returns(T::Enumerable[Class]) }
+        def all_classes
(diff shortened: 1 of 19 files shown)


lib/tapioca/compilers/dsl/base.rb
```diff
@@ -41,6 +43,18 @@ def gather_constants; end
 
         private
 
+        sig { returns(T::Enumerable[Class]) }
+        def all_classes
+          @all_classes = T.let(@all_classes, T.nilable(T::Enumerable[Class]))
+          @all_classes ||= -> { T.cast(ObjectSpace.each_object(Class), T::Enumerable[Class]) }.call
```
paracycle, line 49: This is super ugly, but I couldn't find another way around Sorbet complaining about the `T.cast` call and ivar initialization needing `T.let` instead.

vinistock (approved):

lib/tapioca/compilers/dsl/base.rb
paracycle: I changed it to make an extra `.each` at the end of each enumerable (outside the cast) so that Sorbet is not thrown off and we don't have to do any extra work.

PR #426, opened 2021-08-11
Clean up obsolete references to this repo's master branch

RyanBrushett:
### Motivation

We renamed `master` to `main` so these references need updating.

### Implementation

Find and replace!

### Tests

See automated tests


paracycle (approved):
I wonder if we should not extract the generated RBI links to point to a wiki page or something. But, these changes are good.

PR #427, opened 2021-08-11
Link to a wiki page rather than README from empty RBI

RyanBrushett:
### Motivation

Link to a wiki page about manual gem requires rather than the README on the main branch.

See [this comment](https://github.com/Shopify/tapioca/pull/426#pullrequestreview-727780528) for the suggestion.

### Implementation

- I migrated the "Manual gem requires" section of the README to a new wiki page which you can see at the following link: https://github.com/Shopify/tapioca/wiki/Manual-Gem-Requires
  - Note that I didn't edit any of the content.
- I modified `generator.rb` so empty RBI files would be generated with a wiki link rather than the README link.
- ~I linked the README to the wiki to be complete and so anyone with the previous link will be referred to our wiki.~

### Tests

See included automated tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,31 @@ When you run `tapioca sync` in a project, `tapioca` loads all the gems that are 
 
 ## Manual gem requires
 
-See "[Manual Gem Requires](https://github.com/Shopify/tapioca/wiki/Manual-Gem-Requires)" in our wiki for more information.
+For gems that have a normal default `require` and load all of their constants through such a require, everything works seamlessly. However, for gems that are marked as `require: false` in the Gemfile, or for gems that export optionally loaded types via different requires, where a single require does not load the whole gem code into memory, `tapioca` will not be able to load some of the types into memory and, thus, won't be able to generate complete RBIs for them. For this reason, we need to keep a small external file named `sorbet/tapioca/require.rb` that is executed after all the gems in the Gemfile have been required and before generation of gem RBIs have started. This file is responsible for adding the requires for additional files from gems, which are not covered by the default require.
+
+For example, suppose you are using the class `BetterHtml::Parser` exported from the `better_html` gem. Just doing a `require "better_html"` (which is the default require) does not load that type:
+
+```shell
+$ bundle exec pry
+[1] pry(main)> require 'better_html'
+=> true
+[2] pry(main)> BetterHtml
+=> BetterHtml
+[3] pry(main)> BetterHtml::Parser
+NameError: uninitialized constant BetterHtml::Parser
+from (pry):3:in `__pry__`
+[4] pry(main)> require 'better_html/parser'
+=> true
+[5] pry(main)> BetterHtml::Parser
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -12,31 +12,7 @@ When you run `tapioca sync` in a project, `tapioca` loads all the gems that are
 
 ## Manual gem requires
 
-For gems that have a normal default `require` and load all of their constants through such a require, everything works seamlessly. However, for gems that are marked as `require: false` in the Gemfile, or for gems that export optionally loaded types via different requires, where a single require does not load the whole gem code into memory, `tapioca` will not be able to load some of the types into memory and, thus, won't be able to generate complete RBIs for them. For this reason, we need to keep a small external file named `sorbet/tapioca/require.rb` that is executed after all the gems in the Gemfile have been required and before generation of gem RBIs have started. This file is responsible for adding the requires for additional files from gems, which are not covered by the default require.
```
paracycle, line 15: I think I would still like to keep this in the README for now. It is a unique feature of Tapioca and should be quite prominent in the README IMO. We can think about how we want to change the README afterwards, if you want, but let's keep this here as well, for now.

RyanBrushett: Can do!

mojanjz (approved):

paracycle:
> What's the point of duplicating this entry if we're keeping it in the README?

The wiki link is a link that is supposed to be more stable allowing us to be more free to change the README as we wish. We already got into a case where all the links in existing RBI files are not working now since we changed the main branch name. I would like to avoid any such future breakage.
 
> It seems the previous link would already be directing the users to the corresponding section in the README, what value do we get by moving this around?

The content in the README about this and in the wiki for people coming from an empty RBI can diverge. Right now, we can leave the same content in both places, we can decide how we will change both later.

> I see at least two drawbacks: 1) no more review on the content pushed to the wiki, 2) one content will most likely go out of sync compared to the other.

I believe the wiki is also a git repo, so we should be able to review changes.

paracycle (approved):

Morriar:
> > What's the point of duplicating this entry if we're keeping it in the README?
> 
> The wiki link is a link that is supposed to be more stable allowing us to be more free to change the README as we wish. We already got into a case where all the links in existing RBI files are not working now since we changed the main branch name. I would like to avoid any such future breakage.
> 
> > It seems the previous link would already be directing the users to the corresponding section in the README, what value do we get by moving this around?

It is not. Without a review process who will remind the author of to also update `generator.rb` if the wiki page title changes?

> > The content in the README about this and in the wiki for people coming from an empty RBI can diverge. Right now, we can leave the same content in both places, we can decide how we will change both later.
> 
> > I see at least two drawbacks: 1) no more review on the content pushed to the wiki, 2) one content will most likely go out of sync compared to the other.
> 
> I believe the wiki is also a git repo, so we should be able to review changes.

Then let's wait to enable reviews and to have the need for diverged content? In the current state we just create duplication that will go out of sync.

PR #439, opened 2021-08-15
fix: make CollectionProxy generic types root-namespaced

fsateler:
This avoids errors when you have something like:

```ruby
module Foo
  class Foo < ActiveRecord::Base;
    has_one :bar # This errors with "Unable to resolve constant Bar"
  end
  class Bar < ActiveRecord::Base
  end
end
```

In this case, due to namespacing rules, `Bar` is looked up in `Foo::Foo` instead of the intended `::Foo`.


paracycle (approved):
Thanks, this looks good. I took the liberty of improving it a little and fixing the tests and the documentation. Can you please sign the CLA, so that we can merge the PR?

fsateler:
Done! I think the check needs to be re-triggered.

PR #445, opened 2021-08-18
Remove pry and bundler from gem dependency

miry:
There is no need in pry or bundler in application that depends on tapioca.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -9,8 +9,6 @@ PATH
   remote: .
   specs:
     tapioca (0.4.23)
-      bundler (>= 1.17.3)
-      pry (>= 0.12.2)
       rbi
       sorbet-runtime
       sorbet-static (>= 0.4.4471)


paracycle (changes requested):
This is not correct. Certain functionality of Tapioca relies on using (parts of) Bundler and Pry at runtime. You can also see the same from the test failures.

miry:
@paracycle Let me move those to Gemfile.

miry:
What do you think If I replace Pry::WrappedModule with home made WrapperModule?

paracycle:
Again, my question is which specific problem is this creating? Do you have a real-life example of these dependencies blocking an application from doing anything it needs to do?

miry:
> Do you have a real-life example of these dependencies blocking an application from doing anything it needs to do?

Specifically `tapioca` today - no.

> which specific problem is this creating?

It confuse users(me). Users(me) do not expect to see non standard interactive shell as dependencies.

Issue #413, opened 2021-08-04
Bundler 2.2.25: `materialize': wrong number of arguments (given 2, expected 1) (ArgumentError)


vmsp:
```
vitor@MacBook-Pro project % bundle exec tapioca sync
Removing RBI files of gems that have been removed:

bundler: failed to load command: tapioca (/Users/vitor/.rbenv/versions/2.6.6/bin/tapioca)
Traceback (most recent call last):
	49: from /Users/vitor/.rbenv/versions/2.6.6/bin/bundle:23:in `<main>'
	48: from /Users/vitor/.rbenv/versions/2.6.6/bin/bundle:23:in `load'
	47: from /Users/vitor/.rbenv/versions/2.6.6/lib/ruby/gems/2.6.0/gems/bundler-2.2.25/exe/bundle:37:in `<top (required)>'
	46: from /Users/vitor/.rbenv/versions/2.6.6/lib/ruby/gems/2.6.0/gems/bundler-2.2.25/lib/bundler/friendly_errors.rb:128:in `with_friendly_errors'
	45: from /Users/vitor/.rbenv/versions/2.6.6/lib/ruby/gems/2.6.0/gems/bundler-2.2.25/exe/bundle:49:in `block in <top (required)>'
	44: from /Users/vitor/.rbenv/versions/2.6.6/lib/ruby/gems/2.6.0/gems/bundler-2.2.25/lib/bundler/cli.rb:25:in `start'
	43: from /Users/vitor/.rbenv/versions/2.6.6/lib/ruby/gems/2.6.0/gems/bundler-2.2.25/lib/bundler/vendor/thor/lib/thor/base.rb:485:in `start'
	42: from /Users/vitor/.rbenv/versions/2.6.6/lib/ruby/gems/2.6.0/gems/bundler-2.2.25/lib/bundler/cli.rb:31:in `dispatch'
	41: from /Users/vitor/.rbenv/versions/2.6.6/lib/ruby/gems/2.6.0/gems/bundler-2.2.25/lib/bundler/vendor/thor/lib/thor.rb:392:in `dispatch'
	40: from /Users/vitor/.rbenv/versions/2.6.6/lib/ruby/gems/2.6.0/gems/bundler-2.2.25/lib/bundler/vendor/thor/lib/thor/invocation.rb:127:in `invoke_command'
[...]

PR #446, opened 2021-08-19
Add support for bundler 2.2.25+

RyanBrushett:
### Motivation

Add support for Bundler 2.2.25 and up.
Fix https://github.com/Shopify/tapioca/issues/413

### Implementation

- Keep support for < 2.2.25 by leaving that codepath mostly unchanged.
- Use the current version of the `materialize` method for bundler versions 2.2.25 and up.
- I refactored a bit. If this makes it more confusing, let me know and I can rearrange some things. Also: naming variables is hard.

### Tests

- I added a 2.2.22 and 2.2.25 bundler versions to the test matrix, otherwise relying on the tests that were there.

### DIFF
--- a/lib/tapioca/gemfile.rb
+++ b/lib/tapioca/gemfile.rb
@@ -53,19 +53,26 @@ module Tapioca
 
     sig { returns([T::Array[Gem], T::Array[String]]) }
     def load_dependencies
-      deps = definition.locked_gems.dependencies.values
-
-      missing_specs = T::Array[String].new
-
-      dependencies = definition
-        .resolve
-        .materialize(deps, missing_specs)
+      materialized_dependencies, missing_specs = materialize_deps
+      dependencies = materialized_dependencies
         .map { |spec| Gem.new(spec) }
         .reject { |gem| gem.ignore?(dir) }
         .uniq(&:rbi_file_name)
         .sort_by(&:rbi_file_name)
-
       [dependencies, missing_specs]
+    end
+
+    sig { returns([T::Array[::Gem::Specification], T::Array[String]]) }
+    def materialize_deps
+      deps = definition.locked_gems.dependencies.values
+      missing_specs = T::Array[String].new
+      materialized_dependencies = if definition.resolve.method(:materialize).arity == 1 # Support bundler >= v2.2.25
+        definition.resolve.materialize(deps)
+          .tap { |md| missing_specs = T.cast(md.missing_specs, T::Array[String]) }
+      else
+        definition.resolve.materialize(deps, missing_specs)
+      end
+      [materialized_dependencies, missing_specs]
     end
 
     sig { returns(Bundler::Runtime) }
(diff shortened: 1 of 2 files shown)


RyanBrushett:
Sidenote: Do we need the actions to run on Push and PR or would just having it on Push be OK?

vinistock:
`bin/test` uses `bundle exec` to execute tests. I don't think that is picking up the right bundler version. You'd need to change that CI step to be
```yaml
bundle _${{ bundler_version }}_ exec ...
```
Also, you don't need quotes in the matrix versions.

RyanBrushett:
Tidied up! Thanks Vini.

Those same tests are still failing, will keep digging!
I'm gonna remove the `pull_request` CI option. If someone's strongly opposed, I'll tack it back on.

lib/tapioca/gemfile.rb
```diff
@@ -53,21 +53,29 @@ def require
...
-
       [dependencies, missing_specs]
     end
 
+    sig { returns([T::Array[::Gem::Specification], T::Array[String]]) }
+    def materialize_deps
+      deps = definition.locked_gems.dependencies.values
+      missing_specs = T::Array[String].new
+      materialized_dependencies = T::Array[Gem].new
+      if ::Gem.loaded_specs["bundler"].version < ::Gem::Version.create("2.2.25")
```
paracycle, line 70: Can we just check for the arity of the method instead of the Bundler version, please?

RyanBrushett: Ah sure, good call.

lib/tapioca/gemfile.rb
```diff
@@ -53,21 +53,29 @@ def require
...
+    def materialize_deps
+      deps = definition.locked_gems.dependencies.values
+      missing_specs = T::Array[String].new
+      materialized_dependencies = T::Array[Gem].new
+      if definition.resolve.method(:materialize).arity == 1 # Support bundler versions >= 2.2.25
+        materialized_dependencies = definition.resolve.materialize(deps)
+        missing_specs = materialized_dependencies.missing_specs
+      else
+        materialized_dependencies = definition.resolve.materialize(deps, missing_specs)
+      end
```
paracycle, line 75:
```suggestion
      materialized_dependencies = if definition.resolve.method(:materialize).arity == 1 # Support bundler versions >= 2.2.25
        definition.resolve.materialize(deps)
          .tap { |md| missing_specs = md.missing_specs }
      else
        definition.resolve.materialize(deps, missing_specs)
      end
```

spec/tapioca/cli/gem_spec.rb
```diff
@@ -51,6 +51,7 @@ module Baz; end
       class Baz::Role
         include ::SmartProperties
         extend ::SmartProperties::ClassMethods
+        extend ::SmartProperties::ModuleMethods
```
paracycle, line 54: How come these changed when our Gemfile.lock has not changed? This seems to be a problem with how we are testing. We really should not be having smart properties' version change from underneath us.

paracycle: Ok, I fixed this with #449, so rebasing on `main` should render this unnecessary.

paracycle (approved):

RyanBrushett:
I'll fix the branch protection and ⛵ .

PR #450, opened 2021-08-23
Remove pull_request from ci options

RyanBrushett:
### Motivation

I originally had this as part of https://github.com/Shopify/tapioca/pull/446 but I extracted it to make backporting easier.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,6 @@
 name: CI
 
-on: [push, pull_request]
+on: push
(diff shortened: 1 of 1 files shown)


KaanOzkan (approved):

rafaelfranca:
Doesn't this break when the PR is open by someone that can't push to this repository? https://github.com/Shopify/tapioca/pull/473 seems to be a case of a PR that don't have CI because the GitHub action was not triggered

PR #453, opened 2021-08-23
Merge generated RBIs with gem RBIs

### LINKED ISSUE #139: [Feature request] Include sigs from gem rbi files
Hi! I've been really happy using tapioca for my company's codebase, but have run into a snag on more recent versions of sorbet.

We have an internal gem that contains our protobuf definitions and is used by a variety of other apps. If your not familiar, protobufs have a compilation step that take `*.proto` files which define protobuf messages (essentially structs) and output `*_pb.rb` files that define corresponding ruby classes. I've been using [this protoc plugin](https://github.com/coinbase/protoc-gen-rbi) to generate rbi files at compile time and putting them in the gem's `rbi/` folder.

This had been working pretty well; sorbet picks up the rbi files from the gem and typechecks the proto classes in our apps. The rbis stay in sync because they're auto-generated at compile time. 

Unfortunately, a [recent change in sorbet](https://github.com/sorbet/sorbet/commit/998805307213c0649f73f1d97b2240eda331a426) prevents sorbet from using gem rbis when tapioca is being used, with the rationale that tapioca copies sigs from gems when you run `tapioca generate`. As far as I can tell, this is the case for inline sigs, but does not cover our use case of generated `rbi` files.

I've read some of the discussion on the sorbet slack and understand that this hasn't been a priority to implement, but I want to present this use case where copying gem rbis would be useful.

If you think this is a worthwhile feature and feasible, I'm happy to spend some time working on it.

mojanjz:
### Motivation
closes #139

pending a new release of the RBI gem. I'm pointing to a local clone of the gem for now.

As explained in this [comment](https://github.com/Shopify/tapioca/issues/139#issuecomment-898006987), we want to be able to merge generated RBIs with RBIs that are included in the gem. In a case of a conflict, where the gem has both the `/rbi` folder and signatures in the code base, the generated RBI takes precedence over the included RBI.

### Implementation

Added an extra step to `compile_gem_rbi` to merge the generated content with the copied RBIs if the `rbi` folder exists.

### Tests
Added tests to `gem_spec` not sure if we are maintaining `generator_spec` as well or not. If yes, I'll add tests to that as well.


lib/tapioca/generator.rb
```diff
@@ -559,6 +560,38 @@ def compile_gem_rbi(gem)
...
+      copied_rbi = copy_gem_rbi(gem)
+      display_conflict_warnings(generated_rbi, copied_rbi)
+      final_rbi = RBI::Rewriters::Merge.merge_trees(generated_rbi, copied_rbi, keep: RBI::Rewriters::Merge::Keep::LEFT)
+      final_rbi.string
+    end
+
+    sig { params(gem: Gemfile::Gem).returns(RBI::Tree) }
+    def copy_gem_rbi(gem)
+      rbi = RBI::Tree.new
+      Dir.glob("#{gem.full_gem_path}/rbi/*.rbi").each do |file|
```
paracycle, line 577: Can we move the logic of finding all the exported RBI files from a gem into the `Gem` class? Something like:
```suggestion
      gem.exported_rbi_files.each do |file|
```
would be much easier to read and would be more encapsulated.

lib/tapioca/generator.rb
```diff
@@ -550,6 +550,7 @@ def compile_gem_rbi(gem)
         say("Done (empty output)", :yellow)
       else
         content << rbi_body_content
+        content = merge_with_copied_rbi(gem, content)
```
paracycle, line 553: While this is fine, I'd rather not do this at this stage. By the time `compiler.compile(gem)` is done, the RBI tree it has built has already been rewritten and transformed. That means, we need to parse it again to do the merge. But, I also suspect that we might be losing some of our transformations we are doing inside `.compile` to serialize the RBI file as a result as well.

In my opinion, it would be cleaner for `compiler.compile(gem, exported_rbi_tree)` be the call and for `SymbolGenerator#generate` to do the actual tree merges, and then apply the RBI transformations only after the merge is done.

I am not sure how we can return/display the merge conflicts in that case, but maybe conflicts can be yielded to a block supplied to the `.compile` method.

So, roughly this `compile_gem_rbi` method might look something like this:
```ruby
def compile_gem_rbi(gem)
  exported_rbi_tree = copy_gem_rbi(gem)
  compiler = Compilers::SymbolTableCompiler.new

  # some other code....

  rbi_body_content = compiler.compile(gem, exported_rbi_tree) do |conflicts|
    # output conflicts as warning here
  end
  # the rest of the method
end
```

What would be even much better, would be if `Gem` actually had a way of returning the exported RBI tree. Something like `gem.exported_rbi_tree` reader that memoized the result. If that existed, then you wouldn't even need to pass an extra argument to `compile`, `SymbolGenerator#generate` could just ask `gem` for it.

lib/tapioca/generator.rb
```diff
@@ -559,6 +560,38 @@ def compile_gem_rbi(gem)
...
+      rbi
+    end
+
+    sig { params(generated_rbi: RBI::Tree, copied_rbi: RBI::Tree).void }
+    def display_conflict_warnings(generated_rbi, copied_rbi)
+      rewriter = RBI::Rewriters::Merge.new
+      rewriter.merge(generated_rbi)
+      conflicts = rewriter.merge(copied_rbi)
+      indent do
+        say("\n#{conflicts.join("\n")}") unless conflicts.empty?
```
mojanjz, line 591: even with using this indent command, I wasn't able to make it match the rest of the commands. Any suggestions?

lib/tapioca/generator.rb
```diff
@@ -559,6 +560,38 @@ def compile_gem_rbi(gem)
...
+      final_rbi = RBI::Rewriters::Merge.merge_trees(generated_rbi, copied_rbi, keep: RBI::Rewriters::Merge::Keep::LEFT)
+      final_rbi.string
+    end
+
+    sig { params(gem: Gemfile::Gem).returns(RBI::Tree) }
+    def copy_gem_rbi(gem)
+      rbi = RBI::Tree.new
+      Dir.glob("#{gem.full_gem_path}/rbi/*.rbi").each do |file|
+        gem_rbi = RBI::Parser.parse_file(file)
+        display_conflict_warnings(rbi, gem_rbi)
```
paracycle, line 579: Can we add some conflicts to the `bar` gem RBI files, so that we can expect to see these warnings in the test output, as well?

paracycle (changes requested):

spec/tapioca/cli/gem_spec.rb
```diff
@@ -105,6 +128,32 @@ def fizz; end
...
+          assert_path_exists("#{outdir}/bar@0.3.0.rbi")
+          assert_equal(BAR_RBI, File.read("#{outdir}/bar@0.3.0.rbi"))
+        end
+
+        it "must generate a gem RBI and resolves conflicts with copied gem RBIs by keeping the generated RBI" do
+          output = execute("gem", "fizz")
+
+          assert_includes(output, <<~OUTPUT)
+            Processing 'fizz' gem:
+              Compiling fizz, this may take a few seconds...#{"     "}
```
Morriar, line 148: What is this `#{"     "}`? Is it from the `indent` block?

spec/support/repo/sorbet/config
```diff
@@ -1,2 +0,0 @@
---dir
```
Morriar, line 1: I don't think this changes should be committed?

spec/support/repo/sorbet/tapioca/require.rb
```diff
@@ -1,4 +0,0 @@
-# typed: true
```
Morriar, line 1: I don't think this changes should be committed?

lib/tapioca/generator.rb
Morriar: Do you have an example of how you want to display this?

Note: I'm not sure a simple white line is enough for the warning to be visible. The output is already verbose and the line might be lost in the middle of the other lines. You could be using some colours as seen here: https://github.com/Shopify/tapioca/blob/92f2a10cf7f603ec1a00d77ad7624c2735542d28/lib/tapioca/generator.rb#L52

Maybe even a prefix such has `<yellow+bold>Warning:</yellow+bold> rest of the line`.

lib/tapioca/generator.rb
```diff
@@ -559,6 +565,34 @@ def compile_gem_rbi(gem)
       end
     end
 
+    sig { params(gem: Gemfile::Gem, rbi: RBI::Tree).returns(RBI::Tree) }
+    def merge_with_exported_rbi(gem, rbi)
+      return rbi unless gem.has_exported_rbi_files?
+
+      copied_rbi = gem.exported_rbi_tree do |conflicts|
+        say("\n #{conflicts.join("\n")}", :yellow)
+      end
```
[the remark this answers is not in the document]
Morriar: Maybe we should rename `exported_rbi_tree` in something more explicit or add a few comments here?

RyanBrushett:
Rebased. Will keep nudging this along as I get time.

lib/tapioca/gemfile.rb
```diff
@@ -145,6 +145,32 @@ def contains_path?(path)
...
+
+      sig { params(block: T.proc.params(arg: T::Array[RBI::Rewriters::Merge::Conflict]).void).returns(RBI::Tree) }
+      def exported_rbi_tree(&block)
+        rbi = RBI::Tree.new
+        rewriter = RBI::Rewriters::Merge.new
+        conflicts = []
+        exported_rbi_files.each do |file|
+          gem_rbi = RBI::Parser.parse_file(file)
+          rbi = RBI::Rewriters::Merge.merge_trees(rbi, gem_rbi, keep: RBI::Rewriters::Merge::Keep::NONE)
+          conflict = rewriter.merge(gem_rbi)
```
Morriar, line 166: Are we merging this thing twice?

`RBI::Rewriters::Merge.merge_trees` already calls `RBI::Rewriters::Merge.new` and `RBI::Rewriters::Merge#merge`.

I believe the whole method could be rewritten to something like this: 

```rb
      sig { returns(RBIMergeResult) }
      def exported_rbi_tree
        rewriter = RBI::Rewriters::Merge.new(keep: RBI::Rewriters::Merge::Keep::NONE)
        all_conflicts = []

        exported_rbi_files.each do |file|
          gem_rbi = RBI::Parser.parse_file(file)
          file_conflicts(rewriter.merge(gem_rbi))
          all_conflicts.concat(file_conflicts)
        end

        RBIMergeResult.new(tree: rewriter.tree, conflicts: conflicts)
      end
```

But looking at this I think I should just make `RBI::Rewriters` API better so it can handle multiple files merging and conflicts better.

lib/tapioca/generator.rb
```diff
@@ -534,12 +536,16 @@ def compile_gem_rbi(gem)
       say("Compiling #{gem_name}, this may take a few seconds... ")
 
       strictness = config.typed_overrides[gem.name] || "true"
-      rbi_body_content = compiler.compile(gem)
+      rbi = compiler.compile(gem)
+      rbi.comments << RBI::Comment.new("typed: #{strictness}") unless rbi.empty?
+      rbi.nest_singleton_methods!
```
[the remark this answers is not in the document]
Morriar: I think we can remove this as well? It will be nested by the rewriter called below.

lib/tapioca/generator.rb
```diff
@@ -534,12 +536,16 @@ def compile_gem_rbi(gem)
...
+      rbi = Tapioca::RBIMerger.merge_with_exported_rbi!(rbi, gem, Thor::Shell::Color.new)
```
Morriar, line 542: Why the `!` suffix? None of the argument will be modified right?

Morriar: I'm not sure to understand why the `merge_with_exported_rbi` is extracted in a separate class here. The separation seems very artificial. For proof I'd cite the strong coupling with `Shell::Color.new`.

As I see it it would make more sense to keep it here, so it would be moved to the `Gem` generator once #504 is merged? 

If we do extract it as a separate service, it should also be used when we merge the gem's RBIs from the `rbi/` dir.

lib/tapioca/generator.rb
```diff
@@ -559,6 +565,15 @@ def compile_gem_rbi(gem)
       end
     end
 
+    sig { params(rbi: RBI::Tree).returns(String) }
+    def transform_rbi(rbi)
```
Morriar, line 569: What do you think about moving this to the tree here: https://github.com/Shopify/tapioca/blob/main/lib/tapioca/rbi_ext/model.rb.

sorbet/config
```diff
@@ -1,4 +1,6 @@
 --dir
 .
 --ignore=/vendor
+--ignore=/spec/support/gems/fizz
+--ignore=/spec/support/gems/bar
```
Morriar, line 5: Do we want to ignore the full gems or just the `rbi/` directories?

```suggestion
--ignore=/spec/support/gems/fizz/rbi
--ignore=/spec/support/gems/bar/rbi
```

lib/tapioca/rbi_merger.rb
```diff
@@ -0,0 +1,34 @@
...
+      end
+
+      private
+
+      sig { params(generated_rbi: RBI::Tree, copied_rbi: RBI::Tree, shell: Thor::Shell::Color).void }
+      def display_conflict_warnings(generated_rbi, copied_rbi, shell)
+        rewriter = RBI::Rewriters::Merge.new
+        rewriter.merge(generated_rbi)
+        conflicts = rewriter.merge(copied_rbi)
+        shell.say(conflicts.join("\n"), :yellow) unless conflicts.empty?
```
Morriar, line 30: You have access to [both nodes](https://github.com/Shopify/rbi/blob/main/lib/rbi/rewriters/merge_trees.rb#L85-L86) and so their [location](https://github.com/Shopify/rbi/blob/main/lib/rbi/model.rb#L15). I believe we could do better and tell the user which files/lines are creating the conflict.

spec/tapioca/cli/gem_spec.rb
```diff
@@ -120,6 +144,32 @@ def fizz; end
...
+          assert_equal(BAR_RBI, File.read("#{outdir}/bar@0.3.0.rbi"))
+        end
+
+        it "must generate a gem RBI and resolves conflicts with exported gem RBIs by keeping the generated RBI" do
+          output = execute("gem", "fizz")
+
+          compiling_spec = "Compiling fizz, this may take a few seconds..."
+          assert_includes(output, compiling_spec)
+
+          conflict_spec = "Conflicting definitions for `::Fizz::<self>#baz(a, b)`"
```
Morriar, line 166: A more precise message with [these infos](https://github.com/Shopify/rbi/blob/main/lib/rbi/loc.rb#L8-L12) would be way more useful to solve the problems.

spec/tapioca/compilers/symbol_table_compiler_spec.rb
```diff
@@ -40,7 +40,12 @@ def compile
       spec = Bundler::StubSpecification.from_stub(stub)
       gem = Tapioca::Gemfile::GemSpec.new(spec)
 
-      Tapioca::Compilers::SymbolTableCompiler.new.compile(gem)
+      rbi = Tapioca::Compilers::SymbolTableCompiler.new.compile(gem)
+      rbi.nest_singleton_methods!
+      rbi.nest_non_public_methods!
+      rbi.group_nodes!
+      rbi.sort_nodes!
+      rbi.string
```
Morriar, line 48: Ditto about moving this here: https://github.com/Shopify/tapioca/blob/main/lib/tapioca/rbi_ext/model.rb.

lib/tapioca/generator.rb
RyanBrushett: > Why the ! suffix? None of the argument will be modified right?

Oversight on my part. At some point it was calling a bang method but indeed, not anymore. I can definitely just move it back to the other file though 👍

lib/tapioca/gemfile.rb
```diff
@@ -145,6 +145,32 @@ def contains_path?(path)
...
+          conflicts << conflict unless conflict.blank?
```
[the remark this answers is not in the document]
paracycle: Actually I think this should be a `reduce` operation.

lib/tapioca/generator.rb
```diff
@@ -559,6 +565,31 @@ def compile_gem_rbi(gem)
...
+
+      RBI::Rewriters::Merge.merge_trees(rbi, copied_rbi, keep: RBI::Rewriters::Merge::Keep::LEFT)
+    end
+
+    sig { params(generated_rbi: RBI::Tree, copied_rbi: RBI::Tree).void }
+    def display_conflict_warnings(generated_rbi, copied_rbi)
+      rewriter = RBI::Rewriters::Merge.new
+      rewriter.merge(generated_rbi)
+      conflicts = rewriter.merge(copied_rbi)
+      say(conflicts.join("\n"), :yellow) unless conflicts.empty?
```
RyanBrushett, line 590: The fact that this and the method above now report conflicts in a way that is formatted differently and containing different information needs to be cleaned up certainly.
I'm away for a while so left this commit as a WIP.

vinistock (approved):

lib/tapioca/cli.rb
```diff
@@ -143,6 +143,10 @@ def dsl(*constants)
     option :doc,
       type: :boolean,
       desc: "Include YARD documentation from sources when generating RBIs. Warning: this might be slow"
+    option :exported_gem_rbis,
+      type: :boolean,
+      default: true,
```
paracycle, line 148: Can we drop the default value from here? If we have this, we will get into the #566 again, where adding a `exported_gem_rbis: false` in the config file will have no effect.
```suggestion
```

paracycle (changes requested):

paracycle (approved):

PR #463, opened 2021-08-27
AASM generator

kddnewton:
This commit adds a generator for the AASM state machine gem. AASM allows you to define a state machine with states, events, and transitions. Each of these generates various things in the owning class.

* For each state `xxx`
  * A constant `STATE_XXX` with the value that is stored in the state column
  * A method `xxx?` that allows you to query for if the owning class is in that state
* For each event `xxx`
  * A method `xxx` that triggers that event
  * A method `xxx!` that triggers that event and potentially raises an error
  * A method `xxx_without_validation!` that triggers that event and skips validation
  * A method `may_xxx?` that runs the validation but does not trigger that event

This gem relies very heavily on `instance_eval`. For example, in the follow snippet:

```ruby
class StateMachine
  include AASM
  extend T::Sig

  aasm do
    # This block is evaluated in the context of a new AASM::Base object.
    # We _could_ shim this class to get the right binding for the aasm
    # method, but see below.

    state :sleeping, initial: true
    state :running, :cleaning

    event :run do
      # This block is evaluated in the context of a new
      # AASM::Core::Event object. We could also shim this class to get
      # the right binding for the event method, but see below.

      before do
        # This block is evaluated all of the way back up in the
        # StateMachine class, which means that before_run here will
        # actually evaluate properly. So even if we were to shim the
        # aasm and event methods, there's no way to shim it such that
        # this method call will evaluate without either
        # T.bind(T.attached_class) which doesn't exist or forcing the
        # consumers to always T.bind(self, StateMachine) within each
        # callback.
        before_run
      end

      transitions from: :sleeping, to: :running
    end
  end

  private

  sig { void }
  def before_run; end
end
```

To get around the problems mentioned in the snippet above, we generate a private PrivateAASMMachine constant within each class that includes a state machine where we properly bind the callback methods. You can see the spec for a detailed layout of what gets generated.

@RyanBrushett I ended up keeping around the private classes just because they made it more of a complete implementation here. Appears to work.

### DIFF
--- a/spec/tapioca/compilers/dsl/aasm_spec.rb
+++ b/spec/tapioca/compilers/dsl/aasm_spec.rb
@@ -22,19 +22,6 @@ class Tapioca::Compilers::Dsl::AASMSpec < DslSpec
   end
 
   describe("#decorate") do
-    it("generates empty RBI file if this is no state machine") do
-      add_ruby_file("content.rb", <<~RUBY)
-        class StateMachine
-        end
-      RUBY
-
-      expected = <<~RBI
-        # typed: strong
-      RBI
-
-      assert_equal(expected, rbi_for(:StateMachine))
-    end
-
     it "generates empty RBI when AASM is included but no AASM call has been made" do
       add_ruby_file("content.rb", <<~RUBY)
         class StateMachine
--- a/lib/tapioca/compilers/dsl/aasm.rb
+++ b/lib/tapioca/compilers/dsl/aasm.rb
@@ -49,15 +49,9 @@ module Tapioca
             T::Array[String]
           )
 
-        sig { override.params(root: RBI::Tree, constant: Module).void }
+        sig { override.params(root: RBI::Tree, constant: T.all(::AASM::ClassMethods, Class)).void }
         def decorate(root, constant)
-          # Make sure the thing that's being passed in here actually included
-          # AASM
-          return unless constant.respond_to?(:aasm)
-
-          # Using T.unsafe here because at this point we know that the object
-          # includes AASM
-          aasm = T.unsafe(constant).aasm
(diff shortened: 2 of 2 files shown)


paracycle (approved):
This looks good to me, I just have a small test addition suggestion.

RyanBrushett (approved):
Tested this IRL with a couple of projects and it worked great. Thanks!

spec/tapioca/compilers/dsl/aasm_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+      add_ruby_file("content.rb", <<~RUBY)
+        class StateMachine
+          include AASM
+        end
+      RUBY
+
+      expected = <<~RBI
+        # typed: strong
+      RBI
+    end
```
RyanBrushett, line 48: @paracycle is this what you had in mind?

spec/tapioca/compilers/dsl/aasm_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+      RUBY
+
+      expected = <<~RBI
+        # typed: strong
+      RBI
+
+      assert_equal(expected, rbi_for(:StateMachine))
+    end
+
+    it "generates RBI when AASM is included but no AASM call has been made" do
```
RyanBrushett, line 38: @paracycle Is this what you had in mind?

RyanBrushett: My instinct says this should be empty RBI in this case... 👀

kddnewton:
Thanks for the follow-up @RyanBrushett it was so nice to come back from vacation and see this merged :)

PR #464, opened 2021-08-30
Disambiguate Struct definitions from Struct subclasses

paracycle:
### Motivation

Struct definition is currently failing for classes defined like:
```ruby
class Foo < Struct
end
```

which end up not defining a struct definition, but subclassing `Struct` instead. Only instances of `Foo` will be struct definitions themselves. Thus, Tapioca fails when trying to call `Foo.members`, since the method does not exist on `Foo`.

### Implementation

What separates just subclassing `Struct` and a struct definition is the existence of the `members` method on the constant.

### Tests

Expanded existing tests to check for this case.

### DIFF
--- a/lib/tapioca/compilers/symbol_table/symbol_generator.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_generator.rb
@@ -304,6 +304,19 @@ module Tapioca
 
         sig { params(constant: Class).returns(T::Boolean) }
         def struct_definition?(constant)
+          # Make sure this is constant defined by `Struct.new` and not just
+          # by subclassing `Struct`.
+          #
+          # class Foo < Struct
+          # end
+          #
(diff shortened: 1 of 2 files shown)


vinistock (approved):

KaanOzkan (approved):

lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -304,6 +304,19 @@ def compile_type_variable_declarations(tree, constant)
 
         sig { params(constant: Class).returns(T::Boolean) }
         def struct_definition?(constant)
+          # Make sure this is constant defined by `Struct.new` and not just
```
Morriar, line 307:
```suggestion
          # Make sure this constant is defined by `Struct.new` and not just
```
?

PR #466, opened 2021-08-30
Improve return type of `materialize_deps`

paracycle:
### Motivation

The existing return type was problematic for the generated Tapioca RBI file since Tapioca gem generator drops the leading `::` anchors from constants declared in `sig`s. Thus `Gem::Specification` ends up pointing to `Tapioca::Gemfile::Gem` instead.

### Implementation

Referring to the `Spec` alias is more correct.

### Tests

No extra tests.

### DIFF
--- a/lib/tapioca/gemfile.rb
+++ b/lib/tapioca/gemfile.rb
@@ -62,7 +62,7 @@ module Tapioca
       [dependencies, missing_specs]
     end
 
-    sig { returns([T::Array[::Gem::Specification], T::Array[String]]) }
+    sig { returns([T::Array[Spec], T::Array[String]]) }
     def materialize_deps
       deps = definition.locked_gems.dependencies.values
       missing_specs = T::Array[String].new


KaanOzkan (approved):
These are not identical due to alias covering more types. I assume we are okay with that, I don't have enough context here.

paracycle:
> These are not identical due to alias covering more types. I assume we are okay with that, I don't have enough context here.

The return type is actually `Bundler::SpecSet` but the RBI in Sorbet seems to have messed up the declaration of that. It is an `Enumerable`, so needs to define an `Elem` type member, but it is not marked as generic, so does not have the `[]` method, etc. But, really `SpecSet` is an `Array` with multiple kinds of things, each of which can be a `Bundler::StubSpecification`, `Bundler::RemoteSpecification`, etc. So `T::Array[Spec]` works just fine and we don't have a need for much more correctness here either.

PR #468, opened 2021-08-30
Halt DSL generation if there are migrations pending

### LINKED ISSUE #443: Active Record generators should refuse to run if there are pending migrations
Since AR generators reflect the state of the database, what Tapioca produces could end up being wrong with respect to the state of the application (which could be assuming that new columns, associations are in place). We've observed this to be a common case of confusion, where people run DSL generation without having ran the migrations fully and expect different results.

The solution is to explicitly check for pending migrations and refuse to run the whole DSL pipeline if there are any pending migrations.

vinistock:
Closes #443

### Motivation

If there are migrations pending, then the output of DSL generators might be incorrect. The best option is to halt DSL generation and prompt the user to run the migrations.

### Implementation

I used the built-in rake task from Rails to check if there are pending migrations. The task already aborts the process and prints the list of migrations if there are any.

Using the task also means that this should work regardless of Rails version, since we're not trying to implement the pending migration checker on our end.

This is what the ouput looks like
Screen Shot 2021-08-30 at 4 28 31 PM

### Tests

Included tests.


lib/tapioca/generator.rb
```diff
@@ -692,5 +694,13 @@ def report_diff_and_exit_if_out_of_date(diff, command)
         exit(1)
       end
     end
+
+    sig { void }
+    def abort_if_pending_migrations!
+      return unless defined?(Rails)
+
+      Rails.application.load_tasks
```
Morriar, line 702: What if `Rails` is defined but we're not in an application? 

Should we protect this behind a `return unless File.exist?("config/application.rb")` like its done for [`load_rails`](https://github.com/Shopify/tapioca/blob/ccd7c75305f4c3c5d0ad7f10d1bd63544f14e565/lib/tapioca/loader.rb#L31)?

paracycle: Yeah. For example, that is exactly the case in our support repo. `Rails` is defined by not in the way you think. So the added test really fails in this PR.

Does eager loading the Rails application not load tasks, already? Do we really need this line?

vinistock: I tested with an application, rake tasks are not loaded during eager load. Probably because the tasks as typically invoked from a CLI context and not used from inside the application running. Without that line the conditional below always fails because the task is never defined.

spec/support/repo/lib/database.rb
```diff
@@ -0,0 +1,27 @@
+# typed: true
+# frozen_string_literal: true
+
+# We need this workaround otherwise the `todo_spec` thinks
+# Rake::DSL is a todo symbol
```
Morriar, line 5: I think you can put the definition here: https://github.com/Shopify/tapioca/blob/main/spec/support/repo/config/shim.rbi

vinistock: Nice! I didn't know it was there.

PR #472, opened 2021-08-31
Fix root namespace

KaanOzkan:

### DIFF
--- a/lib/tapioca/sorbet_ext/name_patch.rb
+++ b/lib/tapioca/sorbet_ext/name_patch.rb
@@ -10,6 +10,7 @@ module T
         end
       end
 
+      T::Types::Simple::Private::Pool.instance_variable_set(:@cache, ObjectSpace::WeakMap.new)
       prepend NamePatch
     end
   end
--- a/lib/tapioca/sorbet_ext/name_patch.rb
+++ b/lib/tapioca/sorbet_ext/name_patch.rb
@@ -6,7 +6,7 @@ module T
     class Simple
       module NamePatch
         def name
-          @name ||= ::Tapioca::Reflection.name_of(@raw_type).freeze
+          @name ||= ::Tapioca::Reflection.qualified_name_of(@raw_type).freeze
         end
       end
 
--- a/spec/tapioca/compilers/dsl/sidekiq_worker_spec.rb
+++ b/spec/tapioca/compilers/dsl/sidekiq_worker_spec.rb
@@ -85,13 +85,13 @@ class Tapioca::Compilers::Dsl::SidekiqWorkerSpec < DslSpec
         # typed: strong
 
         class NotifierWorker
-          sig { params(customer_id: Integer).returns(String) }
+          sig { params(customer_id: ::Integer).returns(String) }
           def self.perform_async(customer_id); end
 
-          sig { params(interval: T.any(DateTime, Time), customer_id: Integer).returns(String) }
+          sig { params(interval: T.any(DateTime, Time), customer_id: ::Integer).returns(String) }
           def self.perform_at(interval, customer_id); end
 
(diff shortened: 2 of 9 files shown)


spec/tapioca/compilers/symbol_table_compiler_spec.rb
```diff
@@ -2712,6 +2712,8 @@ def c
       RUBY
 
       output = template(<<~RBI)
+        ::Foo::FooAttachedClass = Foo::FooAttachedClass
```
KaanOzkan, line 2715: Extraneous constant definition

paracycle: Ah, I think I know why this might be. I'll have a look at this tomorrow.

lib/tapioca/sorbet_ext/name_patch.rb
```diff
@@ -6,10 +6,11 @@ module Types
     class Simple
       module NamePatch
         def name
-          @name ||= ::Tapioca::Reflection.name_of(@raw_type).freeze
+          @name ||= ::Tapioca::Reflection.qualified_name_of(@raw_type).freeze
         end
       end
 
+      T::Types::Simple::Private::Pool.instance_variable_set(:@cache, ObjectSpace::WeakMap.new)
```
paracycle, line 13: The problem apparently was that existing instances of `T::Types::Simple` has memoized their names in the `@name` ivar, so they were holding on to that even after we prepend the patch.

The fix is to use a different memoizing ivar in the prepended module, so that they all start fresh. However, that causes an infinite loop descending into the `qualified_name_of` method on `Reflection` since the method has a signature which tries to do the same thing recursively.

Thus, we need a copy of the `qualified_name_of` method here without a signature to fix everything:

```suggestion
          @qualified_name ||= qualified_name_of(@raw_type).freeze
        end

        private

        NAME_METHOD = Module.instance_method(:name)

        def qualified_name_of(constant)
          name = NAME_METHOD.bind(constant).call
          return if name.nil?

          if name.start_with?("::")
            name
          else
            "::#{name}"
          end
        end
      end
```

KaanOzkan: This fixed the `T.nilable` conversions. However, there are still some types that are unaffected by the patch in the tests. Not sure if it's a blocker for merging this PR.

spec/tapioca/compilers/dsl/sidekiq_worker_spec.rb
```diff
@@ -85,13 +85,13 @@ def perform(customer_id)
         # typed: strong
 
         class NotifierWorker
-          sig { params(customer_id: Integer).returns(String) }
+          sig { params(customer_id: ::Integer).returns(String) }
           def self.perform_async(customer_id); end
 
-          sig { params(interval: T.any(DateTime, Time), customer_id: Integer).returns(String) }
+          sig { params(interval: T.any(DateTime, Time), customer_id: ::Integer).returns(String) }
```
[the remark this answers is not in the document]
paracycle: Hmm, this is interesting. I would have expected this to have been fixed too.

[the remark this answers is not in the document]
jeffcarbs: The reason these aren't changing is because they're hard-coded in the code: https://github.com/Shopify/tapioca/blob/main/lib/tapioca/compilers/dsl/sidekiq_worker.rb#L59.

The `Integer` param is derived from the live ruby code:
https://github.com/Shopify/tapioca/blob/main/lib/tapioca/compilers/dsl/sidekiq_worker.rb#L59

I wonder if there should be some helper that allows us to specify these manual types via code and have it get compiled to a string.

PR #473, opened 2021-09-01
Add readme directions on migrating to Tapioca

peternixey:
### Motivation
On initially installing Tapioca I had no idea that it needed a migration in order to work from an existing Sorbet installation. This signals that need directly from the Readme

### Implementation
Updated readme

### Tests
n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Tapioca helps simplify your setup too. Gems such as `sorbet-typed`, `sorbet-rail
 
 ## 1. Preparing an existing Sorbet-installation for Tapioca
 
-***If you are not yet using Sorbet then skip straight to [step 2](#install_tapioca_and_generate_rbi_files)***
+***If you are not yet using Sorbet then skip straight to [step 2](#2-install-tapioca-and-generate-rbi-files)***
 
 The RBI files that Tapioca generates **should not be added** on top of those that Sorbet has already generated for you. Tapioca needs to run afresh. Tapioca also supercedes the need for gems like `sorbet-rails` or `sorbet-typed`. In order to start using Tapioca you will need to remove these gems and regenerate all of your automatically generated RBI files.
 
@@ -59,7 +59,8 @@ Then run `bundle install`
 bundle exec tapioca init
 
 # Generate RBI files for gems
-# (Generate only for specific gems using `tapioca gem [gem...]`)
+# (Generate only for specific gems using `tapioca gem 
+gem...]`)
 bundle exec tapioca gem
 
 # Generate the list of all unresolved constants
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Then run `bundle install`
 
 ### Generate the RBI files for gems
 
-````bash
+```bash
 # Generate sorbet/config and sorbet/tapioca/require.rb if required
 # (they won't be touched if they already exist)
 bundle exec tapioca init
--- a/README.md
+++ b/README.md
@@ -102,6 +102,8 @@ bin/tapioca dsl
 
 While generating RBI files you might realize that some shims have become obsolete or incorrect. Make sure to edit or remove existing shims that are causing errors. You can read about DSL RBI generators supplied by `tapioca` in [the manual](manual/generators.md).
 
+(Note - there are still a few rough patches around this WRT Rails - [please see this interim fix](https://github.com/Shopify/tapioca/issues/179#issuecomment-738948320))
+
 ### Run Tapioca todo (optional)
 
 (If there are still errors for missing definitions, you may want to try bringing back any shims you saved for later)
--- a/README.md
+++ b/README.md
@@ -59,8 +59,7 @@ Then run `bundle install`
 bundle exec tapioca init
 
 # Generate RBI files for gems
-# (Generate only for specific gems using `tapioca gem 
-gem...]`)
+# (Generate only for specific gems using `tapioca gem gem...]`)
 bundle exec tapioca gem
 
 # Generate the list of all unresolved constants
--- a/README.md
+++ b/README.md
@@ -12,10 +12,35 @@ As yet, no gem exports type information in a consumable format and it would be a
 
 When you run `tapioca sync` in a project, `tapioca` loads all the gems that are in your dependency list from the Gemfile into memory. It then performs runtime introspection on the loaded types to understand their structure and generates an appropriate RBI file for each gem with a versioned filename.
 
-Tapioca helps simplify your setup too. Gems such as `sorbet-typed`, `sorbet-rails` are ways to provide accurate typing information for DSLs and gems. However, Tapioca aims to provide the complete tooling to generating RBIs for gems and DSLs, and does not require being combined with other gems, such as `sorbet-rails`, which can be removed from your gemfile.
+Tapioca helps simplify your setup too. Gems such as `sorbet-typed`, `sorbet-rails` are *localized* solutions that provide typing information for DSLs and gems. However, Tapioca aims to provide the *complete tooling* to generating RBIs for gems and DSLs. Tapioca does not require being combined with other gems, such as `sorbet-rails`, which can be removed from your gemfile.
 
 
-## Installing Tapioca in a project with no existing RBI files (i.e. a fresh Sorbet project)
+### How does tapioca compare to "srb rbi gems" ?
+
+[Please see the detailed answer on our wiki](https://github.com/Shopify/tapioca/wiki/How-does-tapioca-compare-to-%22srb-rbi-gems%22-%3F)
+
+# Installing Tapioca
+
+## 1. Preparing an existing Sorbet-installation for Tapioca
+
+***If you are not yet using Sorbet then skip straight to [step 2](#install_tapioca_and_generate_rbi files)***
+
+The RBI files that Tapioca generates **should not be added** on top of those that Sorbet has already generated for you. Tapioca needs to run afresh. Tapioca also supercedes the need for gems like `sorbet-rails` or `sorbet-typed`. In order to start using Tapioca you will need to remove these gems and regenerate all of your automatically generated RBI files.
+
+### Remove any current RBIs
+
+The easiest way to remove existing RBI files is by removing the entire `sorbet/rbi` folder.
+
+**Note**: if your application has shims (hand written RBI files), keep them saved somewhere in case you need to restore some of them. Chances are, fewer shims will be needed after transitioning to Tapioca. Additionally, don't forget to keep your `sorbet/config` file too. 
+
+### Remove any other solutions
+
+Remove `sorbet-rails` or other such alternative solutions from the `Gemfile`. As mentioned, Tapioca is meant to be used on its own. 
+
+If the previously used alternatives had their own configuration files, you should remove them as well.
+
+## 2. Install Tapioca and generate RBI files
+*Star there in a fresh sorbet installation.*
 
 ### Add Tapioca to your application's `Gemfile`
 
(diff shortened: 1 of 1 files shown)


Morriar (commented):
What do you think about moving this new section right after this one: https://github.com/Shopify/tapioca/blob/8f653b36e4529b93868cf6c07d4c359aa734baee/README.md#how-does-tapioca-compare-to-srb-rbi-gems- ?

README.md
```diff
@@ -6,53 +6,143 @@
 
 Tapioca is a library used to generate RBI (Ruby interface) files for use with [Sorbet](https://sorbet.org). RBI files provide the structure (classes, modules, methods, parameters) of the gem/library to Sorbet to assist with typechecking.
 
+### Why use Tapioca
```
Morriar, line 9:
```suggestion
### Why use Tapioca?
```

README.md
```diff
@@ -6,53 +6,143 @@
...
 
 When you run `tapioca sync` in a project, `tapioca` loads all the gems that are in your dependency list from the Gemfile into memory. It then performs runtime introspection on the loaded types to understand their structure and generates an appropriate RBI file for each gem with a versioned filename.
 
-## Manual gem requires
+Tapioca helps simplify your setup too. Gems such as `sorbet-typed`, `sorbet-rails` are ways to provide accurate typing information for DSLs and gems. However, Tapioca aims to provide the complete tooling to generating RBIs for gems and DSLs, and does not require being combined with other gems, such as `sorbet-rails`, which can be removed from your gemfile.
```
Morriar, line 15: Said like that it sounds like Tapioca does not provide _accurate_ typing information contrary to other solutions?

README.md
```diff
@@ -6,53 +6,143 @@
...
 
-For gems that have a normal default `require` and load all of their constants through such a require, everything works seamlessly. However, for gems that are marked as `require: false` in the Gemfile, or for gems that export optionally loaded types via different requires, where a single require does not load the whole gem code into memory, `tapioca` will not be able to load some of the types into memory and, thus, won't be able to generate complete RBIs for them. For this reason, we need to keep a small external file named `sorbet/tapioca/require.rb` that is executed after all the gems in the Gemfile have been required and before generation of gem RBIs have started. This file is responsible for adding the requires for additional files from gems, which are not covered by the default require.
 
-For example, suppose you are using the class `BetterHtml::Parser` exported from the `better_html` gem. Just doing a `require "better_html"` (which is the default require) does not load that type:
+## Installing Tapioca in a project with no existing RBI files (i.e. a fresh Sorbet project)
```
Morriar, line 18: Let's put back the section `## How does tapioca compare to "srb rbi gems" ?` here, before jumping to the installation procedure.

README.md
```diff
@@ -6,53 +6,143 @@
...
-[2] pry(main)> BetterHtml
-=> BetterHtml
-[3] pry(main)> BetterHtml::Parser
-NameError: uninitialized constant BetterHtml::Parser
-from (pry):3:in `__pry__`
-[4] pry(main)> require 'better_html/parser'
-=> true
-[5] pry(main)> BetterHtml::Parser
-=> BetterHtml::Parser
+### 1. Add this line to your application's `Gemfile`
```
Morriar, line 20:
```suggestion
### Add Tapioca to your application's `Gemfile`
```

README.md
```diff
@@ -6,53 +6,143 @@
...
+
+```ruby
+group :development do
+  gem 'tapioca', require: false
+end
 ```
+and run `bundle install`
 
-In order to make sure that `tapioca` can reflect on that type, we need to add the line `require "better_html/parser"` to the `sorbet/tapioca/require.rb` file. This will make sure `BetterHtml::Parser` is loaded into memory and a type annotation is generated for it in the `better_html.rbi` file. If this extra `require` line is not added to `sorbet/tapioca/require.rb` file, then the definition for that type will be missing from the RBI file.
+### 2, Initialize folder structure
```
Morriar, line 29:
```suggestion
### Initialize the folder structure
```

README.md
```diff
@@ -6,53 +6,143 @@
...
 
-If you ever run into a case, where you add a gem or update the version of a gem and run `tapioca sync` but don't have some types you expect in the generated gem RBI files, you will need to make sure you have added the necessary requires to the `sorbet/tapioca/require.rb` file.
+Command: `tapioca init`
 
-You can use the command `tapioca require` to auto-populate the `sorbet/tapioca/require.rb` file with all the requires found
-in your application. Once the file generated, you should review it, remove all unnecessary requires and commit it.
+This will create the `sorbet/config` and `sorbet/tapioca/require.rb` files for you, if they don't exist. If any of the files already exist, they will not be changed.
 
-## How does tapioca compare to "srb rbi gems" ?
+### 3. Generate for gems
```
Morriar, line 35:
```suggestion
### Generate gems RBIs
```

README.md
```diff
@@ -6,53 +6,143 @@
...
 
-[Please see the detailed answer on our wiki](https://github.com/Shopify/tapioca/wiki/How-does-tapioca-compare-to-%22srb-rbi-gems%22-%3F)
+Command: `tapioca generate [gems...]`
 
-## Installation
+This will generate RBIs for the specified gems and place them in the RBI directory.
 
-Add this line to your application's `Gemfile`:
+### 4. Generate for all gems in Gemfile
```
Morriar, line 41:
```suggestion
or:
```

README.md
```diff
@@ -6,53 +6,143 @@
...
 
-```ruby
+Command: `tapioca sync`
```
Morriar, line 43:
```suggestion
Command: `tapioca gem`
```

README.md
```diff
@@ -6,53 +6,143 @@
...
+
+This will sync the RBIs with the gems in the Gemfile and will add, update, and remove RBIs as necessary.
+
+### 5. Generate the list of all unresolved constants
```
Morriar, line 47:
```suggestion
### Generate the list of all unresolved constants
```

README.md
```diff
@@ -6,53 +6,143 @@
...
+
+Command: `tapioca todo`
+
+This will generate the file `sorbet/rbi/todo.rbi` defining all unresolved constants as empty modules.
+
+### 6. Generate DSL RBI files
```
Morriar, line 53:
```suggestion
### Generate DSL RBI files
```

README.md
```diff
@@ -6,53 +6,143 @@
...
+
+Command: `tapioca dsl [constant...]`
+
+This will generate DSL RBIs for specified constants (or for all handled constants, if a constant name is not supplied). You can read about DSL RBI generators supplied by `tapioca` in [the manual](manual/generators.md).
+
+## Installing Tapioca into a project that already uses Sorbet (and has RBI files)
```
Morriar, line 59: We should probably link to this page of the wiki: https://github.com/Shopify/tapioca/wiki/Migrating-to-Tapioca.

README.md
```diff
@@ -6,53 +6,143 @@
...
+
+Here are some things that do not need to be run for the migration (or ever in some cases).
+
+1. `bin/tapioca sync`: this command is meant to be used **after** the initial setup is complete to update RBIs for gems that have been upgraded. It does not need to be used as part of the migration process, but should be used later
+2. `bundle exec srb rbi hidden-definitions`: not necessary at all, ever
+3. `bundle exec srb rbi suggest-typed`: not necessary at all, ever. Prefer `bundle exec spoom bump`
+
+Done! Your application should be all set and type checking should pass.
+
+## How does tapioca compare to "srb rbi gems" ?
```
Morriar, line 138: If we're going to document all of this here, we should also add a section explaining how to keep the RBIs up-to-date when the Gemfile changes and the files related to DSLs change.

Morriar (changes requested):
First, thanks a lot for taking care of this. Our README desperately needed some love ❤️ 

I'm not sure a hard separation between fresh Sorbet project and projects using `srb rbi` is relevant. Many sections / steps are common to both routes and we'll end up repeating a lot of content.

What do you think about merging the two approaches but marking some steps as _optional_ if they are related to `sorbet-rails` or `srb rbi`?

peternixey:
Hi @Morriar - sorry for the delayed response, I was away on holiday. 

I went through your changes, all of which seemed very reasonable. I've done a complete revamp of the page here to combine the two different flows into one and then to add in the links and clarity you suggested:

https://github.com/peternixey/tapioca/blob/main/README.md

I wrote it all on the assumption that most people will find Tapioca *after* they find Sorbet and so the primary use-case is for an existing sorbet installation. You may know the reality of that better than me though! 

BTW one thing I did was compress a lot of the text down into code blocks which I think makes it easier to scan and consume - you may feel differently though. 

I also wonder whether on a document like this you might spend more time giving me kind feedback than actually doing the edits yourself. If you'd like to take the doc as it is (or as it was) and edit it however you see fit I will not be in the slightest offended and it will probably take you less time than writing me kind tweak-suggestions ;)

So let me hand this over to you and it is yours to merge / bin / use-for-inspriation as helpful!

Best wishes,
Peter

README.md
```diff
@@ -6,53 +6,140 @@
...
 ```ruby
 group :development do
   gem 'tapioca', require: false
 end
 ```
+Then run `bundle install`
+
+### Generate the RBI files for gems
+
+````bash
```
rafaelfranca, line 56:
```suggestion
```bash
```

README.md
```diff
@@ -6,53 +6,140 @@
...
+# (Generate only for specific gems using `tapioca gem 
+gem...]`)
+bundle exec tapioca gem
+
+# Generate the list of all unresolved constants
+# This will generate the file `sorbet/rbi/todo.rbi`
+# defining all unresolved constants as empty modules
+bundle exec tapioca todo
+
+````
```
rafaelfranca, line 71:
```suggestion
```
```

File: README.md [hunk 15]
```diff
@@ -6,53 +6,140 @@
...
+
+### Iterate RBI generation to ensure all gems are covered
+
+The first step may result in some typechecking errors due to missing definitions. Let's pick them up and dela with them:
+
+````bash
```
rafaelfranca, line 77:
```suggestion
```bash
```

README.md
```diff
@@ -6,53 +6,140 @@
...
+bundle exec srb tc
+
+# If you notice missing definitions for gems, nudge Tapioca to
+# look for them by running:
+bin/tapioca require
+
+# If this doesn't add the requisite gems, add them manually to:
+# sorbet/tapioca/require.rb
+# repeat as necessary
+````
```
rafaelfranca, line 88:
```suggestion
```
```

rafaelfranca (commented):
While I can understand a lot of users will arive to tapioca after knowing about Sorbet I think our primary flow needs to for users that didn't setup anything yet and extract the migration "guide" to a different document linked in the README.

We can still link to the document before we explain the "pristine flow" but having both flows in the same document when users will either fall in one or another category could be confusing.

peternixey:
@rafaelfranca you guys obviously know your audience much better than me but I’m just interested in the logic as to why most people would come to Tapioca with a pristine setup.

Surely one finds out about Tapioca after trying Sorbet and then realising that it has its limitations. I could understand if Tapioca was part of the official Sorbet setup guide / package but as far as I know it’s not.

To assuage my curiosity, what’s the context in which a user would discover Tapioca prior to discovering Sorbet?

peternixey:
I’ve just realised that of course Ruby 3 uses RBI files too so presumably that would be the other (and potentially main) use case - is that right?

rafaelfranca:
Ruby 3 uses RBS files, not RBIs, so that is another migration setup.

> To assuage my curiosity, what’s the context in which a user would discover Tapioca prior to discovering Sorbet?

Inside Shopify mostly.

But my interest of moving out the migration steps is mainly because of the perspective of the library itself. As a library, Tapioca has no knowledge on your setup, so it assumes you have a pristine setup, that is the path we mostly care about. Of course how users land here will vary, and that is fine, we should document some of those cases. I just don't think having everything in the README will make it easier to read. The README should only include the pristine way. From there we can link to other pages, sometimes maintained by different people, for the various setups that people might have before arriving here.

peternixey:
Hi @rafaelfranca and thanks for your answers. I'd like to clarify what would be most helpful to you guys from this next. I

I'm getting conflicting messages from your side though so I'd prefer to either wrap up or hand it over. @Morriar had originally suggested combining the Sorbet-based installation with the pristine installation and so that's the way that I took it. 

I'm not worried if you'd prefer to leave this but if it's helpful then there's no point in wasting the few hours that I've spent revising it. If it's a hassle tell me then I'll leave it, if it's an asset then let's make the most of the couple of hours I've spent on it so far:

Can I suggest the following course of action:

1. You tell me whether or not you'd like to tweak what I've got
2. You decide whether you'd prefer to do it or you'd like me to (for small changes it's far more sensible for you to do them)
3. You decide on one person who's code reviewing so as to avoid bouncing between different points of opinion. 

FWIW I would suggest that the bulk of people outside of Shopify will find Tapioca via Sorbet so it's probably a big thing to factor into your instructions and accounting for it will hugely boost people's initial impressions of Tapioca. 

I can offer you another hour of my time but only take it if you actually want this PR otherwise let's just park it. 

Best
Peter

bmulholland:
Typo: dela -> deal

PR #476, opened 2021-09-01
Revert non-nil default value inference for SmartProperties

paracycle:
### Motivation

Mostly reverts #338.

The problem with the approach we took in #338 was that even though the properties would have an initial default value, Smart Properties has no protections against those properties being assigned a `nil` value.

For example:
```ruby
class Foo
  include SmartProperties
  property :bar, default: "42"
end

foo = Foo.new
foo.bar # => "42"
foo.bar = nil
foo.bar # => nil
```

I've seen a few places in Core that explicitly check for `nil` even though the property had a default value, so I think a nilable type is safer overall.

/cc @samuelgiles 

### Implementation

Got rid of the default check for deciding on nilability. Left the other changes that simplified the code path from #338.

### Tests

Updated tests.

### DIFF
--- a/lib/tapioca/compilers/dsl/smart_properties.rb
+++ b/lib/tapioca/compilers/dsl/smart_properties.rb
@@ -53,10 +53,10 @@ module Tapioca
       #   sig { params(published: T.nilable(T::Boolean)).returns(T.nilable(T::Boolean)) }
       #   def published=(published); end
       #
-      #   sig { returns(T::Boolean) }
+      #   sig { returns(T.nilable(T::Boolean)) }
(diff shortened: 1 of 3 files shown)


paracycle:
> Ah shame but makes sense 👍

Yeah, indeed it is a shame. However, I see no reason for SP to not check for nil and use default value if nil is supplied during a set operation. @samuelgiles If you are interested in making that improvement in SP, we can make the switch back to non-nilable for fields with default values. 

I suggest that you discuss this on SP if you feel strongly about it. I am afraid I don't have the time to follow this up any further.

PR #479, opened 2021-09-02
Include YARD documentation in the generated RBIs for methods

### LINKED ISSUE #437: Include documentation from original source
If a definition contains Yard documentation in the source, we should add it to the generated RBI so it's surfaced by LSP extensions.

vinistock:
Closes #437

### Motivation

Because we generate RBIs without including comments, we lose the documentation that is included in the original source. It's a good improvement to keep that documentation, so that we can surface that on hovering, for example.

Also, take a look at the RBI generate for `pry-doc` which was included in this PR. All the YARD documentation is there!

**Note**: this PR only handles method definitions, but we should also include any documentation available for classes and modules, which can be delivered in a second PR.

### Implementation

The core of the implementation is relatively simple.
- Use pry-doc to get the documentation of a method
- For each line of documentation, add an `RBI::Comment` to the method we are compiling

The somewhat tricky part is adjusting `Pry::Code` parsing so that it catches both documentation written immediately above the method or above the method's signature. For example,

```ruby
class Foo
  # This comment comes immediately above the method definition and so Pry::Doc can find it
  # @return [void]
  def bar; end
end

class AnotherFoo
  # This comment comes before sig and not def and so Pry::Code ignores it
  # @return [void]
  sig { void }
  def another_bar; end
end
```

In order to get around this, we need to patch the code helpers from `method_source`. The logic of the method boils down to this
- We receive `lines` which is the file's source code starting from the first line all the way to the line immediately before the method definition. i.e.: `source[0..index_of_method_def - 1]`
- We go through each line appending the comments to a buffer and reset the buffer if we find anything that is not a comment

This resetting of the buffer is what causes the documentation to be empty. So, I added another branch to the conditional and now the logic is
- If comment, append to buffer
- If it's a sig and there are no other comments after the sig, then we found the comment section for the method we are looking for and the buffer is already complete, so just return it early
- If it's any other code or if it's a sig from a previous method, reset the buffer

### Tests

Added tests.


lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -655,6 +659,13 @@ def compile_method(tree, symbol_name, constant, method, visibility = RBI::Public
           tree << rbi_method
         end
 
+        sig { params(method: T.any(UnboundMethod, Method)).returns(String) }
+        def yard_docs(method)
+          Pry::Method.new(method).doc
+        rescue StandardError
```
vinistock, line 665: Some weird methods inside `pry` cause errors. Like, for example, they have a `SourceLocation#[](*)` method that has no receiver (I don't know how), and that breaks its own code. So, I think it's fine to leave this as a best effort and if we can't get the docs, just compile the method without it.

sorbet/tapioca/config.yml
```diff
@@ -1,2 +1,4 @@
 ---
 exclude: []
+typed_overrides:
+  "pry-doc": "false"
```
vinistock, line 4: `Pry-doc` re-defines a bunch of methods from other pry-related libraries and type checking fails.

lib/tapioca/compilers/symbol_table/symbol_generator.rb
Morriar: Shouldn't we return `nil` instead to avoid creating comments with empty strings?

lib/tapioca/method_source_ext/code_helpers.rb
```diff
@@ -0,0 +1,38 @@
+# typed: true
+# frozen_string_literal: true
+
+module MethodSource
```
Morriar, line 4: We should rely on `yard-sorbet` instead: https://github.com/stripe/yard-sorbet/blob/master/lib/yard-sorbet/sig_handler.rb#L19

Morriar (commented):
Can you also update the [issue](https://github.com/Shopify/tapioca/issues/437) to list the missing parts please? 🙏

lib/tapioca/compilers/symbol_table/symbol_generator.rb
vinistock: We invoke `lines` on the empty string, which returns an empty array and then when we invoke `each` it won't perform any loops. This is basically what happens

```ruby
"".lines -> []
[].each {...} -> nothing happens
```

lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -2,6 +2,8 @@
 # frozen_string_literal: true
 
 require "pathname"
+Tapioca.silence_output { require "pry-doc" }
```
vinistock, line 5: This is necessary because `pry-doc` prints directly to `STDOUT` if it doesn't find the exact folder of documentation for the current Ruby version.

spec/tapioca/compilers/symbol_table_compiler_spec.rb
```diff
@@ -2281,6 +2282,8 @@ def inherited(s); end
         class Baz
           abstract!
 
+          # Don't need to silence warnings via without_ruby_warnings when calling
```
vinistock, line 2285: This one is interesting. `pry-doc` will actually include documentation for methods created dynamically using `define_method`. This is coming from Sorbet.

vinistock:
This should be ready for another round of reviews. I added documentation for modules, classes and constants as well, since it was trivial after switching to YARD.

Regarding performance, this seemed to slow down tests on my machine by quite a lot, but not for generating RBIs for an actual app. It might be because every test example spawns a new tapioca process and needs to parse YARD docs again, but I don't we'll be able to escape that.

I also compared parsing an entire directory or individual gem files (current implementation) and parsing directories is very significantly slower.

lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -21,6 +24,8 @@ def initialize(gem, indent = 0)
           @seen = Set.new
           @alias_namespace ||= Set.new
           @symbol_queue = T.let(symbols.sort.dup, T::Array[String])
+
+          gem.files.each { |path| YARD.parse(path.to_s, [], Logger::Severity::ERROR) }
```
Morriar, line 28: Is it really an error for Tapioca purpose?

```suggestion
          gem.files.each { |path| YARD.parse(path.to_s, [], Logger::Severity::WARN) }
```

lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -154,16 +159,22 @@ def compile_object(tree, name, value)
             name_of(klass)
           end
 
+          comments = documentation_comments(name)
+
           if klass_name == "T::Private::Types::TypeAlias"
-            tree << RBI::Const.new(name, "T.type_alias { #{T.unsafe(value).aliased_type} }")
+            constant = RBI::Const.new(name, "T.type_alias { #{T.unsafe(value).aliased_type} }")
+            constant.comments.concat(comments) unless comments.empty?
+            tree << constant
```
Morriar, line 167: You should be able to pass it directly to the constructor (and so avoid initializing an empty array just to concat to it):
```suggestion
            tree << RBI::Const.new(name, "T.type_alias { #{T.unsafe(value).aliased_type} }", comments: comments)
```

See the initializer hereL https://github.com/Shopify/rbi/blob/main/lib/rbi/model.rb#L303.

lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -154,16 +159,22 @@ def compile_object(tree, name, value)
...
             return
           end
 
           return if klass_name&.start_with?("T::Types::", "T::Private::")
 
           type_name = klass_name || "T.untyped"
+          constant = RBI::Const.new(name, "T.let(T.unsafe(nil), #{type_name})")
+          constant.comments.concat(comments) unless comments.empty?
```
Morriar, line 175:
```suggestion
          constant = RBI::Const.new(name, "T.let(T.unsafe(nil), #{type_name})", comments: comments)
```

lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -178,6 +189,9 @@ def compile_module(tree, name, constant)
               RBI::Module.new(name)
             end
 
+          comments = documentation_comments(name)
+          scope.comments.concat(comments) unless comments.empty?
+
```
Morriar, line 194: Same here, you can move those lines before the initializations of the `Class` and `Module` and pass the comments to the `new` call.

lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -631,6 +645,11 @@ def compile_method(tree, symbol_name, constant, method, visibility = RBI::Public
           end
 
           rbi_method = RBI::Method.new(method_name, is_singleton: constant.singleton_class?, visibility: visibility)
+
+          separator = constant.singleton_class? ? "." : "#"
+          comments = documentation_comments("#{symbol_name}#{separator}#{method_name}")
+          rbi_method.comments.concat(comments) unless comments.empty?
+
```
Morriar, line 652:
```suggestion
          separator = constant.singleton_class? ? "." : "#"
          comments = documentation_comments("#{symbol_name}#{separator}#{method_name}")
          
          rbi_method = RBI::Method.new(method_name, is_singleton: constant.singleton_class?, visibility: visibility, comments: comments)
```

spec/spec_helper.rb
```diff
@@ -6,6 +6,7 @@
 require "minitest/spec"
 require "minitest/hooks/default"
 require "minitest/reporters"
+require "byebug"
```
Morriar, line 9: Why do we need this?

lib/tapioca/compilers/symbol_table/symbol_generator.rb
vinistock: Using warn is the default and it prints a lot of stuff to the screen related to YARD finding undocumented classes. We need to turn it off or else the output looks horrible.

spec/spec_helper.rb
vinistock: Every time when debugging I forget that we don't require byebug for tests and need to add it. We have it in the Gemfile, we are just not requiring it in the spec helper.

vinistock:
As per @Morriar's suggestion, I have added a CLI option to ignore docs. I followed the same naming used in bundler and rubygems, which is `--doc, --no-doc`, but let me know if you prefer a more descriptive argument name.

lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -11,16 +13,20 @@ class SymbolGenerator
         include(Reflection)
 
         IGNORED_SYMBOLS = ["YAML", "MiniTest", "Mutex"]
+        IGNORED_COMMENTS = [":nodoc:", "typed:", "frozen_string_literal:"]
```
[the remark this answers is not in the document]
vinistock: TIL

Morriar (approved):
Should we add a test that contains magic comments?

vinistock:
Caught two special comments we had missed: `rubocop:` and `:doc:`. Added both to the list and generated Tapioca's own RBIs including documentation in the latest commit.

PR #481, opened 2021-09-03
Stop monkey-patching core classes

paracycle:
### Motivation

Monkey patching `Class` to add `descendants` causes a confusion with the `Class#descendants` method that Active Support adds and the signature on this method leaks out into the Tapioca generated RBI which causes type-checking problems in Core.
    
Monkey patching `String` to add `underscore` causes a confusion with the `String#underscore` method that Active Support adds and causes problems like https://discourse.shopify.io/t/file-naming-inflection-issues-with-tapioca-dsl-verify-in-ci/19663 (since the method later on gets replaced by the AS version and starts using application inflectors, etc, which we don't want, we just want a simple deterministic CamelCase to under_score implementation)

### Implementation

`descendants` is arguably a reflection method, so I moved it to `Tapioca::Reflection` module, which is already included in the Base DSL class and SymbolGenerator class, as `descendants_of(klass)`. The added benefit is that through type parameters, we get better typing on this method, thus we need less casts.
    
I moved the `underscore` method onto the `Generator` class and also duplicated it into the DSL test base class.

### Tests

No added tests.


spec/dsl_spec.rb
```diff
@@ -46,7 +45,19 @@ def target_class_name
 
   sig { returns(String) }
   def target_class_file
-    target_class_name.underscore
+    underscore(target_class_name)
+  end
+
+  sig { params(class_name: String).returns(String) }
```
mutecipher, line 51: This seems to be replication of the previous declaration of this method in `lib/tapioca/generator.rb`. Would it be better to make this a helper module somewhere we can inject instead?

paracycle: Indeed it is, but since this is a test base class, I wasn't too worried about refactoring to a helper class/module. The real usage is in `generator.rb` only.

RyanBrushett (approved):

vinistock (approved):

PR #491, opened 2021-09-08
Add Active Record fixture DSL generator

### LINKED ISSUE #436: Implement ActiveRecord::TestFixtures DSL generator

vinistock:
Closes #436

### Motivation

Rails dynamically generates methods for test fixtures based on YAML files. These methods only exist as the tests are running, so Sorbet does not know about them. This DSL generator loads all fixtures into memory and then produces an RBI for `ActiveSupport::TestCase` - from which all tests inherit from - including all the fixture methods that are dynamically created.

### Implementation

- Create a fake test class to be able to load fixtures
- Load them into memory and create the class cache to be able to retrieve the right types
- Go through each one of the tables and create a corresponding fixture method

We return early if there are no fixtures, because it's not necessary to produce an empty RBI for `ActiveSupport::TestCase`.

### Tests

See included tests.


spec/tapioca/compilers/dsl/active_record_fixtures_spec.rb
```diff
@@ -0,0 +1,170 @@
...
+      RBI
+
+      assert_equal(expected, rbi_for("ActiveSupport::TestCase"))
+    end
+  end
+
+  private
+
+  sig { void }
+  def fake_rails_app
```
Morriar, line 161: I'm surprised we didn't need something like this before?

I think most of the tests are playing with the Rails API to create the necessary context instead of creating a complete app like [here](https://github.com/Shopify/tapioca/blob/main/spec/tapioca/compilers/dsl/active_record_scope_spec.rb#L95-L102). Could we take the same approach for this generator?

Also couldn't this be done in the `before` block?

lib/tapioca/compilers/dsl/active_record_fixtures.rb
```diff
@@ -0,0 +1,87 @@
...
+      class ActiveRecordFixtures < Base
+        extend T::Sig
+
+        class FixtureLoader
+          include ActiveRecord::TestFixtures
+        end
+
+        sig { override.params(root: RBI::Tree, constant: T.class_of(ActiveSupport::TestCase)).void }
+        def decorate(root, constant)
+          T.unsafe(FixtureLoader).fixture_path = Rails.root.join("test", "fixtures")
```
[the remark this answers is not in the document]
vinistock: Because they are added to the class by using `cattr_*` inside a concern's `included` block. I added a shim with the method definitions.

sorbet/rbi/shims/fixture_loader.rbi
```diff
@@ -0,0 +1,16 @@
+# typed: true
+
+module Tapioca
+  module Compilers
+    module Dsl
+      class ActiveRecordFixtures < Base
+        class FixtureLoader
+          def self.fixture_path=(fixture_path); end
+          def self.fixture_path; end
+
```
Morriar, line 10: Why this blank line?

spec/tapioca/compilers/dsl/active_record_fixtures_spec.rb
```diff
@@ -0,0 +1,170 @@
...
+    base_folder = Pathname.new(tmp_path("lib"))
+
+    config_class = Struct.new(:root)
+    config = config_class.new(base_folder)
+    app_class = Struct.new(:config)
+    app = app_class.new(config)
+    Rails.application = app
```
Morriar, line 168:
```suggestion
    Rails.application = app_class.new(config)
```

spec/tapioca/compilers/dsl/active_record_fixtures_spec.rb
Morriar: It might be because the before only requires `active_record` and I think the block is ran before loading the generator class which brings `rails` too late.

Maybe you can also `require "rails"` in the before block.

vinistock: Requiring rails did the trick 👍.

Morriar (approved):

lib/tapioca/compilers/dsl/active_record_fixtures.rb
```diff
@@ -0,0 +1,87 @@
...
+      class ActiveRecordFixtures < Base
+        extend T::Sig
+
+        class FixtureLoader
+          include ActiveRecord::TestFixtures
+        end
+
+        sig { override.params(root: RBI::Tree, constant: T.class_of(ActiveSupport::TestCase)).void }
+        def decorate(root, constant)
+          FixtureLoader.fixture_path = Rails.root.join("test", "fixtures")
```
paracycle, line 47: As far as I know, this path is configurable, so should we try to understand what the configuration is?

lib/tapioca/compilers/dsl/active_record_fixtures.rb
```diff
@@ -0,0 +1,87 @@
...
+        sig { params(mod: RBI::Scope, name: String, return_type: String).void }
+        def create_fixture_method(mod, name, return_type)
+          mod.create_method(
+            name,
+            parameters: [
+              RBI::TypedParam.new(
+                param: RBI::RestParam.new("fixture_names"),
+                type: "Symbol"
+              ),
+            ],
```
paracycle, line 80: Let's use our helpers:
```suggestion
            parameters: [create_rest_param("fixture_names", type: "Symbol")],
```

sorbet/rbi/shims/fixture_loader.rbi
```diff
@@ -0,0 +1,15 @@
+# typed: true
+
+module Tapioca
+  module Compilers
+    module Dsl
+      class ActiveRecordFixtures < Base
+        class FixtureLoader
```
paracycle, line 7: Why do we need this shim? We should be able to generate `ActiveRecord::TestFixtures` with the proper `mixes_in_class_modules` annotation.

spec/tapioca/compilers/dsl/active_record_fixtures_spec.rb
```diff
@@ -0,0 +1,168 @@
...
+      RBI
+
+      assert_equal(expected, rbi_for("ActiveSupport::TestCase"))
+    end
+  end
+
+  private
+
+  sig { void }
+  def fake_rails_app
```
paracycle, line 160: Can we rename this method to be `define_fake_rails_app` or something, to make it clear what it is doing?

sorbet/rbi/shims/fixture_loader.rbi
vinistock: Can Tapioca currently pick up Active Support special attributes invoked from within an `included` block? I think this is something we're currently missing.

[This is where the methods are created](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/test_fixtures.rb#L19-L27).

lib/tapioca/compilers/dsl/active_record_fixtures.rb
```diff
@@ -0,0 +1,82 @@
...
+          cache = ActiveRecord::FixtureSet::ClassCache.new({}, ActiveRecord::Base)
+          table_names = FixtureLoader.fixture_table_names
+          return if table_names.empty?
+
+          root.create_path(constant) do |mod|
+            table_names.each do |table_name|
+              accessor_name = table_name.tr("/", "_")
+              model_class = cache[table_name]
+
+              create_fixture_method(mod, accessor_name, model_class.name)
```
paracycle, line 59: I don't think we can do `model_class.name` as the return type here since the return type is dependent on the number of parameters supplied to the fixture method:
```ruby
posts(:first_post, :last_post) => [#<Post:0x111>, #<Post:0x222>]
```
so this will have to be `T.untyped`, I am afraid.

lib/tapioca/compilers/dsl/active_record_fixtures.rb
```diff
@@ -0,0 +1,82 @@
...
+        end
+
+        sig { override.params(root: RBI::Tree, constant: T.class_of(ActiveSupport::TestCase)).void }
+        def decorate(root, constant)
+          FixtureLoader.fixture_path = Rails.root.join("test", "fixtures")
+          FixtureLoader.fixtures(:all)
+
+          cache = ActiveRecord::FixtureSet::ClassCache.new({}, ActiveRecord::Base)
+          table_names = FixtureLoader.fixture_table_names
+          return if table_names.empty?
```
paracycle, line 52: We don't actually need to go through table names to figure out what the generated method names will be since the methods are actually already defined on `FixtureLoader`'s anonymous ancestors:
```suggestion
          methods = FixtureLoader.ancestors.reject(&:name).map do |mod|
            [mod.private_instance_methods(false), mod.instance_methods(false)]
          end.flatten

          return if methods.empty?
```

Since we can't type the return value of the fixture methods, this will be enough for us.

As an extra benefit, if we do this, I suspect that we won't even have to setup tables or even define classes for the models in our test cases.

lib/tapioca/compilers/dsl/active_record_fixtures.rb
paracycle: Unfortunately we can't, since the two types are not compatible. Imagine if I called `.map` on the return value, the model would not have that method, so Sorbet will complain about the model type not having the method.

lib/tapioca/compilers/dsl/active_record_fixtures.rb
vinistock: Very nice! This cleaned up the tests and the code so much 👍

paracycle (commented):
Nice! Final round of review.

PR #496, opened 2021-09-10
Add `BaseGenerator` and test suite

mutecipher:
### Motivation

Part of an ongoing refactor I'm working on to break up `lib/tapioca/generator.rb` into smaller classes. This provides a
base class for further classes to be built upon, which abstract definitions of `error_handler` and `generate` methods.
It also provides some helper methods that will be used my future classes (or extracted into a module).

### Implementation

Pull shared logic into a base class that contains abstract methods for `generate` and `error_handler`.

### Tests

Attached to the PR. Provides additional testing for supplemental methods used in `lib/tapioca/generate.rb`.


lib/tapioca/generators/base_generator.rb
```diff
@@ -0,0 +1,110 @@
+# typed: strict
+# frozen_string_literal: true
+
+require "active_support/all"
```
RyanBrushett, line 4: Is this needed? I see underscore but [generator has its own underscore](https://github.com/Shopify/tapioca/pull/481) and I think `require "pathname"` can take care of the path stuff.

spec/tapioca/generators/base_generator_spec.rb
```diff
@@ -0,0 +1,170 @@
...
+          constants = ["Foo", "Bar", "Baz"]
+          config = ConfigBuilder.from_options(:dsl, {})
+          generator = Tapioca::Generators::BaseGenerator.new(config)
+          result = generator.constantize(constants)
+          expected = []
+          assert_equal(expected, result)
+        end
+
+        it "returns an array of filepaths when constants are found" do
+          skip
```
RyanBrushett, line 109: Couple of skips here, can you add a comment for why they're skipped?

RyanBrushett (commented):
Couple of quick comments before I sign off, will 👀 again Monday.

lib/tapioca/generators/base_generator.rb
```diff
@@ -0,0 +1,110 @@
...
+
+module Tapioca
+  module Generators
+    class BaseGenerator
```
Morriar, line 8: Do we need to repeat `Generator` here? 

```suggestion
    class Base
```

For example in Rails most base class are just called `Base`, the namespace being enough to disambiguate.

lib/tapioca/generators/base_generator.rb
```diff
@@ -0,0 +1,110 @@
...
+
+      sig { returns(Config) }
+      attr_reader :config
+
+      sig { params(config: Config).void }
+      def initialize(config)
+        @config = config
+        @bundle = T.let(nil, T.nilable(Gemfile))
+        @compiler = T.let(nil, T.nilable(Compilers::SymbolTableCompiler))
+        @loader = T.let(nil, T.nilable(Loader))
```
Morriar, line 22: Is this going to be the base class of all generators? These attributes seems strongly coupled with gem RBIs.

spec/tapioca/generators/base_generator_spec.rb
```diff
@@ -0,0 +1,170 @@
...
+require "generator_spec"
+require "tapioca/generators/base_generator"
+
+module Tapioca
+  module Generators
+    class BaseGeneratorSpec < GeneratorSpec
+      describe "#rbi_header" do
+        it "displays when config.file_header is true" do
+          config = ConfigBuilder.from_options(:dsl, {})
+          generator = Tapioca::Generators::BaseGenerator.new(config)
```
Morriar, line 14: If the class is marked `abstract` I don't think we should instantiate it.

If we want to test the base services, it would be better to create a dummy subclass that we can instantiate:

```rb
class DummyGenerator < BaseGenerator; end
```

lib/tapioca/generators/base_generator.rb
```diff
@@ -0,0 +1,110 @@
...
+    class BaseGenerator
+      extend(T::Sig)
+      extend(T::Helpers)
+
+      abstract!
+
+      sig { returns(Config) }
+      attr_reader :config
+
+      sig { params(config: Config).void }
```
Morriar, line 17: `Config` seems to be a concept closely related to a CLI tool. Why not decoupling this and passing all the config options as keyword parameter?

This will have at least two benefits:
1. Better separation of concerns (since a generator will be unable to access config value unrelated to its purpose)
2. Simplify the way we test the generators (no need to create a config object)

lib/tapioca/generators/base_generator.rb
```diff
@@ -0,0 +1,110 @@
...
+          # DO NOT EDIT MANUALLY
+          # This is an autogenerated file for #{reason}.
+          # Please instead update this file by running `#{command}`.
+        HEAD
+
+        sigil = <<~SIGIL if strictness
+          # typed: #{strictness}
+        SIGIL
+
+        if config.file_header
```
Morriar, line 58: It's weird that the `command`, `reason`, `strictness` come from the method parameters but the `file_header` comes from the config.

I think `file_header` should be another keyword argument.

lib/tapioca/generators/base_generator.rb
```diff
@@ -0,0 +1,110 @@
...
+      end
+
+      sig { abstract.void }
+      def generate; end
+
+      sig { abstract.params(error: String).void }
+      def error_handler(error); end
+
+      sig { params(command: String, reason: T.nilable(String), strictness: T.nilable(String)).returns(String) }
+      def rbi_header(command, reason: nil, strictness: nil)
```
Morriar, line 47: We should move this as an extension to [`RBI::File`](https://github.com/Shopify/rbi/blob/main/lib/rbi/model.rb#L125) under [`rbi_ext`](https://github.com/Shopify/tapioca/tree/main/lib/tapioca/rbi_ext)

spec/tapioca/generators/base_generator_spec.rb
```diff
@@ -0,0 +1,170 @@
...
+            # DO NOT EDIT MANUALLY
+            # This is an autogenerated file for .
+            # Please instead update this file by running `bin/tapioca dsl`.
+
+          RUBY
+          assert_equal(expected, result)
+        end
+
+        it "doesn't display when config.file_header is false" do
+          config = ConfigBuilder.from_options(:dsl, { "file_header" => false })
```
Morriar, line 26: I think the `config` being repeated in each test case here is a symptom of the bad separation of concerns it brings when required to build a `Generator`.

lib/tapioca/generators/base_generator.rb
```diff
@@ -0,0 +1,110 @@
...
+
+      sig { params(constant_names: T::Array[String]).returns(T::Array[Module]) }
+      def constantize(constant_names)
+        constant_map = constant_names.map do |name|
+          [name, Object.const_get(name)]
+        rescue NameError
+          [name, nil]
+        end.to_h
+
+        unprocessable_constants = constant_map.select { |_, v| v.nil? }
```
Morriar, line 80: I find it super surprising that calling a method called `constantize` on a few strings would have the side effect of removing files from my file system.

Let's split this into a second method with an explicit name.

spec/tapioca/generators/base_generator_spec.rb
```diff
@@ -0,0 +1,170 @@
...
+          generator.remove(fake_filename)
+          refute_path_exists(fake_filename)
+        end
+
+        it "removes the file if it exists" do
+          filename = repo_path / "disposable.rb"
+          File.write(filename, <<~RUBY)
+            # typed: true
+            module Disposable
+            end
```
Morriar, line 159:
```suggestion
            # typed: true
            
            module Disposable; end
```

spec/tapioca/generators/base_generator_spec.rb
mutecipher: So previous behaviour was to just exit the application when constants weren't found or were un-processable. I want to leave that behaviour to the CLI itself rather than a generator. So wanted to get some feedback before moving forward with a fleshed out solution. In this case, I've simply removed the `exit(1)` call from the original method, but haven't made any other changes.

My proposed solution would be to just return an empty array (as described in the test) for un-processable constants, and potentially log it using the `error_handler` implemented by the subclass.

lib/tapioca/generators/base_generator.rb
mutecipher: I was doing this initially to maintain compatability with the existing implementation. But that was the game plan for future changes, as IIRC, `Config` will be going through some changes as well.

mutecipher:
Closing in favour of https://github.com/Shopify/tapioca/pull/504

PR #499, opened 2021-09-13
Replace all `sync` in headers with `gem`

paracycle:
### Motivation

It turns out that I forgot to replace the `bin/tapioca sync` commands in the generated gem RBI file headers.

### Implementation

- Replace all `bin/tapioca sync` header or output strings with `bin/tapioca gem` instead.
- Start adding the gem name to the command for a single file so that regeneration is more easy, i.e. `bin/tapioca gem foo`, instead of just `bin/tapioca gem`.

### Tests

Updated existing tests.


RyanBrushett (approved):

Morriar (approved):
I'll take me years to stop using `tapioca sync` 😂

PR #500, opened 2021-09-15
Load the Rails application before loading the gems

Morriar:
### Motivation

Sometimes gems depends on things that will be loaded by the Rails application so we need to make sure we load the application before the gems.

### Tests

See automated tests.


paracycle (approved):
🙏

RyanBrushett (approved):

PR #503, opened 2021-09-17
Move symbol_generator to typed strict

Morriar:
### Motivation

Increase strictness in this file to make it easier to maintain going forward.

### DIFF
--- a/lib/tapioca/compilers/symbol_table/symbol_generator.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_generator.rb
@@ -1,4 +1,4 @@
-# typed: true
+# typed: strict
 # frozen_string_literal: true
 
 require "pathname"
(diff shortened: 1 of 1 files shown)


lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -447,7 +457,7 @@ def method_missing(symbol, *args)
             # is the list of all dynamically extended modules because of that
             # constant. We grab that value by deleting the key for the original
             # constant.
-            T.must(mixins_from_modules.delete(constant)),
+            mixins_from_modules.delete(constant),
```
Morriar, line 460: The value was `untyped` so `T.must` is flagged as redundant.

vinistock (approved):

PR #504, opened 2021-09-17
Extract `Generators` from `generator.rb`

mutecipher:
### Motivation

`tapioca`'s current use of one monolithic `Generator` class is proving to be troublesome and tough to maintain. This is my initial step towards decoupling and simplifying that class into different smaller classes. This will allow for smaller future changes to these individual classes, as well as permit for further decoupling down the road. 

### Implementation

I created a `Tapioca::Generators` namespace which will contain separate `Generators`. All generators in this namespace will inherit from `Base` which defines an interface for how generators should perform going forward. 

~Each has a `#generate` method as well as a `#handle_error` method. This defines the contract that generators should hold going forward which will allow for extensions/plugins to be integrated much easier in the future.~ Removed in https://github.com/Shopify/tapioca/pull/504/commits/e0b92d8bfaf860976ea6b1b9e350283d9d55bb9f

This current step simply extracts the `Dsl`, `Todo`, `Require` and `Gem` logic from `generator.rb` and pastes them in their individual classes. _Note: There is currently some overlap in logic that will be simplified and extracted in future PRs._ After doing that, I made the necessary changes in `Cli` to use these updated generators and made the necessary changes to ensure the current `Cli` integration tests continued to pass without any modification to the tests themselves.

### Tests

As noted above, I didn't add any additional tests in this case, as I simply copied the logic from `generator.rb` to the respective new files. However, in future PRs, unit tests for each of the individual generators will be added to test the individual logic of those classes respectively.

### DIFF
--- a/lib/tapioca/generators/base.rb
+++ b/lib/tapioca/generators/base.rb
@@ -1,7 +1,7 @@
 # typed: strict
 # frozen_string_literal: true
 
-# NOTE: Temporary measure to get pairity with existing behaviour. Thor should be abstracted out so Generators
+# NOTE: Temporary measure to get parity with existing behaviour. Thor should be abstracted out so Generators
 # do NOT need to know about the CLI interface.
 require "thor"
 


lib/tapioca/config.rb
```diff
@@ -35,6 +35,8 @@ def outpath
     DEFAULT_GEMDIR = T.let("#{DEFAULT_RBIDIR}/gems", String)
     DEFAULT_TODOSPATH = T.let("#{DEFAULT_RBIDIR}/todo.rbi", String)
 
+    DSL_COMPILERS_PATH = T.let("#{__dir__}/compilers/dsl", String)
+
```
mutecipher, line 39: Initially this path was being generated in `generator.rb` which gave me a fun headache trying to determine why compilers weren't being found after the copy/pasta. 

Decided that since this is a static path, this is probably the best place for it to live.

lib/tapioca/generators/dsl.rb
```diff
@@ -0,0 +1,366 @@
...
+          Rake::Task["db:abort_if_pending_migrations"].invoke
+        end
+      end
+
+      sig { void }
+      def load_dsl_generators
+        say("Loading DSL generator classes... ")
+
+        Dir.glob([
+          "#{Tapioca::Config::DSL_COMPILERS_PATH}/*.rb",
```
mutecipher, line 114: This proved to be a fun bug to hunt down. 

In `generator.rb` it was listed as 
```ruby
"#{dir}/compilers/dsl/*.rb"
```
which was silently failing in the background of all the integrations tests after migrating the logic to this file (we're now one level higher). After determining the cause, I opted to extract this into a `Config` constant.

Morriar (commented):
Can't we also extract the `Requires` generator as a class?

lib/tapioca/generators/gem.rb
```diff
@@ -0,0 +1,372 @@
...
+      sig { void }
+      def perform_removals
+        say("Removing RBI files of gems that have been removed:", [:blue, :bold])
+        puts
+
+        anything_done = T.let(false, T::Boolean)
+
+        gems = removed_rbis
+
+        shell.indent do
```
mutecipher, line 222: The `indent` method was previously accessible due to our subclassing of `Thor::Shell`. So now I've patched it to call the instance of `shell` that lives on `Thor::Base` in `Generators::Base`.

lib/tapioca/generators/base.rb
```diff
@@ -0,0 +1,33 @@
...
+      attr_reader :config
+
+      abstract!
+
+      sig { params(config: Tapioca::Config).void }
+      def initialize(config)
+        @config = config
+      end
+
+      sig { abstract.void }
```
Morriar, line 26: Let's not introduce this right now if it's only to override it with empty implementations.

lib/tapioca/generators/gem.rb
```diff
@@ -0,0 +1,372 @@
...
+          .to_h
+      end
+
+      sig { params(message: String, color: T.any(Symbol, T::Array[Symbol])).void }
+      def say_error(message = "", *color)
+        force_new_line = (message.to_s !~ /( |\t)\Z/)
+        # NOTE: This is a hack. We're no longer subclassing from Thor::Shell::Color
+        # so we no longer have access to the prepare_message call.
+        # We should update this to remove this.
+        buffer = shell.send(:prepare_message, *T.unsafe([message, *T.unsafe(color)]))
```
mutecipher, line 352: Similarly, since we no longer subclass `Thor::Shell::Color`, the `prepare_message` protected method is no longer available to us. I had to do a small hack here to make it work as it did previously, but will need to re-evaluate in a future PR. 

My hope is to extract this kind of logic out to a `Logger` class that can abstract this away, and handle it for us.

lib/tapioca/generators/base.rb
Morriar: Or can't we pass all the options to the subclasses constructors so `generate` doesn't need different parameters and we can override it?

spec/spec_helper.rb
```diff
@@ -6,6 +6,7 @@
 require "minitest/spec"
 require "minitest/hooks/default"
 require "minitest/reporters"
+require "byebug"
```
mutecipher, line 9: I'm constantly importing and using `byebug` during debugging, so I included it in our `spec_helper`. It seems like a useful place to have it included, however, if there's strong opposition, I can purge this.

lib/tapioca/generators/dsl.rb
```diff
@@ -0,0 +1,366 @@
...
+
+          exit(1)
+        end
+
+        generator_map.values
+      end
+
+      sig do
+        params(constant_name: String, contents: String, outpath: Pathname, quiet: T::Boolean)
+          .returns(T.nilable(Pathname))
```
Morriar, line 185:
```suggestion
        params(
          constant_name: String, 
          contents: String, 
          outpath: Pathname, 
          quiet: T::Boolean
        ).returns(T.nilable(Pathname))
```

lib/tapioca/generators/base.rb
```diff
@@ -0,0 +1,33 @@
...
+      def generate; end
+
+      sig { abstract.params(error: Error).void }
```
Morriar, line 29: Same question here, why introduce it now if it's not used in subclasses?

lib/tapioca/generators/todo.rb
```diff
@@ -0,0 +1,74 @@
...
+        File.write(todos_path, content)
+
+        say("Done", :green)
+
+        say("All unresolved constants have been written to #{name}.", [:green, :bold])
+        say("Please review changes and commit them.", [:green, :bold])
+      end
+
+      sig { params(command: String, reason: T.nilable(String), strictness: T.nilable(String)).returns(String) }
+      def rbi_header(command, reason: nil, strictness: nil)
```
Morriar, line 53: It's used in multiple places, should we move it in the base class?

lib/tapioca/generators/base.rb
mutecipher: Either of these sounds like what I was thinking of. But you're right, the abstract methods weren't needed yet, and was too early of an optimization. I removed them from this PR and can address these types of changes as I go back and clean up the code, etc.

lib/tapioca/generators/todo.rb
mutecipher: I'm going to handle that in the next smaller PR where I can extract some shared logic to `Base` where it makes sense, as well as more appropriate classes.

As you mentioned in https://github.com/Shopify/tapioca/pull/496#discussion_r707401996, this would probably be better suited in `Rbi::File` as an extension or something to that effect.

But I wanna keep the scope of this PR focused and not start making too many out of scope changes.

Morriar:
> @KaanOzkan Honestly, kept it around due to the deprecated calls in `generators.rb`. I didn't want to include purging those in this PR, so I figured I'll do a separate PR which purges the deprecated commands and the file itself.

I agree with Kaan, I was expecting to see some code being removed. As it it's just creating duplication.

What's blocking to remove the code moved from generator to each class? Can't we change the deprecated commands to use the new generator classes?

mutecipher:
> > @KaanOzkan Honestly, kept it around due to the deprecated calls in `generators.rb`. I didn't want to include purging those in this PR, so I figured I'll do a separate PR which purges the deprecated commands and the file itself.
> 
> I agree with Kaan, I was expecting to see some code being removed. As it it's just creating duplication.
> 
> What's blocking to remove the code moved from generator to each class? Can't we change the deprecated commands to use the new generator classes?

@Morriar @KaanOzkan I disagree. IMO this PR is already getting past my comfort zone for bloat. While, yes, we have integration tests in place to assert behaviour, we should be doing this individually before the complete removal IMO as it's safer and easier to review.

RyanBrushett (commented):
I'm a big fan of very small PRs for sure.

Can we queue up a PR to do removal as well, based on this one? How many files need to be deleted?

Morriar (changes requested):
> What's blocking to remove the code moved from generator to each class? Can't we change the deprecated commands to use the new generator classes?

If we can't answer this question right now, it might be sign of something very wrong with the refactoring done in this PR. I'd prefer for it to do an actual cleaning rather than just duplicating code in the hope that, in a few PRs, things might finally be cleaner.

As per,

> There's already a lot going on in this PR so would like to keep it as is, and make smaller focused PRs for the other changes.

and

> I'm a big fan of very small PRs for sure.

In this case, let's do one PR by generator we extract instead of extracting all of them in the same PR. I'd prefer the code to be cleaned as we go instead of waiting the end of the refactoring we might have to revert because it doesn't work.

PR #509, opened 2021-09-21
Extract generator for requires

Morriar:
### Motivation

We want to extract each generator from the current `Generator`, because:

* better design: later we'll be able to introduce a common ancestor to all generators
* easier testing: we will be able to unit test each generator in isolation
* better separation of concerns: by breaking the coupling between generators

### Implementation

We extracted the code related to `requires` generation in it's own generator class.

### Tests

See automated tests (unchanged).

### DIFF
--- a/lib/tapioca/generators/require.rb
+++ b/lib/tapioca/generators/require.rb
@@ -31,8 +31,7 @@ module Tapioca
         # it with the new one found in the client code and remove the old ones.
         File.delete(@requires_path) if File.exist?(@requires_path)
 
-        content = String.new
-        content << "# typed: true\n"
+        content = +"# typed: true\n"
         content << "# frozen_string_literal: true\n\n"
         content << rb_string
 


Morriar:
We'll follow up on this pull-request by extracting more generators into their own classes 👍

paracycle:
> This new hierarchy might at some point become very similar to what we already have under `compilers`.
> 
> * What do you think about the generator hierarchy being parallel with the one under `compilers`?
> * Do you foresee some confusion between `generator` and `compiler`?
> * Should we merge both hierarchies and keep only one concept?

Let's keep separate hierarchies for now, since I don't think we should merge them. A gem RBI generator could rely on multiple "compiler"s (like the symbol table one, or YARD annotation one or something else) to generate the gem RBI, for example.

lib/tapioca/generators/require.rb
```diff
@@ -0,0 +1,51 @@
...
+          say("Nothing to do", :green)
+          return
+        end
+
+        # Clean all existing requires before regenerating the list so we update
+        # it with the new one found in the client code and remove the old ones.
+        File.delete(@requires_path) if File.exist?(@requires_path)
+
+        content = String.new
+        content << "# typed: true\n"
```
vinistock, line 35:
```suggestion
        content = +"# typed: true\n"
```

vinistock (approved):
Should the specs for this generator be extracted as well?

PR #510, opened 2021-09-21
Extract `Todo` generator from main `Generator`

mutecipher:
### Motivation

We want to extract the `todo` generator from the larger `Generator` class. This will allow for better design (smaller more isolated classes), and easier testability.

### Implementation

* Extracts `Todo` generator from `Generator`,
* Removes coupling between the `Config` and the new `Todo` class,
* Updates `Cli` to make use of new generator

### Tests

See automated tests (no change).


vinistock (approved):
Should we extract the specs too?

mutecipher:
> Should we extract the specs too?

Those are the integration tests that exist right now. Unit tests will be added after all the classes have been extracted.

Morriar (commented):
Let's rebase this on `main` since https://github.com/Shopify/tapioca/pull/509 has been merged.

RyanBrushett (approved):
Nice! Small PR. 🍏 assuming it's rebased.

PR #511, opened 2021-09-21
Extract `Dsl` generator from `Generator` class

mutecipher:
### Motivation

We want to extract each generator from the current Generator, because:

better design: later we'll be able to introduce a common ancestor to all generators
easier testing: we will be able to unit test each generator in isolation
better separation of concerns: by breaking the coupling between generators

### Implementation

* Extract `build_dsl` from `Generator` class into isolated class
* Decouple parameters from `Config` file
* Update `Cli` to use new `Dsl` class

### Tests

See automated tests (no changes).

### DIFF
--- a/lib/tapioca/config.rb
+++ b/lib/tapioca/config.rb
@@ -29,7 +29,6 @@ module Tapioca
     TAPIOCA_CONFIG = T.let("#{TAPIOCA_PATH}/config.yml", String)
 
     DEFAULT_COMMAND = T.let("bin/tapioca", String)
-    DEFAULT_COMPILER_PATH = T.let("#{__dir__}/compilers/dsl", String)
     DEFAULT_POSTREQUIRE = T.let("#{TAPIOCA_PATH}/require.rb", String)
     DEFAULT_RBIDIR = T.let("#{SORBET_PATH}/rbi", String)
     DEFAULT_DSLDIR = T.let("#{DEFAULT_RBIDIR}/dsl", String)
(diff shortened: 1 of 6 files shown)


lib/tapioca/config.rb
```diff
@@ -29,6 +29,7 @@ def outpath
     TAPIOCA_CONFIG = T.let("#{TAPIOCA_PATH}/config.yml", String)
 
     DEFAULT_COMMAND = T.let("bin/tapioca", String)
+    DEFAULT_COMPILER_PATH = T.let("#{__dir__}/compilers/dsl", String)
```
mutecipher, line 32: As mentioned in a previous PR, this is a static path that never changes, and was being dynamically generated in the `Generator` class. Moved it here as it makes more sense and won't be affected by the file moving to a different namespace.

lib/tapioca/generators.rb
```diff
@@ -0,0 +1,5 @@
+# typed: true
```
mutecipher, line 1: Moved to a top-level `generators` namespace to make for easier updates in the future. Can be updated in the future to dynamically pull in all generators.

paracycle (approved):
Looks good to me

lib/tapioca/config.rb
Morriar: I'm not sure about this one:
* Having it under `Config` implies this is something configurable.
* Having `default` in the name implies this is a default value that may be changed.
* It doesn't say anything about `dsl`

What do you think about moving it under `Tapioca::Compilers::Dsl::COMPILERS_PATH`?

PR #513, opened 2021-09-22
Extract `Gem` from `Generator` class

mutecipher:
### Motivation

We want to extract each generator from the current Generator, because:

* better design: later we'll be able to introduce a common ancestor to all generators
* easier testing: we will be able to unit test each generator in isolation
* better separation of concerns: by breaking the coupling between generators

### Implementation

* Extracted `build_gem_rbis` and `sync_rbis_with_gemfile` into `Gem` class from `Generator`
* Extracted `Config` dependency from `Gem` class
* Updated `Cli` to use the new `Gem` class
* Fixed use of deprecated commands in `gem_spec.rb`
* Removed `generator.rb` file (now empty)
* Removed deprecated `sync` and `generate` commands from `Cli`
* Updated `internals.rb` to use the new `generators.rb`

### Tests

See automated tests (removed `sync_spec.rb` and `generate_spec.rb`)


lib/tapioca/generators.rb
```diff
@@ -1,6 +1,9 @@
 # typed: true
 # frozen_string_literal: true
 
+require "thor" # TODO: Remove me when logging logic has been abstracted.
```
mutecipher, line 4: `generator.rb` previously contained `require "thor"` which was removed in this PR. Moved it to this file so it will be included in subsequent generators until it has been moved out into a logger.

spec/tapioca/cli/gem_spec.rb
```diff
@@ -278,9 +278,10 @@ class Foo::Secret; end
           File.delete(repo_path / "sorbet/tapioca/require.rb")
         end
       end
+
       describe("sync") do
         it "must perform no operations if everything is up-to-date" do
-          execute("generate")
+          execute("gem")
```
mutecipher, line 284: This was previously using the deprecated `generate` command, which I'm guessing was due to copy/pasta? Regardless, I updated to use the `gem` command and tests are passing as expected.

spec/tapioca/cli/gem_spec.rb
```diff
@@ -338,7 +339,7 @@ class Foo::Secret; end
         end
 
         it "must respect exclude option" do
-          execute("generate")
+          execute("gem")
```
mutecipher, line 342: Same here.

vinistock (approved):

PR #514, opened 2021-09-23
Move shared generator logic into `Base` class

mutecipher:
### Motivation

As part of the ongoing effort to cleanup/decouple the `Generators` in Tapioca. I've moved shared logic into a `Base` 
class. 

### Implementation

* Moved up the `T::Sig` and `Thor::Shell` includes
* Moved up the `default_command` value that is shared amongst all
  generators
* Made an abstract `generate` method to ensure that all `Generators` subclasses will follow that design

### Tests

See automated tests (no changes).

### DIFF
--- a/lib/tapioca/generators/base.rb
+++ b/lib/tapioca/generators/base.rb
@@ -1,0 +1,23 @@
+# typed: strict
+# frozen_string_literal: true
+
+module Tapioca
+  module Generators
+    class Base
+      extend T::Sig
+      extend T::Helpers
+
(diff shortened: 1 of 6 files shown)


lib/tapioca/generators/base.rb
```diff
@@ -0,0 +1,23 @@
+# typed: strict
+# frozen_string_literal: true
+
```
Morriar, line 3: Should we move the `require "thor" # TODO: Remove me when logging logic has been abstracted.` here?

vinistock (approved):

Morriar (approved):

Issue #250, opened 2021-03-08
Attribute accessors defined in concerns are not generated


vinistock:
When declaring attribute accessors within an `included` block of an `ActiveSupport::Concern`, the resulting methods are not generated in the signatures. For example,
```rb
# The concern
module Taggable
  extend ActiveSupport::Concern

  included do
    attr_accessor :tag
  end
end

# A class including the concern
class Article
  include Taggable
end

# Methods exist in `Article`, but are not generated in the RBIs
article = Article.new
article.tag = "my tag"
article.tag => "my tag"

# Expected RBI
class Article
  def tag=(tag); end
  
  def tag; end
end
```

PR #515, opened 2021-09-23
Handle class attributes used inside concerns

vinistock:
### Motivation

Sorbet is not able to identify class attributes created inside `included` blocks in concerns, which leads to false positives for method missing.

E.g.:
```ruby
module Concern
  extend ActiveSupport::Concern

  included do
    class_attribute :title
  end
end

class Article
  include Concern
end

Article.title # => Method missing! (false positive)
```

Tapioca can find these attributes when we find dynamic inclusion and extension and output the complete RBIs.

### Implementation

- Moved the original logic into `DynamicMixinCompiler` and modified it to keep track of class attributes
- Started generating class attribute related RBIs inside `GeneratedClassMethods` and `GeneratedInstanceMethods`
- Re-generated Tapioca's own RBIs

### Next steps

This PR will handle gems that create class attribute inside concerns, but applications can also do the same. We can leverage the same logic in a DSL generator to handle that case as well and produce the correct RBIs for concerns.

### Tests

See included tests.


lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -199,9 +199,14 @@ def compile_body(tree, name, constant)
           compile_methods(tree, name, constant)
           compile_module_helpers(tree, constant)
           compile_mixins(tree, constant)
-          compile_mixes_in_class_methods(tree, constant)
           compile_props(tree, constant)
           compile_enums(tree, constant)
+
+          unless constant.is_a?(Class)
+            analyzer = DynamicMixinCompiler.new(constant)
```
paracycle, line 206: Maybe we should stop calling it an `analyzer` now.

lib/tapioca.rb
```diff
@@ -21,4 +21,5 @@ class Error < StandardError; end
 require "tapioca/constant_locator"
 require "tapioca/compilers/dsl/base"
 require "tapioca/helpers/active_record_column_type_helper"
+require "tapioca/helpers/dynamic_mixin_compiler"
```
paracycle, line 24: I don't think this is a helper anymore, it has actually turned into a compiler itself. So maybe it should live under `tapioca/compilers/` too.

lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -496,13 +433,9 @@ def collect_mixed_in_class_methods(constant, dynamic_extends)
...
 
-        sig { params(tree: RBI::Tree, constant: Module).void }
-        def compile_mixes_in_class_methods(tree, constant)
-          return if constant.is_a?(Class)
-
-          dynamic_extends, dynamic_includes = collect_dynamic_mixins_of(constant)
-
-          dynamic_includes
+        sig { params(tree: RBI::Tree, constant: Module, mixin_analyzer: DynamicMixinCompiler).void }
+        def compile_mixes_in_class_methods(tree, constant, mixin_analyzer)
```
paracycle, line 437: Now that I think about it, I think we can move this method to `DynamicMixinCompiler` as well.

We need to:

1. Get rid of old Sorbet support and bump our min Sorbet dependency in the gemspec to `> 0.5.6200` https://github.com/Shopify/tapioca/blob/main/lib/tapioca/compilers/sorbet.rb/#L20-L22
2. That means we don't need the whole `collect_mixed_in_class_methods` method, we can just grab the code in here: https://github.com/Shopify/tapioca/blob/main/lib/tapioca/compilers/symbol_table/symbol_generator.rb/#L471-L473 and inline it into this function.
3. After calling this method on `mixin_compiler`, we can  then use the `dynamic_includes` and `dynamic_extends` attributes of `mixin_compiler` and add their symbols to the queue.

So I propose that we change `compile_body` to do:
```ruby
def compile_body(tree, name, constant)
  # ...
  compile_props(tree, constant)
  compile_enums(tree, constant)
  compile_dynamic_mixins(tree, constant)
end
```
and then, we can have:
```ruby
def compile_dynamic_mixins(tree, constant)
  # this is only valid for pure modules
  return if constant.is_a?(Class)

  mixin_compiler = DynamicMixinCompiler.new(constant)
  # we don't need to pass `constant` since the mixin compiler already has it, it can store it.
  mixin_compiler.compile_mixes_in_class_methods(tree) 
  mixin_compiler.compile_class_attributes(tree)

  (mixin_compiler.dynamic_includes + mixin_compiler.dynamic_extends).each do |mod|
    add_to_symbol_queue(name_of(mod))
  end
end
```

vinistock: I really like this refactor and it's indeed easier than we initially thought to move all of this inside the new mixin compiler. However, I did move the entire `collect_mixed_in_class_methods` inside as well. Changing the Sorbet dependency requirement to `0.5.6200` is a pretty big bump considering that we are currently on `0.4.4471`.

What do you think?

vinistock:
I have bumped the Sorbet requirement, applied the requested changes and made three fixes:
- started invoking `super` is defined inside re-definition of `class_attribute`
- added missing instance attribute checks for the `empty?` method in `DynamicMixinCompiler`
- started rejecting names that begin with `<#` in `Reflection#name_of`, in order to properly ignore anonymous classes

lib/tapioca/compilers/dynamic_mixin_compiler.rb
```diff
@@ -0,0 +1,199 @@
...
+
+        if instance_predicate
+          class_attribute_predicates.concat(attrs)
+
+          if instance_reader
+            instance_attribute_predicates.concat(attrs)
+          end
+        end
+
+        super(*attrs, **kwargs) if defined?(super)
```
paracycle, line 82: Can't we just do:
```suggestion
        super if defined?(super)
```
to pass all the arguments implicitly?

lib/tapioca/compilers/dynamic_mixin_compiler.rb
```diff
@@ -0,0 +1,199 @@
...
+
+    @instance_attribute_readers = T.let(instance_attribute_readers, T::Array[Symbol])
+    @instance_attribute_writers = T.let(instance_attribute_writers, T::Array[Symbol])
+    @instance_attribute_predicates = T.let(instance_attribute_predicates, T::Array[Symbol])
+  end
+
+  sig { returns(T::Boolean) }
+  def empty_attributes?
+    @class_attribute_readers.empty? && @class_attribute_writers.empty? && @class_attribute_predicates.empty? &&
+      @instance_attribute_readers.empty? && @instance_attribute_writers.empty? && @instance_attribute_predicates.empty?
```
paracycle, line 130: Do we really need instance attr checks? Actually, do we even need any class attr predicate check? The latter 4 cannot have elements if the first two are empty, so I guess it is enough to check:
```suggestion
    @class_attribute_readers.empty? && @class_attribute_writers.empty?
```

paracycle (commented):
Looks great, I have a couple of minor comments

lib/tapioca/compilers/dynamic_mixin_compiler.rb
paracycle: Hmmm, yeah, it could work differently indeed.

paracycle:
Fixes #250

vinistock:
@paracycle I think that in order to fix #250 we could use the same strategy, but we'd need to also patch `attr_reader`, `attr_accessor` and `attr_writer` to keep track of the generated methods. In the current state, we'll only catch `class_attribute`.

PR #516, opened 2021-09-24
Extract `Init` generator from `Cli`

mutecipher:
### Motivation

Make the `Cli` class as minimal as possible.

### Implementation

* Moved `Init` logic into it's own generator class
* Separated `generate_binstub` and `generate_binstub!` to differentiate
  the destructive behaviour in a more Ruby way
* Updated string output to use `Config::DEFAULT_COMMAND` when
  appropriate and updated the relevant tests

### Tests

See automated tests (small changes included).


lib/tapioca/generators/init.rb
```diff
@@ -0,0 +1,84 @@
...
+      def generate_binstub!
+        installer.generate_bundler_executable_stubs(spec, { force: true })
+        shell.say_status(:force, @default_command, :yellow)
+      end
+
+      sig { void }
+      def generate_binstub
+        installer.generate_bundler_executable_stubs(spec)
+        shell.say_status(:create, @default_command, :green)
+      end
```
mutecipher, line 71: Note: Separated these into two methods to be more Ruby-like to display the destructive nature of the `force` usage.

lib/tapioca/generators/init.rb
```diff
@@ -0,0 +1,84 @@
...
+
+      sig { void }
+      def create_post_require
+        @file_writer.create_file(@default_postrequire, skip: true) do
+          <<~CONTENT
+            # typed: true
+            # frozen_string_literal: true
+
+            # Add your extra requires here (`#{@default_command} require` can be used to boostrap this list)
```
mutecipher, line 56: This change led to changes in two different tests as it previously just stated `tapioca require` which must have slipped through our previous updates.

vinistock (approved):

lib/tapioca/generators/init.rb
```diff
@@ -0,0 +1,84 @@
+# typed: strict
+# frozen_string_literal: true
+
+module Tapioca
+  module Generators
+    class Init < Base
+      sig do
+        params(
+          file_writer: Thor::Actions,
```
Morriar, line 9: Isn't `Thor::Actions` a [module](https://rubydoc.info/github/wycats/thor/master/Thor/Actions)? Why can't we include it instead of using composition here?

lib/tapioca/cli.rb
```diff
@@ -29,9 +29,13 @@ class Cli < Thor
 
     desc "init", "initializes folder structure"
     def init
-      create_config
-      create_post_require
-      generate_binstub
+      generator = Generators::Init.new(
+        file_writer: self,
```
Morriar, line 33: This dependency is a bummer, especially for test isolation.

Morriar (commented):
It seems we use `File.write` instead of `Thor::Actions#create_file` in most places: 
* https://github.com/Shopify/tapioca/blob/main/lib/tapioca/generators/gem.rb#L162
* https://github.com/Shopify/tapioca/blob/main/lib/tapioca/generators/dsl.rb#L226

I'm not convinced of the necessity to include for `Thor::Actions` if this is only to abstract the `skip` behaviour in one place that could use a simple `if`.

If we do want to use this one instead. Let's move it as an include to the `Base` class and change the other uses so we don't need to add it here.

lib/tapioca/generators/init.rb
mutecipher: Yeah, we discussed this in the Slack thread. But, basically, it overrides the `initialize` method and injects `ClassMethods` into the class. So composition, in this case, is the better solution. As @paracycle notes below, decoupling this behaviour can be addressed in a separate PR.

PR #526, opened 2021-09-28
Add `rbi_ext/header.rb` module

mutecipher:
### Motivation

Progress towards cleaning out unrelated code from generators.

### Implementation

* Create an `RBI` extension, `Header`, which includes the `rbi_header` method.
* Update the references to use the new module

### Tests

See automated tests (no change).


vinistock:
Would it make sense to require and include the module in `Base` so that we don't need to do it on every generator?

mutecipher:
> Would it make sense to require and include the module in `Base` so that we don't need to do it on every generator?

> Could this be a method in `Base` instead?

So it's not currently being used in all generators, only `Dsl`, `Gem` and `Todo`.

So it's not necessarily needed for _every_ generator. So I don't think that's the right choice as it stands right now.

Morriar:
> So it's not necessarily needed for every generator.

But most of them right?

> So I don't think that's the right choice as it stands right now.

Can you elaborate? What would be the pros and cons of each approach?

PR #528, opened 2021-09-29
Add class_attributes DSL generator

vinistock:
### Motivation

In #515, we started handling `class_attribute` invocations that happen inside gems, but we are not currently generating the same RBIs for `class_attribute` usage inside the app's own code.

This PR leverages the same `DynamicMixinCompiler` to find and generate RBIs for class attributes inside an application.

### Implementation

The new DSL generator just delegates the hard work to the `DynamicMixinCompiler`.

**Questions**
- Are we missing something in `gather_constants`?
- Should we also include modules that have `inherited` and `extended` methods defined, since those can also be used to invoke `class_attribute`?

Btw, I tried using `respond_to?` instead of checking the methods array, but that breaks because some other gems patch `respond_to?`.

### Tests

See included tests.


KaanOzkan (approved):

Morriar (approved):

vinistock:
Changed the module matching to be a little bit smarter. It will now match any module that has overridden `Module.included`, which exists as a private method by default.

spec/tapioca/compilers/dsl/mixed_in_class_attributes_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+          extend ActiveSupport::Concern
+        end
+
+        module SomeOtherModule
+        end
+      RUBY
+
+      assert_includes(gathered_constants, "ManualIncluded")
+      assert_includes(gathered_constants, "Concern")
+      refute_includes(gathered_constants, "SomeOtherModule")
```
paracycle, line 29: I prefer checking for strict array equality here since that would catch it if some other module were to sneak into this list. Moreover `gathered_constants` does a built-in sort to make that easy.

```suggestion
      assert_equal(["ManualIncluded", "Concern"], gathered_constants)
```

spec/tapioca/compilers/dsl/mixed_in_class_attributes_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+
+    it("gathers modules with private included hooks") do
+      add_ruby_file("file.rb", <<~RUBY)
+        module PrivateIncluded
+          def self.included(base); end
+          private_class_method :included
+        end
+      RUBY
+
+      assert_includes(gathered_constants, "PrivateIncluded")
```
paracycle, line 40: Same comment as above:
```suggestion
      assert_equal(["PrivateIncluded"], gathered_constants)
```

paracycle (approved):
This looks really good. Small comment about test assertion but LGTM.

vinistock:
@paracycle in the latest commit I excluded classes from the generator, since we're not interested in them, which makes the list more accurate.

However, the issue with using `assert_equal` is that the generator matches many other modules not added by the test, like `T::Props` or `Thor::Base`. The list has about 20 modules or so.

Do you think we should match against the entire list?

PR #529, opened 2021-09-29
Update `Symbol(TableCompiler|Generator)` to use `RBI::File` over `String`

mutecipher:
### Motivation

This was a blocker for migrating all generators to use `RBI::File`.

`SymbolGenerator` and `SymbolTableCompiler` will now generate `RBI::Tree`s as opposed to `String`s.

### Implementation

Update the `sig`s and modify the implementation so that the transformations on the `RBI::Tree` take place in the
generator rather than the compiler.

Taken from Mojan's in-progress PR (see https://github.com/Shopify/tapioca/pull/453/commits/3543e3973cea81dad744465d710952a653758ab4)

### Tests

See automated tests (minor updates).

### DIFF
--- a/lib/tapioca/compilers/symbol_table/symbol_generator.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_generator.rb
@@ -41,12 +41,9 @@ module Tapioca
           gem.parse_yard_docs if include_doc
         end
 
-        sig { returns(RBI::Tree) }
-        def generate
-          rbi = RBI::Tree.new
-
+        sig { params(rbi: RBI::Tree).returns(RBI::Tree) }
+        def generate(rbi)
           generate_from_symbol(rbi, T.must(@symbol_queue.shift)) until @symbol_queue.empty?
-
           rbi
         end
 
--- a/lib/tapioca/compilers/symbol_table/symbol_generator.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_generator.rb
@@ -41,10 +41,9 @@ module Tapioca
           gem.parse_yard_docs if include_doc
         end
 
-        sig { params(rbi: RBI::File).returns(RBI::Tree) }
+        sig { params(rbi: RBI::File).void }
         def generate(rbi)
           generate_from_symbol(rbi.root, T.must(@symbol_queue.shift)) until @symbol_queue.empty?
-          rbi.root
         end
 
         private
--- a/lib/tapioca/compilers/symbol_table/symbol_generator.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_generator.rb
@@ -41,17 +41,13 @@ module Tapioca
           gem.parse_yard_docs if include_doc
         end
 
-        sig { returns(String) }
+        sig { returns(RBI::Tree) }
         def generate
           rbi = RBI::Tree.new
 
           generate_from_symbol(rbi, T.must(@symbol_queue.shift)) until @symbol_queue.empty?
 
-          rbi.nest_singleton_methods!
-          rbi.nest_non_public_methods!
-          rbi.group_nodes!
-          rbi.sort_nodes!
-          rbi.string
+          rbi
         end
 
         private
--- a/lib/tapioca/compilers/symbol_table/symbol_generator.rb
+++ b/lib/tapioca/compilers/symbol_table/symbol_generator.rb
@@ -41,10 +41,10 @@ module Tapioca
           gem.parse_yard_docs if include_doc
         end
 
-        sig { params(rbi: RBI::Tree).returns(RBI::Tree) }
+        sig { params(rbi: RBI::File).returns(RBI::Tree) }
         def generate(rbi)
-          generate_from_symbol(rbi, T.must(@symbol_queue.shift)) until @symbol_queue.empty?
-          rbi
+          generate_from_symbol(rbi.root, T.must(@symbol_queue.shift)) until @symbol_queue.empty?
+          rbi.root
         end
 
         private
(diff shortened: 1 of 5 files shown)


lib/tapioca/generators/gem.rb
```diff
@@ -377,6 +379,15 @@ def build_error_for_files(cause, files)
 
         "  File(s) #{cause}:\n  - #{filenames}"
       end
+
+      sig { params(rbi: RBI::Tree).returns(String) }
+      def transform_rbi(rbi)
```
Morriar, line 384: What do you think about moving this as a method of `RBI::Tree` under `rbi_ext`?

lib/tapioca/generators/gem.rb
```diff
@@ -144,7 +144,9 @@ def compile_gem_rbi(gem)
         say("Compiling #{gem_name}, this may take a few seconds... ")
 
         strictness = @typed_overrides[gem.name] || "true"
-        rbi_body_content = compiler.compile(gem, 0, @doc)
+        rbi = compiler.compile(gem, 0, @doc)
+        rbi.nest_singleton_methods!
```
Morriar, line 148: I don't think we need this here, since it's already done in `transform_rbi`.

spec/tapioca/compilers/symbol_table_compiler_spec.rb
```diff
@@ -40,7 +40,12 @@ def compile(include_docs = false)
       spec = Bundler::StubSpecification.from_stub(stub)
       gem = Tapioca::Gemfile::GemSpec.new(spec)
 
-      Tapioca::Compilers::SymbolTableCompiler.new.compile(gem, 0, include_docs)
+      rbi = Tapioca::Compilers::SymbolTableCompiler.new.compile(gem, 0, include_docs)
+      rbi.nest_singleton_methods!
```
Morriar, line 44: Having it under `RBI::Tree` would also simplify this.

Morriar (commented):
I think the title of this PR could be more descriptive.

lib/tapioca/compilers/symbol_table/symbol_generator.rb
```diff
@@ -41,17 +41,13 @@ def initialize(gem, indent = 0, include_doc = false)
           gem.parse_yard_docs if include_doc
         end
 
-        sig { returns(String) }
+        sig { returns(RBI::Tree) }
         def generate
```
paracycle, line 45: Can we change the signature of this method so that it takes an `RBI::Tree` instance (that it then updates and returns)? If we do that, then in the outer generator level we can instantiate an `RBI::File` as early as possible with the provided strictness level, etc and then pass in its root tree here to be modified.

Moreover, an extension method on `RBI::File` in Tapioca can be responsible for doing the nesting and grouping and sorting before calling the actual `RBI::File#string` method, like what @Morriar is asking for below.

mutecipher: I'm not 100% sure I get what you're asking for. You want the generate to accept an `RBI::Tree` and return an `RBI::Tree`?

paracycle: Yes, exactly. My idea is for code to flow like something like this:
```ruby
rbi_file = RBI::File.new(strictness: "true")
SymbolTableCompiler.new.generate(rbi_file.root)
File.write("foo.rbi", rbi_file.to_s) # where `to_s` calls `transform_rbi` on the `root` internally
```
We can then make file operations be another abstraction later on:
```
rbi_file = Tapioca:RBIFile.new(filename, strictness: "true")
SymbolTableCompiler.new.generate(rbi_file.root)
rbi_file.save
```

mutecipher: Perfect. That makes much more sense. Will start on implementing that now. 👍🏻

lib/tapioca/rbi_ext/model.rb
```diff
@@ -86,6 +86,11 @@ def create_method(name, parameters: [], return_type: "T.untyped", class_method:
       self << method
     end
 
+    sig { override.returns(String) }
+    def to_s
```
Morriar, line 90: Can we use something else dans `to_s`? Maybe `transformed_string`?

Morriar (changes requested):

lib/tapioca/rbi_ext/model.rb
```diff
@@ -4,6 +4,24 @@
 require "rbi"
 
 module RBI
+  class File
+    extend T::Sig
+
+    sig { returns(String) }
+    def transformed_string
+      transform_rbi!
+      root.string
```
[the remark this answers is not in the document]
mutecipher: `string` on `RBI::File` returns a different value that `string` on `RBI::Tree`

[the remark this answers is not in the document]
Morriar: Can you figure out why? What is the "weird formatting"?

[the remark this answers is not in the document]
mutecipher: Adds all comments and sigils to every file and reformats some of the signature outputs to be single-line etc.

Morriar:
Could we call `tranform_rbi!` and return `tree.root` in the tests instead
of this method so the tests stay unchanged for now and we get the right
behavior in the app?

Or we can delay this change until you merge your PR about `rbi_header` but
we should add a TODO here to not forget to change it.

lib/tapioca/generators/gem.rb
```diff
@@ -144,7 +144,9 @@ def compile_gem_rbi(gem)
         say("Compiling #{gem_name}, this may take a few seconds... ")
 
         strictness = @typed_overrides[gem.name] || "true"
-        rbi_body_content = compiler.compile(gem, 0, @doc)
+        rbi = RBI::File.new(strictness: strictness)
```
Morriar, line 147: Aren't we going to add the strictness twice if it's also done by `rbi_header`?

Morriar: Don't you get the same result if you do not pass any strictness to the `RBI::File`?

```suggestion
        rbi = RBI::File.new
```

PR #530, opened 2021-10-01
Clean up File IO in generators

mutecipher:
### Motivation

We were currently using two different approaches to writing files in generators. `File.write` and `Thor::Action.create_file`. We want to unify this to be consistent for all generators.

### Implementation

Make a `create_file` method in the `Base` generator, and update references/tests for the updated line output.

### Tests

See automated tests (minor changes).

### DIFF
--- a/lib/tapioca/generators/base.rb
+++ b/lib/tapioca/generators/base.rb
@@ -10,6 +10,10 @@ module Tapioca
       extend T::Sig
       extend T::Helpers
 
+      class FileWriter < Thor
+        include Thor::Actions
+      end
+
       # TODO: Remove me when logging logic has been abstracted
       include Thor::Base
 
@@ -17,6 +21,7 @@ module Tapioca
 
       sig { params(default_command: String).void }
       def initialize(default_command:)
+        @file_writer = T.let(FileWriter.new, FileWriter)
         @default_command = default_command
       end
 
@@ -37,6 +42,17 @@ module Tapioca
 
         $stderr.print(buffer)
         $stderr.flush
+      end
+
+      sig do
+        params(
+          path: T.any(String, Pathname),
+          content: String,
+          skip: T::Boolean
+        ).void
+      end
+      def create_file(path, content, skip: false)
+        @file_writer.create_file(path, skip: skip) { content }
       end
     end
   end
--- a/lib/tapioca/generators/base.rb
+++ b/lib/tapioca/generators/base.rb
@@ -48,11 +48,12 @@ module Tapioca
         params(
           path: T.any(String, Pathname),
           content: String,
-          skip: T::Boolean
+          skip: T::Boolean,
+          verbose: T::Boolean
         ).void
       end
-      def create_file(path, content, skip: false)
-        @file_writer.create_file(path, skip: skip) { content }
+      def create_file(path, content, skip: false, verbose: true)
+        @file_writer.create_file(path, skip: skip, verbose: verbose) { content }
       end
     end
   end
--- a/spec/tapioca/cli/dsl_spec.rb
+++ b/spec/tapioca/cli/dsl_spec.rb
@@ -33,7 +33,7 @@ module Tapioca
           Loading DSL generator classes... Done
           Compiling DSL RBI files...
 
-          Wrote: #{outdir}/post.rbi
+                create  #{outdir}/post.rbi
 
           Done
           All operations performed in working directory.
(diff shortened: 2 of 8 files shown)


lib/tapioca/generators/base.rb
```diff
@@ -10,13 +10,18 @@ class Base
       extend T::Sig
       extend T::Helpers
 
+      class FileWriter < Thor
+        include Thor::Actions
+      end
```
KaanOzkan, line 15: Are we okay with FileWriter having access to all Thor::Actions instead of just `create_file`?

spec/tapioca/cli/dsl_spec.rb
```diff
@@ -33,7 +33,7 @@ class DslSpec < CliSpec
           Loading DSL generator classes... Done
           Compiling DSL RBI files...
 
-          Wrote: #{outdir}/post.rbi
+                create  #{outdir}/post.rbi
```
KaanOzkan, line 36: I assume we can't customize the "create" string 😕.

lib/tapioca/generators/base.rb
mutecipher: I would say "yes", if only because we can update other helpers to take advantage of the built in `Thor::Actions`.

i.e. `create_directory`, etc.

spec/tapioca/cli/dsl_spec.rb
mutecipher: Unfortunately, no. But in this case I think it's a bit better looking because it's at least consistent.

lib/tapioca/generators/base.rb
KaanOzkan: 👍
Yeah I don't love the name of the class since it can do more. But it's a nitpick.

lib/tapioca/generators/base.rb
```diff
@@ -10,13 +10,18 @@ class Base
...
+
       # TODO: Remove me when logging logic has been abstracted
       include Thor::Base
 
       abstract!
 
       sig { params(default_command: String).void }
       def initialize(default_command:)
```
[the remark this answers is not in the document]
mutecipher: I'm not sure that would serve any significant use case for now. That basically takes us back to the previous implementation, just moving the writer to a nested class.

spec/tapioca/cli/todo_spec.rb
```diff
@@ -37,7 +37,10 @@ def foo
         File.delete(repo_path / "file.rb")
 
         assert_equal(<<~CONTENTS, output)
-          Compiling sorbet/rbi/todo.rbi, this may take a few seconds... Done
+          Finding all unresolved constants, this may take a few seconds... Done
+
+                create  sorbet/rbi/todo.rbi
```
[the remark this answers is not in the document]
mutecipher: That doesn't have any effect on the `Todo` generator. We don't hide any output under flags in the generation of the `todo.rbi`.

lib/tapioca/generators/base.rb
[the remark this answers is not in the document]
Morriar: It allows you to pass a custom `FileWriter`, for example for unit testing.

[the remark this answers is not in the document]
paracycle: It will make it easier to test these classes in isolation, though. In tests, we can pass in a fake `FileWriter` and assert operations performed after the call, without even touching the file system.

spec/tapioca/cli/todo_spec.rb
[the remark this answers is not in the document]
Morriar: I mean here: https://github.com/Shopify/tapioca/pull/530/files#diff-585ba67372fcead01370ead6886cf98e8d43dce4b7520c8083357e8a5577d0c2R41 should we pass `verbose: true` so the `create ...` message is not displayed?

spec/tapioca/cli/dsl_spec.rb
```diff
@@ -225,13 +225,13 @@ def body=(body); end
...
           Processing: Baz::Role
-          Wrote: sorbet/rbi/dsl/baz/role.rbi
+                create  sorbet/rbi/dsl/baz/role.rbi
           Processing: Job
-          Wrote: sorbet/rbi/dsl/job.rbi
+                create  sorbet/rbi/dsl/job.rbi
           Processing: Namespace::Comment
-          Wrote: sorbet/rbi/dsl/namespace/comment.rbi
+                create  sorbet/rbi/dsl/namespace/comment.rbi
           Processing: Post
```
paracycle, line 233: Should we also change these lines to `say_status(:processing, "Post", :yellow)` so that they line up better with the other output?

paracycle (commented):
The PR is good overall, agreed with @Morriar's points inline and I've also raised one suggestion which might make the output cleaner.

lib/tapioca/generators/base.rb
```diff
@@ -10,13 +10,18 @@ class Base
...
+
       # TODO: Remove me when logging logic has been abstracted
       include Thor::Base
 
       abstract!
 
-      sig { params(default_command: String).void }
-      def initialize(default_command:)
+      sig { params(default_command: String, file_writer: Thor::Actions).void }
+      def initialize(default_command:, file_writer: FileWriter.new)
```
Morriar, line 23: You'll need to reflect this change in all subclasses initializers.

mutecipher: It's using the default value which is a new instance of `FileWriter` which shouldn't require any changes to existing subclasses.

paracycle (approved):
Nice, thank you!

PR #538, opened 2021-10-06
Fixes for `sorbet` breaking change

mutecipher:
### Motivation

* Bump `sorbet` and `sorbet-runtime` to v0.5.9204
* Bump `spoom` to v1.1.4
* Bump `rbi` to v0.0.6
To unblock https://github.com/Shopify/tapioca/pull/532

### Implementation

`bundle update rbi spoom sorbet sorbet-runtime`

### Tests

N/A

---

Note: CI is currently breaking due to a change in Sorbet that is causing `spoom` to break. Fix for that is up at https://github.com/Shopify/spoom/pull/100

### DIFF
--- a/sorbet/rbi/shims/bundler.rbi
+++ b/sorbet/rbi/shims/bundler.rbi
@@ -1407,24 +1407,6 @@ end
 class Bundler::InstallHookError
 end
 
-class Bundler::Installer
-  def generate_bundler_executable_stubs(spec, options=T.unsafe(nil)); end
-
-  def initialize(root, definition); end
-
-  def post_install_messages(); end
-
-  def run(options); end
-end
-
-class Bundler::Installer
-  def self.ambiguous_gems(); end
-
-  def self.ambiguous_gems=(ambiguous_gems); end
-
-  def self.install(root, definition, options=T.unsafe(nil)); end
-end
-
 class Bundler::InvalidOption
   def status_code(); end
 end


paracycle (changes requested):
Doesn't this transitively bump the Sorbet version dependency in Tapioca? I don't want us to target a day old version of Sorbet.

mutecipher:
> Doesn't this transitively bump the Sorbet version dependency in Tapioca? I don't want us to target a day old version of Sorbet.

It wasn't for `sorbet-runtime` because neither `spoom` nor `rbi` had a specific version declaration. So was giving a version mismatch without.

paracycle:
> Can't we still merge this change and release Tapioca 0.6.0 which would require this version of `sorbet-runtime` while releasing more 0.5.X versions for backward compatible patches?

We could do that, but when there is a cleaner solution to this, which is to make the runtime `requires_ancestor` method accept normal parameter arguments, why should be make our dependency stricter? Stricter dependencies will mean a lot of work when teams inside and outside of Shopify are migrating to v6.0. That implies that adoption will be slow, when we want the exact opposite.

> I believe we already did this in the past for breaking changes on the static checker or in Tapioca's API.

We did do that, but that was mostly because we dropped support for Ruby 2.4 and 2.5. We know for sure that no project at Shopify uses those Ruby versions and they are old enough to drop at this point. This is not really the same situation, though. We are dropping support for 2 day old Sorbet versions all of a sudden.

PR #539, opened 2021-10-06
Bump `spoom` to v1.1.4

mutecipher:
- Bump `spoom` to v1.1.4
- Typecheck fixes


sorbet/rbi/shims/bundler.rbi
```diff
@@ -1410,8 +1410,6 @@ end
 class Bundler::Installer
   def generate_bundler_executable_stubs(spec, options=T.unsafe(nil)); end
 
-  def generate_standalone_bundler_executable_stubs(spec); end
-
```
mutecipher, line 1414: Looks like we have an RBI defined for this already, and our shim signature doesn't match.

PR #555, opened 2021-10-21
Support ActiveRecord::Enum for abstract classes

etiennebarrie:
### Motivation

ActiveRecord::Enum can be defined on abstract classes. Also [ActiveRecord::Enum can be inherited](https://github.com/rails/rails/blob/v6.1.4.1/activerecord/test/cases/enum_test.rb#L481), extending the list of potential values for the attribute.

### Implementation

Remove the guard against abstract class.

### Tests

Tweak test showing that abstract classes were ignored. Add test confirming inherited enums have methods declared for the values they define.

### DIFF
--- a/lib/tapioca/compilers/dsl/active_record_enum.rb
+++ b/lib/tapioca/compilers/dsl/active_record_enum.rb
@@ -80,7 +80,7 @@ module Tapioca
 
         sig { override.returns(T::Enumerable[Module]) }
         def gather_constants
-          descendants_of(::ActiveRecord::Base).reject(&:abstract_class?)
+          descendants_of(::ActiveRecord::Base)
         end
 
         private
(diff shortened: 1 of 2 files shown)


paracycle:
> * Why aren't we generating the superclass of the class we generate?

This is since these RBI files are enriching the actual class/module defined in the code base, and since that code is the actual source of truth, there is no need to define superclasses, etc in these RBI files.

paracycle:
> not sure why we ignore abstract classes.

This is probably a mistake. We've had to ignore abstract AR classes for the column, association generators since trying to read those properties tried to do database accesses which would fail for abstract models.

Your example clearly shows that the same assumption does not hold for abstract classes for enums.

I think I have a fix for this that I will push on this branch.

paracycle:
Btw, your test was not exactly correct, since we don't expect `Conversation` to declare all the `enum` attributes itself. We expect it to inherit the ones that its parent class declares. So the test needs to check the output of both `Conversation` and `AbstractConversation`, which actually passes, since `AbstractConversation` is not marked as an `abstract_class`

vinistock (approved):

paracycle:
Thank you @etiennebarrie!

Issue #536, opened 2021-10-05
Add an `ErrorHandler` class


mutecipher:
As part of implementing a solution for https://github.com/Shopify/tapioca/issues/238, I'm proposing that we move the error handling capabilities into it's own class to separate concern from the generators themselves, as the behaviour is generic enough that it could be used elsewhere in the codebase.

An implementation of the class, 
* should handle `registering` errors as they're encountered during `dsl` or `gem` generation.
* should print be able to hold information regarding the issue encountered, and 
  * if we understand what's going on, how to deal with it.

---

As @paracycle, noted in https://github.com/Shopify/tapioca/issues/238. We should provide the following information:
* which generator caused an error
* for which constant
* with what error message and stack trace

mutecipher:
FYI, closing this. This technically was completed with #553. If we want to elaborate, extract, etc. We can open another issue.

PR #558, opened 2021-10-22
Update `Tapioca::Error` class

mutecipher:
### Motivation

As part of #536 and #238, we want to update the error messaging for DSL compilation in `tapioca`. This introduces an updated `Tapioca::Error` class which includes additional fields for `generator` (the Generator which errored out) as well as `help_message` guiding messaging we can provide tips in when we know what likely went wrong.

### Implementation

Moved `Tapioca::Error` from `lib/tapioca.rb` to it's own file, `lib/tapioca/error.rb`. Extended the `initializer` method to accept additional fields, `generator` and `help_message`. Added some helper methods for formatting better output (still some more work could be done here, but this is a decent starting point).

I envision this will be used as a general case for `tapioca` related error messaging. I figure that DSL compilers could subclass this class for narrower, more scoped, errors in their respective classes to keep the errors close to the code.

**Considerations**:
* I'm on the fence whether we should define `help_messages` at runtime, or have them predefined for subclasses of `Tapioca::Error`
  * Runtime provides more flexibility, while predefined may have to be more general depending where it occurs
  * Keeping it predefined does have the benefit of maintaining a straightforward `StandardError` API which may be nice
* The `formatted_message` output is still _semi_-WIP right now. There's room for improvement, however, without definitive use cases it's harder to estimate. 

### Tests

See automated tests.

### DIFF
--- a/lib/tapioca.rb
+++ b/lib/tapioca.rb
@@ -13,8 +13,6 @@ module Tapioca
   ensure
     $VERBOSE = original_verbosity
   end
-
-  class Error < StandardError; end
 end
 
 require "tapioca/reflection"
--- a/lib/tapioca/error.rb
+++ b/lib/tapioca/error.rb
@@ -1,0 +1,59 @@
+# typed: strict
+# frozen_string_literal: true
+
+module Tapioca
+  class Error < StandardError
+    extend T::Sig
+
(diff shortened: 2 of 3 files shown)


mutecipher:
Closing in favour of https://github.com/Shopify/tapioca/pull/578

PR #575, opened 2021-10-29
Add an extension of PrivateRelation for "where chain"

### LINKED ISSUE #574: Generated AR Relations types don't support WhereChain methods
Loving the new AR Relations types that are generated by Sorbet, noticed something that's currently missing / slightly wrong with them.

When using the `#where` method, it'll return a [WhereChain](https://github.com/rails/rails/blob/v6.1.4/activerecord/lib/active_record/relation/query_methods.rb#L17) instance when called without arguments which can then be used to chain things like [`where.not`](https://edgeapi.rubyonrails.org/classes/ActiveRecord/QueryMethods/WhereChain.html#method-i-not) and the new methods [`where.missing`](https://edgeapi.rubyonrails.org/classes/ActiveRecord/QueryMethods/WhereChain.html#method-i-associated) and [`where.associated`](https://edgeapi.rubyonrails.org/classes/ActiveRecord/QueryMethods/WhereChain.html#method-i-missing).

The current approach has `not` as a method on PrivateRelation which isn't quite correct because Rails doesn't support `User.not(name: 'ryan')` (which currently passes type checking) but requires you to do `User.where.not(name: 'ryan')`.

Creating a new `PrivateWhereChainRelation` method that subclassed `PrivateRelation` would allow us to contain just the methods that are available in this context, and have `where` be typed to return an instance of `PrivateWhereChainRelation`. The methods themselves could then return a regular `PrivateRelation` as `User.where.not().not()` is not valid.

ryanwilsonperkin:
### Motivation
Fixes #574 

Rails 6.1 introduced an API for `Post.where.missing(:author)` that can
be used as syntactic sugar on top of joining and checking the associated
id is nil.

This leverages the existing QueryMethods::WhereChain class which we
haven't previously used in Tapioca.

WhereChain should be returned from `#where` and should give the ability
to do where.missing, where.not (and soon, where.associated). The result
should give back a plain PrivateRelation because
where.missing().missing() is invalid.

### Implementation
Introduced a new subclass of both `PrivateRelation` and `PrivateAssociationRelation` which appends a "WhereChain" suffix and implements `not` and `missing`, and adds the ability to pick up `associated` automatically once we update to the version of Rails that it is released in.

### Tests
Tested!

### DIFF
--- a/lib/tapioca/compilers/dsl/active_record_relations.rb
+++ b/lib/tapioca/compilers/dsl/active_record_relations.rb
@@ -465,13 +465,6 @@ module Tapioca
           def create_relation_methods
             create_relation_method("all")
             create_relation_method(
-              "not",
-              parameters: [
-                create_param("opts", type: "T.untyped"),
-                create_rest_param("rest", type: "T.untyped"),
-              ]
-            )
-            create_relation_method(
               "where",
               parameters: [
                 create_rest_param("args", type: "T.untyped"),
--- a/lib/tapioca/compilers/dsl/active_record_relations.rb
+++ b/lib/tapioca/compilers/dsl/active_record_relations.rb
@@ -297,6 +297,15 @@ module Tapioca
           def create_relation_where_chain_methods(klass)
             WHERE_CHAIN_QUERY_METHODS.each do |method_name|
               case method_name
+              when :not
+                klass.create_method(
+                  method_name.to_s,
+                  parameters: [
+                    create_param("opts", type: "T.untyped"),
+                    create_rest_param("rest", type: "T.untyped"),
+                  ],
+                  return_type: RelationClassName
+                )
               when :missing
                 klass.create_method(
                   method_name.to_s,
@@ -337,6 +346,15 @@ module Tapioca
           def create_association_relation_where_chain_methods(klass)
             WHERE_CHAIN_QUERY_METHODS.each do |method_name|
               case method_name
+              when :not
+                klass.create_method(
+                  method_name.to_s,
+                  parameters: [
+                    create_param("opts", type: "T.untyped"),
+                    create_rest_param("rest", type: "T.untyped"),
+                  ],
+                  return_type: AssociationRelationClassName
+                )
               when :missing
                 klass.create_method(
                   method_name.to_s,
--- a/lib/tapioca/compilers/dsl/active_record_relations.rb
+++ b/lib/tapioca/compilers/dsl/active_record_relations.rb
@@ -228,7 +228,7 @@ module Tapioca
               .grep_v(/=$/) # end with "=""
               .grep_v(/(?<!uniq)!$/) # end with "!" except for "uniq!"
           end, T::Array[Symbol])
-          QUERY_WHERE_CHAIN_METHODS = T.let(
+          WHERE_CHAIN_QUERY_METHODS = T.let(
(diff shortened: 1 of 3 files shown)


ryanwilsonperkin:
Note to self: this needs docs

ryanwilsonperkin:
An alternative less-correct-but-simpler approach would be to implement `missing` and `associated` on the PrivateRelation/PrivateAssociationRelation directly

lib/tapioca/compilers/dsl/active_record_relations.rb
```diff
@@ -219,13 +219,19 @@ def generate
...
+            query_methods -= [:where]
             # Remove the methods that ...
             query_methods
               .grep_v(/_clause$/) # end with "_clause"
               .grep_v(/_values?$/) # end with "_value" or "_values"
               .grep_v(/=$/) # end with "=""
               .grep_v(/(?<!uniq)!$/) # end with "!" except for "uniq!"
           end, T::Array[Symbol])
+          WHERE_CHAIN_QUERY_METHODS = T.let(
+            ActiveRecord::QueryMethods::WhereChain.instance_methods(false),
```
paracycle, line 232: While we don't have tests for it, we do try to keep support for Rails 5.2 as much as we  can. If this is a class added in Rails 6, maybe we should check that it is defined first?

lib/tapioca/compilers/dsl/active_record_relations.rb
```diff
@@ -260,7 +266,9 @@ def create_classes_and_includes
             end
 
             create_relation_class
+            create_relation_where_chain_class
```
paracycle, line 269: Could we run these as part of `create_relation_class` and `create_association_relation_class` respectively? I feel like it will keep this top level flow much easier to read.

lib/tapioca/compilers/dsl/active_record_relations.rb
```diff
@@ -614,17 +666,29 @@ def create_common_method(name, parameters: [], return_type: nil)
...
+              parameters: T::Array[RBI::TypedParam],
+              relation_return_type: String,
+              association_return_type: String,
+            ).void
+          end
+          def create_relation_method(
+            name,
+            parameters: [],
+            relation_return_type: RelationClassName,
+            association_return_type: AssociationRelationClassName
```
paracycle, line 681: Nice refactor! ❤️

paracycle (commented):
Thank you so much for this. I am also a little concerned about the opposite mistake we are doing here with supporting `where(...).not`, etc, but I guess we don't have many choices and that is the lesser error we could be making.

I left a couple of small inline comments, mostly around code org.

lib/tapioca/compilers/dsl/active_record_relations.rb
ryanwilsonperkin: Yeah sure thing

ryanwilsonperkin:
Noticed one issue while testing this out on Services-DB: the `Element = type_member(fixed: ...)` does not get picked up properly by the where chain subclass, and need to be added to that class as well.

paracycle:
> Noticed one issue while testing this out on Services-DB: the `Element = type_member(fixed: ...)` does not get picked up properly by the where chain subclass, and need to be added to that class as well.

Yeah, that would be the case. Good catch!

Morriar (approved):

ryanwilsonperkin:
👍🏻 Thanks folks!

PR #576, opened 2021-11-01
Move test helpers to `lib/tapioca`

mutecipher:
### Motivation

We want to have our test helpers be reusable for people extending `tapioca` capabilities. This moves our test helpers
to `lib/tapioca/helpers` and updates the existing references.

### Implementation

- Move test helper files to `lib/helpers`
- Update existing helper references

### Tests

See automated tests.


lib/tapioca/helpers/test/isolation.rb
```diff
@@ -0,0 +1,37 @@
...
+      module Isolation
+        extend T::Sig
+        require "thread"
+
+        sig { returns(T::Boolean) }
+        def self.forking_env?
+          !ENV["NO_FORK"] && Process.respond_to?(:fork)
+        end
+
+        def run
```
mutecipher, line 21: This was the root cause of `Minitest` failing to run locally and on CI (Thanks to @RyanBrushett for figuring it out). This method should technically return `Mintest::Result`, however, doing that raised some type errors since it doesn't return a `Minitest::Result` and parsing into a `Minitest::Result` caused _every_ test to run in isolation causing some messy output. 

Took @paracycle's suggestion from Slack and just removed the signature from this specific method. It did require lowering the strictness on this file from `strict` to `true`. But the test suite is working as expected again.

vinistock (approved):

KaanOzkan (approved):

lib/tapioca/ruby_ext/forking_patch.rb
```diff
@@ -0,0 +1,42 @@
+# typed: strict
+# frozen_string_literal: true
+
+module Forking
```
paracycle, line 4: These should not be in the `ruby_ext` folder, since they are actually internal modules of `Isolation`. Can we move these back please?

PR #582, opened 2021-11-01
Add support for dsl parallelization

### LINKED ISSUE #577: Refactor DSL generation to allow parallelizing the work
We started parallelizing RBI generation in #559. However, we had to revert the parallelization for DSLs in #573, because the DSL generator is not safe for parallelization.

We'll need to refactor the way we find out the stale DSL RBIs file to remove and bring back parallelism to DSL generators.

vinistock:
Closes #577 

### Motivation

In order to run the `dsl` generator in parallel, we need some way of keeping track of which files we should not purge.

### Implementation

By commit:
1. Allow the parallelized block by executor to return a value. That value is serialize using `Marshal` and written/read through a pipe so that the workers can return something from the block. The executor then returns an array of all of the values that were returned by each iteration in the block
2. Revert the commit that removed parallelization from DSL #573
3. Make the changes to return the filenames from the parallelized block, so that we can have the right information and purge them later

### Tests

See included tests.

### DIFF
--- a/lib/tapioca/compilers/dsl_compiler.rb
+++ b/lib/tapioca/compilers/dsl_compiler.rb
@@ -22,16 +22,24 @@ module Tapioca
           requested_constants: T::Array[Module],
           requested_generators: T::Array[T.class_of(Dsl::Base)],
           excluded_generators: T::Array[T.class_of(Dsl::Base)],
-          error_handler: T.nilable(T.proc.params(error: String).void)
+          error_handler: T.nilable(T.proc.params(error: String).void),
+          number_of_workers: T.nilable(Integer),
         ).void
       end
-      def initialize(requested_constants:, requested_generators: [], excluded_generators: [], error_handler: nil)
+      def initialize(
+        requested_constants:,
+        requested_generators: [],
+        excluded_generators: [],
+        error_handler: nil,
+        number_of_workers: nil
+      )
         @generators = T.let(
           gather_generators(requested_generators, excluded_generators),
           T::Enumerable[Dsl::Base]
         )
         @requested_constants = requested_constants
         @error_handler = T.let(error_handler || $stderr.method(:puts), T.proc.params(error: String).void)
+        @number_of_workers = number_of_workers
       end
 
       sig { params(blk: T.proc.params(constant: Module, rbi: RBI::File).void).void }
@@ -45,7 +53,10 @@ module Tapioca
           ERROR
         end
 
-        constants_to_process.sort_by { |c| c.name.to_s }.each do |constant|
+        Executor.new(
+          constants_to_process.sort_by { |c| c.name.to_s },
+          number_of_workers: @number_of_workers
+        ).run_in_parallel do |constant|
           rbi = rbi_for_constant(constant)
           next if rbi.nil?
 
--- a/lib/tapioca/compilers/dsl_compiler.rb
+++ b/lib/tapioca/compilers/dsl_compiler.rb
@@ -44,8 +44,8 @@ module Tapioca
 
       sig do
         type_parameters(:T).params(
-          blk: T.proc.params(constant: Module, rbi: RBI::File).returns(T.nilable(T.type_parameter(:T)))
-        ).returns(T.nilable(T::Array[T.nilable(T.type_parameter(:T))]))
+          blk: T.proc.params(constant: Module, rbi: RBI::File).returns(T.type_parameter(:T))
+        ).returns(T.nilable(T::Array[T.type_parameter(:T)]))
       end
       def run(&blk)
         constants_to_process = gather_constants(requested_constants)
@@ -71,7 +71,7 @@ module Tapioca
           report_error(msg)
         end
 
-        result
+        result&.compact
       end
 
       private
--- a/lib/tapioca/compilers/dsl_compiler.rb
+++ b/lib/tapioca/compilers/dsl_compiler.rb
@@ -42,7 +42,11 @@ module Tapioca
         @number_of_workers = number_of_workers
       end
 
-      sig { params(blk: T.proc.params(constant: Module, rbi: RBI::File).void).void }
+      sig do
+        type_parameters(:T).params(
+          blk: T.proc.params(constant: Module, rbi: RBI::File).returns(T.nilable(T.type_parameter(:T)))
+        ).returns(T.nilable(T::Array[T.nilable(T.type_parameter(:T))]))
+      end
       def run(&blk)
         constants_to_process = gather_constants(requested_constants)
 
@@ -53,7 +57,7 @@ module Tapioca
           ERROR
         end
 
-        Executor.new(
+        result = Executor.new(
           constants_to_process.sort_by { |c| c.name.to_s },
           number_of_workers: @number_of_workers
         ).run_in_parallel do |constant|
@@ -66,6 +70,8 @@ module Tapioca
         generators.flat_map(&:errors).each do |msg|
           report_error(msg)
         end
+
+        result
       end
 
       private
--- a/lib/tapioca/compilers/dsl_compiler.rb
+++ b/lib/tapioca/compilers/dsl_compiler.rb
@@ -45,7 +45,7 @@ module Tapioca
       sig do
         type_parameters(:T).params(
           blk: T.proc.params(constant: Module, rbi: RBI::File).returns(T.type_parameter(:T))
-        ).returns(T.nilable(T::Array[T.type_parameter(:T)]))
+        ).returns(T::Array[T.type_parameter(:T)])
       end
       def run(&blk)
         constants_to_process = gather_constants(requested_constants)
@@ -71,7 +71,7 @@ module Tapioca
           report_error(msg)
         end
 
-        result&.compact
+        result.compact
       end
 
       private
--- a/lib/tapioca/executor.rb
+++ b/lib/tapioca/executor.rb
@@ -33,7 +33,7 @@ module Tapioca
     end
     def run_in_parallel(&block)
       # If we only have one worker selected, it's not worth forking, just run sequentially
-      return @queue.map { |item| block.call(item) }.compact if @number_of_workers == 1
+      return @queue.map { |item| block.call(item) } if @number_of_workers == 1
 
       read_pipes = []
       write_pipes = []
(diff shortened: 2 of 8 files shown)


lib/tapioca/executor.rb
```diff
@@ -25,8 +25,14 @@ def initialize(queue, number_of_workers: nil)
...
+      type_parameters(:T).params(
+        block: T.proc.params(item: T.untyped).returns(T.type_parameter(:T))
+      ).returns(T.nilable(T::Array[T.type_parameter(:T)]))
+    end
     def run_in_parallel(&block)
+      read, write = IO.pipe
+
       # If we only have one worker selected, it's not worth forking, just run sequentially
       if @number_of_workers == 1
         block.call(@queue.shift) until @queue.empty?
```
paracycle, line 38: Lol, this code path should also be returning the result of block calls. I guess this is why the other tests are failing on CI.

lib/tapioca/executor.rb
```diff
@@ -38,11 +44,29 @@ def run_in_parallel(&block)
       workers = (0...@number_of_workers).map do
         items = @queue.shift(@items_per_worker)
 
-        fork { block.call(items.shift) until items.empty? }
+        fork do
+          read.close
+          result = items.map { |item| block.call(item) }.compact
+
+          packed = [Marshal.dump(result)].pack("m")
+          write.puts("-#{packed}") unless result.empty?
```
paracycle, line 52: Can we extract the delimiter (`-` in this case) to be a constant somewhere so that we can use the same constant below when parsing? It would also be a convenient place to note that we chose `-` since it is not in the Base64 alphabet that we are using to serialize the data on the pipe.

lib/tapioca/executor.rb
```diff
@@ -38,11 +44,29 @@ def run_in_parallel(&block)
...
       workers.each { |pid| Process.waitpid(pid) }
+
+      result = read.read
+      read.close
+
+      if result
+        result
+          .split("-")
+          .drop(1)
+          .flat_map { |item| T.unsafe(Marshal.load(item.unpack1("m"))) }
```
paracycle, line 68: It would be great to add comment here as to what this is really doing. The mechanics are easy to understand but the why is a little harder. Extracting the delimiter will help a little with the explanation too.

Maybe something like:
```suggestion
        result
          .split("-") # Split the output at delimiter that separates each worker's output
          .drop(1) # Drop the leading empty space
          .flat_map { |item| T.unsafe(Marshal.load(item.unpack1("m"))) } # Unpack the data and merge
```

spec/tapioca/cli/dsl_spec.rb
```diff
@@ -364,6 +374,24 @@ module GeneratedFoo; end
...
 
+      it "removes stale RBIs properly when running in parallel" do
+        FileUtils.mkdir_p("#{outdir}/baz")
+        FileUtils.mkdir_p("#{outdir}/namespace")
+
+        FileUtils.touch("#{outdir}/baz/role.rbi")
+        FileUtils.touch("#{outdir}/job.rbi")
+        FileUtils.touch("#{outdir}/post.rbi")
+        FileUtils.touch("#{outdir}/namespace/comment.rbi")
+
```
paracycle, line 385: Can we still add a couple of stale RBI files here in any case? I know it does not affect the case we are currently testing for, but it might catch against something we might end up introducing at some point in the future.

lib/tapioca/compilers/dsl_compiler.rb
```diff
@@ -55,6 +70,8 @@ def run(&blk)
         generators.flat_map(&:errors).each do |msg|
           report_error(msg)
         end
+
+        result
```
paracycle, line 74: I _think_we might be able to get rid of all the `T.nilable` shenanigans in the signature of this method, if we:
```suggestion
        result.compact
```
instead. That should ensure that we return an array on non-nilable items.

lib/tapioca/compilers/dsl_compiler.rb
```diff
@@ -22,19 +22,31 @@ class DslCompiler
...
         )
         @requested_constants = requested_constants
         @error_handler = T.let(error_handler || $stderr.method(:puts), T.proc.params(error: String).void)
+        @number_of_workers = number_of_workers
       end
 
-      sig { params(blk: T.proc.params(constant: Module, rbi: RBI::File).void).void }
+      sig do
+        type_parameters(:T).params(
+          blk: T.proc.params(constant: Module, rbi: RBI::File).returns(T.nilable(T.type_parameter(:T)))
```
paracycle, line 47: I don't think the block passed to this methods really returns anything that is nilable. It is just the block we pass to `Executor` inside this method returns either this type parameter or `nil`.
```suggestion
          blk: T.proc.params(constant: Module, rbi: RBI::File).returns(T.type_parameter(:T))
```

paracycle (commented):
Couple of comments!

lib/tapioca/executor.rb
vinistock: Added some explanation before these invocations. Let me know what you think.

lib/tapioca/executor.rb
```diff
@@ -25,24 +28,60 @@ def initialize(queue, number_of_workers: nil)
...
       # queue
       workers = (0...@number_of_workers).map do
         items = @queue.shift(@items_per_worker)
 
-        fork { block.call(items.shift) until items.empty? }
+        fork do
+          read.close
+          result = items.map { |item| block.call(item) }.compact
+
+          # We mapped the result of invoking the paralllized block into an array. In order to return the array from the
```
paracycle, line 53:
```suggestion
          # We mapped the result of invoking the parallelized block into an array. In order to return the array from the
```

paracycle (commented):
💯

paracycle (approved):

vinistock:
Although the tests pass, this is not ready to go yet. For some reason when trying to run this on applications hangs after completing the DSL generation. I'll try to discover why, but it seems that workers get stuck on `executor.rb:61` while writing to the pipe.

vinistock:
The problem that caused DSL to hang is fixed. The important changes of the latest commit are:

1. Read from the pipe before waiting for the processes to finish
2. Make sure to close all the created pipes both inside the forked worker and in the main processes. Not doing this leads to "too many files open" and breaks the test suite. I had to fix the isolation helper too

tasks/generator_documentation.rake
```diff
@@ -69,6 +69,8 @@ task generate_dsl_documentation: :yard_for_generate_documentation do
   end
 
   def load_registry
+    require "tapioca/compilers/dsl/config.rb"
```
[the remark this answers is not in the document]
vinistock: Done!

lib/tapioca/executor.rb
```diff
@@ -25,24 +28,62 @@ def initialize(queue, number_of_workers: nil)
...
       # If we only have one worker selected, it's not worth forking, just run sequentially
-      if @number_of_workers == 1
-        block.call(@queue.shift) until @queue.empty?
-        return
-      end
+      return @queue.map { |item| block.call(item) }.compact if @number_of_workers == 1
+
+      # Create an IO pipe to communicate the return value of the parallelized block back from the workers to the main
+      # process. It's important to only create pipes if we're running with more than one worker, otherwise the tests
+      # fail with a "too many open files" error.
```
Morriar, line 42: > otherwise the tests fail with a "too many open files" error

This is concerning. How do we reach the `$ ulimit -n` with only one worker? Are we missing a `close` somewhere? Or could this be related to the `IsolationHelper` we use?

lib/tapioca/executor.rb
```diff
@@ -25,24 +28,62 @@ def initialize(queue, number_of_workers: nil)
...
-        fork { block.call(items.shift) until items.empty? }
+        fork do
+          read.close
+          result = items.map { |item| block.call(item) }.compact
+
+          # We mapped the result of invoking the parallelized block into an array. In order to return the array from the
+          # worker back to the main process, we encode it in Base64, append a separator in the beginning and write it to
+          # the pipe. The separator helps us split the results that are coming from the multiple workers. It looks
+          # something like this:
+          # -absbasd13231-asbasd123123
```
Morriar, line 58: Since all workers write to the same pipe, isn't there a slight chance that two workers write at the same time and mangle the output?

```
worker1 writes"-1111"
worker2 writes "-2222"

output: "-11-222211"
```

Morriar: I _think_ it depends on the size of what you write and if it goes over the buffer size (which I _think_ is system dependent)?

paracycle: AFAIK, the default behaviour of Ruby IO objects is to not buffer on the Ruby side and write straight to the underlying IO: https://ruby-doc.org/core-3.0.2/IO.html#sync-method

As for the underlying pipe implementation, I think what you said is true:
> Write requests of {PIPE_BUF} bytes or less shall not be interleaved with data from other processes doing writes on the same pipe. Writes of greater than {PIPE_BUF} bytes may have data interleaved, on arbitrary boundaries, with writes by other processes, whether or not the O_NONBLOCK flag of the file status flags is set.

_From: https://pubs.opengroup.org/onlinepubs/9699919799/functions/write.html_

So this really depends on the definition for `PIPE_BUF` which seems to be fairly generous for the platforms we are concerned with: https://arto.s3.amazonaws.com/notes/posix#pipe-buf

paracycle: Alternatively, we could create a pipe-per-fork and read from them sequentially in the parent process. That would be resilient to this kind of behaviour and we won't have to worry about splitting the read data, etc either.

vinistock: I have switched the implementation to use multiple pipes, I think that's nicer and we can avoid the separator. It also makes sure the return is no longer nilable, which simplifies our types a bit.

lib/tapioca/executor.rb
```diff
@@ -25,24 +26,54 @@ def initialize(queue, number_of_workers: nil)
...
 
-        fork { block.call(items.shift) until items.empty? }
+        # Each worker has their own pair of pipes, so that we can read the result from each worker separately
+        read, write = IO.pipe
+        read_pipes << read
+        write_pipes << write
+
+        fork do
+          read.close
+          result = items.map { |item| block.call(item) }.compact
```
paracycle, line 53: Now that I think about this, maybe we should not be doing a `compact` here and just return an array of whatever the block call returned. There could be instances in the future where we might want to know what was returned even if it was `nil`. Just noting for the future, though, this is good as it stands now.

KaanOzkan (approved):

PR #584, opened 2021-11-02
Add `MissingConstantError` messaging

mutecipher:
Part of: #238

### Motivation

We want to advise the end user when associations cannot be generated due to a missing constant in the defined association.

See also: https://discourse.shopify.io/t/bin-tapioca-dsl-yields-uninitialized-constant/22411

### Implementation

Capture the raised `MissingConstantError` during `populate_association` and add a clearer error message to the error buffer.

### Tests

See automated tests.

### DIFF
--- a/lib/tapioca/compilers/dsl/active_record_associations.rb
+++ b/lib/tapioca/compilers/dsl/active_record_associations.rb
@@ -151,12 +151,12 @@ module Tapioca
               populate_single_assoc_getter_setter(mod, constant, association_name, reflection)
             end
           rescue SourceReflectionError
-            add_error(<<~MSG)
+            add_error(<<~MSG.strip)
               Cannot generate association `#{reflection.name}` on `#{constant}` since the source of the through association is missing.
             MSG
           rescue MissingConstantError
-            add_error(<<~MSG)
-              Cannot generate association `#{reflection.name}` on `#{constant}` since the constant `#{reflection.name.capitalize}` does not exist.
+            add_error(<<~MSG.strip)
+              Cannot generate association `#{reflection.name}` on `#{constant}` since the constant `#{reflection.class_name}` does not exist.
             MSG
           end
         end
--- a/lib/tapioca/compilers/dsl/active_record_associations.rb
+++ b/lib/tapioca/compilers/dsl/active_record_associations.rb
@@ -254,7 +254,7 @@ module Tapioca
         end
         def type_for(constant, reflection)
           return "T.untyped" if !constant.table_exists? || polymorphic_association?(reflection)
-          raise MissingConstantError unless Object.const_defined?(reflection.class_name)
+          raise MissingConstantError unless reflection.class_name.safe_constantize
 
           T.must(qualified_name_of(reflection.klass))
         end
@@ -266,6 +266,7 @@ module Tapioca
           ).returns(String)
         end
         def relation_type_for(constant, reflection)
+          raise MissingConstantError unless reflection.class_name.safe_constantize
           "ActiveRecord::Associations::CollectionProxy" if !constant.table_exists? ||
                                                             polymorphic_association?(reflection)
 
--- a/lib/tapioca/compilers/dsl/active_record_associations.rb
+++ b/lib/tapioca/compilers/dsl/active_record_associations.rb
@@ -104,6 +104,7 @@ module Tapioca
         include Helper::ActiveRecordConstants
 
         SourceReflectionError = Class.new(StandardError)
+        MissingConstantError = Class.new(StandardError)
 
         ReflectionType = T.type_alias do
           T.any(::ActiveRecord::Reflection::ThroughReflection, ::ActiveRecord::Reflection::AssociationReflection)
@@ -153,7 +154,7 @@ module Tapioca
             add_error(<<~MSG)
               Cannot generate association `#{reflection.name}` on `#{constant}` since the source of the through association is missing.
             MSG
-          rescue NameError
+          rescue MissingConstantError
             add_error(<<~MSG)
               Cannot generate association `#{reflection.name}` on `#{constant}` since the constant `#{reflection.name.capitalize}` does not exist.
             MSG
@@ -253,6 +254,7 @@ module Tapioca
         end
         def type_for(constant, reflection)
           return "T.untyped" if !constant.table_exists? || polymorphic_association?(reflection)
+          raise MissingConstantError unless Object.const_defined?(reflection.class_name)
 
           T.must(qualified_name_of(reflection.klass))
         end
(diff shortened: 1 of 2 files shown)


lib/tapioca/compilers/dsl/active_record_associations.rb
```diff
@@ -153,6 +153,10 @@ def populate_associations(mod, constant)
             add_error(<<~MSG)
               Cannot generate association `#{reflection.name}` on `#{constant}` since the source of the through association is missing.
             MSG
+          rescue NameError
```
KaanOzkan, line 156: Would it be better to create a custom error object similar to `SourceReflectionError` when we explicitly try to load the constant instead of catching all `NameErrors` within this method? I'm afraid down the line it will hide a real `NameError` bug. But that might be too low probability to care.

mutecipher: Yeah, we'd need to come up with some way of checking the actual class name without calling `reflection.klass` as that's the actual call that raises when it doesn't exist.

We can always add a ticket for this on the backlog for improvements. If other reviewers are fine with that.

Morriar: Shouldn't it be based on the `reflection.name` instead?

Something like:

```rb
raise ReflectionConstantError unless Object.const_defined?(reflection.name.to_s)
```

spec/tapioca/compilers/dsl/active_record_associations_spec.rb
```diff
@@ -407,7 +407,42 @@ def commenters_attributes=(attributes); end
...
+
+            module GeneratedAssociationMethods; end
+          end
+        RBI
+
+        assert_equal(expected, rbi_for(:Post))
+
+        assert_equal(1, generated_errors.size)
+        assert_equal(<<~MSG, generated_errors.first)
+          Cannot generate association `author` on `Post` since the constant `Author` does not exist.
```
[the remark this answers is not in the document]
Morriar: Or even better if we can get the location where this association is defined?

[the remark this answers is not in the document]
mutecipher: > Is it possible to include the type of association in the message? Something like
> 
> ```
> Cannot generate association `has_one: :author` on `Post` since the constant `Author` does not exist.
> ```

We don't have that specific information when the error is raised. It's failing before we get that far along because we're attempting to build the association via reflection. The code actually blows up when trying to determine that relation in https://github.com/Shopify/tapioca/blob/87bef08e088ea3b7496283dbf866b98425f6bebd/lib/tapioca/compilers/dsl/active_record_associations.rb#L172
The `type_for` method calls `relflection.klass` trying to build the association, and that's where the `NameError` gets raised.

> Or even better if we can get the location where this association is defined?

I don't intend to put it into the error message, but that was the idea behind the updated `Tapioca::Error` class, to print that information, and the error message together rather than jumbled together in error messages.

spec/tapioca/compilers/dsl/active_record_associations_spec.rb
```diff
@@ -407,7 +407,42 @@ def commenters_attributes=(attributes); end
...
+        RUBY
+
+        expected = <<~RBI
+          # typed: strong
+
+          class Post
+            include GeneratedAssociationMethods
+
+            module GeneratedAssociationMethods; end
+          end
```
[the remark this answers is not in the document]
mutecipher: I'm not too sure off hand, I haven't dug into the logic of that. So can't really speak to how difficult it would be.

lib/tapioca/compilers/dsl/active_record_associations.rb
mutecipher: Weird. `safe_constantize` seems to be causing failures locally for me, but `reflection.class_name` is working fine. Will push up the change.

spec/tapioca/compilers/dsl/active_record_associations_spec.rb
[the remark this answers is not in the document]
paracycle: We normally try to avoid this, but in this case, it is fine, since there was an error in `Post` that needs to be fixed. Once the error is fixed, we would either not generate this file at all or will be generating proper association methods for it.

lib/tapioca/compilers/dsl/active_record_associations.rb
```diff
@@ -153,6 +153,10 @@ def populate_associations(mod, constant)
...
+            add_error(<<~MSG)
+              Cannot generate association `#{reflection.name}` on `#{constant}` since the constant `#{reflection.name.capitalize}` does not exist.
```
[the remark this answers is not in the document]
paracycle: Oh, we should also change this to use `class_name` as well:
```suggestion
              Cannot generate association `#{reflection.name}` on `#{constant}` since the constant `#{reflection.class_name}` does not exist.
```

mutecipher:
^^ Rebasing off `main`

Was getting failures for `docs`? :wtf:

spec/tapioca/compilers/dsl/active_record_associations_spec.rb
```diff
@@ -407,7 +407,42 @@ def commenters_attributes=(attributes); end
...
+              create_table :posts do |t|
+                t.references(:author)
+              end
+            end
+          end
+        RUBY
+
+        add_ruby_file("post.rb", <<~RUBY)
+          class Post < ActiveRecord::Base
+            has_one :author
```
paracycle, line 423: Can we also add the cases for `has_many` and `belongs_to` and `has_and_belongs_to_many`?
```suggestion
            has_one :author
            has_many :comments
            belongs_to :blog
            has_and_belongs_to_many :commenters
```

lib/tapioca/compilers/dsl/active_record_associations.rb
```diff
@@ -249,6 +254,7 @@ def populate_collection_assoc_getter_setter(klass, constant, association_name, r
         end
         def type_for(constant, reflection)
           return "T.untyped" if !constant.table_exists? || polymorphic_association?(reflection)
+          raise MissingConstantError unless Object.const_defined?(reflection.class_name)
```
paracycle, line 257: Can we use the same mechanism as Active Record to do the `const_defined?` check? If Rails ever changes how it does that internally, we might end up not matching in behaviour. That's why I'd suggested `safe_constantize` instead:
```suggestion
          raise MissingConstantError unless reflection.class_name.safe_constantize
```

lib/tapioca/compilers/dsl/active_record_associations.rb
```diff
@@ -153,6 +154,10 @@ def populate_associations(mod, constant)
             add_error(<<~MSG)
               Cannot generate association `#{reflection.name}` on `#{constant}` since the source of the through association is missing.
             MSG
+          rescue MissingConstantError
+            add_error(<<~MSG)
```
paracycle, line 158: We probably want to `strip` the error message, since as it is now, it has a dangling `\n` at the end due to the heredoc usage:
```suggestion
            add_error(<<~MSG.strip)
```

We probably want to do the same this for the other error message too, btw.

mutecipher:
^^ Rebase with documentation fix off `main`

vinistock (approved):

PR #585, opened 2021-11-03
Refactor cli tests

Morriar:
### Motivation

Closes https://github.com/Shopify/tapioca/issues/432.

This is a major refactor of cli-related tests to not depend on real files and directory but dynamically create mocks for gems and projects we want to run Tapioca on.

### Previous approach

With the previous approach cli tests required a test project with real files under `spec/support/repo` and a mock gems written under `spec/support/gems`. This came with a few limitations:

* We were not running Tapioca in a real life setup and files were written in a custom location different from the one generally used by projects running Tapioca which at some point led to break the normal RBI output process and not catching it in tests (and one more bug I found while refactoring)
* Everything was executed in the context of Tapioca and its Bundle which means loading/installing things we might not want (for example `CliSpec` was bringing the whole test hardness for `DslSpec` while including `SpecHelper`)
* We had to add shims and special treatment for the support files (think Sorbet and Rubocop) since the files are committed in the Tapioca project
* As to limit the number of support files we were cramming as much features in the mock gems / repo as we could leading to less atomic and slower tests
* We didn't distinguish between `stdout` and `stderr` leading to error messages being printed in the wrong stream (see the many `FIXME` I added)
* We generally lacked the ability (or the motivation) to represent complex setups to test edge cases

### New approach

With the new approach we provide a framework to create complex setups (projects and required gems) on the fly and try to fix all previous limitations at the cost of a slightly more verbose (but also more flexible) testing framework. Here's a simple example showing how the mocks are used:

```rb
# Create a mock gem which uses `smart_properties`
foo = mock_gem("foo", "1.0.0", ["smart_properties"]) do
  write("lib/foo.rb", <<~RB)
    require "smart_properties"

    class Role
      include SmartProperties
      property :title, accepts: String
    end
  RB
end

# Create a mock project which depends on `foo`
project = mock_project do
  require_mock_gem(foo)
  require_real_gem("smart_properties", "1.15.0")
  bundle install

  write("lib/main.rb", <<~RB)
    require "foo"
  RB
end

# Run `tapioca` in the context of `project`
out, err, status = @project.tapioca("dsl")

assert_empty(err)

assert_includes(out, <<~OUT)
  create  sorbet/rbi/dsl/foo.rbi
OUT

assert(status)

# Cleanup
foo.destroy
project.destroy
```

While this might seem tedious, keep in mind that a lot of this can be factorized inside `before`/`after` blocks to minimize the amount of code to write for each test.

#### Outcomes

Since we're much more intentional in what we install/load/generate, another outcome is a major improvement in testing time, especially for the `gem_spec` test:

|                        | before | after |
|--------------|--------|-----|
| `gem_spec`  | 6m 32s | 1m 40s |
| `dsl_spec`    | 34s | 21s |

Another outcome is that all the cli-related tests are now `typed: strict` and every instance variable is now typed which lets us benefit from all the power of autocompletion with the Sorbet LSP integration.

### Implementation

The main pieces are:

* `MockDir`: a `TMPDir`-like used to create mocks of gems and projects
* `MockGem`: a mock gem in which we can add lib files and gemspec dependencies
* `MockProject`: a project context in which we can create lib files, include mock gems and run `bundle exec tapioca`
* `SpecWithProject`: A Spec superclass that creates a MockProject before running the tests and deletes it after

Disclaimer: I had to be a bit creative with the `before(:all)` and `after(:all)` to allow using them while keeping the instance variables compatible with `typed: strict`. Some cleaning can be done later if we teach Sorbet how to handle them properly.

### Tests

It's all about them. All previous tests are still here, I also added a few more around gem and dsl RBI generation.

### Next steps

I see at least two more things we can do to clean and speed up tests:

1. Refactor the `describe` blocks to better benefit from `before` and `after` hooks and create customized contexts for each kind of test without re-running `bundle install` or other commands again and again. Right now the `describe` blocks are organized around the Tapioca sub-command being tested which prevent this change.
2. Avoid generating RBIs for Tapioca's own dependencies during tests since we never assert anything about their presence or their content and most of the time spent is related to loading/generating those.

### Review

I strongly recommend reading this PR commit by commit as the diff might be a bit intimidating. There is now functional change in this PR but a tiny fix in d912b9757d8a52e5d9387f2fd7e7946f7ac25fb0.


spec/helpers/mock_project.rb
```diff
@@ -0,0 +1,83 @@
...
+        [out, err, T.must(status.success?)]
+      end
+    end
+
+    # Run a Tapioca `command` with `bundle exec` in this project context (unbundled env)
+    sig { params(command: String).returns([String, String, T::Boolean]) }
+    def tapioca(command)
+      exec_command = ["tapioca", command]
+      exec_command << "--workers=1" if command.start_with?(/dsl|gem/) && !command.match?("--workers")
+      bundle_exec(exec_command.join(" "))
```
KaanOzkan, line 80: Why do we need to run with one worker in tests?

KaanOzkan (commented):
😍

spec/helpers/mock_project.rb
vinistock: Because the output order changes when running with multiple forks and a lot of our tests expect a specific order in the output. It would require changing the expectations.

vinistock (approved):
This is great

mutecipher (approved):

spec/spec_with_project.rb
```diff
@@ -0,0 +1,133 @@
...
+    attr_reader :project
+
+    # around do |&block|
+    #   puts "AROUND"
+    #   files_before = project.files
+    #   super(&block)
+    #   files_after = project.files
+
+    #   assert_equal(files_before, files_after)
+    # end
```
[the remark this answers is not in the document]
Morriar: 🤦

spec/helpers/mock_project.rb
Morriar: > the output order changes when running with multiple forks

Maybe we should change this in subsequent PRs, I'm sure we can get the tests to run a bit faster with multiple workers.

PR #587, opened 2021-11-03
Generate typed store accessors in their own module

Morriar:
### Motivation

As to better represent the real behaviour here: https://github.com/rails/rails/blob/57fe7dfce9fc751dd521ec8e175924eb1480b5d5/activerecord/lib/active_record/store.rb#L131.

One problem with the current implementation occurs when models redefine the method to change its type. The type used by Sorbet then depends on the load order (last loaded wins):

```rb
# User redefinition in the model Ruby file
class MyModel
  sig { returns(TypeB) }
  def value; end
end

# Definition in RBI as generated by Tapioca
class MyModel
  sig { returns(TypeA) }
  def value; end
end

T.reveal_type(MyModel.new.value) # TypeA, because loaded last wins
```

By using a included module we avoid this indeterministic behaviour:

```rb
# User redefinition in the model Ruby file
class MyModel
  sig { returns(TypeB) }
  def value; end
end

# Definition in RBI as generated by Tapioca
class MyModel
  include StoreAccessors

  module StoreAccessors
    sig { returns(TypeA) }
    def value; end
  end
end

T.reveal_type(MyModel.new.value) # always TypeB
```

Please note that this kind of redefinition is type unsafe (and works only because we didn't specify `override`):

```rb
sig { params(accessor: MyModel::StoreAccessor).void }
def foo(accessor)
  bar(accessor.value) # Will break at runtime if called with `MyModel`
end

sig { params(value: TypeA).void }
def bar(string)
  # Something expecting a `TypeA`
end

foo(MyModel.new) # Boom!
```

### Tests

See updated test.

cc. @rafaelfranca

### DIFF
--- a/lib/tapioca/compilers/dsl/active_record_typed_store.rb
+++ b/lib/tapioca/compilers/dsl/active_record_typed_store.rb
@@ -34,53 +34,57 @@ module Tapioca
       # # post.rbi
       # # typed: true
       # class Post
-      #   sig { params(review_date: T.nilable(Date)).returns(T.nilable(Date)) }
-      #   def review_date=(review_date); end
+      #   include StoreAccessors
       #
-      #   sig { returns(T.nilable(Date)) }
-      #   def review_date; end
+      #   module StoreAccessors
+      #     sig { params(review_date: T.nilable(Date)).returns(T.nilable(Date)) }
+      #     def review_date=(review_date); end
       #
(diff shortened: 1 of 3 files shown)


paracycle (approved):
Perfect, thank you!

vinistock (approved):

rafaelfranca (approved):

Issue #551, opened 2021-10-20
Consider updating `Generators::Gem#remove` to use `Thor::Actions#remove_file`


mutecipher:
Right now there's some inconsistency between creating and removing RBI files. Should look at moving to use `Thor::Actions#remove_file` for consistent output and behaviour. Also, worth looking into similar behaviour for `def move` if `Thor::Actions` has a synonymous helper.

PR #593, opened 2021-11-04
Replace `remove` method and use `remove_file`

mutecipher:
Closes: #551

### Motivation

Maintain consistency with our file IO in the application. Since we use `Thor` helpers for creating files, we should do the same for deleting files.

### Implementation

Removed custom logic for removing files and instead updated to use `Thor::Actions#remove_file` in its place.

### Tests

See automated tests.

### DIFF
--- a/lib/tapioca/generators/base.rb
+++ b/lib/tapioca/generators/base.rb
@@ -56,6 +56,16 @@ module Tapioca
       def create_file(path, content, force: true, skip: false, verbose: true)
         @file_writer.create_file(path, force: force, skip: skip, verbose: verbose) { content }
       end
+
+      sig do
+        params(
+          path: T.any(String, Pathname),
+          verbose: T::Boolean
+        ).void
+      end
+      def remove_file(path, verbose: true)
+        @file_writer.remove_file(path, verbose: verbose)
+      end
     end
   end
 end
--- a/lib/tapioca/generators/dsl.rb
+++ b/lib/tapioca/generators/dsl.rb
@@ -172,7 +172,8 @@ module Tapioca
         unless unprocessable_constants.empty?
           unprocessable_constants.each do |name, _|
             say("Error: Cannot find constant '#{name}'", :red)
-            remove(dsl_rbi_filename(name))
+            filename = dsl_rbi_filename(name)
+            remove_file(filename) if File.file?(filename)
           end
 
           exit(1)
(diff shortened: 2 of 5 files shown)


lib/tapioca/generators/dsl.rb
```diff
@@ -172,7 +172,8 @@ def constantize(constant_names)
         unless unprocessable_constants.empty?
           unprocessable_constants.each do |name, _|
             say("Error: Cannot find constant '#{name}'", :red)
-            remove(dsl_rbi_filename(name))
+            filename = dsl_rbi_filename(name)
+            remove_file(filename) if File.file?(filename)
```
mutecipher, line 176: Added this extra check here so that we don't attempt to delete a file that doesn't exist and match the current test output.

spec/tapioca/cli/dsl_spec.rb
```diff
@@ -102,10 +102,10 @@ module GeneratedFoo; end
           Compiling DSL RBI files...
 
           Error: Cannot find constant 'NonExistent::Foo'
-          -- Removing: #{outdir}/non_existent/foo.rbi
+                remove  #{outdir}/non_existent/foo.rbi
           Error: Cannot find constant 'NonExistent::Bar'
           Error: Cannot find constant 'NonExistent::Baz'
-          -- Removing: #{outdir}/non_existent/baz.rbi
+                remove  #{outdir}/non_existent/baz.rbi
```
mutecipher, line 108: This is the test in relation to the above logic check.

KaanOzkan (approved):

lib/tapioca/generators/base.rb
```diff
@@ -56,6 +56,16 @@ def say_error(message = "", *color)
...
       end
+
+      sig do
+        params(
+          path: T.any(String, Pathname),
+          verbose: T::Boolean
+        ).void
+      end
+      def remove_file(path, verbose: true)
+        @file_writer.remove_file(path, verbose: verbose)
```
vinistock, line 67: Does this expect interactive input by the user? That is, do we need the `force` flag as we need for `create_file`?

lib/tapioca/generators/dsl.rb
vinistock: Should we use `File.exist?` instead? What's the difference between that and `file?`?

PR #595, opened 2021-11-04
Exclude anonymous or value constants from DSL generation

### LINKED ISSUE #592: ActiveRecord generators break on anonymous classes
Hello! First off, thanks for an excellent library, I've been getting deep into it over the last week and am so impressed by the quality of the project.

I'm getting breakages in the ActiveRecord DSL generators (`ActiveRecordColumns`, `ActiveRecordRelations`, and `ActiveRecordScope`) if, at any point before running them, I invoke `ActiveRecord::Base.connection.schema_migration`. This in turn [generates](https://github.com/rails/rails/blob/37c46e859bc53175dba913e61b60042aee216ab1/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb#L157) an anonymous class:

```rb
Class.new(ActiveRecord::SchemaMigration) do
  define_singleton_method(:name) { schema_migration_name }
  define_singleton_method(:to_s) { schema_migration_name }

  self.connection_specification_name = spec_name
end
```

The issue is that there is now a `MyModel::SchemaMigration` class in `ObjectSpace` (where MyModel is the base model in my application) that is both a) a descendant of `ActiveRecord::Base`, meaning Tapioca will try to generate RBI for it, and b) anonymous, meaning Tapioca won't know where to write it (and will raise an [exception](https://github.com/Shopify/tapioca/blob/4169604043d0cd0a043b156cf63b1bfcf1093f59/lib/tapioca/rbi_ext/model.rb#L57), which is the error I'm actually encountering).

One solution might be to additionally filter constants in the `gather_constants` [method](https://github.com/Shopify/tapioca/blob/4169604043d0cd0a043b156cf63b1bfcf1093f59/lib/tapioca/compilers/dsl/active_record_columns.rb#L130) of each ActiveRecord generator to ones that `Tapioca::Reflection` can find a name for:
```rb
def gather_constants
  descendants_of(::ActiveRecord::Base).reject(&:abstract_class?).select {|c| Tapioca::Reflection.name_for(c)}
end
```

It might even make sense to do this for all generators? Creating classes at runtime with `Class.new` shouldn't blow up Tapioca if a generator happens to find that class in its implementation of `gather_constants`.

paracycle:
### Motivation

Fix #592

### Implementation

The whole idea of DSL generation revolves around being able to generate files that are named after the constants that are being decorated, so it makes sense to filter out any constants that were collected that are anonymous.

Moreover, while we enforce that the constants that we expect to be collected are supposed to be classes/modules, we never actually enforce that at runtime, in the absence of Sorbet runtime, so I've added a filter to exclude value constants from the list of collected constants as well.

### Tests

No added tests, but I added an anonymous Sidekiq Worker class in the test repo to trigger a failure if we don't filter anonymous classes.

### DIFF
--- a/lib/tapioca/compilers/dsl_compiler.rb
+++ b/lib/tapioca/compilers/dsl_compiler.rb
@@ -49,6 +49,8 @@ module Tapioca
       end
       def run(&blk)
         constants_to_process = gather_constants(requested_constants)
+          .select { |c| Reflection.name_of(c) && Module === c } # Filter anonymous or value constants
+          .sort_by! { |c| T.must(Reflection.name_of(c)) }
 
         if constants_to_process.empty?
           report_error(<<~ERROR)
@@ -58,7 +60,7 @@ module Tapioca
         end
 
         result = Executor.new(
-          constants_to_process.sort_by { |c| c.name.to_s },
+          constants_to_process,
           number_of_workers: @number_of_workers
         ).run_in_parallel do |constant|
           rbi = rbi_for_constant(constant)
--- a/spec/support/repo/lib/job.rb
+++ b/spec/support/repo/lib/job.rb
@@ -6,3 +6,8 @@ class Job
   def perform(foo, bar)
   end
 end
+
+# This is to make sure we don't fail on anonymous constants
+Class.new do
+  include Sidekiq::Worker
+end


mutecipher:
Looks like you got a merge conflict after the DSL parallelization PR. ^^

KaanOzkan (approved):

PR #605, opened 2021-11-10
Add support for Rails generators

michaelherold:
### Motivation

Rails generators are application code, though not part of your main Rails application. The files are not automatically loaded when starting your app because Railties loads them when you run the generate command.

However, since it's part of your application, it's helpful to be able to enable strong typing within those files. Without knowing about the Thor-based DSL, you're left writing shims yourself.

Having support within Tapioca means that you won't have to manage this manually any more.

### Implementation

I load the two most likely candidate files. There are others that we could load, for example:

- `rails/generators/active_model`
- `rails/generators/erb`
- `rails/generators/generated_attribute`

I didn't want to get into the rabbit hole of trying to load all of them because that would be brittle between Rails versions. Do you think that's the correct choice?

Also, in order to avoid generating RBIs for built-in classes, I opted for pattern matching on the class name. That doesn't feel great, but I wasn't sure what would be a better approach. There are two main base classes, `Rails::Generators::NamedBase` and `Rails::Generators::Base`, but there are also some in gems other than `railties`. Do you have suggestions on a better solution?

~I had to add some RBIs for `railties` for type-checking since I use them in the tests. It appears that a prior version of Tapioca (or maybe Parlour?) wrote the existing file and that it used to include comments. Instead of including the big, messy diff, I included only the pieces that I needed. This leaves the file looking inconsistent because there are chunks in the new style and the rest is in the old. Is that okay, or should I rewrite the whole file?~ I rewrote the whole file.

### Tests

Yep, there are tests for corner cases that I could think of.


lib/tapioca/compilers/dsl/rails_generators.rb
```diff
@@ -0,0 +1,131 @@
...
+      # ~~~
+      #
+      # this compiler will produce the RBI file `service_generator.rbi` with the following content:
+      #
+      # ~~~rbi
+      # # service_generator.rbi
+      # # typed: strong
+      #
+      # class ServiceGenerator
+      #   sig { returns(::String)}
```
Morriar, line 34: I'm not familiar with Rails generator but could this be nil? Or will the generator error before?

lib/tapioca/compilers/dsl/rails_generators.rb
```diff
@@ -0,0 +1,131 @@
...
+        )
+
+        sig do
+          override
+            .params(
+              root: RBI::Tree,
+              constant: T.class_of(::Rails::Generators::Base)
+            )
+            .void
+        end
```
Morriar, line 56: I think we can inline this:

```suggestion
        sig { override.params(root: RBI::Tree, constant: T.class_of(::Rails::Generators::Base)).void }
```

lib/tapioca/compilers/dsl/rails_generators.rb
```diff
@@ -0,0 +1,131 @@
...
+              !name.match?(BUILT_IN_MATCHER) &&
+              const < ::Rails::Generators::Base
+          end
+        end
+
+        private
+
+        sig do
+          params(klass: RBI::Tree, argument: T.any(Thor::Argument, Thor::Option)).void
+        end
```
Morriar, line 87:
```suggestion
        sig { params(klass: RBI::Tree, argument: T.any(Thor::Argument, Thor::Option)).void }
```

lib/tapioca/compilers/dsl/rails_generators.rb
```diff
@@ -0,0 +1,131 @@
...
+        def generate_methods_for_argument(klass, argument)
+          type = type_for(argument)
+
+          klass.create_method(
+            argument.name,
+            parameters: [],
+            return_type: type
+          )
```
Morriar, line 95:
```suggestion
          klass.create_method(
            argument.name,
            parameters: [],
            return_type: type_for(argument)
          )
```

lib/tapioca/compilers/dsl/rails_generators.rb
michaelherold: Arguments are required, options are not. So this will always be a String, not `T.nilable(String)`. The only nilable DSL methods are `class_options` without a default.

To better understand the difference, arguments are akin to positional arguments while options are akin to keyword arguments. As an example:

    bin/rails g my_generator <argument> [--<class_option> [value]]

I could add an option without a default to demonstrate a nilable option as part of this if you'd like. It's covered by the test suite.

lib/tapioca/compilers/dsl/rails_generators.rb
```diff
@@ -0,0 +1,131 @@
...
+              constant: T.class_of(::Rails::Generators::Base)
+            )
+            .void
+        end
+        def decorate(root, constant)
+          base_class = base_class_for(constant)
+          arguments = constant.arguments - base_class.arguments
+          class_options = constant.class_options.reject do |name, option|
+            base_class.class_options[name] == option
+          end
```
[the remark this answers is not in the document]
michaelherold: The check below could be part of `gather_constants`, but we would have to repeat the filtering in `#decorate`. This filters out class options that the parent defines, because we don't need to define them as part of the child class's RBI (since they are already part of the parent's).

I opted to only perform the filter once and include the early return below so we don't have to perform repeat work.

What do you think?

michaelherold (commented):
I resolved the style nits and left the other two threads open with comments. Thanks!

KaanOzkan (commented):
> Instead of including the big, messy diff, I included only the pieces that I needed

Feel free to include the big messy diff for `railtie.rbi`, since that's what tapioca outputted. 

However, since [this gem RBI update PR](https://github.com/Shopify/tapioca/pull/647) hasn't touched `railtie` I'm thinking that you didn't supply the `--doc` option? If that's the case please supply that and commit the RBI as a whole.

paracycle (approved):
I am not in love with the `BUILT_IN_MATCHER` gymnastics, but I guess this is the best way of excluding internal classes, for now.

Thanks for working on this and putting this together. Much appreciated.

PR #609, opened 2021-11-11
Sanitize type aliases

vinistock:
### Motivation

Because we weren't sanitizing type alias values, a proc type alias that returns void would be printed as `<VOID>`.
```ruby
# This code
MyType = T.type_alias { T.proc.params(val: T.untyped).void }

# Became this RBI
MyType = T.type_alias { T.proc.params(val: T.untyped).returns(<VOID>) }
```

We already handle such cases in other parts of the symbol generator, but we simply didn't do it for type aliases.

### Implementation

Use our method for sanitizing types before inserting the type alias value.

### Tests

See included tests.


paracycle (approved):
Amazing, thank you.

mutecipher (approved):

PR #613, opened 2021-11-16
Bump `rbi` to v0.0.7

Morriar:
### Motivation

Unblocks #453.

### DIFF
--- a/lib/tapioca/rbi_ext/model.rb
+++ b/lib/tapioca/rbi_ext/model.rb
@@ -37,7 +37,7 @@ module RBI
 
     sig { void }
     def set_empty_body_content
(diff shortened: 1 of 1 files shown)


vinistock:
Do the code changes work with previous RBI versions? That is, is it enough to upgrade RBI in the `Gemfile.lock` or do we need to add a version constraint in the dependency?

Morriar:
> Do the code changes work with previous RBI versions? That is, is it enough to upgrade RBI in the `Gemfile.lock` or do we need to add a version constraint in the dependency?

Good call, I gated the version in the Gemspec 👍

vinistock (approved):

mutecipher (approved):

Issue #618, opened 2021-11-17
Parallel RBI generation is not deterministic because of autoload


vinistock:
When generating RBIs, we require the bundle/application ahead of time in the main process, but that's not sufficient to load the entire code.

When we delegate generation to the `Executor`, the `SymbolTableCompiler` will autoload constants for the first time inside a forked worker. This means that the other workers will not know about the autoloaded constant, which leads to RBIs that are not deterministic.

It's possible to verify that this is the case by inspecting the `$LOADED_FEATURES` outside and inside of the forked workers. This prints different loaded features from the workers, which is most likely the source of the non-deterministic behavior when running in parallel.
```ruby
# lib/tapioca/generators/gem.rb

def generate
  # ...

  current_features = $LOADED_FEATURES
  Executor.new(.....) do |gem|
    puts $LOADED_FEATURES - current_features
    # ...
  end
  # ...
end
```

PR #619, opened 2021-11-18
Set default number of workers to one

vinistock:
### Motivation

Until we can figure out a way of fixing #618, we should default to sequential RBI generation and warn users that using more than one worker can cause trouble.

### Implementation

Just changed the defaults and added a warning when using a different number of workers.

### Tests

Current tests are enough.

### DIFF
--- a/lib/tapioca/cli.rb
+++ b/lib/tapioca/cli.rb
@@ -82,8 +82,7 @@ module Tapioca
     option :workers,
       aliases: ["-w"],
       type: :numeric,
-      default: nil,
-      desc: "Number of parallel workers to use when generating RBIs"
+      desc: "EXPERIMENTAL: Number of parallel workers to use when generating RBIs"
     def dsl(*constants)
       current_command = T.must(current_command_chain.first)
       config = ConfigBuilder.from_options(current_command, options)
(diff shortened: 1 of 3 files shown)


lib/tapioca/cli.rb
```diff
@@ -82,7 +82,7 @@ def todo
     option :workers,
       aliases: ["-w"],
       type: :numeric,
-      default: nil,
+      default: 1,
       desc: "Number of parallel workers to use when generating RBIs"
```
Morriar, line 86:
```suggestion
      desc: "EXPERIMENTAL: Number of parallel workers to use when generating RBIs"
```

lib/tapioca/cli.rb
```diff
@@ -82,7 +82,7 @@ def todo
     option :workers,
       aliases: ["-w"],
       type: :numeric,
-      default: nil,
+      default: 1,
```
Morriar, line 85: From https://github.com/Shopify/tapioca/pull/566, it seems this isn't the right way to fix the default configuration values?

vinistock: Removed the default CLI option according to that PR.

mutecipher (approved):

PR #622, opened 2021-11-22
Add spin bootstrap script

mutecipher:
### Motivation

`spin`-ify the `tapioca` gem for development on Shopify's internal environment.

### Implementation

Add a `.spin/bootstrap` script to pull down the `libsqlite3-dev` library and install dependencies.

### Tests

N/A

### DIFF
--- a/.spin/bootstrap
+++ b/.spin/bootstrap
@@ -1,0 +1,7 @@
+#!/usr/local/env bash
+
+# Install some dependencies
+apt-get install libsqlite3-dev
+
+# Configure environment
+bin/setup
--- a/.spin/bootstrap
+++ b/.spin/bootstrap
@@ -1,7 +1,7 @@
 #!/usr/local/env bash
 
 # Install some dependencies
-apt-get install libsqlite3-dev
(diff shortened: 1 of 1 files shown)


.spin/bootstrap
```diff
@@ -0,0 +1,7 @@
+#!/usr/local/env bash
+
+# Install some dependencies
+apt-get install libsqlite3-dev
```
Morriar, line 4: Not sure what is the common way to do it with `spin` but since the install is from a script it might make sense to pass the `-y` flag:

```suggestion
apt-get install -y libsqlite3-dev
```

Morriar (approved):

.spin/bootstrap
mutecipher: Good catch. Added `--no-install-recommends` as well to avoid `apt` pulling unnecessary packages. This is a similar pattern used in `Dockerfile` for Codespaces (see [here](https://github.com/microsoft/vscode-dev-containers/blob/19adbac72304efc3f7b09970a3bac3c5a68a4c27/containers/ruby/.devcontainer/Dockerfile#L10-L11)).

PR #627, opened 2021-11-23
Compare modules by identity for Tapioca::Compilers::Dsl::Base#processable_constants

etiennebarrie:
This prevents exceptions from badly-behaved modules that implement `eql?` in a manner incompatible with `Object#eql?`.

```ruby
bad_module = Object.new
def bad_module.eql?() end # incompatible with Object#eql?

hash_conflict = Object.new
hash_conflict.define_singleton_method(:hash) { bad_module.hash }

require "set"
set = Set.new
set << hash_conflict
set.include?(bad_module) rescue p $!
set.compare_by_identity
p set.include?(bad_module)
```

There's nothing specific about modules, but since Module doesn't implement eql? for its instances, we can assume it's equivalent to comparing by identity.

### Motivation

We have a `bin/tapioca dsl --verify` in our CI and it randomly fails. I tracked it down to a misbehaved `Dry::Logic::Predicates` implementing `eql?` (at the singleton level) in a manner incompatible with `Object#eql?`, see https://github.com/dry-rb/dry-logic/issues/73#issuecomment-971595351.

While ultimately the fix should be in there, I felt like this was acceptable given we know we're dealing with modules which don't define `eql?` and `hash`, hence we know our set can use comparison by identity.

### Implementation

Use `Set#compare_by_identity` on the set of `processable_constants`.
Define the shim because it's missing from Sorbet.

### Tests

I didn't see any tests specifically for `Tapioca::Compilers::Dsl::Base#processable_constants` or `Tapioca::Compilers::Dsl::Base#handles?` so I left it out.

### DIFF
--- a/sorbet/rbi/shims/set.rbi
+++ b/sorbet/rbi/shims/set.rbi
@@ -1,0 +1,5 @@
+# typed: strict
+
+class Set
+  def compare_by_identity; end
+end


sorbet/rbi/shims/set.rbi
```diff
@@ -0,0 +1,5 @@
+# typed: strict
+
+class Set
+  def compare_by_identity; end
```
Morriar, line 4: Good catch, I opened a PR upstream to define this method: https://github.com/sorbet/sorbet/pull/4906.

paracycle (approved):
Makes sense, thank you.

PR #628, opened 2021-11-23
Move `include Thor::Base` to `CliHelper`

mutecipher:
### Motivation

Now that logging helpers have been moved out, I am able to move the `include Thor::Base` to a more appropriate place with those helpers.

### Implementation

Lingering `TODO` from `Generator` refactor. Now that logging helpers have been extracted to `CliHelper`, the `include Thor::Base` (which was included for those logging methods) can live with them.

### Tests

See automated tests. (No change)

### DIFF
--- a/lib/tapioca/helpers/cli_helper.rb
+++ b/lib/tapioca/helpers/cli_helper.rb
@@ -8,7 +8,7 @@ module Tapioca
     extend T::Sig
     extend T::Helpers
 
-    requires_ancestor { Thor::Shell }
+    include Thor::Base
 
     sig { params(message: String, color: T.any(Symbol, T::Array[Symbol])).void }
     def say_error(message = "", *color)
--- a/lib/tapioca/generators/base.rb
+++ b/lib/tapioca/generators/base.rb
@@ -12,7 +12,6 @@ module Tapioca
       end
 
       include CliHelper
-      include Thor::Base # TODO: Remove me when logging logic has been abstracted
 
       abstract!
 


lib/tapioca/helpers/cli_helper.rb
```diff
@@ -8,7 +8,7 @@ module CliHelper
     extend T::Sig
     extend T::Helpers
 
-    requires_ancestor { Thor::Shell }
+    include Thor::Base
```
Morriar, line 11: It doesn't make sense to include `Base` here. This module is a helper that (for now) only requires the client to bring `Shell`. It shouldn't bring [options](https://github.com/rails/thor/blob/master/lib/thor/base.rb#L35) nor anything related to command handling.

Morriar (commented):
From what I understand in the TODO message, we wanted to provide an abstraction for logging that could be implemented using `Thor`. I'm not sure the current PR actually fixes this TODO.

Morriar (commented):
This [comment](https://github.com/Shopify/tapioca/pull/628#pullrequestreview-813851716) hasn't been answered nor addressed. Are we still planing to provide an abstraction?

Plus I don't understand why we want to move the include in the helper. `Thor::Shell` will be provided by the classes including `CliHelper`.

mutecipher:
Ahh, my bad. Misread the comment. 

> From what I understand in the TODO message, we wanted to provide an abstraction for logging that could be implemented using `Thor`. I'm not sure the current PR actually fixes this TODO.

We were previously discussing the idea of removing the `Thor` dependency all together. However, we've since changed gears and seemingly leaned more into `Thor` when it comes to File IO. So the `TODO` itself is, IMO, out-of-date.

I think it's safe to keep `Thor` for now, unless we decide to go a different route when it comes to logging (i.e. Kafka, or some other such aggregate).

Morriar:
Ok, in that case let's just remove the TODO from here but keep the include in `Generators::Base` 👍 

If we're going that direction, do we still need the `FileWriter` abstraction?

mutecipher:
> Ok, in that case let's just remove the TODO from here but keep the include in `Generators::Base` 👍
> 
> If we're going that direction, do we still need the `FileWriter` abstraction?

I think the `FileWriter` is probably the right direction still, it abstracts the `Thor::Actions` dependency from the `Generator` and keeps that logic well confined.

I'd still like to do something similar with logging (down the road) where we maintain a simple API and can change the backend of it as necessary without needing to update the whole codebase.

RyanBrushett (approved):

RyanBrushett:
Can you edit the PR title so it better reflects what we did here? Like "Clean up a TODO" or something to that effect.

PR #637, opened 2021-11-29
Properly load reflections for namespaced has_many relations

jflinter:
### Motivation
This follows up https://github.com/Shopify/tapioca/pull/634 to properly load reflections for namespaced has_many relations (vs just has_one relations).

### Implementation
It's a pretty straightforward port of the logic from #634.

### Tests
Updated the test suite to cover the new functionality.

### DIFF
--- a/lib/tapioca/compilers/dsl/active_record_associations.rb
+++ b/lib/tapioca/compilers/dsl/active_record_associations.rb
@@ -328,9 +328,9 @@ module Tapioca
           ).returns(String)
         end
         def relation_type_for(constant, reflection)
-          raise MissingConstantError unless reflection.class_name.safe_constantize
-          "ActiveRecord::Associations::CollectionProxy" if !constant.table_exists? ||
-                                                            polymorphic_association?(reflection)
+          validate_reflection!(reflection)
+          return "ActiveRecord::Associations::CollectionProxy" if !constant.table_exists? ||
+                                                                   polymorphic_association?(reflection)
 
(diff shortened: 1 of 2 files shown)


jflinter:
I actually think tests are failing because I uncovered a bug - if you look carefully in the [prior implementation](https://github.com/Shopify/tapioca/pull/637/files#diff-8d6f9eab330d7ad3102c9c4c7a8f58099275a00a48880b883d41fc0079ed36dbL332), there isn't actually a return in the error case (i.e., line 332 is a no-op). This PR fixes that issue at the cost of breaking some tests that seem to have been depending on the wrong behavior.

jflinter:
(Updated the tests - it's nice, the generated RBI is a lot better now!)

Morriar (approved):

jflinter:
Hmm, I've tried signing the CLA a bunch of times, at different times of the day, etc and get the github error every time. Is there a way to do it manually or out-of-band?

paracycle:
> Hmm, I've tried signing the CLA a bunch of times, at different times of the day, etc and get the github error every time. Is there a way to do it manually or out-of-band?

@jflinter I've been looking at our CLA system that hosts that form and I was able to recreate the problem you were having and I think I've solved your problem. Could you please try again? Thank you 🙏

jflinter:
@paracycle that seems to have worked, thanks! Want to re-run the CLA workflow?

paracycle:
Thanks again for a great PR and apologies for the run-around with the CLA process.

jflinter:
No problem! Happy to be able to get off our fork of tapioca now!

PR #639, opened 2021-11-30
Eager load constants registered for `autoload` after loading gems

### LINKED ISSUE #618: Parallel RBI generation is not deterministic because of autoload

paracycle:
### Motivation

Fixes #618

We have been getting non-deterministic and order-dependent generation for gem RBI files due to constants registered for `autoload`. Since the act of loading those constants can have other side effect such as triggering other requires, etc, we were getting a change in behaviour depending on exactly when the autoload was triggered.

### Implementation

The implementation adds an `AutoloadTracker` class that hooks `Module#autoload` and collects all constant names that have been registered for autoload. After gem loading is finished, Tapioca asks `AutoloadTracker` to eager load all the constants registered as such.

In order to handle cases where an autoload operation registers another autoload, `AutoloadTracker` treats its collection of constant names as a queue and the eager load operation pops the first item in the queue and works until there are no items left.

### Tests

Added a test that model a specific case we knew about `yard` and `rake` that was because of autoload and was resulting in the need for adding `require "rake/tasklib"` in `tapioca/require.rb` files manually.


spec/tapioca/compilers/symbol_table_compiler_spec.rb
```diff
@@ -386,7 +386,7 @@ class Array
           def foo_int; end
         end
 
-        class Module
+        class Range
```
vinistock, line 389: Why did these turn into `Range`?

vinistock (approved):

lib/tapioca/autoload_tracker.rb
```diff
@@ -0,0 +1,32 @@
...
+    def autoload(module_name, filename)
+      AutoloadTracker.register("#{self}::#{module_name}")
+      super
+    end
+
+    sig { void }
+    def self.eager_load_all!
+      until @constant_names_registered_for_autoload.empty?
+        # Grab the next constant name
+        constant_name = T.must(@constant_names_registered_for_autoload.shift)
```
Morriar, line 19: If you want to avoid the comments here you could extract these lines as methods:

```rb
sig { returns(String) }
def pop_next_constant_name
  T.must(@constant_names_registered_for_autoload.shift)
end

sig { params(constant_name: String).void }
def trigger_constant_autoload(constant_name)
  Reflection.constantize(constant_name, inherit: true)
end
```

spec/tapioca/compilers/symbol_table_compiler_spec.rb
Morriar: From the commit comment: 

> We sometimes patch `Module` inside Tapioca so it is not a good candidate for testing this behaviour.

I guess we say the autoload hook appearing in the test output.

paracycle: Since we are prepending a module inside of `Module` in `AutoloadTracker`, this test started picking up that new module. So I opted to go for a class that we don't modify inside of Tapioca.

lib/tapioca/autoload_tracker.rb
paracycle: I think the code is readable enough in its current form, but I wanted to make it explicit that we were treating the array as a queue with the comments.

PR #641, opened 2021-11-30
Add a command to clean shim files automatically

### LINKED ISSUE #435: Check useless shims after RBI/DSL generation

Morriar:
### Motivation

As we get more precise content generated from the `gem` and `dsl` commands, we can expect that some shim definition are not useful anymore.

This PR adds a `clean-shims` command that remove the definitions that are no longer required from the shim RBI files.

Closes #435:

### Implementation

This work is mostly based on the [`RemoveKnownDefinitions tree rewriter`](https://github.com/Shopify/rbi/pull/100):

1. We parse the gem RBIs and index their content
2. We parse the DSL RBIs and index their content
3. We parse the shim RBIs then for each definition in the shims
   * If the definition is present in the index we remove it from the tree
   * Then, if a scope is already present in the index and is empty in the shim, we remove it from the tree
   * Then, if a file is empty we delete it

Some considerations about what we remove and what we keep:
* We preserve methods that have the same names but different parameters like `def foo; end` vs. `def foo(x); end`. In those cases, there is most likely something the developer wanted that was not generated properly in the RBI generated (and it might have been set as `typed: false`).
* We preserve nodes that have a signature in the shims but not in the generated RBIs (or a different signature in the generated RBIs).
* We also preserve different attribute accessors such as `attr_reader :foo` vs. `attr_accessor :foo, :bar`.
* We discard nodes that have a different comment between the generated RBIs and the shim as to be sure to remove nodes like this one:

   ```
   # TODO remove this once Tapioca is fixed
   class SomethingThatWasMissingInTheGeneratedRBIs; end
   ```

Please note that for project with _a lot_ of gem/dsl RBI files this command might take some time to complete since we would need to parse a lot of content.

### Example

Be the following project:

```rb
# sorbet/rbi/gems/foo@1.0.0.rbi
class Foo
  attr_reader :foo
end

# sorbet/rbi/gems/bar@2.0.0.rbi
module Bar
  def bar; end
end

# sorbet/rbi/shims/foo.rbi
class Foo
  def foo; end
end

# sorbet/rbi/shims/bar.rbi
module Bar
  def foo; end
  def bar; end
end
```

And run `tapioca clean-shims`:

```
$ tapioca clean-shims

Loading gem RBIs from sorbet/rbi/gems...  Done
Loading dsl RBIs from sorbet/rbi/dsl...  Done
Cleaning shim RBIs from sorbet/rbi/shims...
  Deleted ::Bar#bar() at sorbet/rbi/shims/bar.rbi:3:2-3:14 (duplicate from sorbet/rbi/gems/bar@2.0.0.rbi:2:2-2:14)
  Deleted ::Foo#foo() at sorbet/rbi/shims/foo.rbi:2:2-2:18 (duplicate from sorbet/rbi/gems/foo@1.0.0.rbi:2:2-2:18)
  Deleted ::Foo at sorbet/rbi/shims/foo.rbi:1:0-3:3 (duplicate from sorbet/rbi/gems/foo@1.0.0.rbi:1:0-3:3)
  Deleted empty file sorbet/rbi/shims/foo.rbi
Done
```

This will have deleted the `foo.rbi` shim and cleaned the `bar.rbi` one:

```rbi
# sorbet/rbi/shims/bar.rbi
module Bar
  def foo; end
end
```

### Tests

See automate tests.

### DIFF
--- a/spec/tapioca/cli/clean_shims_spec.rb
+++ b/spec/tapioca/cli/clean_shims_spec.rb
@@ -1,0 +1,266 @@
+# typed: true
+# frozen_string_literal: true
+
+require "spec_with_project"
+
+module Tapioca
+  class CleanShimsTest < SpecWithProject
+    describe "tapioca clean-shims" do
+      before(:all) do
+        @project.bundle_install
+      end
+
+      after do
+        @project.remove("sorbet/rbi")
+      end
+
+      it "does nothing when there is no duplicate to clean" do
+        @project.write("sorbet/rbi/gems/foo@1.0.0.rbi", <<~RBI)
+          class Foo
+            attr_reader :foo
+          end
+        RBI
+
+        @project.write("sorbet/rbi/shims/foo.rbi", <<~RBI)
+          class Foo
+            attr_reader :bar
+          end
+        RBI
+
+        out, _, status = @project.tapioca("clean-shims")
+
+        assert_equal(<<~OUT, out)
+          Loading gem RBIs from sorbet/rbi/gems...  Done
+          Loading dsl RBIs from sorbet/rbi/dsl...  Done
+          Cleaning shim RBIs from sorbet/rbi/shims... Done (nothing to do)
+        OUT
+
+        assert(status)
+
+        assert_project_file_equal("sorbet/rbi/shims/foo.rbi", <<~RBI)
+          class Foo
+            attr_reader :bar
+          end
+        RBI
+      end
+
+      it "cleans unecessary shims" do
+        @project.write("sorbet/rbi/gems/foo@1.0.0.rbi", <<~RBI)
+          class Foo
+            attr_reader :foo
+          end
+        RBI
+
+        @project.write("sorbet/rbi/gems/bar@2.0.0.rbi", <<~RBI)
+          module Bar
+            def bar; end
+          end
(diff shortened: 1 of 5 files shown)


spec/tapioca/cli/clean_shims_spec.rb
```diff
@@ -0,0 +1,261 @@
+# typed: true
+# frozen_string_literal: true
+
+require "spec_with_project"
+
+module Tapioca
+  class CleanShimsTest < SpecWithProject
```
vinistock, line 7: Can we add an example for a shim that only gets partially deleted? Something like
```ruby
# shim
# Both methods are generated by Tapioca, but because one has a signature, we only deleted `baz`
class Foo
  sig { params(arg: Integer).void }
  def bar(arg); end

  def baz; end
end

# generated
class Foo
  def bar(arg); end
  def baz; end
end

# Which should then produce the following shim after cleanup
class Foo
  sig { params(arg: Integer).void }
  def bar(arg); end
end
```

vinistock (approved):

paracycle (approved):
This looks great! I am not in love with the `clean-shims` command name, but LGTM and maybe we can find a better name before release.

PR #647, opened 2021-12-03
Update gem RBIs for `tapioca`

mutecipher:
### Motivation

Update gem RBIs for `tapioca`

### Implementation

Ran `bundle exec tapioca gem --all --doc`

### Tests

See automated tests.


Morriar:
It seems we have some type errors to solve

PR #650, opened 2021-12-06
Bump `google-protobuf` gem for `dev` compatibility

KaanOzkan:
### Motivation

`dev up` fails on tapioca because current version of `google-protobuf` gem doesn't support Apple Silicon

### Implementation

`bundle update google-protobuf`
`bundle exec tapioca gem google-protobuf`


Morriar (approved):

spec/tapioca/compilers/dsl/protobuf_spec.rb
```diff
@@ -32,10 +32,7 @@ class Tapioca::Compilers::Dsl::ProtobufSpec < DslSpec
         Cart = Google::Protobuf::DescriptorPool.generated_pool.lookup("MyCart").msgclass
       RUBY
 
-      assert_equal(
-        ["Cart", "Google::Protobuf::Map", "Google::Protobuf::RepeatedField"],
```
KaanOzkan, line 36: I don't love losing the check for these but 1. I didn't want to add a minimum version dependency for `google-protobuf` for users and 2. Didn't want to list all excludable constants apart from `Map` and `RepeatedField`. However, that's also a possibility.

paracycle: You can always add:
```ruby
assert_includes(gathered_constants, "Google::Protobuf::Map")
assert_includes(gathered_constants, "Google::Protobuf::RepeatedField")
``` 
as additional checks.

paracycle (approved):

spec/tapioca/compilers/dsl/protobuf_spec.rb
```diff
@@ -15,7 +15,7 @@ class Tapioca::Compilers::Dsl::ProtobufSpec < DslSpec
         end
       RUBY
 
-      assert_equal([], gathered_constants)
+      assert_equal([], gathered_constants.reject { |constant| constant.start_with?("Google::Protobuf") })
```
Morriar, line 18: I think we can simplify this with something like

```suggestion
      assert(gathered_constants.all? { |constant| constant.start_with?("Google::Protobuf") })
```

PR #654, opened 2021-12-06
Rename and group tracker classes

paracycle:
### Motivation

Now that we have 3 tracker classes, it makes sense to group and refactor them into a single namespace.

### Implementation

File moves, renames and namespace addition mostly.

### Tests

No additional tests.


KaanOzkan (approved):

lib/tapioca/trackers/constant_tracker.rb
```diff
@@ -0,0 +1,42 @@
...
+
+require "set"
+
+module Tapioca
+  module Trackers
+    # Registers a TracePoint immediately upon load to track points at which
+    # classes and modules are opened for definition. This is used to track
+    # correspondence between classes/modules and files, as this information isn't
+    # available in the ruby runtime without extra accounting.
+    module ConstantTracker
```
Morriar, line 12: `Tapioca::Trackers::ConstantTracker` => `Tracker` seems redundant in this name and the repetition does not bring much value. Will it bring more context to call the module `Tapioca::Trackers::ConstantOrigin`?

And same for the other trackers?

paracycle: Great points! I've done the following renames:
- `Tapioca::Trackers::AutoloadTracker` => `Tapioca::Trackers::Autoload`
- `Tapioca::Trackers::MixinTracker` => `Tapioca::Trackers::Mixin`
- `Tapioca::Trackers::ConstantTracker` => `Tapioca::Trackers::ConstantDefinition` (since this one tracks definition (or redefinition) of constants)

lib/tapioca/trackers.rb
```diff
@@ -0,0 +1,6 @@
+# typed: strict
+# frozen_string_literal: true
+
+require "tapioca/trackers/autoload_tracker"
+require "tapioca/trackers/constant_tracker"
+require "tapioca/trackers/mixin_tracker"
```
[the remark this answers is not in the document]
paracycle: Great catch. I've done the proper ordering, reverted the test change that should have told me this was happening and added a long comment explaining why mixin tracker should be loaded before any other trackers. Thank you!

lib/tapioca/trackers/mixin_type.rb
```diff
@@ -0,0 +1,14 @@
+# typed: strict
+# frozen_string_literal: true
+
+module Tapioca
+  module Trackers
+    class MixinType < T::Enum
```
Morriar, line 6: Could we nest this under `Trackers::Mixin::Type`?

Morriar (approved):

PR #659, opened 2021-12-07
Refactor config infrastructure to be integrated with Thor

### LINKED ISSUE #490: Consider standardizing gem/dsl configuration
It seems like tapioca has two main "modes": `gem` and `dsl`. The configuration for gem happens at the top level (e.g. `exclude` CLI arg or in config.yml) whereas the config for dsl happens more verbosely (`exclude-generators` CLI arg or in config.yml). Also, some configs seem to only apply to `gem` (e.g. prerequire, postrequire, `sorbet/tapioca/require.rb`: https://github.com/Shopify/tapioca/blob/b8e60c65b5c2dbbc75654564b9975120d7de5d82/lib/tapioca/generator.rb#L232)

I know this is due to how tapioca has organically grown but before the 1.0.0 it might be worth reconsidering. Thinking about it from first principles, I think a config.yml like this might make sense:
```yml
gem:
  exclude:
    - some_gem
    - some_other_gem
  prerequire: path/to/some_file.rb
dsl:
  exclude:
    - SomeGenerator
    - SomeOtherGenerator
```

Then maybe you can turn the single [Config object](https://github.com/Shopify/tapioca/blob/f5380aa0b566eed041b94d119e696eb91a57ca76/lib/tapioca/config.rb) into two: one for gem and one for dsl. This would help enforce that you can't pass a gem-only config into dsl.

The CLI could actually be simpler, since you wouldn't need `exclude` and `exclude-generators`. `exclude` would just mean different things in each context.

### `require.rb`

I was also thinking about this message about writing custom DSL generators: https://github.com/Shopify/tapioca/issues/459#issuecomment-906596036. What about moving `tapioca/require.rb` to `tapioca/gem/require.rb` and having an analogous `tapioca/dsl/require.rb` and if you wanted to load any additional DSL generators you'd do so there? Personally I think that having actual code live in the `sorbet/` folder is a little weird so we have our custom DSL generators similar to how that comment suggests: they live in `lib` and then simply require them in a file called `tapioca/dsl/require.rb`.

paracycle:
### Motivation

Fixes #490 

Our existing config implementation was hacky and [error-prone](#566), and we also wanted it to support options for more than one command (since it was originally designed at a time when `gem` was the only real command).

### Implementation

Instead of relying on a strongly typed struct class like `Config` to read the supplied options, we decided to instead work with Thor and not against it. We've accepted that getting strong types for options will be hard, which has allowed us to hook into the `options` call on Thor, in order to perform the `default` -> `config` -> `command line` supplied options merge, in that order.

The major problem to solve was how Thor works with default values supplied to options. Thor sets an option value to the default value if the user has not supplied the option on the command line. This prevents the application from being able to tell apart the cases when the option was supplied by the user but it was supplied as the default value, or if the option was not supplied at all, and it was set to the default value.

We need to be able to differentiate between the two cases, so that we can override default, non-supplied values by the values supplied in a config file.

To be able to pull this off, we intercept the `initialize` method of the `Thor` subclass and filter out the default values supplied to options. We store those default values in a separate hash, so that when the application asks for the options for the command, we can do the merging of the 3 levels of values in the correct way and resolve the option to the right value.

I've also:
- Moved some class level config options down to the corresponding commands.
- Changed some of the option flag names that didn't make much sense, and made them more uniform.
- Added a section in the `README` about how to use the configuration file.
- Updated the help output for various commands that we have documented.

### Tests

No additional tests

Co-authored-by: Alexandre Terrasa <583144+Morriar@users.noreply.github.com>


README.md
```diff
@@ -55,9 +55,10 @@ end
 and do not forget to execute `tapioca` using `bundler`:
 
 ```shell
-$ bundle exec tapioca
+$ bundle exec tapioca help
```
Morriar, line 58: Note for later: find a way to check/automate this

README.md
```diff
@@ -156,26 +164,47 @@ Usage:
...
+                                                 # Default: 1
+  -c, [--config=<config file path>]              # Path to the Tapioca configuration file
+                                                 # Default: sorbet/tapioca/config.yml
+  -V, [--verbose], [--no-verbose]                # Verbose output for debugging purposes
 
 generate RBIs for dynamic methods
 ```
+## Configuration
+
+Tapioca has support for loading command defaults from a configuration file. The default configuration
```
Morriar, line 185:
```suggestion
Tapioca supports loading command defaults from a configuration file. The default configuration
```

README.md
```diff
@@ -156,26 +164,47 @@ Usage:
...
-- `--generate-command [command]`: **[DEPRECATED]** The command to run to regenerate RBI files (used in header comment of the RBI files), defaults to the current command.
-- `--typed-overrides [gem:level]`: Overrides typed sigils for generated gem RBIs for gem `gem` to level `level` (`level` can be one of `ignore`, `false`, `true`, `strict`, or `strong`, see [the Sorbet docs](https://sorbet.org/docs/static#file-level-granularity-strictness-levels) for more details).
+For example, if you always want to generate gem RBIs with inline documentation, then you would create the file `sorbet/tapioca/config.yml` as:
+
+```yaml
+gem:
+  docs: true
+```
+
+Additionally, if you always wanted to exclude the `AASM` and `ActiveRecordFixtures` DSL compilers in your DSL RBI generation runs, your config file would look like this:
```
Morriar, line 198:
```suggestion
Additionally, if you always want to exclude the `AASM` and `ActiveRecordFixtures` DSL compilers in your DSL RBI generation runs, your config file would then look like this:
```

lib/tapioca.rb
```diff
@@ -15,6 +22,25 @@ def self.silence_warnings(&blk)
...
   class Error < StandardError; end
+
+  SORBET_PATH = T.let("sorbet", String)
+  SORBET_CONFIG = T.let("#{SORBET_PATH}/config", String)
+  TAPIOCA_PATH = T.let("#{SORBET_PATH}/tapioca", String)
+  TAPIOCA_CONFIG = T.let("#{TAPIOCA_PATH}/config.yml", String)
+
+  DEFAULT_COMMAND = T.let("bin/tapioca", String)
+  DEFAULT_POSTREQUIRE = T.let("#{TAPIOCA_PATH}/require.rb", String)
+  DEFAULT_RBIDIR = T.let("#{SORBET_PATH}/rbi", String)
```
Morriar, line 33: We use `SORBET_PATH` a few line before. Should we rename it to `SORBET_DIR` or rename this one `DEFAULT_RBI_PATH`?

Also, let's add an underscore:
```suggestion
  DEFAULT_RBI_DIR = T.let("#{SORBET_PATH}/rbi", String)
```

lib/tapioca.rb
```diff
@@ -15,6 +22,25 @@ def self.silence_warnings(&blk)
...
+  DEFAULT_DSLDIR = T.let("#{DEFAULT_RBIDIR}/dsl", String)
```
Morriar, line 34:
```suggestion
  DEFAULT_DSL_DIR = T.let("#{DEFAULT_RBI_DIR}/dsl", String)
```

lib/tapioca.rb
```diff
@@ -15,6 +22,25 @@ def self.silence_warnings(&blk)
...
+  DEFAULT_GEMDIR = T.let("#{DEFAULT_RBIDIR}/gems", String)
```
Morriar, line 35:
```suggestion
  DEFAULT_GEM_DIR = T.let("#{DEFAULT_RBI_DIR}/gems", String)
```

lib/tapioca.rb
```diff
@@ -15,6 +22,25 @@ def self.silence_warnings(&blk)
...
+  DEFAULT_SHIMDIR = T.let("#{DEFAULT_RBIDIR}/shims", String)
```
Morriar, line 36:
```suggestion
  DEFAULT_SHIM_DIR = T.let("#{DEFAULT_RBI_DIR}/shims", String)
```

lib/tapioca.rb
```diff
@@ -15,6 +22,25 @@ def self.silence_warnings(&blk)
...
+  DEFAULT_TODOSPATH = T.let("#{DEFAULT_RBIDIR}/todo.rbi", String)
```
Morriar, line 37: Again, `PATH` vs `DIR`.

```suggestion
  DEFAULT_TODOS_PATH = T.let("#{DEFAULT_RBI_DIR}/todo.rbi", String)
```

Should we use `PATH` (or maybe `FILE`?) for a file and `DIR` for a directory?

lib/tapioca.rb
```diff
@@ -15,6 +22,25 @@ def self.silence_warnings(&blk)
...
 
   class Error < StandardError; end
+
+  SORBET_PATH = T.let("sorbet", String)
+  SORBET_CONFIG = T.let("#{SORBET_PATH}/config", String)
+  TAPIOCA_PATH = T.let("#{SORBET_PATH}/tapioca", String)
+  TAPIOCA_CONFIG = T.let("#{TAPIOCA_PATH}/config.yml", String)
+
+  DEFAULT_COMMAND = T.let("bin/tapioca", String)
+  DEFAULT_POSTREQUIRE = T.let("#{TAPIOCA_PATH}/require.rb", String)
```
Morriar, line 32: Maybe we should rename it:
```suggestion
  DEFAULT_POSTREQUIRE_PATH = T.let("#{TAPIOCA_PATH}/require.rb", String)
```

Or:
```suggestion
  DEFAULT_POSTREQUIRE_FILE = T.let("#{TAPIOCA_PATH}/require.rb", String)
```

lib/tapioca/cli.rb
```diff
@@ -4,104 +4,113 @@
...
-        default_command: Config::DEFAULT_COMMAND
+        sorbet_config: SORBET_CONFIG,
+        default_postrequire: DEFAULT_POSTREQUIRE,
+        default_command: DEFAULT_COMMAND
       )
       generator.generate
     end
 
     desc "require", "generate the list of files to be required by tapioca"
+    option :postrequire, type: :string, default: DEFAULT_POSTREQUIRE
```
Morriar, line 32: Depending on the answer to the previous questions:

```suggestion
    option :postrequire_path, type: :string, default: DEFAULT_POSTREQUIRE_PATH
```

lib/tapioca/cli.rb
```diff
@@ -4,104 +4,113 @@
...
+        sorbet_config_path: SORBET_CONFIG,
+        default_command: DEFAULT_COMMAND
       )
       Tapioca.silence_warnings do
         generator.generate
       end
     end
 
     desc "todo", "generate the list of unresolved constants"
+    option :todos_path,
```
Morriar, line 45: Or

```suggestion
    option :todos_file,
```

lib/tapioca/cli.rb
```diff
@@ -4,104 +4,113 @@
...
+      type: :string,
+      default: DEFAULT_TODOSPATH
+    option :file_header,
+      type: :boolean,
+      desc: "Add a \"This file is generated\" header on top of each generated RBI file",
```
Morriar, line 50: Should we extract this as a constant?

lib/tapioca/cli.rb
```diff
@@ -4,104 +4,113 @@
...
         generator.generate
       end
     end
 
     desc "dsl [constant...]", "generate RBIs for dynamic methods"
-    option :generators,
+    option :outdir,
+      aliases: ["--out", "-o"],
+      banner: "directory",
+      desc: "The output directory for generated RBI files",
```
Morriar, line 67:
```suggestion
      desc: "The output directory for generated DSL RBI files",
```

lib/tapioca/cli.rb
```diff
@@ -114,60 +123,75 @@ def dsl(*constants)
     end
 
     desc "gem [gem...]", "generate RBIs from gems"
+    option :outdir,
+      aliases: ["--out", "-o"],
+      banner: "directory",
+      desc: "The output directory for generated RBI files",
```
Morriar, line 129:
```suggestion
      desc: "The output directory for generated gem RBI files",
```

lib/tapioca/cli.rb
```diff
@@ -114,60 +123,75 @@ def dsl(*constants)
...
       banner: "file",
-      desc: "A file to be required after Bundler.require is called"
+      desc: "A file to be required after Bundler.require is called",
+      default: DEFAULT_POSTREQUIRE
     option :exclude,
       aliases: ["-x"],
       type: :array,
       banner: "gem [gem ...]",
-      desc: "Excludes the given gem(s) from RBI generation"
+      desc: "Excludes the given gem(s) from RBI generation",
```
Morriar, line 153: Let's be consistent with the options for `dsl`
```suggestion
      desc: "Exclude the given gem(s) from RBI generation",
```

lib/tapioca/cli.rb
```diff
@@ -4,104 +4,113 @@
...
+      type: :boolean,
+      desc: "Add a \"This file is generated\" header on top of each generated RBI file",
+      default: true
+    option :only,
       type: :array,
-      aliases: ["--gen", "-g"],
       banner: "generator [generator ...]",
-      desc: "Only run supplied DSL generators"
-    option :exclude_generators,
+      desc: "Only run supplied DSL generators",
```
Morriar, line 76: Let's be consistent with the options for `gems`

```suggestion
      desc: "Only run supplied DSL generator(s)",
```

lib/tapioca/cli.rb
```diff
@@ -4,104 +4,113 @@
...
+      default: []
+    option :exclude,
       type: :array,
       banner: "generator [generator ...]",
-      desc: "Exclude supplied DSL generators"
+      desc: "Exclude supplied DSL generators",
```
Morriar, line 81:
```suggestion
      desc: "Exclude supplied DSL generator(s)",
```

lib/tapioca/cli.rb
```diff
@@ -114,60 +123,75 @@ def dsl(*constants)
...
       banner: "gem [gem ...]",
-      desc: "Excludes the given gem(s) from RBI generation"
+      desc: "Excludes the given gem(s) from RBI generation",
+      default: []
     option :typed_overrides,
       aliases: ["--typed", "-t"],
       type: :hash,
       banner: "gem:level [gem:level ...]",
-      desc: "Overrides for typed sigils for generated gem RBIs"
+      desc: "Overrides for typed sigils for generated gem RBIs",
```
Morriar, line 159:
```suggestion
      desc: "Override for typed sigils for generated gem RBIs",
```

lib/tapioca/cli.rb
```diff
@@ -114,60 +123,75 @@ def dsl(*constants)
...
+      default: DEFAULT_OVERRIDES
     option :verify,
       type: :boolean,
-      default: false,
-      desc: "Verifies RBIs are up-to-date"
+      desc: "Verifies RBIs are up-to-date",
```
Morriar, line 163:
```suggestion
      desc: "Verify RBIs are up-to-date",
```

Morriar, line 163:
```suggestion
      desc: "Verify RBIs are up-to-date",
```

lib/tapioca/helpers/config_helper.rb
```diff
@@ -0,0 +1,85 @@
...
+      @command_name = T.let(command.name, String)
+      @merged_options = T.let(nil, T.nilable(Thor::CoreExt::HashWithIndifferentAccess))
+      @defaults = T.let(nil, T.nilable(Thor::CoreExt::HashWithIndifferentAccess))
+
+      # Filter command options unless we are handling the help command.
+      # This is so that the defaults are printed
+      filter_defaults(command_options) unless command_name == "help"
+
+      # Call original initialize method
+      super
```
Morriar, line 27: Dr Obvious is obvious 😂 

```suggestion
      super
```

lib/tapioca/helpers/config_helper.rb
```diff
@@ -0,0 +1,85 @@
...
+    attr_reader :command_name
+
+    sig { params(args: T.untyped, local_options: T.untyped, config: T.untyped).void }
+    def initialize(args = [], local_options = {}, config = {})
+      # Store current command
+      command = config[:current_command]
+      command_options = config[:command_options]
+      @command_name = T.let(command.name, String)
+      @merged_options = T.let(nil, T.nilable(Thor::CoreExt::HashWithIndifferentAccess))
+      @defaults = T.let(nil, T.nilable(Thor::CoreExt::HashWithIndifferentAccess))
```
Morriar, line 20: Can't we initialize those with a `Thor::CoreExt::HashWithIndifferentAccess.new` directly? Why do we need a `nil`?

lib/tapioca/helpers/config_helper.rb
```diff
@@ -0,0 +1,85 @@
...
+
+      if File.exist?(config_file)
+        config = YAML.load_file(config_file, fallback: {})
+      end
+
+      Thor::CoreExt::HashWithIndifferentAccess.new(config[command_name] || {})
+    end
+
+    sig do
+      params(options: T.nilable(Thor::CoreExt::HashWithIndifferentAccess))
```
Morriar, line 70: Is it really supposed to be `nilable`? It's already a variadic: `*options`

```suggestion
      params(options: T.nilable(Thor::CoreExt::HashWithIndifferentAccess))
```

vinistock (approved):

lib/tapioca/cli.rb
paracycle: I chose to not change this, since the option name is long and I feel like we don't have much to gain by adding `_file` or `_path` at the end.

lib/tapioca/helpers/config_helper.rb
paracycle: Yes, because someone can pass `nil` as an argument: `merge_options(nil, {}, {foo: 1})`

lib/tapioca/helpers/config_helper.rb
paracycle: I did that for `@defaults` but `@merged_options` is being used for memoization, so I didn't change that.

PR #660, opened 2021-12-09
Allow DSL generators to change behaviour based on if `ActiveRecordRelations` is enabled or not

paracycle:
### Motivation

If we release Tapioca in its current state, we will have to deal with 1000s of type errors, all related to the new `ActiveRecordRelations` change, before we can bump the version in Core. In order to decouple the Tapioca version bump from clearing all those errors, the best way forward is to be able to disable the `ActiveRecordRelation` generator.

However, the existence of that generator has changed the behaviour of other generators as well.  Primarily, whereas `ActiveRecordAssociations` generator was generating collection association method returns as the fake generic `CollectionProxy`, it is now doing it using the internal `PrivateCollectionProxy`. Similarly, whereas `ActiveRecordScope` generator was generating `T.untyped` as the return value of scope method, it is not generating them as `PrivateRelation` or `PrivateAssociationRelation` instead.

So that behaviour of those generators have to change depending on if `ActiveRecordRelation` has been enabled or not.

### Implementation

The implementation is best followed by reading the commits one by one, but in a nutshell:

- We now pass in the `DslCompiler` instance to each generator, so that a generator can make queries about the context it is running in.
- That allows us to add a method to query if a generator with a given name is enabled or not.
- Using that power, we can change the behaviour of `ActiveRecordAssociations` and `ActiveRecordScope` accordingly.

We also had to update how DSL generator specs work internally to be able to test a similar change in behaviour as well.

### Tests

Updated tests to exercise both relations enabled and disabled cases. Also added a few new test cases that exercise code paths that were never exercised before.

### DIFF
--- a/lib/tapioca/compilers/dsl_compiler.rb
+++ b/lib/tapioca/compilers/dsl_compiler.rb
@@ -22,7 +22,7 @@ module Tapioca
           requested_constants: T::Array[Module],
           requested_generators: T::Array[T.class_of(Dsl::Base)],
           excluded_generators: T::Array[T.class_of(Dsl::Base)],
-          error_handler: T.nilable(T.proc.params(error: String).void),
+          error_handler: T.proc.params(error: String).void,
           number_of_workers: T.nilable(Integer),
         ).void
       end
@@ -30,7 +30,7 @@ module Tapioca
         requested_constants:,
         requested_generators: [],
         excluded_generators: [],
-        error_handler: nil,
+        error_handler: $stderr.method(:puts).to_proc,
         number_of_workers: nil
       )
         @generators = T.let(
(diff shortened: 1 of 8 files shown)


lib/tapioca/compilers/dsl_compiler.rb
```diff
@@ -76,6 +76,19 @@ def run(&blk)
...
 
+      sig { params(generator_name: String).returns(T::Boolean) }
+      def generator_enabled?(generator_name)
+        # Try to find built-in tapioca generator first, then globally defined generator. The
+        # explicit `break` ensures the class is returned, not the `potential_name`.
+        generator_klass = ["Tapioca::Compilers::Dsl::#{generator_name}", generator_name].find do |potential_name|
+          break Object.const_get(potential_name)
+        rescue NameError
+          # Skip if we can't find generator by the potential name
+        end
```
Morriar, line 87: Should we extract this so it can be share with `constantize_generators`? https://github.com/Shopify/tapioca/blob/d63482a222f0cd5af4e8a765bcc6b9bfa801ec1a/lib/tapioca/generators/dsl.rb#L191-L197

Morriar (approved):

vinistock (approved):

lib/tapioca/compilers/dsl_compiler.rb
paracycle: Done!

Issue #610, opened 2021-11-15
Need a better way to handle required belongs_to associations.


thomasdziedzic-census:
slack context: https://sorbet-ruby.slack.com/archives/CHN2L03NH/p1636647565028400

I'm trying to migrate our code base from using sorbet & sorbet-rails to using tapioca for all rbi generation.

Currently tapioca always generates T.nilable belongs_to associations where as previously in our code base, our rbis generated via sorbet-rails were not nilable. This has caused a lot of additional T.must's in our code base (1000s).

This created some friction so I forked the [active_record_associations generator](https://github.com/Shopify/tapioca/blob/v0.5.4/lib/tapioca/compilers/dsl/active_record_associations.rb) and added code to it to only generate nilable code if `optional: true`.

It seems like others have encountered this friction as well for instance https://github.com/Shopify/tapioca/issues/174

While digging into what sorbet-rails does, I found that it has a couple of bugs open for non-nilable belongs_to associations carrying nil values: https://github.com/chanzuckerberg/sorbet-rails/issues/253 & https://github.com/chanzuckerberg/sorbet-rails/issues/356

So it seems like tapioca might be too strict, and sorbet-rails might be too open :) Is there a middle road here?

drop-george:
This is one of the biggest hurdles I'm seeing right now in moving from `srb rbi` to `tapioca`.

rafaelfranca:
The types generated by sorbet is correct. All `belongs_to` associations can have `nil` value.

Given:

```ruby
class MyModel < ApplicationRecord
  belongs_to :foo
end
```

This would be a valid runtime behavior in any Rails application

```ruby
MyModel.new.foo # => nil
```

That is an invalid model, but for the purpose of type checking, at certain points of a lifetime of a program, all `belongs_to` associations can be `nil`, so the correct signature for those methods are really `T.nilable`. The fact that sorbet-rails already generated wrong types is a bummer.

rafaelfranca:
I don't think there is a middle road here. We can't teach Sorbet about conditional runtime behavior (that is what required belongs_to associations are).

The opposite of this is to generate non-nilable types, but have programs failing at runtime because the type checker didn't caught the cases were the association retuned `nil`, which defeats the reason why someone would use a type checker.

Our choice is to require some friction, in order to achieve type safety.

philsmy:
[...]

PR #664, opened 2021-12-13
Generate strong ActiveRecord assocations for StrongAssociationGeneration descendants

jflinter:
### Motivation
https://github.com/Shopify/tapioca/issues/610

(This is arguably the biggest blocker in my org's sorbet migration - the hundreds of `T.must`s that would need to be added absent a change like this are annoying and result in less readable code.)

I'm happy to make any directional changes here but wanted to get the ball rolling!

### Implementation
This follows the approach laid out in https://github.com/jflinter/tapioca/blob/2ca267efa51e7cb5863eb010142c2185614c3a76/lib/tapioca/helpers/active_record_column_type_helper.rb#L61 (although it inverts the logic somewhat).

Notably this doesn't do anything towards actually providing a runtime implementation of `StrongAssociationGeneration` - while I think that's is the right direction, it feels beyond the scope of this change.

### Tests
Updated tests. Notably, the added reference to `column_hash` in `ActiveRecordAssociations` means that tests that didn't define a schema.rb would break. Adding those caused a few types to improve along the way.


paracycle:
@jflinter I think this is not the way we want to go, tbh. The reason why we had a test for `StrongTypeGeneration` in `ActiveRecordColumns` was because we actually had module with that exact name that would hook into `before_initialize` to make sure that you couldn't create a model instance without initializing all of its non-nilable attributes. Once we were able to enforce, at runtime, that we couldn't create instances of models with non-nil attributes, we were also able to generate stronger types for those attributes.

The situation is not exactly the same for associations though. How could a theoretical `StringAssocationGenerator` module be able to enforce that associations are indeed not-nil at runtime?

Our team does realize that the defaults we have chosen, while type-correct, might not be the defaults that the community wants. For that reason, I think we would instead like to start adding a way to pass configuration options to DSL compilers, so that they can exhibit different behaviours that are controllable by the end user. How does that sound to you?

jflinter:
Yeah, I agree with the idea of making this a configuration option instead. The way I look at it is that a `StrongAssociationGeneration` module or whatever _is_ a form of configuration, just a configuration that's defined in an unconventional way. For us, our code already makes the assumption that associations with required columns will always be present, so extending that module would simply be a way of opting modules into the desired behavior.

Is there an open issue for generator configuration, or a spec that's been discussed? I'd love to try to get it moving if it wouldn't be too disruptive.

PR #668, opened 2021-12-17
Remove mixin filtering

paracycle:
### Motivation

It seems mixin filtering is not mature enough yet, and the way we are filtering mixins will end up resulting in RBIs that have some critical mixins missing completely. So in order to not block the 0.6.0 release, I've decided to revert the mixin filtering instead and to focus on implementing it more completely after 0.6.0.

### Implementation

Partial revert of #569

### Tests

Reverting the test changes in #569


RyanBrushett (approved):

Morriar (approved):
😢

paracycle:
> 😢

Not to worry, I have it working but it will have to wait for 0.6.1 😉

PR #672, opened 2021-12-20
CollectionProxy is generic, requires type argument

ryanwilsonperkin:
### Motivation
Fix issue with the generated types introduced in v0.6.0 for polymorphic has_many relations

While integrating this new release into a service, we noticed the following error message:

```
Malformed type declaration. Generic class without type arguments ActiveRecord::Associations::CollectionProxy
```

This was happening because we are opting in to the new relations typing, and we had a polymorphic has_many relationship which triggered this particular condition.

### Implementation
`CollectionProxy` is generic so it requires an argument to be passed. I updated the test to reflect this and then the code to match.

Updates the previous implementation from https://github.com/Shopify/tapioca/commit/217a1d025e351378180478ec3c25ad45c2a1e7c7

### Tests
Yep! There was an existing test but it asserted on incorrect behaviour, and has now been fixed.

### DIFF
--- a/lib/tapioca/compilers/dsl/active_record_associations.rb
+++ b/lib/tapioca/compilers/dsl/active_record_associations.rb
@@ -335,7 +335,7 @@ module Tapioca
 
           if relations_enabled
             if polymorphic_association
-              "ActiveRecord::Associations::CollectionProxy"
+              "ActiveRecord::Associations::CollectionProxy[T.untyped]"
             else
               "#{qualified_name_of(reflection.klass)}::#{AssociationsCollectionProxyClassName}"
             end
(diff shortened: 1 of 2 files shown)


paracycle:
I don't think this is correct. If AR relations are enabled then `CollectionProxy` shouldn't be generic anyway. That was a hack that we had in place which we are now getting rid of.

ryanwilsonperkin:
Very interesting, that doesn't seem to be the case in our service at the moment, it reports the above error. Here's the gem versions we're using:

- activerecord: 6.1.4.1
- tapioca: 0.6.0
- shopify-types 0.5.6
- sorbet 0.5.9426

paracycle:
> Edit: Scratch that, it appears we'd hardcoded our own shim that was setting it generic. I'll remove that now

Yes. That is something that should be removed, indeed. Sorry for the confusion.

PR #679, opened 2021-12-26
Start testing against Ruby 3.1

paracycle:
### Motivation

Now that Ruby 3.1 is released, we should start targeting it.

### Implementation

Add Ruby 3.1 to test matrix and fix problems.

### Tests

No added tests.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -36,4 +36,9 @@ group(:development, :test) do
   gem("aasm", require: false)
   gem("bcrypt", require: false)
   gem("xpath", require: false)
+
+  # net-smtp was removed from default gems in Ruby 3.1, but is used by the `mail` gem.
+  # So we need to add it as a dependency until `mail` is fixed:
+  # https://github.com/rails/rails/blob/0919aa97260ab8240150278d3b07a1547489e3fd/Gemfile#L178-L191
+  gem("net-smtp", "0.3.1", require: false)
 end
--- a/lib/tapioca/gemfile.rb
+++ b/lib/tapioca/gemfile.rb
@@ -99,6 +99,9 @@ module Tapioca
       sig { returns(String) }
       attr_reader :full_gem_path, :version
 
+      sig { returns(T::Array[Pathname]) }
+      attr_reader :files
+
       sig { params(spec: Spec).void }
       def initialize(spec)
         @spec = T.let(spec, Tapioca::Gemfile::Spec)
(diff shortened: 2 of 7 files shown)


spec/tapioca/cli/gem_spec.rb
```diff
@@ -831,74 +831,6 @@ module Foo
           assert_empty(err)
           assert(status)
         end
-
-        it "generates abstract classes properly" do
```
KaanOzkan, line 835: Why was this removed?

paracycle: Because it was the wrong location for the test. We already have the same thing tested in `symbol_compiler_spec`, there was no need to test the same thing in a CLI test, and it was causing some problem that I don't remember now, tbh.

KaanOzkan (approved):

lib/tapioca/gemfile.rb
```diff
@@ -176,14 +165,52 @@ def exported_rbi_tree
...
+
       sig { returns(T::Boolean) }
       def default_gem?
         @spec.respond_to?(:default_gem?) && @spec.default_gem?
       end
 
       sig { returns(Pathname) }
       def ruby_lib_dir
-        Pathname.new(RbConfig::CONFIG["rubylibdir"])
+        RbConfig::CONFIG["rubylibdir"]
```
vinistock, line 190: Is this method used anywhere other than in `resolve_to_ruby_lib_dir`? Could it be a local variable there?

vinistock (approved):

Gemfile
```diff
@@ -36,4 +36,9 @@ group(:development, :test) do
   gem("aasm", require: false)
   gem("bcrypt", require: false)
   gem("xpath", require: false)
+
+  # net-smtp was removed from default gems in Ruby 3.1, but is used by the `mail` gem.
+  # So we need to add it as a dependency until `mail` is fixed:
+  # https://github.com/rails/rails/blob/0919aa97260ab8240150278d3b07a1547489e3fd/Gemfile#L178-L191
+  gem("net-smtp", "0.3.1", require: false)
```
rafaelfranca, line 43: Which version of Rails are you testing against? If it is 7.0 you will not need this in a few minutes.

paracycle: We are still testing against Rails 6.1, yet. We will have to add Rails 7 support soon.

PR #683, opened 2021-12-29
Ignore interpolated strings in RequiresCompiler

brasic:
### Motivation

If a require uses an interpolated argument it should not be included in the compiled file for the following reasons:
 - The output of `collect_requires` is later filtered by `name_in_project?`, which compares require arguments with file paths.  This strategy won't work with interpolated strings.
 - The regular expression isn't smart enough to correctly handle the full semantics of interpolated strings, which allows for things like interior quotes.  That means that an instance of e.g. `require "#{ENV["SOMETHING"]}"` will be transformed into `require "#{ENV["` which is invalid syntax.

### Implementation
I considered altering the regular expression to use a negative lookahead assertion against the sequence `#{` but the regex is already fairly obscure.  It seemed simpler to add a post-processing step to remove any indicators of interpolation.

### Tests
Included


brasic:
rebased onto current main!

paracycle (approved):

PR #687, opened 2022-01-05
Update CI workflow

paracycle:
### Motivation

- Fix Ruby 3.0 being resolved as 3.1 unless quoted: https://bibwild.wordpress.com/2021/12/28/github-action-setup-ruby-needs-to-quote-3-0-or-will-end-up-with-ruby-3-1/
- Remove Bundler version matrix from CI, in favour of a test that runs with Bundler 2.2.22 explicitly.

### Implementation

Read commit by commit

### Tests

One extra test case for Bundler 2.2.22 added.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [ 2.6, 2.7, 3.0 ]
+        ruby: [ "2.6", "2.7", "3.0" ]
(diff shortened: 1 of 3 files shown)


RyanBrushett (commented):
Ahh much cleaner. I dig it.

Is that first commit (https://github.com/Shopify/tapioca/pull/687/commits/eee0d72ab9f58a1f0371a8d2d76bea5d314aebda) still necessary?

RyanBrushett (approved):
Approve w/ commit cleanup. Thanks Ufuk!

PR #697, opened 2022-01-10
Nested generic interfaces with overrides create invalid gem RBI output

jeffcarbs:
### Motivation

We introduced the use of nested generic interfaces in an internal gem which is causing tapioca to produce invalid RBI.

### Implementation

Not sure how to fix, this PR is just adding a reproduction via test case. Open to suggestions or someone just adding a commit here that fixes it.

_Edit by @paracycle_: The fix was to delay the computation of generic type names and to not cache them. The problem with missing type names was because of an early run of the super method signature in `Root#children` when the signature for `OtherRoot#children` was being evaluated. However, at that point, the type member for `Root` has not been bound to its name `"Elem"` yet, so the name that was constructed very early in generic type creation was incomplete.

By delaying the generic type name construction to as late as possible and never caching it, we can allow the `Root#children` method signature to evaluate the name for `Node[Elem]` after the type member has been bound to its name.

### Tests

See included test

### DIFF
--- a/spec/tapioca/compilers/symbol_table_compiler_spec.rb
+++ b/spec/tapioca/compilers/symbol_table_compiler_spec.rb
@@ -2762,7 +2762,7 @@ class Tapioca::Compilers::SymbolTableCompilerSpec < Minitest::HooksSpec
             []
           end
 
-          sig { returns(T::Array[Node[Integer]]) }
+          sig { override.returns(T::Array[Node[Integer]]) }
           def non_abstract_but_overriden_children
             []
           end
@@ -2782,8 +2782,11 @@ class Tapioca::Compilers::SymbolTableCompilerSpec < Minitest::HooksSpec
 
           Elem = type_member(fixed: Integer)
 
-          def children(*args, &blk); end
-          def non_abstract_but_overriden_children(*args, &blk); end
+          sig { override.returns(T::Array[Node[Integer]]) }
+          def children; end
+
+          sig { override.returns(T::Array[Node[Integer]]) }
+          def non_abstract_but_overriden_children; end
         end
 
         module Root
--- a/spec/tapioca/compilers/symbol_table_compiler_spec.rb
+++ b/spec/tapioca/compilers/symbol_table_compiler_spec.rb
@@ -2719,6 +2719,98 @@ class Tapioca::Compilers::SymbolTableCompilerSpec < Minitest::HooksSpec
       assert_equal(output, compile)
     end
 
+    it("compiles nested generic interfaces") do
+      add_ruby_file("generic.rb", <<~RUBY)
+        module Root
+          extend T::Sig
+          extend T::Generic
+          Elem = type_member
+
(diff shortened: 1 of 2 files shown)


spec/tapioca/compilers/symbol_table_compiler_spec.rb
```diff
@@ -2719,6 +2719,98 @@ def inherited(s); end
...
+          extend T::Generic
+
+          interface!
+
+          Elem = type_member
+
+          sig { abstract.returns(T::Array[Node[Elem]]) }
+          def abstract_but_not_overridden_children; end
+
+          sig { abstract.returns(T::Array[Node[Elem]]) }
```
jeffcarbs, line 2800: Tapioca produces incorrect RBI here:
```
sig { abstract.returns(T::Array[Node[]]) }
```

spec/tapioca/compilers/symbol_table_compiler_spec.rb
```diff
@@ -2719,6 +2719,98 @@ def inherited(s); end
...
+          def children; end
+
+          sig { returns(T::Array[Node[Elem]]) }
+          def non_abstract_but_overriden_children; end
```
jeffcarbs, line 2804: Tapioca produces incorrect RBI here:
```
sig { returns(T::Array[Node[]]) }
```

spec/tapioca/compilers/symbol_table_compiler_spec.rb
```diff
@@ -2719,6 +2719,98 @@ def inherited(s); end
...
+
+          interface!
+
+          Elem = type_member
+
+          sig { abstract.returns(T::Array[Node[Elem]]) }
+          def abstract_but_not_overridden_children; end
+
+          sig { abstract.returns(T::Array[Node[Elem]]) }
+          def children; end
```
jeffcarbs, line 2801: Tapioca produces incorrect RBI here:
```
sig { abstract.returns(T::Array[Node[]]) }
```

spec/tapioca/compilers/symbol_table_compiler_spec.rb
```diff
@@ -2719,6 +2719,98 @@ def inherited(s); end
...
+      add_ruby_file("generic.rb", <<~RUBY)
+        module Root
+          extend T::Sig
+          extend T::Generic
+          Elem = type_member
+
+          interface!
+
+          sig { abstract.returns(T::Array[Node[Elem]]) }
+          def children; end
```
jeffcarbs, line 2732: When something `include Root` and redefines this method, tapioca starts producing the following sig:
```
sig { abstract.returns(T::Array[Node[Elem]]) }
```

jeffcarbs:
@paracycle - thanks for the fix! I think I've signed/agreed to the CLA so not sure why that check isn't green but happy to do whatever to unblock this.

paracycle:
Closing in favour of #701.

PR #701, opened 2022-01-11
Improve generic type variable tracking

paracycle:
### Motivation

A test case submitted by @jeffcarbs in #697 surfaced the weakness of how we were doing late name binding for type variables. I was not satisfied with the changes I made on that PR, and wanted to find a better way, which is this PR that improves type variable detection immensely.

### Implementation

There are 2 main changes made in this PR to properly run the added test:

1. We add a `<=` implementation for the generic types that we create to act like they are always covariant or contravariant. Sorbet runtime asserts co/contravariance of types in `in` and `out` positions, and fails with an error if the assertions are not satisfied. Since we are not interested in detecting these case, but more in getting the code to run, we act like everything is fine to pass these checks.
2. One of the problems we've had was how (and when) to figure out that a type variable (for example, constructed by `type_member`) is bound to a specific constant name (for example, to `Elem`). We were resorting to scanning the subconstants of a constant as we were processing it, but that made the process dependent on when we processed a certain constant and would break when there was a type variable access to a constant that was not processed yet.

   However, we know that modules are automatically bound to the name of the constant they are assigned to at the time of assignment. For example:

   ```ruby
   foo = Module.new
   foo.name # => nil
   Elem = foo
   foo.name # => "Elem"
   ```

   So, the realization is that if we make `Tapioca::TypeVariable` instances modules, then we wouldn't have to worry about this since the name binding would be done by Ruby automatically at assignment time.

   All we have to do is to make `Tapioca::TypeVariable` subclass from `Module` so that its instances are modules, and to special case that module when we are processing constants, which is exactly what this PR does.

### Tests

Added the test contributed by @jeffcarbs

### DIFF
--- a/lib/tapioca/sorbet_ext/generic_name_patch.rb
+++ b/lib/tapioca/sorbet_ext/generic_name_patch.rb
@@ -23,17 +23,29 @@ module T
       def type_member(variance = :invariant, fixed: nil, lower: T.untyped, upper: BasicObject)
         # `T::Generic#type_member` just instantiates a `T::Type::TypeMember` instance and returns it.
         # We use that when registering the type member and then later return it from this method.
-        type_member = Tapioca::TypeMember.new(variance, fixed, lower, upper)
-        Tapioca::GenericTypeRegistry.register_type_variable(self, type_member)
-        type_member
+        Tapioca::TypeVariable.new(
+          Tapioca::TypeVariable::Type::Member,
+          variance,
+          fixed,
+          lower,
+          upper
+        ).tap do |type_variable|
+          Tapioca::GenericTypeRegistry.register_type_variable(self, type_variable)
+        end
       end
 
       def type_template(variance = :invariant, fixed: nil, lower: T.untyped, upper: BasicObject)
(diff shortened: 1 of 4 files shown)


vinistock (approved):

lib/tapioca/sorbet_ext/generic_name_patch.rb
```diff
@@ -60,60 +85,62 @@ def name
...
+      # module gets bound to a name under that name, as well.
+      unless constant_name
+        constant_name = bind_and_get_name_pre_3_0
+      end
 
-    sig { returns(T.untyped) }
-    attr_reader :fixed, :lower, :upper
+      # If we still don't have a name, fallback to `T.untyped`
+      unless constant_name
+        return "T.untyped"
```
Morriar, line 140: This feels weird considering the name of the method `constant_name`. Shouldn't we be returning `nil` and let the client code put `T.untyped` in place of the type variable?

paracycle: Good call, since if we ever fall back to `T.untyped`, we would actually generate bad code like:
```ruby
T.untyped = type_member
```

This really should return `nil`, indeed.

PR #702, opened 2022-01-12
Add information about typed_overrides in ver >= 0.6.0

mszyndel:
### Motivation

Readme is missing information about the change to `typed_overrides` in `sorbet/tapioca/config.yml`

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ and supplying an alternative configuration file path.
 
 A configuration file must be a well-formed YAML file with top-level keys for the various Tapioca commands. Keys under each such top-level command should be the underscore version of a long option name for that command and the value for that key should be the value of the option.
 
-For example, if you always want to generate gem RBIs with inline documentation and you want to generate `activesupport` RBIs with `typed: false`, then you would create the file `sorbet/tapioca/config.yml` as:
+For example, if you always want to generate gem RBIs with inline documentation and you want to generate `activesupport` RBI with `typed: false`, then you would create the file `sorbet/tapioca/config.yml` as:
 
 ```yaml
 gem:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -206,6 +206,14 @@ dsl:
   - ActiveRecordFixtures
 ```
 
+Starting with version `0.6.0` `typed_overrides` need to be defined under the top-level `gem` key
```
KaanOzkan, line 209: Since tapioca is below version `1.0` we tend to be less cautious with these things. You can see that on top of the README we have ```⚠️ Note: This software is currently under active development. The API and interface should be considered unstable until a v1.0.0 release.```

So I think a better solution is to remove this line and instead combine the `yml` example with the first `gem` example above:
```
For example, if you always want to generate gem RBIs with inline documentation and you want to generate `activesupport` RBIs with `typed: false`, then you would create the file `sorbet/tapioca/config.yml` as:
```

KaanOzkan: I still think there's value in giving an example with `typed_overrides` since it's so common.

README.md
```diff
@@ -188,11 +188,13 @@ and supplying an alternative configuration file path.
 
 A configuration file must be a well-formed YAML file with top-level keys for the various Tapioca commands. Keys under each such top-level command should be the underscore version of a long option name for that command and the value for that key should be the value of the option.
 
-For example, if you always want to generate gem RBIs with inline documentation, then you would create the file `sorbet/tapioca/config.yml` as:
+For example, if you always want to generate gem RBIs with inline documentation and you want to generate `activesupport` RBIs with `typed: false`, then you would create the file `sorbet/tapioca/config.yml` as:
```
Morriar, line 191:
```suggestion
For example, if you always want to generate gem RBIs with inline documentation and you want to generate `activesupport` RBI with `typed: false`, then you would create the file `sorbet/tapioca/config.yml` as:
```

Morriar (approved):

paracycle:
Closing in favour of #710

PR #703, opened 2022-01-12
Drop pry dependency

casperisfine:
Unless I'm mistaken it's meant as a development / test dependency, not runtime.


KaanOzkan:
I remember we used to rely on `pry` but the tests are passing which makes me think that's no longer the case

paracycle:
I am afraid we actually have a runtime dependency on `Pry` still: https://github.com/Shopify/tapioca/blob/main/lib/tapioca/compilers/symbol_table/symbol_generator.rb#L669

I _think_ we can get rid of this soon and even if we can't, it is not code we can't replicate, but we still have to depend on it explicitly today.

Issue #357, opened 2021-06-15
Generated RBI skipping included modules.


vaporyhumo:
I'm building a gem where some of the classes are including sorbet modules, particularly `T::Props` and `T::Props::Constructor`; these modules are not getting included when running `bin/tapioca generate my_gem` on other projects.

Here is an extract of the file generated by `bin/parlour` when run on the gems directory:
```ruby
module Muina
  class Params
    include T::Props
    include T::Props::Constructor

    sig { params(params: T.any(T::Hash[T.untyped, T.untyped], ActionController::Parameters)).returns(T.attached_class) }
    def self.extract(params); end
  end
end
```

and here is the matching extract from the .rbi file being generated by tapioca:

```ruby
module Muina; end
class Muina::Params
  class << self
    def extract(params); end
  end
end
``` 

So I'm getting a lot of missing methods errors. Is there something I'm not getting right?

For now the way to go seems to be adding a script to copy the parlour's rbi file manually, and skipping it in tapioca (?).

I **can** provide the whole repo, and will gladly add any missing information upon request.

paracycle:
Very early on in the project, we were not filtering any of the modules that are included/extended into constants, but then we ended up in situations where we generated things like:
```ruby
class Foo
  include T::Private::Final
end
```
which would fail type-checking since `T::Private::Final` was never declared in the Sorbet RBI file, since it is a private constant.

Because of this, we started filtering all module includes/extends where the module name starts with `"T::"`. You can see that happening [here](https://github.com/Shopify/tapioca/blob/e724ef9348710f1f8a20a9972a2afe080ad04459/lib/tapioca/compilers/symbol_table/symbol_generator.rb/#L372) and [here](https://github.com/Shopify/tapioca/blob/e724ef9348710f1f8a20a9972a2afe080ad04459/lib/tapioca/compilers/symbol_table/symbol_generator.rb/#L372)

Now, it seems we might have gone too far with that filtering, which is what is causing the problem you are having. On the other hand, [you can see that](https://github.com/sorbet/sorbet/blob/master/rbi/sorbet/tprops.rbi#L3-L4) those includes were never really designed to have been used outside of Sorbet runtime itself. 

So I am torn between how Tapioca should behave here.

vaporyhumo:
[...]

PR #705, opened 2022-01-14
Allow extended/included T::Props* to be present in .rbi

CraigChilds94:
### Motivation
_This is my first PR on this project._

- As per the issue outlined in https://github.com/Shopify/tapioca/issues/357 there are cases where code will include `T::Props` modules, in the event of this, missing methods errors will occur as Tapioca is not outputting these in the `.rbi` files.

- The current implementation filters out included modules starting with `T::`, this does not allow for this circumstance. 

- As per https://github.com/sorbet/sorbet/blob/master/rbi/sorbet/tprops.rbi the "private-ness" of these modules is in question, but as per the related issue it appears to be something we wish to support.

Closes https://github.com/Shopify/tapioca/issues/357

### Implementation
- It replaces the filter predicate of `T::` with a Regex which allows `T::Props`

### Tests
- Test and updated tests in [spec/tapioca/compilers/symbol_table_compiler_spec.rb](https://github.com/Shopify/tapioca/blob/3631c249e1c12485ef118453159e9097656733d1/spec/tapioca/compilers/symbol_table_compiler_spec.rb)


lib/tapioca/compilers/dynamic_mixin_compiler.rb
```diff
@@ -174,7 +174,7 @@ def compile_class_attributes(tree)
 
   sig { params(tree: RBI::Tree).returns([T::Array[Module], T::Array[Module]]) }
   def compile_mixes_in_class_methods(tree)
-    includes = dynamic_includes.select { |mod| (name = name_of(mod)) && !name.start_with?("T::") }
+    includes = dynamic_includes.select { |mod| (name = name_of(mod)) && !name.start_with?(/T\:\:(?!Props)/) }
```
paracycle, line 177: This looks good, but can we extract this logic out to a method and make the check more explicit please? I find the regexp highly unreadable myself. I'd rather see a `filtered_mixin?(mixin_name)` method like:
```ruby
def filtered_mixin?(mixin_name)
  if mixin_name.start_with?("T::Props")
    false # don't filter T::Props and subconstants since they have semantic value
  elsif mixin_name.start_with?("T::")
    true # rest of T:: namespace is mostly private constants
  else
    false # nothing else is filtered
  end
end
```
where we can add the reason why things are filtered or not, as well.

CraigChilds94:
@paracycle Thanks, I was trying to avoid introducing duplicate code but if it's a non-issue then this is an okay approach :)

paracycle:
@CraigChilds94 Thank you for avoiding duplication, but I'd rather have duplication than extract a module prematurely. It is easier to spot duplication and find semantic extractions as they accumulate, than it is to figure out if an extracted module should be refactored into something else.

Issue #644, opened 2021-12-03
Find duplicates in shim files


Morriar:
Consider this:

```
# sorbet/rbi/shims/a.rbi

class Foo; end
```

```
# sorbet/rbi/shims/b.rbi

class Foo; end
```

We should clean or at least warn the user about it.

Note: this is not done with `clean-shims` because we currently do not index shim files.

PR #706, opened 2022-01-14
Replace `clean-shims` by `check-shims`

Morriar:
### Motivation

Closes https://github.com/Shopify/tapioca/issues/644.

The previous approach of automatically deleting duplicated entries was not ideal because:

1. it didn't look at duplicated RBIs in the same shim files
2. it didn't look at duplicated RBIs between shim files
3. it automatically deleted the entry from the shims without knowing if it was the right entry to delete

For all these reasons I think it would be better to offer a more precise duplicates detection mechanism and let the end-user decide which definitions should be deleted.

### Implementation

Logic:

1. Index all the definitions from shims, gems and dsl RBIs
2. Find properties that have more than one definition in RBIs including at least one in shims
3. Discard definitions that have a signature (having a duplicate for a property in a shim to add a signature is ok)
4. Display error messages for the remaining ones

Example output with manually added duplicated entries:

image

The diff is easier to review commit by commit.

### Tests

See automated tests.


KaanOzkan:
Are you saying generated RBI could evolve overtime to something else?

I was thinking that an edge case is when a shim node has the exact same signature with a gem/dsl node. With this implementation we wouldn't consider them duplicates. However, if a shim is defining the same method name, arguments and same signature then it must be identical to the generated gem/dsl RBI and is a duplicate.

So I think the question is whether it's worth supporting this edge case by adding more complexity to the `shims_have_duplicates` method.

vinistock:
The PR looks good to me. My only concern is: should we somehow make it clear in the error that devs should be deleting the duplicates from the shims and not from the auto-generated RBIs (DSL and gems)?

I feel like that could be a source of confusion. Devs might think they need to remove the duplicate from the DSL/gem RBI, which will then continue failing since Tapioca will just re-generate those back in.

Morriar:
> I was thinking that an edge case is when a shim node has the exact same signature with a gem/dsl node. With this implementation we wouldn't consider them duplicates. 

Oh good point, if the signatures are identical we can remove the shim. I'll see if I can get this working 👍

Morriar:
> Should we somehow make it clear in the error that devs should be deleting the duplicates from the shims and not from the auto-generated RBIs (DSL and gems)?

@vinistock I added a last line to the error message:

```
Please remove the duplicated definitions from the sorbet/rbi/shims directory.
```

Let me know if it's clear enough.

KaanOzkan (approved):

PR #709, opened 2022-01-17
Add configuration validation

### LINKED ISSUE #704: Tapioca should check its config file
For example, using:

```yaml
typed_overrides:
  "activesupport": "false"
```

Instead of: 

```yaml
gem:
  typed_overrides:
    "activesupport": "false"
```

Should make Tapioca generate a warning.

This will avoid some frustration as to why an option is not working and warn the user they need to update their config when we update Tapioca.

paracycle:
### Motivation

Fixes #704

### Implementation

The first time `options` method is hit, we validate the Tapioca configuration. The validation makes sure that all top level keys match to some command, all command options match to one of the flags for that command and that all values for command options are of the expected type.

### Tests

Add new tests for configuration failures.

### DIFF
--- a/lib/tapioca/helpers/config_helper.rb
+++ b/lib/tapioca/helpers/config_helper.rb
@@ -6,6 +6,9 @@ require "yaml"
 module Tapioca
   module ConfigHelper
     extend T::Sig
+    extend T::Helpers
+
+    requires_ancestor { Thor }
 
     sig { returns(String) }
     attr_reader :command_name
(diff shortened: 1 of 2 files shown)


lib/tapioca/helpers/config_helper.rb
```diff
@@ -6,6 +6,11 @@
 module Tapioca
   module ConfigHelper
     extend T::Sig
+    extend T::Helpers
+
+    requires_ancestor do
+      Thor
+    end
```
vinistock, line 13: NIT: can we use a one liner here?

lib/tapioca/helpers/config_helper.rb
```diff
@@ -60,9 +65,122 @@ def config_options(options)
...
+
+        validate_config_options(command.options, config_key, config_options || {})
+      end.compact
+
+      unless errors.empty?
+        print_errors(config_file, errors)
+        exit(1)
+      end
+    ensure
+      @validating_config = false
```
vinistock, line 96: Do we want to set this to `false` here? Doesn't that make it so that invoking `config_options` will validate the options every time?

lib/tapioca/helpers/config_helper.rb
```diff
@@ -60,9 +65,122 @@ def config_options(options)
...
+      const :message, String
+      const :colors, T::Array[Symbol]
+    end
+
+    class ConfigError < T::Struct
+      const :message_parts, T::Array[ConfigErrorMessagePart]
+    end
+
+    sig { params(msg: String).returns(ConfigError) }
+    def build_error(msg)
```
vinistock, line 146: It feels strange that we need to scan the string we pass ourselves to this method. Would we be able to simplify this logic with some extra arguments maybe?

lib/tapioca/helpers/config_helper.rb
vinistock: Between different runs we should definitely re-validate configs. I guess what I'm not seeing is: in which scenario will `config_options` be invoked and `@validating_config` will be `true` so that validation is skipped? When do we actually skip the validation?

paracycle: The intention is to not reenter the method if `options` is called again while _inside_ the method. We want to clear that flag as we are exiting, it is not our concern if this method is called multiple times or not.

However, the original `options` method already does memoization on the returned result, so it will take care to not call this again afterwards.

lib/tapioca/helpers/config_helper.rb
paracycle: I found it much less readable to do:
```ruby
say_error("foo bar ", :yellow)
say_error("something ", :bold, :blue)
say_error("something else", :yellow)
```
etc, and instead decided that it should be the responsibility of another method to decide how to colorize and print my message, which is ``"foo bar `something` something else"``.

vinistock (approved):

Morriar (approved):

lib/tapioca/helpers/config_helper.rb
```diff
@@ -60,9 +65,122 @@ def config_options(options)
...
       end
 
+      validate_config!(config_file, config)
+
       Thor::CoreExt::HashWithIndifferentAccess.new(config[command_name] || {})
     end
 
+    sig { params(config_file: String, config: T::Hash[T.untyped, T.untyped]).void }
+    def validate_config!(config_file, config)
+      # To ensure that this is not re-entered, we mark during validation
```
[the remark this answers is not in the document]
paracycle: Yes, [`say_error`](https://github.com/rails/thor/blob/6ba79014c89434de81e14c6a4ddbeb0a300022ce/lib/thor/shell/basic.rb#L114) (or `say`) end up [calling `options[:quiet]`](https://github.com/rails/thor/blob/6ba79014c89434de81e14c6a4ddbeb0a300022ce/lib/thor/shell/basic.rb#L411) to decide to print or not which ends up reentering the `options` method, before we've memoized the merged options hash. That causes an infinite loop. In general, any Thor code we call from this method could end up doing another `options` lookup, so I am protecting this method.

PR #720, opened 2022-01-24
Add `gems` as an alias to the `gem` command

paracycle:
### Motivation

I keep typing `bin/tapioca gems foo bar baz` especially when I am about to generate gem RBIs for multiple gems, so I thought that it would be nice to add `gems` as an alias to the `gem` command.

### Implementation

Use built-in Thor command alias mapping.

### Tests

Add a test to verify that the alias works as intended.

### DIFF
--- a/lib/tapioca/cli.rb
+++ b/lib/tapioca/cli.rb
@@ -255,7 +255,7 @@ module Tapioca
       exit(0)
     end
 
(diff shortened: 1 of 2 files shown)


rafaelfranca (approved):

vinistock (approved):

Morriar (approved):
Should we also add an alias for `dsls`?

PR #730, opened 2022-01-26
Fix Dependabot labels

mutecipher:
### Motivation

The `dependencies` label is already ignored from release notes, so removed the `ignore-for-release` label.

### Implementation

Delete the `ignore-for-release` label from dependabot workflow.

### Tests

See automated tests.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,7 +9,6 @@ updates:
     labels:
       - "dependencies"
       - "ruby"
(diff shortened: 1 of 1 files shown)


KaanOzkan (approved):
`ignore-for-release` doesn't actually seem to be ignored by the way https://github.com/Shopify/tapioca/blob/832155262032107101155ecd9518ecb9428a01a0/.github/release.yml#L4-L7

PR #733, opened 2022-01-26
Extract `sorbet_helper` module

Morriar:
### Motivation

This module can be mixed in when we need to access Sorbet related features instead of calling static methods on it.


rafaelfranca:
I'm not sure this is a good refactoring. Ruby is pretty effective to mask this smell, but you are changing all those classes to inherit from that helper. Does it make sense in terms of design?

Which problem not calling static methods solves?

rafaelfranca (approved):

vinistock (approved):

PR #734, opened 2022-01-27
Merge Compilers::TodoCompiler inside Generator::Compiler

Morriar:
### Motivation

The current separation is artificial, we don't need this indirection, the "compiler" is only two methods and does not servce any purpose on it's own.


lib/tapioca/generators/todo.rb
```diff
@@ -21,55 +21,57 @@ def initialize(todo_file:, file_header:, default_command:, file_writer: FileWrit
...
         say("\nAll unresolved constants have been written to #{name}.", [:green, :bold])
         say("Please review changes and commit them.", [:green, :bold])
       end
 
-      sig { params(command: String, reason: T.nilable(String), strictness: T.nilable(String)).returns(String) }
-      def rbi_header(command, reason: nil, strictness: nil)
-        statement = <<~HEAD
-          # DO NOT EDIT MANUALLY
-          # This is an autogenerated file for #{reason}.
-          # Please instead update this file by running `#{command}`.
```
rafaelfranca, line 59: Doesn't this header still makes sense?

rafaelfranca (approved):

paracycle (approved):

PR #737, opened 2022-01-27
Move dsl helpers together

Morriar:
### Motivation

Those helpers are only meaningful for DSL compilers.


lib/tapioca/rbi_ext/model.rb
```diff
@@ -41,11 +41,6 @@ def set_empty_body_content
       comments << RBI::Comment.new("THIS IS AN EMPTY RBI FILE.")
       comments << RBI::Comment.new("see https://github.com/Shopify/tapioca/wiki/Manual-Gem-Requires")
     end
-
-    sig { returns(T::Boolean) }
-    def empty?
-      root.empty?
-    end
```
KaanOzkan, line 48: This change is present in another PR, let's not forget.

KaanOzkan (approved):

PR #739, opened 2022-01-27
Start using Rails test runner for running tests

paracycle:
### Motivation

This allows us to use line numbers for test running and supplies a much better experience overall.

### Implementation

I also changed the default Rake command to run `bin/test` inline with what the Rails task runner is doing [here](https://github.com/rails/rails/blob/main/railties/lib/rails/test_unit/runner.rb#L33-L37) and [here](https://github.com/rails/rails/blob/main/railties/lib/rails/test_unit/testing.rake#L9-L16).

### Tests

None


vinistock:
I like the idea of using Rails test runners, but I'd expect the default task to be running tests.

mutecipher (approved):

Rakefile
```diff
@@ -1,22 +1,18 @@
...
 
-Rake::TestTask.new do |t|
-  t.libs << "lib"
-  t.libs << "spec"
-  t.warning = false
-  t.test_files = FileList["spec/**/*_spec.rb"]
+desc "Run tests"
+task :test do
+  require "shellwords"
+  test = Array(ENV.fetch("TEST", []))
```
vinistock, line 12: Is it possible to pass more than one test via `TEST`? Should this be a string instead of an array?

vinistock: TIL

vinistock (approved):

Morriar (approved):

PR #747, opened 2022-01-27
Order rubocop rules

Morriar:
Also remove a useless ignored path.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -12,7 +12,6 @@ AllCops:
   SuggestExtensions: false
   Exclude:
   - "vendor/**/*"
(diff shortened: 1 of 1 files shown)


mutecipher (approved):

.rubocop.yml
```diff
@@ -12,18 +12,16 @@ AllCops:
   SuggestExtensions: false
   Exclude:
   - "vendor/**/*"
-  - "spec/support/repo/bin/tapioca"
 
-Style/CaseEquality:
+Gemspec/RequireMFA:
```
rafaelfranca, line 16: This should not be here. This cop is enable disabled in our style guide. You should disable the `NewCops`.

PR #756, opened 2022-01-27
Never generate `T.nilable(T.untyped)` in DSL generators

### LINKED ISSUE #755: DSL generates RBIs with T.nilable(T.untyped) that is now an error
As of https://github.com/sorbet/sorbet/pull/5129, `T.nilable(T.untyped)` is now an error. But the tapioca DSL generators create files that use that.

For example:

```
sorbet/rbi/dsl/active_storage/attachment.rbi:183: T.nilable(T.untyped) is the same as T.untyped https://srb.help/5070
     183 |    sig { returns(T.nilable(T.untyped)) }
                            ^^^^^^^^^^^^^^^^^^^^
  Autocorrect: Use `-a` to autocorrect
    sorbet/rbi/dsl/active_storage/attachment.rbi:183: Replace with T.untyped
     183 |    sig { returns(T.nilable(T.untyped)) }
                            ^^^^^^^^^^^^^^^^^^^^

sorbet/rbi/dsl/active_storage/attachment.rbi:186: T.nilable(T.untyped) is the same as T.untyped https://srb.help/5070
     186 |    sig { params(value: T.nilable(T.untyped)).void }
                                  ^^^^^^^^^^^^^^^^^^^^
  Autocorrect: Use `-a` to autocorrect
    sorbet/rbi/dsl/active_storage/attachment.rbi:186: Replace with T.untyped
     186 |    sig { params(value: T.nilable(T.untyped)).void }
                                  ^^^^^^^^^^^^^^^^^^^^

sorbet/rbi/dsl/active_storage/attachment.rbi:192: T.nilable(T.untyped) is the same as T.untyped https://srb.help/5070
     192 |    sig { returns(T.nilable(T.untyped)) }
                            ^^^^^^^^^^^^^^^^^^^^
  Autocorrect: Use `-a` to autocorrect
    sorbet/rbi/dsl/active_storage/attachment.rbi:192: Replace with T.untyped
     192 |    sig { returns(T.nilable(T.untyped)) }
```

paracycle:
### Motivation

Fixes #755 

### Implementation

Move `as_nilable_type` to the DSL base class and change all usages of `T.nilable` wrapping on bare unknown types use that utility method instead.

### Tests

Updated tests to conform to the new reality.

### DIFF
--- a/lib/tapioca/compilers/dsl/active_record_relations.rb
+++ b/lib/tapioca/compilers/dsl/active_record_relations.rb
@@ -251,6 +251,15 @@ module Tapioca
           sig { returns(String) }
           attr_reader :constant_name
 
+          sig { params(type: String).returns(String) }
+          def as_nilable_type(type)
+            if type.start_with?("T.nilable(", "::T.nilable(") || type == "T.untyped" || type == "::T.untyped"
+              type
+            else
+              "T.nilable(#{type})"
+            end
+          end
+
           sig { void }
           def create_classes_and_includes
             model.create_extend(CommonRelationMethodsModuleName)
@@ -547,7 +556,7 @@ module Tapioca
                   parameters: [
                     create_rest_param("args", type: "T.untyped"),
                   ],
-                  return_type: "T.nilable(#{constant_name})"
+                  return_type: as_nilable_type(constant_name)
                 )
               when :find_by!
                 create_common_method(
@@ -571,7 +580,7 @@ module Tapioca
                 return_type = if method_name.end_with?("!")
                   constant_name
                 else
-                  "T.nilable(#{constant_name})"
+                  as_nilable_type(constant_name)
                 end
 
                 create_common_method(
--- a/lib/tapioca/compilers/dsl/base.rb
+++ b/lib/tapioca/compilers/dsl/base.rb
@@ -176,6 +176,15 @@ module Tapioca
           return_type
         end
 
+        sig { params(type: String).returns(String) }
+        def as_nilable_type(type)
+          if type.start_with?("T.nilable(", "::T.nilable(") || type == "T.untyped" || type == "::T.untyped"
+            type
+          else
+            "T.nilable(#{type})"
+          end
+        end
+
         sig { params(name: String).returns(T::Boolean) }
         def valid_parameter_name?(name)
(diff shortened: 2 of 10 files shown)


lib/tapioca/compilers/dsl/active_record_relations.rb
```diff
@@ -251,6 +251,15 @@ def generate
           sig { returns(String) }
           attr_reader :constant_name
 
+          sig { params(type: String).returns(String) }
```
paracycle, line 254: This inner class does not subclass from `Dsl::Base` so doesn't have access to this method. I added it here for the time being, since this inner class will go away very soon, and this method will go away with it.

mutecipher (approved):

RyanBrushett (approved):

lib/tapioca/compilers/dsl/base.rb
```diff
@@ -176,6 +176,15 @@ def compile_method_return_type_to_rbi(method_def)
           return_type
         end
 
+        sig { params(type: String).returns(String) }
+        def as_nilable_type(type)
+          if type.start_with?("T.nilable(", "::T.nilable(") || type == "T.untyped" || type == "::T.untyped"
+            type
+          else
+            "T.nilable(#{type})"
+          end
```
Morriar, line 185: I think it still possible to get a `T.nilable(T.untyped)` with `as_nilable("T.proc.returns(T.nilable(T.untyped))")`.

We should refactor the helper methods to all pass through a single point where we can check/assert/raise things. In case we generate incorrect RBIs we could stop/skip during the generation and explaining to the user how to report an issue instead of producing invalid RBIs and confused developers.

paracycle: Yes, we can, and that's why I want to build the types like how Sorbet runtime does it, instead of concatenating strings, in the long run. We should be able to do:
```ruby
T.nilable(T.proc.returns(T.nilable(T.untyped))).to_s
# => "T.nilable(T.proc.params().returns(T.nilable(T.untyped)))"
```
and that should always return the correct thing. (Not sure if anyone fixed Sorbet runtime to collapse `T.nilable(T.untyped)` to just `T.untyped` though)

PR #767, opened 2022-01-29
All tests should use `spec_helper` for unified setup

paracycle:
### Motivation

Some tests were doing their own setup inside `spec_with_project` but that meant that any Minitest configuration would have to be duplicated.

However, `spec_with_project` just defines a `Minitest::Spec` subclass, so it can easily be required from `spec_helper` as part of the setup for all tests.

This fixes a problem with CLI tests that were just relying on requiring `spec_with_project` not respecting the line number filter for the runner, since the configuration of Minitest was different there.

### Implementation

- Remove any Minitest setup code from `spec_with_project`
- Require `spec_with_project` in `spec_helper`
- Replace all requires of `spec_with_project` with `spec_helper`

### Tests

No new tests

### DIFF
--- a/spec/tapioca/cli/check_shims_spec.rb
+++ b/spec/tapioca/cli/check_shims_spec.rb
@@ -1,7 +1,7 @@
 # typed: true
 # frozen_string_literal: true
 
-require "spec_with_project"
+require "spec_helper"
 
 module Tapioca
   class CleanShimsTest < SpecWithProject
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -13,6 +13,7 @@ require "tapioca/helpers/test/content"
 require "tapioca/helpers/test/template"
 require "tapioca/helpers/test/isolation"
 require "dsl_spec_helper"
+require "spec_with_project"
 
 Minitest::Reporters.use!(Minitest::Reporters::SpecReporter.new(color: true))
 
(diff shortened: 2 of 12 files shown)


mutecipher (approved):

spec/tapioca/cli/check_shims_spec.rb
```diff
@@ -1,7 +1,7 @@
 # typed: true
 # frozen_string_literal: true
 
-require "spec_with_project"
+require "spec_helper"
```
Morriar, line 4: One problem is that we make all test always require `tapioca/internal` and all what comes with it when it's not always needed. This breaks a bit the isolation.

paracycle: Does it matter? All of the ones that were explicitly using `spec_with_project`, like this one are CLI tests that run in a separate process, no?

Morriar: For example this test will not be isolated the same way: https://github.com/Shopify/tapioca/blob/main/spec/tapioca/gemfile_spec.rb#L5

Not sure if it's okay or not. Maybe we should refactor the test itself. What do you think?

Morriar: I'm good with this. We can always create another isolated spec helper if we need it later 👍

PR #768, opened 2022-01-31
Proposal: gem generation plugin

Morriar:
### Motivation

The `symbol_table_compiler` is a hairy ball that could be divided into smaller chunks.

This PR proposes to extract plugins that would decorate the nodes that are produced `symbol_table_compiler`.

As an example I extracted the `yard_doc` plugin since it was the less coupled part. We can also do it for `requires_ancestor`, `signatures`, `dynamic_mixins`, `enums`, `props` etc.

What do you think?

### DIFF
--- a/lib/tapioca/gem/plugins/base.rb
+++ b/lib/tapioca/gem/plugins/base.rb
@@ -1,0 +1,33 @@
+# typed: strict
+# frozen_string_literal: true
+
+require "pathname"
+
+module Tapioca
+  module Gem
+    module Plugins
+      class Base
+        extend T::Sig
+        extend T::Helpers
+
+        abstract!
+
+        sig { params(cls: RBI::Class).void }
+        def decorate_class(cls)
+        end
+
+        sig { params(mod: RBI::Module).void }
+        def decorate_module(mod)
+        end
+
+        sig { params(const: RBI::Const).void }
+        def decorate_const(const)
+        end
+
+        sig { params(meth: RBI::Method).void }
+        def decorate_method(meth)
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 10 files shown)


paracycle (approved):
Great extraction!

lib/tapioca/gem/plugins/yard_doc.rb
```diff
@@ -0,0 +1,62 @@
...
+
+        sig { params(const: RBI::Const).void }
+        def decorate_const(const)
+          const.comments = documentation_comments(const.fully_qualified_name)
+        end
+
+        sig { override.params(meth: RBI::Method).void }
+        def decorate_method(meth)
+          meth.comments = documentation_comments(meth.fully_qualified_name)
+        end
```
KaanOzkan, line 41: Looks like the only downside is having 4 of these nearly identical method calls on different `RBI` objects for each plugin. In more complicated plugins I assume we can have an "entrypoint" method that is responsible of the internal plugin logic, instead of `SymbolTableCompiler` owning driving it with methods like https://github.com/Shopify/tapioca/blob/53b7307ddb357845f5166a6d57161e5de6300547/lib/tapioca/compilers/symbol_table_compiler.rb#L195

KaanOzkan (approved):
LGTM

lib/tapioca/gem/plugins/base.rb
```diff
@@ -0,0 +1,33 @@
...
+        def decorate_module(mod)
+        end
+
+        sig { params(const: RBI::Const).void }
+        def decorate_const(const)
+        end
+
+        sig { params(meth: RBI::Method).void }
+        def decorate_method(meth)
+        end
```
vinistock, line 29: Should these methods be marked as abstract? Also, could use one liners for the empty bodies? E.g.:
```ruby
def decorate_class(cls); end
```

vinistock (approved):

PR #769, opened 2022-01-31
Clean symbol loader

Morriar:
Clean the way we access symbols:
* extract `SymbolTableParser`
* create `core_symbols`, `engine_symbols` and `gem_symbols`
* memoize `engine_symbols`
* move `SymbolLoader` to top level


paracycle (approved):
😍

lib/tapioca/symbol_loader.rb
```diff
@@ -0,0 +1,81 @@
...
+    class << self
+      extend T::Sig
+      include SorbetHelper
+      include Reflection
+
+      sig { returns(T::Set[String]) }
+      def core_symbols
+        unless @core_symbols
+          output = symbol_table_json_from("-e ''", table_type: "symbol-table-full-json")
+          @core_symbols = T.let(SymbolTableParser.parse_json(output), T.nilable(T::Set[String]))
```
KaanOzkan, line 18: The name `core_symbols` doesn't feel informative enough to me. Is there somewhere else that we use this terminology? If not maybe it's better to use a different name for this?

Morriar: I went with `payload_symbols` instead

KaanOzkan (approved):

PR #771, opened 2022-01-31
Implement support for final method signatures

### LINKED ISSUE #766: Gem RBI generation does not generate `sig(:final)` methods as final

paracycle:
### Motivation

Fixes #766.

### Implementation

Use the new `is_final` flag in `rbi` gem and set it to true for methods that have a signature marked as final.

### Tests

Added an extra test case with final signatures.

### DIFF
--- a/lib/tapioca/compilers/gem/listeners/sorbet_signatures.rb
+++ b/lib/tapioca/compilers/gem/listeners/sorbet_signatures.rb
@@ -60,7 +60,19 @@ module Tapioca
               sig.is_overridable = true
             end
 
+            sig.is_final = signature_final?(signature)
+
             sig
+          end
+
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -48,3 +48,5 @@ group(:development, :test) do
   # https://github.com/rails/rails/blob/0919aa97260ab8240150278d3b07a1547489e3fd/Gemfile#L178-L191
   gem("net-smtp", "0.3.1", require: false)
 end
+
+gem "rbi", github: "Shopify/rbi", branch: "main"
```
paracycle, line 52: I need a release of `rbi` so that I can bump the `.gemspec` dependency version instead of this. Until then this PR will be WIP.

PR #780, opened 2022-02-01
Use root/absolute namespace in compiled RBI

jeffcarbs:
### Motivation

Came across a subtle bug in tapioca RBI generation where absolute namespacing get dropped (e.g. `::String` comes out of RBI as `String`). This causes issues when the referenced constant shares a name with both a global constant and one that lives somewhere inside the current namespace. See the test for an example.

### Implementation

~Unsure how to fix, just wanted to send a reproduction.~ Use `qualified_name_of` in the Simple type name patch.

### Tests

Added new tests and also needed to update existing ones.

### DIFF
--- a/lib/tapioca/compilers/symbol_table_compiler.rb
+++ b/lib/tapioca/compilers/symbol_table_compiler.rb
@@ -131,7 +131,7 @@ module Tapioca
 
       sig { params(event: Gem::SymbolFound).void }
       def on_symbol(event)
-        symbol = event.symbol
+        symbol = event.symbol.delete_prefix("::")
         return if symbol_in_payload?(symbol) && !@bootstrap_symbols.include?(symbol)
 
         constant = constantize(symbol)
--- a/lib/tapioca/sorbet_ext/name_patch.rb
+++ b/lib/tapioca/sorbet_ext/name_patch.rb
@@ -1,9 +1,9 @@
 # typed: true
 # frozen_string_literal: true
 
-# We need sorbet to compile this signature before applying the patch
-# to avoid an infinite loop.
-::Tapioca::Reflection.qualified_name_of(String)
+# We need sorbet to compile the signature for `qualified_name_of` before applying
+# the patch to avoid an infinite loop.
+T::Utils.signature_for_method(::Tapioca::Reflection.method(:qualified_name_of))
 
 module T
   module Types
--- a/lib/tapioca/sorbet_ext/name_patch.rb
+++ b/lib/tapioca/sorbet_ext/name_patch.rb
@@ -1,12 +1,18 @@
 # typed: true
 # frozen_string_literal: true
 
+# We need sorbet to compile this signature before applying the patch
+# to avoid an infinite loop.
+::Tapioca::Reflection.qualified_name_of(String)
+
 module T
   module Types
     class Simple
       module NamePatch
         def name
-          @name ||= ::Tapioca::Reflection.name_of(@raw_type).freeze
+          # Sorbet memoizes this method into the `@name` instance variable but
+          # doing so means that types get memoized before this patch is applied
+          ::Tapioca::Reflection.qualified_name_of(@raw_type)
         end
       end
 
--- a/spec/tapioca/compilers/symbol_table_compiler_spec.rb
+++ b/spec/tapioca/compilers/symbol_table_compiler_spec.rb
@@ -112,7 +112,7 @@ class Tapioca::Compilers::SymbolTableCompilerSpec < Minitest::HooksSpec
         end
 
         Bar::Arr = T.let(T.unsafe(nil), Array)
-        Bar::Foo = T.type_alias { T.any(String, Symbol) }
+        Bar::Foo = T.type_alias { T.any(::String, ::Symbol) }
 
         module Base
           include ::T::Props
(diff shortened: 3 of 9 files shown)


lib/tapioca/sorbet_ext/name_patch.rb
```diff
@@ -6,7 +6,7 @@ module Types
     class Simple
       module NamePatch
         def name
-          @name ||= ::Tapioca::Reflection.name_of(@raw_type).freeze
+          @name ||= ::Tapioca::Reflection.qualified_name_of(@raw_type).freeze
```
jeffcarbs, line 9: This just a stab at fixing which I think is directionally correct but
- doesn't totally solve the problems since I think not all types are `Simple`, some are just the classes.
- breaks a bunch of things 😬

jeffcarbs: Thanks for the pointer, I'll take a look at that PR and play around with it a bit.

jeffcarbs: @paracycle - I think I figured it out, see the latest commit ("Fix NamePatch"). By the time `NamePatch` gets applied, some types have already had their `name` memoized to the non-absolute path. Removing the memoization seems to fix the issue. LMK what you think.

lib/tapioca/compilers/symbol_table_compiler.rb
```diff
@@ -45,7 +45,7 @@ def compile
 
       sig { params(name: T.nilable(String)).void }
       def add_to_symbol_queue(name)
-        @symbol_queue << name unless name.nil? || symbols.include?(name) || symbol_in_payload?(name)
+        @symbol_queue << name.delete_prefix("::") unless name.nil? || symbols.include?(name) || symbol_in_payload?(name)
```
jeffcarbs, line 48: Copied from https://github.com/Shopify/tapioca/pull/472/commits/fb329842c6a3ede2dd6447ad96cf4eb440b87ab1

spec/tapioca/compilers/dsl/active_job_spec.rb
```diff
@@ -99,10 +99,10 @@ def perform(user_id)
 
           class NotifyJob
             class << self
-              sig { params(user_id: Integer).returns(T.any(NotifyJob, FalseClass)) }
+              sig { params(user_id: ::Integer).returns(T.any(NotifyJob, FalseClass)) }
```
jeffcarbs, line 102: The types in `T.any(NotifyJob, FalseClass)` don't get a `::` since that string is hard-coded by the DSL generator: https://github.com/Shopify/tapioca/blob/main/lib/tapioca/compilers/dsl/active_job.rb#L55

Might be worth updating the API to generate `RBI::` types to use real sorbet types and then convert them to strings to things like the NamePatch would apply. Alternatively these DSL generators could be updated to hard-code the `::` or use `::Tapioca::Reflection`.

jeffcarbs:
@paracycle @KaanOzkan - I think this PR fixes the issue that #472 was attempting to fix. Build is green so this should be good for a review.

lib/tapioca/sorbet_ext/name_patch.rb
```diff
@@ -1,12 +1,18 @@
 # typed: true
 # frozen_string_literal: true
 
+# We need sorbet to compile this signature before applying the patch
+# to avoid an infinite loop.
+::Tapioca::Reflection.qualified_name_of(String)
```
jeffcarbs, line 6: Open to suggestions of a less bad way of doing this.

paracycle: Does running `T::Utils.run_all_sig_blocks` work instead? Alternatively, we could mark the sig of `qualified_name_of` as `checked(:never)` as well, if that helps.

lib/tapioca/compilers/symbol_table_compiler.rb
jeffcarbs: Oh true, removing the prefix should be moved to it’s own line above. Will fix.

jeffcarbs: Fixed. I made the code more similar to the commit I referenced above.

jeffcarbs:
@paracycle - wanted to give this a friendly bump since tapioca is currently generating invalid RBI for one of our gems and there's no easy workaround.

lib/tapioca/compilers/symbol_table_compiler.rb
```diff
@@ -140,6 +140,7 @@ def dispatch(event)
       sig { params(event: Gem::SymbolFound).void }
       def on_symbol(event)
         symbol = event.symbol
+        symbol = T.must(symbol[2..-1]) if symbol.start_with?("::")
```
jeffcarbs, line 143: The `SymbolTableCompiler` was refactored recently but I think this is still roughly the equivalent change to: https://github.com/Shopify/tapioca/commit/fb329842c6a3ede2dd6447ad96cf4eb440b87ab1.

paracycle (approved):
This looks good to me. Thanks for working on this and rebasing it on latest `main`.

jeffcarbs:
@paracycle - Thanks for taking a look! I made the `delete_prefix` change you suggested and fixed a new merge conflict that had come up. Build is green so should be good to go :shipit:

paracycle:
Tested this on Shopify Core and it seems to cause no issues. Thanks for pushing this forward 🙏

PR #785, opened 2022-02-02
Proposal: split SymbolTableCompiler in multiple listeners

Morriar:
### Motivation

Split the `SymbolTableCompiler` monolith into smaller, easier to read/maintain/test pieces.

### Implementation

The two main problems with the plugin approach proposed in #768 are:
1. plugins may discover new constants that should be generated (for example finding a new constant while parsing the sig's return type)
2. plugins may need more data than just the RBI node (for example accessing the `::Method`'s signature)

This means that plugins need to call back inside the compiler to ensure newly found constants will be generated as well as a way to exchange information with the compiler. This PR tries to address those two requirements by borrowing concepts from the event-listener, decorator and observer patterns:

The compiler uses an event queue instead of a symbol queue where we can push 3 kinds of events:

1. `symbol` event when we discover a new symbol like `"Foo"` that have to be constantized
2. `constant` event when we discover a new constant like `Foo` that is already constantized
3. `node` event when we add a new RBI node (`Scope`, `Const` or `Method` for now) in the resulting tree

The compiler knows how to handle `symbol` and `constant` events. For `node` events, it calls listeners that will either decorate the existing nodes (to add comments or signatures for example) or create new ones (to add mixins for example).

The listeners can push `symbol` and `constant` events back to the compiler so they will be generated. So if the listener compiling the sorbet signatures can tell the compiler it needs to generate the params and return types.

The listeners can also push `node` events back to the compiler so other listeners can be triggered on the new node.
For example, if a listener adds a new RBI method, the listener responsible for adding the YardDoc on it will be triggered.

Events are used to exchange information. For example the listener creating `RBI::Method` nodes will push a MethodEvent with the sorbet runtime signature and the sanitized parameter names so they can be used by the node listeners down the line.

Theoretically everything could become a listener, I went as far as to extract adding the `RBI::Method` nodes as a listener on newly added RBI::Scopes in the tree. I decided to leave things related to `symbol` and `constant`-handling inside the compiler because generating aliases, consts and scopes are closely tied together right now.

Pros:
* Somewhat clear separation of concerns
* Very flexible design
* Statically typed

Cons:
* Weird coupling between the compiler and listeners to push back events
* Poorly defined separation between what should be in the compiler or in a listener
* Verbosity when we need to add new events (event class + push method + push events + listeners)

### Tests

No functional changes.


lib/tapioca/compilers/gem/events.rb
```diff
@@ -0,0 +1,122 @@
...
+  module Compilers
+    module Gem
+      class Event
+        extend T::Sig
+        extend T::Helpers
+
+        abstract!
+      end
+
+      class NewSymbolFound < Event
```
paracycle, line 16: I think I would prefer the "noun" - "verb" naming for these:
```
SymbolFound
ConstantFound
NodeAdded
ConstNodeAdded
ScopeNodeAdded
MethodNodeAdded
```

lib/tapioca/internal.rb
```diff
@@ -16,6 +16,18 @@
...
+require "tapioca/compilers/gem/listeners/dynamic_mixins"
+require "tapioca/compilers/gem/listeners/methods"
+require "tapioca/compilers/gem/listeners/mixins"
+require "tapioca/compilers/gem/listeners/sorbet_enums"
+require "tapioca/compilers/gem/listeners/sorbet_helpers"
+require "tapioca/compilers/gem/listeners/sorbet_props"
+require "tapioca/compilers/gem/listeners/sorbet_required_ancestors"
+require "tapioca/compilers/gem/listeners/sorbet_signatures"
+require "tapioca/compilers/gem/listeners/sorbet_type_variables"
+require "tapioca/compilers/gem/listeners/yard_doc"
```
paracycle, line 30: Can we move these requires into `tapioca/compilers/gem/listeners.rb` and just require that from here?

paracycle (approved):
This looks great, just some small naming and organizational comments.

lib/tapioca/compilers/gem/events.rb
Morriar: Done in [91306f0](https://github.com/Shopify/tapioca/pull/785/commits/91306f0dcba1f49c616af03b6019520a36890d49) 👍

lib/tapioca/internal.rb
Morriar: Done in [7472541](https://github.com/Shopify/tapioca/pull/785/commits/747254144a45900143805df4abe1cc1ef4b22d72) 👍

lib/tapioca/compilers/gem/listeners/base.rb
```diff
@@ -0,0 +1,50 @@
...
+          def on_const(event)
+          end
+
+          sig { params(event: ScopeNodeAdded).void }
+          def on_scope(event)
+          end
+
+          sig { params(event: MethodNodeAdded).void }
+          def on_method(event)
+          end
```
vinistock, line 45: Should these be marked as abstract and define them in a single line?

vinistock (approved):
Very nice refactor

lib/tapioca/compilers/gem/listeners/base.rb
Morriar: I don't think so because if we make them abstract, your listener will have to implement all of them even if it only needs to react on `on_method` for example.

I prefer for them to no-op by default and just override the ones you need.

Especially if we start adding new events like `on_attr`, `on_sig` etc. We'll need to change all the listeners to add an empty override every time we add a new one.

vinistock (approved):
I don't mind the solution for removing empty nodes. I think running the listener at the end makes sense and fits into the overall design.

PR #786, opened 2022-02-02
Update AASM DSL generator

davebenvenuti:
### Motivation

Hello! I encountered some type errors in a project related to the AASM gem.  More specifically, it appears the DSL generator for AASM only supports one form of their lifecycle callbacks.

```ruby
class Foo
  include AASM
  
  aasm do
    state :state1
    state :state2

    event :do_something do
      transitions from: :state1, to: :state2

      after :some_method
    end

    def some_method
    end
  end
end
```

The above code would produce a typeerror along the lines of `after requires a block parameter, but no block was passed https://srb.help/7021`.  However, [this is in fact valid usage](https://github.com/aasm/aasm#callbacks) of the `after` hook.  Similar to `ActiveRecord` callbacks, they accept either a block or a symbol corresponding to a method name.

### Implementation

This PR updates the generated `sig`s for AASM lifecycle hooks from this:

```ruby
sig { params(block: T.proc.bind(StateMachine).void).returns(T.untyped)  }
```

To this:

```ruby
sig { params(symbol: T.nilable(Symbol), block: T.nilable(T.proc.bind(StateMachine).void)).returns(T.untyped) }  }
```

### Tests

I updated the relevant test.

### DIFF
--- a/lib/tapioca/dsl/compilers/aasm.rb
+++ b/lib/tapioca/dsl/compilers/aasm.rb
@@ -105,7 +105,8 @@ module Tapioca
                   event.create_method(
                     method,
                     parameters: [
-                      create_block_param("block", type: "T.proc.bind(#{name_of(constant)}).void"),
+                      create_opt_param("symbol", type: "T.nilable(Symbol)", default: "nil"),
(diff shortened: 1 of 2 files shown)


kddnewton (approved):
Looks legit to me

PR #787, opened 2022-02-03
Replace usage of `T::Private::Methods.signature_for_method` with public equivalent

jeffcarbs:
### Motivation

I noticed tapioca was using `T::Private::Methods.signature_for_method` which is a "private" module (private in quotes because it's not actually private, just named to try to discourage use). https://github.com/sorbet/sorbet/pull/4041 introduced a utility method that just calls that private method so this PR updates to use that.

### Implementation

- Replace usages of `T::Private::Methods.signature_for_method` with `Reflection.signature_of` where possible
- In `Reflection.signature_of`, switch to using `T::Utils.signature_for_method`

### Tests

No tests since there should be no behavior change.


Morriar (approved):

paracycle (approved):
Love it, thanks!

PR #788, opened 2022-02-03
Expose DSL compiler test helper as a reusable unit

paracycle:
### Motivation

We currently have too much magic inside the `DslSpec` test superclass which makes it hard for other projects to write tests for their own DSL compilers, and it also makes it hard to understand what is going on inside that superclass as well.

### Implementation

Extract logic to a reusable module that should work independently of the testing framework used. This module provides helper methods to setup DSL compiler under test and possibly collaborating DSL compilers that should also be activated. Moreover it provides convenience methods like `rbi_for` and `gathered_constants` that tests can use for assertions.

While I was refactoring tests, I also took the opportunity to make sure our "no errors raised" assertions didn't end up being intrusive across the whole test suite.

### Tests

No new tests.

### DIFF
--- a/lib/tapioca/helpers/test/dsl_compiler.rb
+++ b/lib/tapioca/helpers/test/dsl_compiler.rb
@@ -1,0 +1,122 @@
+# typed: strict
+# frozen_string_literal: true
+
+require "tapioca/helpers/test/content"
+require "tapioca/helpers/test/isolation"
+require "tapioca/helpers/test/template"
+
+module Tapioca
+  module Helpers
+    module Test
+      module DslCompiler
+        extend T::Sig
+        extend T::Helpers
+
+        include Isolation
+        include Content
+        include Template
+
+        requires_ancestor { Kernel }
+
+        sig { params(compiler_class: T.class_of(Tapioca::Compilers::Dsl::Base)).void }
+        def use_dsl_compiler(compiler_class)
+          @context = T.let(CompilerContext.new(compiler_class), T.nilable(CompilerContext))
+        end
+
+        sig { params(compiler_classes: T.class_of(Tapioca::Compilers::Dsl::Base)).void }
+        def activate_other_dsl_compilers(*compiler_classes)
+          context.activate_other_dsl_compilers(compiler_classes)
+        end
+
+        sig { params(constant_name: T.any(Symbol, String)).returns(String) }
+        def rbi_for(constant_name)
+          context.rbi_for(constant_name)
+        end
+
+        sig { returns(T::Array[String]) }
+        def gathered_constants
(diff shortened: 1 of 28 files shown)


lib/tapioca/helpers/test/dsl_compiler.rb
```diff
@@ -0,0 +1,122 @@
...
+
+          sig { returns(T::Array[T.class_of(Tapioca::Compilers::Dsl::Base)]) }
+          attr_reader :other_compiler_classes
+
+          sig { params(compiler_class: T.class_of(Tapioca::Compilers::Dsl::Base)).void }
+          def initialize(compiler_class)
+            @compiler_class = compiler_class
+            @other_compiler_classes = T.let([], T::Array[T.class_of(Tapioca::Compilers::Dsl::Base)])
+            @compiler = T.let(nil, T.nilable(Tapioca::Compilers::Dsl::Base))
+            @pipeline = T.let(nil, T.nilable(Tapioca::Compilers::DslCompiler))
```
Morriar, line 66: Note for later: we need to rename `DslCompiler` -> `DslPipeline`

lib/tapioca/helpers/test/dsl_compiler.rb
```diff
@@ -0,0 +1,122 @@
...
+
+          sig { returns(T::Array[String]) }
+          def gathered_constants
+            compiler.processable_constants.map(&:name).compact.sort
+          end
+
+          sig { params(constant_name: T.any(Symbol, String)).returns(String) }
+          def rbi_for(constant_name)
+            # Make sure this is a constant that we can handle.
+            unless gathered_constants.include?(constant_name.to_s)
```
Morriar, line 87: Should we let `rbi_for` return an empty string for an ungathered/unprocessable constant?

```rb
assert_empty(rbi_for(unprocessable_constant))
```

spec/dsl_spec_helper.rb
```diff
@@ -3,72 +3,60 @@
...
+    @expecting_errors = false
   end
 
-  sig { void }
-  def teardown
-    super
-    T.unsafe(self).subject.errors.clear
+  after do
+    assert_empty(generated_errors) unless @expecting_errors
+    generated_errors.clear
```
Morriar, line 21: If I'm writing my own `after` in my test, doesn't that mean I can't assert anything about errors because the `generated_errors` will be cleared before my hook runs?

paracycle: You can write it in a `before_teardown` method. Alternatively, we can move this to an `after_teardown` method too, so that `after` blocks can still have access to the errors.

PR #794, opened 2022-02-04
Add favor-exported-gem-rbis flag for configuring how rbi merge conflicts are handled

joe-p-thomas:
### Motivation
Working on a project in which one Rails engine (A) has a dependency on another (B). A rbi file has been generated in engine B (using Parlour). But when that exported rbi file is merged with the Tapioca generated one, the generated file takes precedence when there are conflicts. I would like to be able to configure Tapioca to favor the exported rbi file.

### Implementation
Add a new config flag, `favor-exported-gem-rbis`. If this is set, the tree merge strategy for the exported and generate rbi files is swapped.

### Tests
A spec has been added.


joe-p-thomas:
It would give developers the power to create (and reference) rbi files they generate for their own internal gems.

In my case, I have a Rails engine that has "gateway" classes that other engines use to interact with it. The desire is to have every gateway method wrapped with some code (that will be used for tracking). To automate the method wrapping, we made use of meta-programming that re-defines and wraps the gateway methods. But the signatures are not copied over to the redefined methods. Ideally, we would be able to self generate rbi files that handle this scenario, and then export them from the engine.

--- 

Thought through this over the weekend. I am now thinking, instead of a boolean, this config could be a list of gems that this behavior is applied for. And then the rest of the gems still favor the signatures generated from the source (which I can understand is typically the better solution).

joe-p-thomas:
We are taking a different approach and this is no longer needed. Closing PR.

camertron:
I'm really surprised signatures present in a gem's /rbi folder don't take precedence. If someone has gone to the trouble of generating signatures in the /rbi folder of a gem as I have done for [kube-dsl](https://github.com/getkuby/kube-dsl), IMHO they should be honored over anything auto-generated.

@joe-p-thomas, what different approach did you take? I'm currently struggling with the same issue.

PR #795, opened 2022-02-04
Refactor DSL compilers

paracycle:
### Motivation

This is a breaking and backwards-incompatible change to how DSL compilers are structured. We are moving from creating a single instance of each compiler for the whole pipeline run to creating a new compiler instance for each constant that is processed. This allows the compiler interface to be cleaner (since there are less things that need to be passed to `decorate`) and the compiler to hold state if it wants. In particular, the fact that each compiler had to pass `constant` between its methods was a smell for me.

The general interface for a DSL compiler is changing from:
```ruby
class Foo < Tapioca::Dsl::Compiler
  sig { override.params(root: RBI::Tree, constant: T.class_of(FooDsl)).void }
  def decorate(root, constant)
  end

  def gather_constants
  end
end
```
to

```ruby
class Foo < Tapioca::Dsl::Compiler
  Elem  = type_member(fixed: T.class_of(FooDsl))

  sig { override.void }
  def decorate
  end

  def gather_constants
  end
end
```

### Implementation

1. Make the relevant change in the `Compiler` base class 
2. Adapt the compilers to adhere to the new structure
3. Change the `DslCompiler` helper to adapt to the new structure
4. Change how errors are collected now that compiler instances are created and destroyed dynamically.

### Tests

No added tests, refactored existing ones.


lib/tapioca/dsl/compilers.rb
```diff
@@ -0,0 +1,24 @@
...
+  module Dsl
+    module Compilers
+      DIR = T.let(
+        File.expand_path("compilers", __dir__),
+        String
+      )
+      NAMESPACES = T.let(
+        [
+          "#{name}::", # compilers in this namespace
+          "", # compilers that need to be fully namespaced
```
Morriar, line 18: Why is this needed?

paracycle: When resolving a compiler by name, we special case the compilers under the `Tapioca::Dsl::Compilers` namespace so that we can use their class names only. For all other compilers, they need to be fully qualified when passed by name. This array is the array of namespaces we search under given a compiler by name. Thus, this entry is the namespace that we expect 3rd party compilers to be found under. Maybe this should be `"::"`?

Morriar: ^ Can we add this answer as comment? 🙏

lib/tapioca/dsl/compiler.rb
```diff
@@ -9,45 +9,34 @@
...
   module Dsl
-    DSL_COMPILERS_DIR = T.let(File.expand_path("../compilers", __FILE__).to_s, String)
-
     class Compiler
       extend T::Sig
       extend T::Helpers
+      extend T::Generic
 
       include Reflection
+      extend Reflection
```
vinistock, line 18: Is it intentional that we're including and extending the same module? Is there any way we could consolidate these?

vinistock (approved):

lib/tapioca/dsl/compiler.rb
```diff
@@ -56,36 +45,37 @@ def compiler_enabled?(compiler_name)
...
+      def decorate; end
 
       sig { abstract.returns(T::Enumerable[Module]) }
-      def gather_constants; end
+      def self.gather_constants; end
+
+      sig { returns(T::Set[Module]) }
+      def self.processable_constants
+        @processable_constants ||= T.let(
+          Set.new(gather_constants).tap(&:compare_by_identity),
```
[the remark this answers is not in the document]
paracycle: I think when I originally wrote this, I thought that `compare_by_identity` didn't return `self`, so that's why. I can update this.

lib/tapioca/dsl/compiler.rb
paracycle: This is intentional, since we want `gather_constants` methods to have access to `Reflection` methods too, but that method is on the singleton class now.

paracycle: Not sure if there is a better way to do this, though.

vinistock: Yeah, I guess it'll be hard to escape this. Just checking.

Morriar (approved):

paracycle:
> It looks like the PR description has a small typo in the "new" code. I think it should be:
> 
> ```diff
> -def gather_constants
> +def self.gather_constants
> ```
> 
> Might be worth updating for posterity since (like I did) some folks may come to this PR to see how to refactor their custom DSL compilers.

Thanks for pointing this out @jeffcarbs, fixed the code in the description.

ericfirth:
When I did this, I started getting this error in my DSL Generators

```
sorbet/tapioca/generators/devise_methods.rb:10: Type ConstantType declared by parent Tapioca::Dsl::Compiler must be re-declared in DeviseMethods https://srb.help/5014
    10 |class DeviseMethods < Tapioca::Dsl::Compiler
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    sorbet/rbi/gems/tapioca@0.7.0.rbi:586: ConstantType declared in parent here
     586 |  ConstantType = type_member(upper: Module)
```

If I added `ConstantType = type_member(upper: Module)` to the Generator it worked, but I don't really know what that is, why I have to do it and why its not on the above example. Curious if anyone knows.

Morriar:
👋 Hey Eric,

We renamed `Elem` into `ConstantType` in https://github.com/Shopify/tapioca/pull/819.

I believe this is enough:

```rb
class Foo < Tapioca::Dsl::Compiler
  ConstantType = type_member(fixed: T.class_of(FooDsl))

  sig { override.void }
  def decorate
  end

  def self.gather_constants
  end
end
```

Hope this helps.

paracycle:
I updated the example in the description to reflect that the type member name is `ConstantType`.

PR #796, opened 2022-02-07
CLI commands refactor

paracycle:
### Motivation

What we have been calling `Generators` are actually `Commands` that we execute from the `CLI` handler class.

### Implementation

Rename namespace and base class. Encapsulate some default keyword arguments passed to the base instance.

### Tests

No new tests


lib/tapioca/commands.rb
```diff
@@ -0,0 +1,9 @@
+# typed: true
+# frozen_string_literal: true
+
+require_relative "commands/command"
```
vinistock, line 4: Would it make sense to autoload these classes? We probably don't need to have all command classes required at all times, since we can only run one at a time.

lib/tapioca/commands/command.rb
```diff
@@ -16,17 +16,21 @@ class FileWriter < Thor
...
       private
 
+      sig { params(command: Symbol, args: String).returns(String) }
+      def default_command(command, *args)
+        [Tapioca::DEFAULT_COMMAND, command.to_s, *args].join(" ")
+      end
+
+      sig { returns(Thor::Actions) }
+      def file_writer
+        FileWriter.new
```
Morriar, line 31: Should we memoize this?

paracycle: Oh yeah, right. Good point.

lib/tapioca/commands.rb
paracycle: Done

lib/tapioca/commands/command.rb
paracycle: Done

vinistock (approved):

PR #810, opened 2022-02-10
Bump i18n to 1.9.1 as 1.9.0 was yanked

nvasilevski:
### Motivation

https://rubygems.org/gems/i18n

`1.9.0` was yanked for some reason

```
Fetching gem metadata from https://rubygems.org/.........
Your bundle is locked to i18n (1.9.0) from rubygems repository https://rubygems.org/ or installed locally, but that version can no longer be found in
that source. That means the author of i18n (1.9.0) has removed it. You'll need to update your bundle to a version other than i18n (1.9.0) that hasn't
been removed in order to install.
```

So I'm upgrading it to 1.9.1
Just a development change

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -149,7 +149,7 @@ GEM
     globalid (1.0.0)
       activesupport (>= 5.0)
     google-protobuf (3.19.3)
-    i18n (1.9.0)
+    i18n (1.9.1)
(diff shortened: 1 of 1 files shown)


nvasilevski:
🤦 
My `main` branch was way outdated 
Sorry!

PR #811, opened 2022-02-11
Handle attribute_method_matchers rename to attribute_method_patterns

nvasilevski:
### Motivation

Tapioca is failing on Rails edge because of the method rename
https://github.com/rails/rails/pull/44367/files

### Implementation

If I'm not mistaken, we can't to rely on any particular Rails version in this case so taking `respond_to?` approach similar to
https://github.com/Shopify/tapioca/blob/97b2790d9b2bd6700d3ebebbd0d0726a7abf398d/lib/tapioca/dsl/compilers/active_model_attributes.rb#L79-L85

I'm also renaming references to values from `matcher` to `pattern` according to the new terminology. Let me know if you think this is unnecessary.

### Tests

I'm going to run tests against Rails edge but I don't think there is any test I can add to existing ones?


nvasilevski:
Requires some work to properly handle `AttributeMethodPattern` constant rename

Morriar (approved):
I took the liberty to push another commit to inline the, now shorter, signature on `handle_method_pattern?`.

Thanks 👍

KaanOzkan (approved):

Issue #429, opened 2021-08-12
Make require.rb less necessary


paracycle:

vinistock:
Because we sort the requires, it may sometimes result in errors if they need a specific order. Currently, it's a confusing experience for developers, since they often just see a message about missing constants and don't realize what's happening. Ideas for improving the experience:

- Better documentation around `tapioca require`
- Rescue `LoadError` and provide a better tailored error message, mentioning that their `require.rb` file has a mistake and maybe even pointing them to documentation about how to fix it

rafaelfranca:
Can you please explain which problems people have with the require.rb file?

Morriar:
> Can you please explain which problems people have with the `require.rb` file?

I see two main sources of confusion:

1. When a constant or method is missing it generally requires an understanding of the gems internals to know what to require. It puts burden on the user that has to go read the gem's source to find where the constant is declared and which file should be added to Tapioca's `require.rb` file.

2. When the `require.rb` file is auto-populated with `tapioca requires` and an entry is incorrect, users have a hard time figuring what should be removed from the file and why. We've seen occurrences where people were regenerating the `require.rb` file between each gem bump, re-adding the same erroneous requires over and over for example.

[Vini's comment](https://github.com/Shopify/tapioca/issues/429#issuecomment-903905121) explains one more limitation, the ordering of the requires that might also be problematic.

All of this shows that the current approach, while working, might not be the easiest one from a user point of view.

dirceu:
@Morriar you and Ufuk have a branch somewhere with a prototype for this, right? If so, can you link it here?

Morriar:
> @Morriar you and Ufuk have a branch somewhere with a prototype for this, right? If so, can you link it here?

It's very WIP but it's here: https://github.com/Shopify/tapioca/commit/e6e1dba26a3a63c2c33c71732f6073da00eb05d6.

KaanOzkan:
We've prototyped the existing autoload solution linked above and gathered some timings. It was clear that using sorbet's symbol table to identify constants to load was too slow. An early prototype using [RubyIndexer](https://github.com/Shopify/ruby-lsp/blob/main/lib/ruby_indexer/lib/ruby_indexer/index.rb) showed much faster results.

Prototype also revealed that behaviour of the autoload has slightly changed because we're running into stack overflow during loading itself. This wasn't the case when the solution was tested the first time.

Next steps are:
[...]

PR #818, opened 2022-02-16
Need less entries in `require.rb`

paracycle:
### Motivation

Contributes to #429

As things stand right now, we need to add all gems that are marked as `require: false` in the `Gemfile` to the `tapioca/require.rb` file manually. If we made Bundler load those gems by default, we would not need any of those entries.

### Implementation

We are able to get Bundler to load gems by tricking it to think that gems marked as `require: false` are actually `require: true`. We do that by intercepting the `autorequire` attr reader and changing all `[]` (which mean `require: false`) values to `nil` (which mean `require: true`).

### Tests

Added a test with a gem explicitly marked as `require: false`, and verified that the gem is loaded and Tapioca can generate an RBI for it without needing a `tapioca/require.rb` entry.


KaanOzkan (commented):
Will this work with gems that aren't supposed to be loaded by default? For example [fakefs](https://github.com/fakefs/fakefs)?

An alternative solution to having those gems as `require: false` would be to pass them into the `exclude` config. I took a brief look and feels like the `exclude` might be handled [after loading all the dependencies with bundler](https://github.com/Shopify/tapioca/blob/e48d2f9ea7ca89492a60208e692771b1feeb2ba4/lib/tapioca/commands/gem.rb#L67)? If that's the case I don't see how we can circumvent `fakefs` issue with this feature.

paracycle:
@KaanOzkan We implemented the exclusion of excluded gems from being forcibly required and updated the test to check for that too. Do you mind re-reviewing?

KaanOzkan (approved):

PR #819, opened 2022-02-16
Rename `Dsl::Compiler::Elem` into `ConstantType`

Morriar:
### Motivation

Make it more obvious what this type member is about.


paracycle (approved):
Thanks for fixing my bad.

PR #821, opened 2022-02-16
Load nothing from Tapioca by default

paracycle:
### Motivation

The split between top level `tapioca.rb` requires and `tapioca/internal.rb` requires was creating an ordering problem. Also, the default require of Tapioca might be pulling into the requiring much more than we intended, and since we do messy things with Sorbet runtime, that could end up being distruptive.

### Implementation

It is best to not load anything from the default require (we basically define the top-level `Tapioca` namespace and some constants under it, only) and to load everything for the CLI using `tapioca/internal.rb`. This allows us to order the requires so that all the trackers and other runtime patches get loaded before things that depend on them.

**NOTE:** This implies that people who write their own DSL compilers will have to explicitly do a `require "tapioca/dsl/compiler"`.

### Tests

No new tests.

### DIFF
--- a/lib/tapioca/sorbet_ext/generic_name_patch.rb
+++ b/lib/tapioca/sorbet_ext/generic_name_patch.rb
@@ -57,18 +57,6 @@ module T
   module Types
     class Simple
       module GenericPatch
-        def valid?(obj)
-          # Since `Tapioca::TypeVariable` is a `Module`, it will be wrapped by a
-          # `Simple` type. We want to always make type variable types valid, so we
-          # need to explicitly check that `raw_type` is a `Tapioca::TypeVariable`
-          # and return `true`
-          if defined?(Tapioca::TypeVariableModule) && Tapioca::TypeVariableModule === @raw_type
-            return true
-          end
-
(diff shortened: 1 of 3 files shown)


Morriar (approved):
❤️

lib/tapioca/sorbet_ext/generic_name_patch.rb
```diff
@@ -87,10 +75,38 @@ def name
...
 module Tapioca
+  class TypeVariable < ::T::Types::TypeVariable
+    def initialize(type_variable_module)
+      @name = type_variable_module.name
+      super(type_variable_module.variance)
+    end
+
+    def name
+      @name
+    end
```
Morriar, line 107:
```suggestion
    attr_reader :name
```

lib/tapioca/sorbet_ext/generic_name_patch.rb
```diff
@@ -87,10 +75,38 @@ def name
 
       prepend GenericPatch
     end
+
```
Morriar, line 78:
```suggestion
```

PR #833, opened 2022-02-28
Set default if `factory` prop is set

jeffcarbs:
### Motivation

Explained more in https://github.com/Shopify/rbi/pull/122, but basically `T::Struct` fields have a `factory` option, which is basically the same as `default` but takes a proc so the value can be set dynamically. Like default, if a prop/const has a factory set you don't have to pass a value in the initializer.

Tapioca does not handle the `factory` option so a field with that set would seem to be required in the initializer when it actually isn't. This has just come up for us with our first usage of this option in one of our internal gems so we've had to monkey-patch tapioca for now.

### Implementation

I noticed that when we see `immutable: true` we put a Const, even if the field may have been defined via `prop :x, Integer, immutable: true`. It gives the same effect to the type checker while not necessarily being identical to the source code.

I took the same approach here for factory. For type-checking purposes, we just want sorbet to know there's a default set so it will be optional in the initializer. So where we were previously checking if a `default` was defined and setting `default: T.unsafe(nil)` in the RBI, we now also check for `factory` and do the same (set `default: T.unsafe(nil)` in the RBI).

### Tests

Added a test that was red before and green after.


paracycle (approved):
Thanks, this looks reasonable.

PR #837, opened 2022-03-01
Make DSL generation also autocorrect the gems RBI strrictnesses

Morriar:
### Motivation

We introduced this feature in #801 but triggered it only when running `tapioca gem`.

I think it also makes sense to trigger the autocorrect of strictness after a `tapioca dsl` since the generated DSL files might clash with the generated gem files.

### Implementation

* Renamed `RBIHelper` into `SignaturesHelper`
* Extracted `update_gem_rbis_strictnesses` into a new `RBIHelper`
* Added a call to `update_gem_rbis_strictnesses` right after the DSL files generation

### Tests

See added tests.


spec/tapioca/cli/dsl_spec.rb
```diff
@@ -1180,6 +1228,58 @@ class Post
           refute_success_status(result)
         end
       end
+
+      describe "strictness" do
+        it "must turn the strictness of gem RBI files with errors to false" do
+          @project.require_real_gem("smart_properties", "1.15.0")
+          @project.bundle_install
+          # @project.tapioca("init")
```
vinistock, line 1236: Leftovers from debugging?

vinistock (approved):

paracycle (approved):

PR #839, opened 2022-03-01
Fix generic type variable argument serialization

### LINKED ISSUE #838: Invalid type_member inserted into RBI file
We noticed that tapioca is not properly replicating generics within generic type_members. In one of our gems we have a generic like this:

```ruby
InputType = type_member(fixed: Input)
ReturnType = type_member(fixed: Result[OkType, ErrType])
```

in the rbi file generated for this type in a gem which depends on it, the output is:

```ruby
InputType = type_member(fixed: Input)
ReturnType = type_member(fixed: #<Class:0x00007fb9d58fdce8>)
```

It appears to be placing an anonymous class string in the position of the generic type instead of it's actual type.

paracycle:
### Motivation

Fixes #838 

### Implementation

The `GenericNamePatch` is [using string interpolation to serialize the names of `fixed`, `lower` and `upper` arguments](https://github.com/Shopify/tapioca/blob/9297b51bcf560c1416f27c0523a6cd15822cc59d/lib/tapioca/sorbet_ext/generic_name_patch.rb#L158-L160) but we have been only overriding the `name` method on generic types. Since generic types that we create are actually anonymous, calling `to_s` on them returns something like `#<Class:0x00007fb9d58fdce8>` which is both unexpected and wrong.

The implementation makes sure that whoever calls `to_s` on a generic type also gets the expected generic type name.

### Tests

Add a failing test based on the report in #838.

### DIFF
--- a/lib/tapioca/runtime/generic_type_registry.rb
+++ b/lib/tapioca/runtime/generic_type_registry.rb
@@ -111,8 +111,10 @@ module Tapioca
             constant.clone
           end
 
-          # Let's set the `name` method to return the proper generic name
-          generic_type.define_singleton_method(:name) { name }
+          # Let's set the `name` and `to_s` methods to return the proper generic name
+          name_proc = -> { name }
+          generic_type.define_singleton_method(:name, name_proc)
+          generic_type.define_singleton_method(:to_s, name_proc)
 
           # We need to define a `<=` method on the cloned constant, so that Sorbet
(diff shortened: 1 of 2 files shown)


vinistock (approved):

kbacha (approved):
Wow! Thank you for the quick turnaround!

Morriar (approved):

PR #841, opened 2022-03-04
Add an easy way to print `ExecResult` in tests

Morriar:
### Motivation

Just a shortcut so we can do:

```rb
puts result
```

Instead of

```rb
puts result.out
puts result.err
puts result.status
```

When debugging tests.


paracycle (approved):
This will be super helpful indeed.

PR #842, opened 2022-03-04
Check generated RBI files for correctness

### LINKED ISSUE #759: Catch invalid RBIs during the generation
In case we generate incorrect RBIs we could stop/skip during the generation and explain to the user that there was an internal issue and how to report an issue instead of producing the invalid RBIs and confuse the developers when then run `srb tc`.

This may happen in the `rbi` API that is typed with `String` like module, constant, method, param invalid names or invalid types inside `params` or `returns` for example.

With decent regexes we should be able to catch most of these with different approaches:

1. Having a single point to produce such strings for names and types inside Tapioca where we can check them (would involve a lot of refactor in tapioca)
2. Having this done in `rbi` and we create the node instance with a bad value (would be a more generic solution, but we need to handle exceptions properly in Tapioca)
3. Having this done in `rbi` with static typing with new types like `ClassName`, `MethodName` and `Type` (which is basically 2. with extra steps with a more localized exception handling for the client).
4. A combination of 1. and (2. or 3.)

Morriar:
### Motivation

Catches parse errors in generated RBI files and display a helpful error message:

image

Closes #759.

### Implementation

Since we already run `sorbet` on generated files to catch redefinition errors and change the strictness of generated files, I decided to hook on this process to check if we find parse errors (code < 4000) as well.

### Tests

See automated tests.


lib/tapioca/helpers/rbi_helper.rb
```diff
@@ -28,10 +35,60 @@ def update_gem_rbis_strictnesses(gem_names, gem_dir: DEFAULT_GEM_DIR, dsl_dir: D
...
+          Errors:
+          #{parse_errors.map { |error| "  #{error}" }.join("\n")}
+
+          ##########################
+
+        ERR
+      end
+
+      redef_errors = errors.select { |error| error.code == 4010 }
+      update_gem_rbis_strictnesses(redef_errors, gems.map(&:name), gem_dir) if auto_strictness
```
paracycle, line 83: I think the following could be clearer:
```suggestion
      if auto_strictness
        redef_errors = errors.select { |error| error.code == 4010 }
        update_gem_rbis_strictnesses(redef_errors, gems.map(&:name), gem_dir) 
      end
```

lib/tapioca/commands/dsl.rb
```diff
@@ -115,8 +115,13 @@ def execute
...
           if @auto_strictness
             say("")
-            update_gem_rbis_strictnesses([], gem_dir: @gem_dir, dsl_dir: @outpath.to_s)
-            say("")
+            validate_rbi_files(
+              command: default_command(:dsl, @requested_constants.join(" ")),
+              gem_dir: @gem_dir,
+              dsl_dir: @outpath.to_s,
+              auto_strictness: @auto_strictness,
+              compilers: Runtime::Reflection.descendants_of(Tapioca::Dsl::Compiler)
```
paracycle, line 123: I think, we should be able to ask the pipeline exactly which compilers were enabled instead:
```suggestion
              compilers: pipeline.compilers
```

lib/tapioca/commands/dsl.rb
```diff
@@ -115,8 +115,13 @@ def execute
 
           if @auto_strictness
             say("")
-            update_gem_rbis_strictnesses([], gem_dir: @gem_dir, dsl_dir: @outpath.to_s)
-            say("")
+            validate_rbi_files(
+              command: default_command(:dsl, @requested_constants.join(" ")),
```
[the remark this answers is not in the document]
paracycle: I think this is good enough.

paracycle (approved):
Nice! Just a couple of comments.

lib/tapioca/helpers/rbi_helper.rb
```diff
@@ -28,10 +35,62 @@ def update_gem_rbis_strictnesses(gem_names, gem_dir: DEFAULT_GEM_DIR, dsl_dir: D
       errors = Spoom::Sorbet::Errors::Parser.parse_string(res.err)
 
       if errors.empty?
-        say("No error found", [:green, :bold])
+        say("  No error found\n\n", [:green, :bold])
```
paracycle, line 38:
```suggestion
        say("  No errors found\n\n", [:green, :bold])
```

PR #847, opened 2022-03-09
Do not crash when trying to generate DSL RBIs on non-rails apps

Morriar:
### Motivation

Avoid getting a scary stack trace when running `tapioca dsl` in the wrong kind of app:

```
$ bundle exec tapioca dsl
Loading Rails application... Done
Loading DSL compiler classes... Done
Compiling DSL RBI files...

bundler: failed to load command: tapioca
tapioca/lib/tapioca/dsl/compilers/url_helpers.rb:112:in `gather_constants': undefined method `routes' for nil:NilClass (NoMethodError)
```

### Implementation

Check if we are in a Rails app with `Rails.application`.

### DIFF
--- a/spec/tapioca/dsl/compilers/active_record_fixtures_spec.rb
+++ b/spec/tapioca/dsl/compilers/active_record_fixtures_spec.rb
@@ -8,8 +8,34 @@ module Tapioca
     module Compilers
       class ActiveRecordFixturesSpec < ::DslSpec
         describe "Tapioca::Dsl::Compilers::ActiveRecordFixtures" do
-          describe "initialize" do
+          describe "without a Rails app" do
+            it "gathers nothing if not in a Rails application" do
+              add_ruby_file("post_test.rb", <<~RUBY)
+                class PostTest < ActiveSupport::TestCase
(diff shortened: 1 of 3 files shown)


KaanOzkan (approved):

vinistock:
Should we print a message when we don't have a Rails application? My understanding of what will happen is that we're simply not going to process any constants, but maybe we could print a message saying `DSL generators are not yet supported for non-Rails applications`.

vinistock (approved):

spec/tapioca/dsl/compilers/active_record_fixtures_spec.rb
```diff
@@ -8,8 +8,8 @@ module Dsl
     module Compilers
       class ActiveRecordFixturesSpec < ::DslSpec
         describe "Tapioca::Dsl::Compilers::ActiveRecordFixtures" do
-          describe "initialize" do
-            it "gathers only the ActiveSupport::TestCase base class" do
+          describe "without a Rails app" do
```
paracycle, line 11: 🙏

PR #850, opened 2022-03-10
Update `tapioca todo` to handle constants defined in `class << self`

jeffcarbs:
### Motivation

I noticed constants defined inside `class << self`

### Implementation

Constants defined in `class << self` manifest as `T.class_of(TopLevelClass)::Foo` and there was an explicit early return for these cases so I'm not sure if I'm missing something.

### Tests

Updates the main "creates a list of all unresolved constants" test case with some more complicated nesting to reproduce the failure and then fix it.


lib/tapioca/commands/todo.rb
```diff
@@ -74,10 +74,10 @@ def unresolved_constants
           .each_line
           .map do |line|
             next if line.include?("<")
-            next if line.include?("class_of")
-            line.strip.gsub("T.untyped::", "")
+            line.strip.gsub("T.untyped::", "").gsub(/T\.class_of\(([:\w]+)\)/, '\1')
           end
           .compact
+          .sort
```
jeffcarbs, line 80: I added this `sort` so that the tests would produce deterministic results. Happy to remove it if you don't think it's correct to sort these.

lib/tapioca/commands/todo.rb
```diff
@@ -74,10 +74,10 @@ def unresolved_constants
           .each_line
           .map do |line|
             next if line.include?("<")
-            next if line.include?("class_of")
-            line.strip.gsub("T.untyped::", "")
+            line.strip.gsub("T.untyped::", "").gsub(/T\.class_of\(([:\w]+)\)/, '\1')
```
paracycle, line 77: Nitpick: Can we have each `gsub` on its own line and add a comment for each line about what the processing is that they are doing?

Personally, I find that I need to really think about regexps to understand what they are doing and I would have still not understood why they are needed in the first place.

jeffcarbs: Will do 👍

Do you know what the gsub for `T.untyped::` is doing? It was there before and after taking it out the whole suite still passes so I'm not sure what it's doing. Could've been something that has since changed in sorbet 🤔 cc @Morriar since I think you had initially added it.

paracycle: Yeah, maybe @Morriar would remember but I have no idea why we are stripping `T.untyped::` tbh.

Thanks for making the change.

paracycle (approved):

jeffcarbs:
Should be good to go 👍

PR #864, opened 2022-03-15
allow for user-created dsl extensions

deecewan:
### Motivation

I have a need to patch a class so I can track method calls in some DSL. I posted [a message in slack](https://sorbet-ruby.slack.com/archives/C02VD06EQ3U/p1647317966165499?thread_ts=1647312541.927039&cid=C02VD06EQ3U) about it.

The way I'm doing it currently is to have the patch at the top of my compiler, and then re-load the file the constant was defined in. This can be problematic.

This brings extensions into line with compilers, which have a dedicated location in `@tapioca_path` to be stored and loaded.

### Implementation

Copied the code from the `load_dsl_compilers` method, mostly

### Tests

No tests because this functionality doesn't seem to be tested already. I can add tests if they're really wanted.

### DIFF
--- a/lib/tapioca/commands/dsl.rb
+++ b/lib/tapioca/commands/dsl.rb
@@ -378,7 +378,12 @@ module Tapioca
 
       sig { void }
       def load_dsl_extensions
-        Dir["#{__dir__}/../dsl/extensions/*.rb"].sort.each { |f| require(f) }
+        Dir.glob([
+          "#{__dir__}/../dsl/extensions/*.rb",
+          "#{@tapioca_path}/dsl_extensions/**/*.rb",
+        ]).each do |extension|
+          require File.expand_path(extension)
+        end
       end
     end
   end


deecewan:
I've signed the CLA now but can't re-run that check

lib/tapioca/commands/dsl.rb
```diff
@@ -378,7 +378,12 @@ def generate_command_for(constant)
 
       sig { void }
       def load_dsl_extensions
-        Dir["#{__dir__}/../dsl/extensions/*.rb"].sort.each { |f| require(f) }
+        Dir.glob([
+          "#{__dir__}/../dsl/extensions/*.rb",
+          "#{@tapioca_path}/dsl_extensions/**/*.rb",
```
Morriar, line 383: I'm ok with the approach but the name bothers me as it's not matching what we do for `compilers`: https://github.com/Shopify/tapioca/blob/1ae637362417034cf4007e37758fe5fe1263f086/lib/tapioca/commands/dsl.rb#L158-L159

@paracycle should we move the compilers under `sorbet/tapioca/dsl/compilers` so extensions can be under `sorbet/tapioca/dsl/extensions`?

paracycle: With the new loader API, all of this will be gone, tbh, but yeah, I think it is better to have this be:
```suggestion
          "#{@tapioca_path}/dsl/extensions/**/*.rb",
```
instead.

deecewan: I wanted to go with just `extensions` but i thought that may be confusing.

@paracycle i remember seeing talk of the new loader API. is there anywhere I can get more info on it? I've made it work without this patch in my project, so can definitely hold out until the new API drops.

deecewan:
closing, given there is a new loader api coming

PR #866, opened 2022-03-15
Bump `rbi` to v0.0.14

Morriar:
### Motivation

Get a patch for parsing empty `params()` and `returns()` calls as well as better error handling.


vinistock:
Do we want to restrict this in the gemspec?

PR #869, opened 2022-03-18
Start filtering old copies of reloaded constants

### LINKED ISSUE #867: Output a warning when model use in initializers results in duplicate class loading
If a model is used in an initializer, that results in `tapioca dsl` processing the same file/constant twice, changing it each time, and causing a failure in `--verify`. Root cause is that Rails will load the constant during initialization, and then another (same but different) version of the constant later on. Even `ObjectSpace.each_object` will return doubles of the class, and they share no equality at all (not even is_a? or kind_of?).

Per @paracycle: "loading models in initializers is actually a bad practice for many reasons and I believe Rails 7 is stricter about it as well. This is indeed a strange gotcha for Tapioca, which we can make better, but it does point to a problem in the app." In general, models should not be used in initializers. Best practice is to either refer to classes as a string, or wrap the configuration in a `to_prepare` block.

While this is rails best practice, it won't necessarily cause problems anywhere else, and the problem (model use in initializer) is so disconnected from tapioca that it's almost impossible to diagnose. Ideally, tapioca would de-dupe dsl generation based on the file it resolves to. If that's not possible or desireable, adding a warning would be second-best. If possible, the warning should specify which constant is double-loaded.

Warning text could be something like "Class #{dup_class} is loaded to ruby twice, which may cause problems with tapioca. If the class is used in an initializer, change the reference to be a string instead of a constant, or wrap the code in a `to_prepare` block."

paracycle:
### Motivation

Fixes #867 

### Implementation

I decided to show a warning for multiple (different) constants with the same name, and choose the copy that is reachable via `const_get` as the definitive version to generate an RBI for, since we know that is the latest one that was loaded into memory.

I explicitly decided against erroring and refusing to generate anything, since, in some cases, it ends up being hard to figure out what is holding onto the old copies of a reloaded constant. Users can still get the RBI files generated while they are aware of a problem.

### Tests

Added a test to reload a constant with different properties to recreate the problem scenario.


KaanOzkan (approved):

KaanOzkan (commented):
There's a test failure, also can we fix this typo https://github.com/Shopify/tapioca/blob/6262f8758f4e8b24f4797e7a306bd9ba04b0dee3/spec/tapioca/cli/dsl_spec.rb#L537

paracycle:
> There's a test failure, also can we fix this typo
> 
> https://github.com/Shopify/tapioca/blob/6262f8758f4e8b24f4797e7a306bd9ba04b0dee3/spec/tapioca/cli/dsl_spec.rb#L537

Thanks, I fixed both!

PR #877, opened 2022-03-29
Start using compared by identity Set operations for `UrlHelpers#includes_helper?`

paracycle:
### Motivation

    
When doing a difference of arrays for finding interesting ancestors, we need to be careful in case a constant has overridden `Object#hash`. Since array subtractions use `Object#hash` to determine equality, this operation will fail.

### Implementation

Instead, we need to use a Set with compare-by-identity semantics, so that `Object#hash` is never called.

### DIFF
--- a/lib/tapioca/dsl/compiler.rb
+++ b/lib/tapioca/dsl/compiler.rb
@@ -54,7 +54,7 @@ module Tapioca
       sig { returns(T::Set[Module]) }
       def self.processable_constants
         @processable_constants ||= T.let(
-          Set.new(gather_constants).tap(&:compare_by_identity),
+          T::Set[Module].new(gather_constants).compare_by_identity,
           T.nilable(T::Set[Module])
         )
         T.must(@processable_constants)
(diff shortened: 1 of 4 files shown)


lib/tapioca/dsl/compiler.rb
```diff
@@ -54,7 +54,7 @@ def self.gather_constants; end
       sig { returns(T::Set[Module]) }
       def self.processable_constants
         @processable_constants ||= T.let(
-          Set.new(gather_constants).tap(&:compare_by_identity),
+          T::Set[Module].new(gather_constants).compare_by_identity,
```
Morriar, line 57: Is this change relevant in this commit?

vinistock (approved):

sorbet/rbi/shims/set.rbi
```diff
@@ -1,5 +1,6 @@
 # typed: strict
 
 class Set
+  sig { returns(T.self_type) }
```
Morriar, line 4: Should we upstream this change? https://github.com/sorbet/sorbet/blob/master/rbi/stdlib/set.rbi#L174-L178

PR #879, opened 2022-04-01
Check duplicates between shims and Sorbet's RBI payload

Morriar:
### Motivation

Check duplicates between shims files and Sorbet's payload.

It's frequent that we locally fix Sorbet's payload in our shims waiting for the same patch to be merged upstream. The problem is that after a while we forget the local fixes and they stay there for ever. This can even become dangerous when the definition in the payload is later updated and we keep the shadowing local definition.

This PR introduces a way to check for duplicates between the local shims and Sorbet's payload.

### Implementation

This feature uses Sorbet's `--print=payload_sources` to dump the RBIs from its payload so we can index them in the same way we do for gems RBIs and DSL RBIs.

I gated the feature behind a specific Sorbet version like we did for `to_ary_nil_support`.

### Tests

See automated tests.

### DIFF
--- a/sorbet/rbi/shims/bundler.rbi
+++ b/sorbet/rbi/shims/bundler.rbi
@@ -1,6 +1,0 @@
-# typed: true
-
-class Bundler::StubSpecification
-  sig { returns(T::Boolean) }
-  def default_gem?; end
-end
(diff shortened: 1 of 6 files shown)


sorbet/rbi/shims/bundler.rbi
```diff
@@ -1,6 +0,0 @@
-# typed: true
-
-class Bundler::StubSpecification
-  sig { returns(T::Boolean) }
-  def default_gem?; end
```
Morriar, line 5: Found by the feature itself: https://github.com/sorbet/sorbet/blob/master/rbi/stdlib/bundler.rbi#L9957-L9958

vinistock (approved):
Very nice!

PR #885, opened 2022-04-05
Remove support for Ruby 2.6

Morriar:
### Motivation

Ruby 2.6 is [EOL since 2022-03-31](https://www.ruby-lang.org/en/downloads/branches/#:~:text=EOL%20date%3A%202022%2D03%2D31).

Removing support for it allows us to clean some code and enforce good practices in projects using Tapioca.

### Implementation

Also updated `bind(X).call(Y)` to `bind_call(X, Y)`.

### Tests

No functional change.


amomchilov (approved):

egiurleo (approved):

spec/tapioca/dsl/compilers/action_mailer_spec.rb
```diff
@@ -120,9 +120,7 @@ def notify_customer(customer_id); end
                 class NotifierMailer < ActionMailer::Base
                   delegate :notify_customer, to: :foo
 
-                <% if ruby_version(">= 2.7.0") %>
                   module_eval("def notify_admin(...); end")
```
paracycle, line 123: A bit nitpicky, but we shouldn't need the `module_eval` here anymore either, since I believe it was there so that Sorbet would not freak out or something:
```suggestion
                  def notify_admin(...); end
```

paracycle (approved):
God's work, thank you 🙏

MikeMcQuaid:
Hello lovely Sorbet people, thanks for all your great work on this project.

I'm the @Homebrew project leader and have a completely unreasonable request: would you consider keeping Ruby 2.6 support around until a version of macOS ships with a newer (or no) Ruby? We rely on this system version for Homebrew and we're also heavy users of Sorbet.

Feel free to say no and: thanks again for the project!

MikeMcQuaid:
> Do you mean you need to run Tapioca on Ruby 2.6?

@Morriar Yes, sorry!

Issue #881, opened 2022-04-04
Spoom Encoding Issue


ryanquanz:
when running `bin/tapioca gem activerecord` on Rails `7.0.2.3-172dfa74f861c6bfe6ee9ba438dd4094fc1a4401` I get the following error:

```
Requiring all gems to prepare for compiling...  Done

Removing RBI files of gems that have been removed:

  Nothing to do.

  Compiled activerecord
       force  sorbet/rbi/gems/activerecord@7.0.2.3-172dfa74f861c6bfe6ee9ba438dd4094fc1a4401.rbi

Typechecking RBI files...  Done
/Users/ryanquanz/.gem/ruby/3.1.1/gems/spoom-1.1.9/lib/spoom/sorbet/sigils.rb:69:in `write': "\\xC3" from ASCII-8BIT to UTF-8 (Encoding::UndefinedConversionError)
	from /Users/ryanquanz/.gem/ruby/3.1.1/gems/spoom-1.1.9/lib/spoom/sorbet/sigils.rb:69:in `change_sigil_in_file'
	from /Users/ryanquanz/.gem/ruby/3.1.1/gems/sorbet-runtime-0.5.9837/lib/types/private/methods/_methods.rb:272:in `bind_call'
	from /Users/ryanquanz/.gem/ruby/3.1.1/gems/sorbet-runtime-0.5.9837/lib/types/private/methods/_methods.rb:272:in `block in _on_method_added'
	from /Users/ryanquanz/.gem/ruby/3.1.1/gems/tapioca-0.7.1/lib/tapioca/commands/gem.rb:425:in `block in update_strictnesses'
	from /Users/ryanquanz/.gem/ruby/3.1.1/gems/tapioca-0.7.1/lib/tapioca/commands/gem.rb:424:in `each'
	from /Users/ryanquanz/.gem/ruby/3.1.1/gems/tapioca-0.7.1/lib/tapioca/commands/gem.rb:424:in `update_strictnesses'
	from /Users/ryanquanz/.gem/ruby/3.1.1/gems/sorbet-runtime-0.5.9837/lib/types/private/methods/_methods.rb:272:in `bind_call'
	from /Users/ryanquanz/.gem/ruby/3.1.1/gems/sorbet-runtime-0.5.9837/lib/types/private/methods/_methods.rb:272:in `block in _on_method_added'
[...]

PR #886, opened 2022-04-05
Bump spoom to v1.1.11

Morriar:
### Motivation

Fixes https://github.com/Shopify/tapioca/issues/881.
Closes https://github.com/Shopify/tapioca/pull/882.


egiurleo (approved):

vinistock:
Do we want to change the `gemspec` requirement? Otherwise, the issue is only fixed if folks upgrade Spoom on their own.

vinistock (approved):

PR #889, opened 2022-04-07
Fairer queueing

egiurleo:


tapioca.gemspec
```diff
@@ -24,6 +24,7 @@ Gem::Specification.new do |spec|
   spec.metadata["allowed_push_host"] = "https://rubygems.org"
 
   spec.add_dependency("bundler", ">= 1.17.3")
+  spec.add_dependency("parallel")
```
egiurleo, line 27: I thought I commented before requesting review but I never submitted them 🤦🏻‍♀️ 

@morriar can you explain what the difference is between adding a dependency in the Gemfile vs in the gemspec? I assumed parallel went in the gemspec because it's now a required dependency to actually run the gem, but I'm not sure I really understand the distinction.

Also, do you think I should specify a version here?

paracycle: Yes, what you did is correct. The top level idea is this:

1. A gem's `Gemfile` is only for doing development on the gem source code itself. It only brings in things that will be needed as you are working on the gem. Part of the things it brings in are [the items listed inside the `.gemspec` file](https://github.com/Shopify/tapioca/blob/main/Gemfile#L5), since (as I'll mention in the next section) those are also needed for the gem to run.
2. The gem, when packaged and published, needs to declare what other gems it has a "runtime" dependency on, so that tools like Bundler, etc can make sure to install those as well. These are gems that, if they are not present in a user's gem folder, will prevent the main gem from working. Those are listed in the `.gemspec` file as `add_dependency` entries.

In your case, Tapioca would not work on a user's machine if Parallel gem wasn't installed, so that dependency needs to be declared in the gemspec, since that is the only thing that gets packages inside a released gem (along with the actual source code).

Morriar: My understanding it that:

* the `.gemspec` file is used to export the list of declared runtime dependencies. So for example the gems in this file using `add_dependency` will be displayed as runtime dependencies on RubyGems: https://rubygems.org/gems/tapioca. Furthermore, when a project will install `tapioca` as a gem, they will get the dependencies declared in the `.gemspec` as transitive dependencies.

* the `Gemfile` is used for "internal" dependencies. The ones you do _not_ need when executing the gem. This would be gems required to run the tests, to debug, to lint etc. All the need you would only need to contribute to the project. Those dependencies will _not_ be installed as transitive ones when installing `tapioca` in a project.

In your case, since `parallel` is required to execute `tapioca` it _is_ a runtime dependency and should be declared as such in the `.gemspec` as you did 👍 

For the version, it seems you've been using 1.21.0 to run the tests. So it might be safer to add a restriction to require at least this version since we're not sure it works with an older version of the gem.

paracycle: What complicates all of this is the notion of `add_development_dependency` that goes into a `gemspec` and declares optional dependencies that should be used for development. I choose to completely ignore that, since it is a hold over from before the time we had `Gemfile`s and that was the only way a gem could declare things it needed for development as opposed to runtime. I just always use a `Gemfile` for those now and never declare my dev dependencies in the `gemspec`.

egiurleo: I added a version constraint to the parallel gem!

paracycle: I guess you also need to do a `bundle lock` to update the `Gemfile.lock` after that, so that the lockfile also picks up that version bound.

egiurleo: 🤦🏻‍♀️ Love when CI turns red immediately lol!

Morriar (approved):
Amazing, thanks! 👍

Issue #890, opened 2022-04-08
Better mixin generation for gem RBIs


egiurleo:
This issue encompasses multiple problems:

1. We have not had a way of filtering mixins on a constant to generate only the ones that are actually mixed in by the gem we are generating an RBI for. This has been causing issues of attribution and was resulting in Tapioca generating RBIs that would look different in the presence of other gems.

An initial attempt was made in #569 where mixin tracking was introduced with the intention of storing the backtraces of calls to mix in modules. However, this resulted in some RBIs missing critical mixins because it would overmatch mixins to external gems, thus filtering them out and not generating RBIs for them.

2. Another problem is that mixins can be added dynamically (e.g. `Foo.prepend(Bar)`):
* Imagine `Foo` is defined in a gem called foo
* `Bar` is defined in a gem called bar
* In the `bar` gem, the line `Foo.prepend(Bar)` is executed. 
* When generating RBIs for the bar gem, tapioca will not be aware of `Bar` being prepended to `Foo` because it is not generating RBIs for `Foo`, which is defined in a different gem.

This is especially true if `Foo` is a standard library constant. Tapioca only ever generates RBIs for gems that are in the Gemfile, and the standard library does not need to be explicitly added to the gemfile.

There was an attempt to address the second issue in the uk-eg-mixin-fix branch, but it was never finalized. We have since reverted the use of the MixinTracker (#668) and refactored a lot, making the changes on this branch incompatible with what is currently on main.

PR #896, opened 2022-04-12
More effectively determine method ownership for RBI generation

egiurleo:
### Motivation
This issue was discovered by @paracycle while investigating #890, but it is technically unrelated to mixin generation.

While generating RBIs for a method, tapioca implements a check to determine which constant "owns" or defines that method. We only want to generate RBIs for a method on the module or class that defines it, not any that just happen to inherit it.

Historically, we have been checking the `method.owner` and verifying that it matches the class/module for which we are generating RBIs. However, this check misses one particular case: when a method is also defined on a prepended module, it will always appear as if it belongs to the prepended module, even when it is also defined on the class that prepends the module.

This means that tapioca will not generate RBI for that method on the class that defines it.

### Implementation
This code was originally written by @paracycle on commit 80faabc8ace1af686589278c80ff24a5069e7748. I've just cleaned it up and moved it to the correct spot.

Instead of just checking the method owner, this implementation walks up the ancestor chain via the `super_method` method, checking if any of the super methods are owned by the constant. If that is the case, that means the constant also defines the method, and we should generate RBIs for this method on the constant.

### Tests
I have added a test which fails without the changes but passes with the changes included.

### Tophatting
I tested this change by generating gem and dsl RBIs on Shopify core using this branch of tapioca and verifying that the change did not introduce any new typing issues.

### DIFF
--- a/lib/tapioca/gem/listeners/methods.rb
+++ b/lib/tapioca/gem/listeners/methods.rb
@@ -66,7 +66,7 @@ module Tapioca
         end
         def compile_method(tree, symbol_name, constant, method, visibility = RBI::Public.new)
           return unless method
-          return unless method.owner == constant
+          return unless method_owned_by_constant?(method, constant)
           return if @pipeline.symbol_in_payload?(symbol_name) && !@pipeline.method_in_gem?(method)
 
           signature = signature_of(method)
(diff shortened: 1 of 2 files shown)


lib/tapioca/gem/listeners/methods.rb
```diff
@@ -142,6 +142,35 @@ def compile_method(tree, symbol_name, constant, method, visibility = RBI::Public
...
+        def method_defined_by_constant?(method, constant)
+          # Check whether the method is defined on the constant.
+          #
+          # In most cases, it works to check that the constant is the method owner. However,
+          # in the case that a method is also defined in a module prepended to the constant, it
+          # will be owned by the prepended module, not the constant.
+          #
+          # This method implements a better way of checking whether a constant defines a method.
+          # It walks up the ancestor tree via the `super_method` method; if any of the super
+          # methods are owned by the constant, it means that the constant declares the method.
```
Morriar, line 155: Could we have this comment on top of the method as documentation rather than inside?

spec/tapioca/gem/pipeline_spec.rb
```diff
@@ -1210,6 +1210,36 @@ class Bar; end
       assert_equal(output, compile)
     end
 
+    it "compiles method that is also prepended" do
+      add_ruby_file("foo.rb", <<~RUBY)
+        module Foo
+          def bar
+            super
```
Morriar, line 1217: Can we add an example that does not call `super`? 🙏

lib/tapioca/gem/listeners/methods.rb
```diff
@@ -66,7 +66,7 @@ def compile_directly_owned_methods(tree, module_name, mod, for_visibility = [:pu
         end
         def compile_method(tree, symbol_name, constant, method, visibility = RBI::Public.new)
           return unless method
-          return unless method.owner == constant
+          return unless method_defined_by_constant?(method, constant)
```
paracycle, line 69: I know that I had named this method but looking at it again, it feels like `method_owned_by_constant?` would be a more appropriate name, since we are doing ownerships checks.

vinistock (approved):

Morriar (approved):
Amazing thanks! 🎉

PR #897, opened 2022-04-12
Add types to generated RBI files for FrozenRecord

nickpresta:


spec/tapioca/dsl/compilers/frozen_record_spec.rb
```diff
@@ -97,6 +97,111 @@ def last_name?; end
...
+
+                  sig { returns(String) }
+                  def last_name
+                    super
+                  end
+
+                  sig { returns(String) }
+                  def location
+                    super
+                  end
```
nickpresta, line 122: Defining

```ruby
sig { returns(String) }
attr_reader :first_name
```

generates a runtime error in Sorbet when accessed via `Student.first.first_name` since Sorbet thinks this is nilable. Accessing the value via `Student.first["first_name"]` works fine, but I expect most people are using dot notation.

The approach of explicitly defining the method seems like a reasonable tradeoff for type safety.

spec/tapioca/dsl/compilers/frozen_record_spec.rb
```diff
@@ -97,6 +97,111 @@ def last_name?; end
...
+                    def area(grain:); end
+
+                    sig { returns(::String) }
+                    def first_name; end
+
+                    sig { returns(T::Boolean) }
+                    def first_name?; end
+
+                    sig { returns(T.untyped) }
+                    def id; end
```
nickpresta, line 182: Fields that haven't been explicitly defined and typed are still `T.untyped` as before.

lib/tapioca/dsl/compilers/frozen_record.rb
```diff
@@ -72,13 +72,17 @@ def decorate
           attributes = constant.attributes
           return if attributes.empty?
 
+          instance = constant.first
```
nickpresta, line 75: The potential problem here is that if your first record contains fields that subsequent records don't, the type will be generated without `T.nilable`, which is wrong.

In practice, I'm not sure how much this happens though.

lib/tapioca/dsl/compilers/frozen_record.rb
```diff
@@ -72,13 +72,17 @@ def decorate
...
+
           root.create_path(constant) do |record|
             module_name = "FrozenRecordAttributeMethods"
 
             record.create_module(module_name) do |mod|
               attributes.each do |attribute|
+                return_type = instance.attributes[attribute].class.name
```
nickpresta, line 82: This approach isn't sufficient to be a complete type for all the collection types YAML supports (e.g. `Array` and `Hash`) since we aren't (and shouldn't) recursively check every member of the collection to create a union of all possible types.

I think this is "fine" since [Sorbet itself](https://sorbet.org/docs/stdlib-generics) removed this functionality during runtime due to performance problems.

Having these be `T::Array[T.untyped]` and `T::Hash[T.untyped, T.untyped]` seems "OK" and devs can further refine the type with `T.cast` at runtime.

lib/tapioca/dsl/compilers/frozen_record.rb
nickpresta: After some discussion in Slack, @paracycle [suggested we use `ActiveModel::Attributes`](https://shopify.slack.com/archives/CBXRGT3SB/p1649961447540339?thread_ts=1649801826.205049&cid=CBXRGT3SB) to declare the attributes and types without doing any runtime discovery.

This covers the cases where the shape inferred from the first record may be incomplete, and makes it explicit which fields are which types.

However, using `ActiveModel::Attributes` comes with a couple of caveats:

- Arrays, hashes, and symbols cannot be specified, in general. Arrays can be specified if [using PostgreSQL](https://github.com/rails/rails/issues/32324) (associated [test case](https://github.com/rails/rails/blob/8ace32c4ccdc760d17548de777b08c13bc9e07bd/activerecord/test/cases/attributes_test.rb#L239-L269)), but we can't depend on a specific database backend, and hashes don't seem supported at all (although I see some people using `ActiveRecord::Type::Json` which seems like a workaround). This sucks since arrays and hashes are totally fine in YAML and (ideally) shouldn't require any custom types just to get it to work.
- Typing custom types is "fine", but there's no real way to specific anything more specific than what you can return from the `type` method on `ActiveModel::Type::Value`. This results in things like `Array` and `Hash` types, which is "fine", but it would still be preferable to be able to specify the element type.
- Defaults specified using `attribute :my_field, default: 'hi'` don't work in `FrozenRecord`, since the `default_attributes` field should be used instead. So we need to look at `default_attributes` and any dev who specified `attribute :my_field, default: 'hi'` won't actually have their changes reflected. Perhaps we should throw in this case and suggest using `default_attributes`?

The only other option to solve these problems is to define something custom that allows for sorbet types specifically. For example:

```ruby
class Student < FrozenRecord::Base
  extend T::Sig

  self.field_types = {
    first_name: ::String,
    favourite_foods: T::Array[String],
    skills: T::Hash[String, String],
  } 
end
```

but this isn't ideal either, since we're making it up specifically for Tapioca.

---

With all this said, I think using `ActiveModel::Attribute` is better than inferring types and for existing code that isn't specifying the attributes, there are no expected changes. I feel confident that this is an improvement, although not perfect, and sets us up for typing of the [query/finder/calculation methods](https://github.com/byroot/frozen_record#supported-query-methods).

@paracycle I would appreciate any feedback on whether this approach makes sense and if there are any ways to workaround some of the caveats above -- I'm relatively new to Rails so I'm sure I'm missing some context.

PR #902, opened 2022-04-19
[WIP] Support new type variable syntax

KaanOzkan:
This is a prototype.


lib/tapioca/dsl/compiler.rb
```diff
@@ -17,7 +17,7 @@ class Compiler
       include Runtime::Reflection
       extend Runtime::Reflection
 
-      ConstantType = type_member(upper: Module)
+      ConstantType = type_member {{upper: Module}}
```
KaanOzkan, line 20: This is the same formatting that sorbet's autocorrect uses but it currently violates our rubocop rules.

lib/tapioca/sorbet_ext/generic_name_patch.rb
```diff
@@ -20,31 +20,29 @@ def [](*types)
         Tapioca::Runtime::GenericTypeRegistry.register_type(constant, types)
       end
 
-      def type_member(variance = :invariant, fixed: nil, lower: T.untyped, upper: BasicObject)
+      def type_member(variance = :invariant, &blk)
```
KaanOzkan, line 23: Should we instead support both syntaxes at the same time so that devs can update tapioca without migrating their codebases?

lib/tapioca/dsl/compiler.rb
vinistock: What's the violation? Spaces between the curly braces?

KaanOzkan: Good question. It mostly works. It's no longer a typecheck error so Sorbet doesn't try to remove the spaces. Funnily enough sometimes the [BindingConstantWithoutTypeAlias cop](https://github.com/Shopify/rubocop-sorbet/blob/2fce9e3c7be251b67acb8a17fd424faff0fb7e54/lib/rubocop/cop/sorbet/binding_constants_without_type_alias.rb#L18) autocorrect is triggered and we insert an incorrect `T.type_alias` with rubocop which would need fixing.

My biggest concern with this is that it's a 2 step process. If we can't change the rubocop rule we should lobby for adding spaces in Sorbet autocorrect.

vinistock: I think that's going to be a very hard sell, given that the Sorbet folks prefer the alternative style of no spaces around braces. If we can auto-correct this once, let's try to do it.

paracycle: I am not that worried about the Sorbet autocorrect, since it is mostly a helper to get the code semantically right, not necessarily stylistically correct.

In previous autocorrects, as well, we've had to apply the autocorrects and then run RuboCop to fix violations. So, let's do that again.

lib/tapioca/sorbet_ext/generic_name_patch.rb
paracycle: Yes, I would really like this to be backwards compatible, especially since I would like to backport this to the 0.7 series as well, since it is blocking people from updating their Sorbet versions.

lib/tapioca/sorbet_ext/generic_name_patch.rb
```diff
@@ -153,16 +151,18 @@ def name
 
     sig { returns(String) }
     def serialize
-      parts = []
-      parts << ":#{@variance}" unless @variance == :invariant
-      parts << "fixed: #{@fixed}" if @fixed
-      parts << "lower: #{@lower}" unless @lower == T.untyped
-      parts << "upper: #{@upper}" unless @upper == BasicObject
+      positional_arg = @variance unless @variance == :invariant
```
paracycle, line 154: This method should do a Sorbet version check and generate the correct type variable serialization for that version of Sorbet, not just the block syntax for all versions.

paracycle (commented):
Thank you, this is good start. In general, it would be great to have this working both ways, depending on the version of Sorbet this code is working against, so that we can backport this to the 0.7.x series and unblock users.

.rubocop.yml
```diff
@@ -50,3 +50,6 @@ Style/CaseEquality:
 Style/ClassAndModuleChildren:
   Exclude:
   - "spec/tapioca/**/*"
+
+Sorbet/BindingConstantWithoutTypeAlias:
+  Enabled: false
```
paracycle, line 55: This is meant to be temporary and is related to: https://github.com/Shopify/rubocop-sorbet/issues/104

lib/tapioca/dsl/compiler.rb
paracycle: I updated this PR to disable the `BindingConstantWithoutTypeAlias` for now.

Morriar (commented):
Can we add a test showing the different behavior with an old Sorbet version? 🙏

Morriar (approved):
Amazing, thanks! ❤️

lib/tapioca/sorbet_ext/generic_name_patch.rb
```diff
@@ -20,31 +21,47 @@ def [](*types)
         Tapioca::Runtime::GenericTypeRegistry.register_type(constant, types)
       end
 
-      def type_member(variance = :invariant, fixed: nil, lower: T.untyped, upper: BasicObject)
+      def type_member(variance = :invariant, fixed: nil, lower: nil, upper: nil, &blk)
         # `T::Generic#type_member` just instantiates a `T::Type::TypeMember` instance and returns it.
         # We use that when registering the type member and then later return it from this method.
+        hash = if blk
```
vinistock, line 27: Instead of checking if the block is present, should we make the sorbet supports check here?

lib/tapioca/sorbet_ext/generic_name_patch.rb
```diff
@@ -20,31 +21,47 @@ def [](*types)
...
         ).tap do |type_variable|
           Tapioca::Runtime::GenericTypeRegistry.register_type_variable(self, type_variable)
         end
       end
 
-      def type_template(variance = :invariant, fixed: nil, lower: T.untyped, upper: BasicObject)
+      def type_template(variance = :invariant, fixed: nil, lower: nil, upper: nil, &blk)
         # `T::Generic#type_template` just instantiates a `T::Type::TypeTemplate` instance and returns it.
         # We use that when registering the type template and then later return it from this method.
+        hash = if blk
```
vinistock, line 50: Same as my previous comment.

lib/tapioca/sorbet_ext/generic_name_patch.rb
paracycle: This is the processing side of the equation, which can be more relaxed with respect to how bound parameters are passed. Ideally we would like to support mixed usage of bounds being passed as arguments or inside a block, since the `pipeline_spec` is testing exactly that.

For generating side of the equation, we need to be more strict so that the Sorbet version we run against can properly process the generated code.

This is all following the old internet adage of "be liberal with what you accept and conservative with what you send".

vinistock (approved):

PR #903, opened 2022-04-25
Compile RBIs for foreign constants

egiurleo:
### Motivation
Part of the work for #890. Heavily based on work I did with @paracycle on [this branch](https://github.com/Shopify/tapioca/compare/uk-eg-mixin-fix).

Currently, Tapioca only generates RBI for constants that are defined in the gem that's currently being compiled. This is the correct behavior in most cases, except that in Ruby, a gem can add functionality to any constant, even one that it doesn't define.

Let's say there's a gem `foo` that defines the class `Foo`.
Then let's say there's another gem `bar` that defines the module `Bar` and contains the following code:

```
Foo.prepend(Bar)
```

or

```
class Foo
   prepend Bar
end
```

Currently, Tapioca would not recognize that `Foo` had `Bar` mixed in and would generate incomplete RBIs for `Foo`. This is especially true if `Foo` is a constant defined in the Ruby standard library.

We're calling these "foreign constants."

### Implementation
There are a few steps to this implementation:
- Modify the `Mixin` tracker to keep a map between constants and classes that mix them in. The tracker currently keeps a map the other way around, but I think it makes more sense to flip the keys and values in this case (will explain more in next bullet point).
- Introduce a new listener called `ForeignConstants`. This listener takes a constant and uses the `Mixin` tracker to identify all the other constants that have the first constant as a mixin. This will include any constants that are defined in another gem or in the Ruby standard library that have mixed in constants defined in the gem being compiled. All of these constants are then added to the pipeline in a `Gem::ForeignConstantFound` event.
- This event can be processed in a simpler way than the `Gem::ConstantFound` event because there are certain logical branches that don't apply to it and can be skipped entirely. While processing this event, the pipeline knows to skip the `defined_in_gem?` check.
- This will lead to a `ForeignScopeNodeAdded` event being added to the pipeline. The listeners have been modified such that only the `Mixins` listener will accept the `ForeignScopeNodeAdded` event. This way, only mixin RBIs are generated for foreign constants and all other RBIs (method definitions, documentation, etc.) are ignored.

### Tests
I have added multiple pipeline tests and a gem test that fail on main but pass with these changes.

### Tophatting 🎩 

I have tested these changes on Shopify core. It compiles all gem RBIs and results in no type errors!


lib/tapioca/gem/pipeline.rb
```diff
@@ -228,9 +229,9 @@ def compile_object(name, value)
         @root << node
       end
 
-      sig { params(name: String, constant: Module).void }
-      def compile_module(name, constant)
-        return unless defined_in_gem?(constant, strict: false)
+      sig { params(name: String, constant: Module, ignore_defined_in_gem: T::Boolean).void }
+      def compile_module(name, constant, ignore_defined_in_gem: false)
+        return unless defined_in_gem?(constant, strict: false) || ignore_defined_in_gem
```
egiurleo, line 234: I'm not very confident about this flag because of the way it's being passed all the way down from the event into the `compile_module` method. I'm curious what the `strict` keyword is supposed to do and if that expresses a similar concept.

lib/tapioca/runtime/trackers/mixin.rb
```diff
@@ -7,7 +7,8 @@ module Trackers
       module Mixin
         extend T::Sig
 
-        @mixin_map = {}.compare_by_identity
+        @constants_to_mixin_locations = {}.compare_by_identity
```
egiurleo, line 10: I changed the name of this variable from `mixin_map` to `constants_to_mixin_location`. I had a really hard time remembering what it was supposed to do and I think making it clear what it is mapping between would be helpful. Open to thoughts and name suggestions!

lib/tapioca/runtime/trackers/mixin.rb
```diff
@@ -20,26 +21,35 @@ class Type < T::Enum
...
           params(
             constant: Module,
-            mod: Module,
+            mixin: Module,
             mixin_type: Type,
             locations: T.nilable(T::Array[Thread::Backtrace::Location])
           ).void
         end
-        def self.register(constant, mod, mixin_type, locations)
+        def self.register(constant, mixin, mixin_type, locations)
```
egiurleo, line 29: Changed the name of the variable here for clarity.

lib/tapioca/runtime/trackers/mixin.rb
```diff
@@ -20,26 +21,35 @@ class Type < T::Enum
...
         def self.mixin_locations_for(constant)
-          @mixin_map[constant] ||= {
+          @constants_to_mixin_locations[constant] ||= {
             Type::Prepend => {}.compare_by_identity,
             Type::Include => {}.compare_by_identity,
             Type::Extend => {}.compare_by_identity,
           }
         end
+
+        sig { params(mixin: Module).returns(T::Array[Module]) }
```
egiurleo, line 49: Keeping this hash super simple for right now, but I think we could add the actual locations of the mixins in a follow-up PR.

paracycle (commented):
This looks quite complete already. I've added some inline comments.

lib/tapioca/gem/events.rb
```diff
@@ -42,6 +42,23 @@ def initialize(symbol, constant)
       end
     end
 
+    class NotOwnedConstantFound < Event
```
paracycle, line 45: What if this subclassed `ConstantFound`? We are basically using the class as a discriminator on how to act, otherwise, all aspects of this class should be the same as `ConstantFound`.

lib/tapioca/gem/pipeline.rb
```diff
@@ -124,7 +130,7 @@ def dispatch(event)
         case event
         when Gem::SymbolFound
           on_symbol(event)
-        when Gem::ConstantFound
+        when Gem::ConstantFound, Gem::NotOwnedConstantFound
```
paracycle, line 133: If `NotOwnedConstantFound` subclasses `ConstantFound`, we don't need this addition.

lib/tapioca/gem/pipeline.rb
```diff
@@ -142,7 +148,7 @@ def on_symbol(event)
         push_constant(symbol, constant) if constant
       end
 
-      sig { params(event: Gem::ConstantFound).void.checked(:never) }
+      sig { params(event: T.any(Gem::ConstantFound, Gem::NotOwnedConstantFound)).void.checked(:never) }
```
paracycle, line 151: If `NotOwnedConstantFound` subclasses `ConstantFound`, we don't need this `T.any`.

paracycle (commented):
This looks much much cleaner. I think one more pass will be enough. I am still not in love with the name we have of the concept, so it might be good to get feedback from other folks by turning this into a proper PR.

lib/tapioca/gem/pipeline.rb
```diff
@@ -156,7 +162,12 @@ def on_constant(event)
         return if T::Enum === constant # T::Enum instances are defined via `compile_enums`
 
         mark_seen(name)
-        compile_constant(name, constant)
+
+        if event.is_a?(Gem::NotOwnedConstantFound)
+          compile_not_owned_constant(name, T.cast(constant, Module))
```
Morriar, line 167: If you use `event.constant` you won't need to cast it:

```suggestion
          compile_not_owned_constant(name, event.constant)
```

I think we could refactor a bit the code to not use a local variable.

lib/tapioca/gem/pipeline.rb
```diff
@@ -152,11 +158,15 @@ def on_constant(event)
...
+        return if T::Enum === event.constant # T::Enum instances are defined via `compile_enums`
 
         mark_seen(name)
-        compile_constant(name, constant)
+
+        if event.is_a?(Gem::NotOwnedConstantFound)
+          compile_not_owned_constant(name, event.constant)
+        else
+          compile_constant(name, event.constant)
+        end
```
paracycle, line 169: Can we turn this into a `case/when` to make it explicit that we are matching on the type of `event`?
```suggestion
        case event
        when Gem::NotOwnedConstantFound
          compile_not_owned_constant(name, event.constant)
        else
          compile_constant(name, event.constant)
        end
```

lib/tapioca/gem/listeners/not_owned_constants.rb
```diff
@@ -0,0 +1,32 @@
+# typed: strict
+# frozen_string_literal: true
+
+module Tapioca
+  module Gem
+    module Listeners
+      class NotOwnedConstants < Base
```
paracycle, line 7: I am still not a fan of this name, tbh. What do you think of:
- `ForeignConstant` as the name of a constant that we discover that we need to handle but wasn't declared by the current gem.
- `ForeignConstantFound` as the event name

lib/tapioca/gem/listeners/not_owned_constants.rb
```diff
@@ -0,0 +1,32 @@
...
+        sig { override.params(event: ScopeNodeAdded).void }
+        def on_scope(event)
+          mixin = event.constant
+
+          # There are cases where we want to compile RBI for constants not owned by the current
+          # gem - e.g. constants defined in one gem that then have a mixin applied to them in
+          # another gem. This is especially common for Ruby standard library classes.
+          #
+          # We can identify these cases by keeping track of which classes have mixins, and then
+          # adding those classes to the pipeline.
```
paracycle, line 23: This comment is correct but I think we can improve it slightly. What do you think of this:
```suggestion
          # There are cases where we want to process constants not declared by the current
          # gem, i.e. "foreign constant". These are constants defined in another gem 
          # that this gem is applying a mix-in to. This pattern is especially common 
          # for gems that add behaviour to Ruby standard library classes by mixing in
          # modules to them.
          #
          # The way we identify these "foreign constants" is by asking the mixin
          # tracker which constants have mixed in the current module that we are
          # handling. We add all those constants that we discover to the pipeline 
          # to be processed.
```

lib/tapioca/gem/listeners/not_owned_constants.rb
```diff
@@ -0,0 +1,32 @@
...
+        extend T::Sig
+
+        include Runtime::Reflection
+
+        private
+
+        sig { override.params(event: ScopeNodeAdded).void }
+        def on_scope(event)
+          mixin = event.constant
+
```
paracycle, line 17: I think we can return early if `mixin` is a `Class` instance, since no `Class` can be mixed into something else:
```suggestion
          mixin = event.constant

          return if Class === mixin # Classes can't be mixed into anything
```

lib/tapioca/gem/events.rb
```diff
@@ -42,6 +42,19 @@ def initialize(symbol, constant)
...
+    class NotOwnedConstantFound < ConstantFound
+      extend T::Sig
+
+      sig { returns(Module).checked(:never) }
+      attr_reader :constant
+
+      sig { params(symbol: String, constant: Module).void.checked(:never) }
+      def initialize(symbol, constant)
+        super
+        @constant = constant
```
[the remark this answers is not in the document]
paracycle: I wonder if we should do the breakdown of the kinds of constants at push time, so that these become 4 separate events instead:
1. `AliasConstantFound`
2. `ValueConstantFound`
3. `ModuleConstantFound`
4. `ForeignModuleConstantFound < ModuleConstantFound`

All we would need to do is to move the logic in `compile_constant` inside `push_constant` and the handling becomes much simpler. WDYT @Morriar?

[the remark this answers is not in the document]
Morriar: Yeah I like it!

Another option to keep the original design is to use a cast:

```rb
class ForeignConstantFound < ConstantFound
  extend T::Sig

  sig { override.returns(Module) }
  def constant
    T.cast(@constant, Module)
  end

  sig { params(symbol: String, constant: Module).void }
  def initialize(symbol, constant)
    super
  end
end
```

I'm good with both approaches 👍

egiurleo:
Spent some time pairing with @Morriar to debug the errors that were occurring when trying to compile gem types on Shopify core. This is what we discovered:
- The errors were occurring because tapoica loads all of Shopify core when compiling gems. This means that when core mixes in certain gems (e.g. `ActiveModel`), the mixin tracker will track those classes from core. The `NotOwnedConstants` listener will then add those classes from core to the pipeline, and tapioca will try to compile them. Obviously this is not desired behavior.
- I think the actual issue is that the mixin tracker is casting too wide of a net. It's tracking even the mixins that happen outside of the gem we're currently compiling. This seems related to the other issue that #890 is trying to address, so it might make sense to try to address that issue first, then revisit this PR.

My next step will be to write a failing test to make sure the issue is captured in a repeatable way.

lib/tapioca/gem/listeners/foreign_constants.rb
```diff
@@ -0,0 +1,38 @@
...
+        include Runtime::Reflection
+
+        private
+
+        sig { override.params(event: ScopeNodeAdded).void }
+        def on_scope(event)
+          mixin = event.constant
+          return if Class === mixin # Classes can't be mixed into other constants
+
+          # There are cases where we want to process constants not declared by the current
```
egiurleo, line 19: @paracycle Updated the comment here! It is almost identical to what you suggested with a couple of small wording changes.

lib/tapioca/gem/listeners/foreign_constants.rb
```diff
@@ -0,0 +1,38 @@
...
+          # gem, i.e. "foreign constant". These are constants defined in another gem to which
+          # this gem is applying a mix-in. This pattern is especially common for gems that add
+          # behavior to Ruby standard library classes by mixing in modules to them.
+          #
+          # The way we identify these "foreign constants" is by asking the mixin tracker which
+          # constants have mixed in the current module that we are handling. We add all the
+          # constants that we discover to the pipeline to be processed.
+          Runtime::Trackers::Mixin.constants_with_mixin(mixin).each do |constant, locations|
+            location = locations.first
+            next if location.nil? || !@pipeline.gem.contains_path?(location)
```
egiurleo, line 29: I was experiencing an issue where running `tapioca gem` on Shopify core was generating RBIs for ALL mixins, not just mixins performed in the gem currently being compiled. This is my fix for this issue. The mixin tracker keeps track of the location where this mixin took place, and checks that it occurred within the gem before generating RBIs.

Two points:
- Currently, the locations are an array of file paths, so I took the first one here, which I believe should correspond to the "closest" location where the mixin took place? Let me know if that sounds right for now.
- I'm going to create a follow-up PR that changes the notion of what a mixin location is, based on the work @paracycle did in [this branch](https://github.com/Shopify/tapioca/compare/uk-eg-mixin-fix#) -- basically, taking the most recent "require" location as the correct location. I wanted to save that for another PR because this one is already quite big.

paracycle: It is going to change soon, as you said, but up to this point, our thinking was "this is coming from this gem if any of the locations match the gem require path prefix". So we used to have this method that wasn't being used, that we can bring back again for this: https://github.com/Shopify/tapioca/commit/bdf5d45cd74f329fd438dd8a9b54fd7f4205a39d

lib/tapioca/gem/listeners/foreign_constants.rb
```diff
@@ -0,0 +1,46 @@
...
+          # There are cases where we want to process constants not declared by the current
+          # gem, i.e. "foreign constant". These are constants defined in another gem to which
+          # this gem is applying a mix-in. This pattern is especially common for gems that add
+          # behavior to Ruby standard library classes by mixing in modules to them.
+          #
+          # The way we identify these "foreign constants" is by asking the mixin tracker which
+          # constants have mixed in the current module that we are handling. We add all the
+          # constants that we discover to the pipeline to be processed.
+          Runtime::Trackers::Mixin.constants_with_mixin(mixin).each do |constant, locations|
+            next unless mixed_in_by_gem?(locations)
```
egiurleo, line 28: @paracycle Here's the use of that method. As far as I could tell, we were never actually using the `mixin_type` from the tracker, so I've just been storing the constant and locations. Let me know if this looks alright!

paracycle: Yes, for this part of the work, this much is enough. 

However, later on we will also have to revert https://github.com/Shopify/tapioca/pull/668 at which point, we will have start deciding, for a given mixin and a given mixin type (`include`, `extend`, `prepend`), is that happening from inside this gem or not. At that point, we will need that logic.

tasks/readme.rake
```diff
@@ -1,6 +1,12 @@
 # typed: true
 # frozen_string_literal: true
 
+# Temporary fix for CI
+# A patch as been upstreamed, we're waiting for it to be merged.
+# See https://github.com/rubyconfig/config/pull/319
+# TODO: remove this one the patch is released.
+require "rails"
```
egiurleo, line 8: This is a patch to help fix CI that will be merged soon (#965). Once it's merged I'll rebase on main.

paracycle (approved):
Looks great! Thanks for all the work you put into this.

The `should_skip?` mechanism is a little bit kludgy but I think we can refine it afterwards.

egiurleo:
@paracycle Yay! 100% agreed that `should_skip?` is kludgy (great word) but I am totally happy to make it better later.

I'll merge this once #965 or something similar is merged.

lib/tapioca/gem/events.rb
```diff
@@ -42,6 +42,20 @@ def initialize(symbol, constant)
...
+      extend T::Sig
+
+      sig { override.returns(Module) }
+      def constant
+        T.cast(@constant, Module)
+      end
+
+      sig { params(symbol: String, constant: Module).void }
+      def initialize(symbol, constant)
+        super
```
vinistock, line 55: Why do we need to define this method if all it does is invoke `super`?

lib/tapioca/gem/listeners/foreign_constants.rb
```diff
@@ -0,0 +1,81 @@
...
+          # There are cases where we want to process constants not declared by the current
+          # gem, i.e. "foreign constant". These are constants defined in another gem to which
+          # this gem is applying a mix-in. This pattern is especially common for gems that add
+          # behavior to Ruby standard library classes by mixing in modules to them.
+          #
+          # The way we identify these "foreign constants" is by asking the mixin tracker which
+          # constants have mixed in the current module that we are handling. We add all the
+          # constants that we discover to the pipeline to be processed.
+          Runtime::Trackers::Mixin.constants_with_mixin(mixin).each do |constant, locations|
+            next if defined_by_application?(constant)
```
vinistock, line 28: Do we need both of these checks? If it's mixed in by a gem, it can't be defined by the application, right?

paracycle: We need both of them since our application code is split into a lot gems. But this part is temporary code until the next PR which will have better attribution to code that triggered a specific tracked event.

lib/tapioca/gem/listeners/base.rb
```diff
@@ -42,6 +44,15 @@ def on_scope(event)
         sig { params(event: MethodNodeAdded).void }
         def on_method(event)
         end
+
+        sig { params(event: NodeAdded).returns(T::Boolean) }
+        def should_skip?(event)
```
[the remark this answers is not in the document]
egiurleo: I like that! Would you be okay with me doing that in a follow-up PR?

lib/tapioca/gem/events.rb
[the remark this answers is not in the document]
egiurleo: I created an issue to follow up on this: #977

Morriar (approved):

PR #908, opened 2022-04-25
Prevent Sorbet and Sorbet runtime version drift in tests

paracycle:
### Motivation

Since we have not been pinning the versions of the dependencies of
Tapioca inside a mock project, we were getting version drift between
what versions of Sorbet and Sorbet runtime Tapioca was using and the
ones that the mock project ended up using.

This is creating problems when a newer version of Sorbet runtime has
breaking changes, like the one introduced here: https://github.com/sorbet/sorbet/commit/99a7135942149c34e76daf0566ff549b68cb61ad#diff-ad43eb0a6b94929e24a6ad3121a0a0629095756b1d7019a3fe8b7dfbdab3cf21R14
which was causing failures like the ones here: https://github.com/Shopify/tapioca/runs/6164393894?check_suite_focus=true#step:8:541

### Implementation

Since we know that any mock project will have dependencies on `sorbet` and `sorbet-runtime` through its dependency on `tapioca`, I decided to add explicit dependencies on those two gems for a mock project with versions fixed to the versions that Tapioca uses.

### Tests

No new tests, existing ones should pass.


spec/helpers/mock_project.rb
```diff
@@ -31,10 +31,16 @@ def require_mock_gem(gem, require: nil)
     sig { params(name: String, version: T.nilable(String), require: T.nilable(T.any(FalseClass, String))).void }
     def require_real_gem(name, version = nil, require: nil)
       line = +"gem \"#{name}\""
+
+      gemfile = read("Gemfile")
+      lines = gemfile.lines.delete_if { |gemfile_line| gemfile_line.start_with?(line) }
```
Morriar, line 36: This is not ideal. We'll spend a loot of time reading the Gemfile for each test suite.
It seems this is only required for the `check-shim` test where we pass a custom version of the Sorbet right?

What do you think about removing this and creating the project manually in the `check-shims` test instead?

```rb
@project = MockProject.new("#{TEST_TMP_PATH}/#{spec_name}/project")
@project.gemfile(project.tapioca_gemfile)
```

We could even extract `#{TEST_TMP_PATH}/#{spec_name}/project` as a service to make it easier.

paracycle: Ok, I changed this to make the `mock_project` method take an optional `sorbet_dependency` keyword parameter that is `true` by default, and making just this test reinitialize the mock project with that parameter value passed as `false`.

Morriar (approved):

PR #913, opened 2022-04-29
Add descriptive message for unexpected DSL exceptions

paracycle:
### Motivation

Currently, when a DSL compiler raises an error while running a `decorate` operation, it is not very obvious which compiler failed and during the processing of which constant. The failing DSL compiler is kind of obvious if one looks at the backtrace, but the constant that it is failing on is very hard to understand if one is not running DSL generation in verbose mode.

### Implementation

This PR handles unexpected exceptions, shows a friendly error message for them and re-raises the exception to keep the current behaviour.

### Tests

Added a test with a failing DSL compiler.

### DIFF
--- a/lib/tapioca/dsl/pipeline.rb
+++ b/lib/tapioca/dsl/pipeline.rb
@@ -152,6 +152,9 @@ module Tapioca
 
           compiler = compiler_class.new(self, file.root, constant)
           compiler.decorate
+        rescue
+          $stderr.puts("Error: `#{compiler_class.name}` failed to generate RBI for `#{constant}`")
+          raise # This is an unexpected error, so re-raise it
         end
 
         return if file.root.empty?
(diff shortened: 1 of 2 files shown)


lib/tapioca/dsl/pipeline.rb
```diff
@@ -152,6 +152,9 @@ def rbi_for_constant(constant)
 
           compiler = compiler_class.new(self, file.root, constant)
           compiler.decorate
+        rescue
+          $stderr.puts("ERROR: `#{compiler_class.name}` failed to generate RBI for `#{constant}`")
```
Morriar, line 156: In other places we've been using `Error`:

```suggestion
          $stderr.puts("Error: `#{compiler_class.name}` failed to generate RBI for `#{constant}`")
```

vinistock (approved):

AntoineToussaint:
Best PR ever!

PR #920, opened 2022-05-03
Enable documentation generation by default

Morriar:
### Motivation

* This feature makes the whole VS Code experience so much better
* It's been battle tested for quite a long now and works well
* It's slow only for a very limited number of gems, let's have people disable it if needed instead

### Implementation

`!false`

### Tests

See automated tests.


paracycle (approved):
🙏

PR #928, opened 2022-05-10
Bump version on main to latest

rwstauner:
### Motivation

Want to try out "main" but shopify-types already specifies later version.


paracycle:
Fixed by https://github.com/Shopify/tapioca/commit/2c7a1dd89660a34d3ec5ad9320637a4d8b431a1b

PR #929, opened 2022-05-12
Add support for :oneof fields on protobuf messages

andrewn617:
### Motivation
Google Protobuf messages support `oneof` field types:

```
add_message "MyCart" do
  oneof :contact_info do
    optional :phone_number, :int32, 1
    optional :email, :string, 2
  end
end
```

`MyCart` will have a `#contact_info` method that returns the name of the optional that has been used (or nil).

This PR adds a method definition to RBIs for messages with `oneof` fields.

### Implementation
`Descriptor` provides an `each_oneof` method. I iterate over the `oneof` descriptors and create an RBI method for each.

### Tests
👍

### DIFF
--- a/lib/tapioca/dsl/compilers/protobuf.rb
+++ b/lib/tapioca/dsl/compilers/protobuf.rb
@@ -83,6 +83,7 @@ module Tapioca
               create_type_members(klass, "Key", "Value")
             else
               descriptor = T.let(T.unsafe(constant).descriptor, Google::Protobuf::Descriptor)
+              descriptor.each_oneof { |oneof| create_oneof_method(klass, oneof) }
               fields = descriptor.map { |desc| create_descriptor_method(klass, desc) }
               fields.sort_by!(&:name)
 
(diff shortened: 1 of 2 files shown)


plcstpierre (approved):

RayLLiu (approved):

asoules (approved):
🏆

paracycle (approved):
I wasn't sure how `oneof` fields worked, and it was not clear to me why the method would be returning a nilable `String`. So I looked up the [documentation](https://developers.google.com/protocol-buffers/docs/reference/ruby-generated#oneof) for it and understood that the `oneof` field holds the name of the field that ends up being populated.

Here is a code snippet that describes the situation for everyone's sake:
```ruby
cart = MyCart.new

# Fields have their defaults.
raise unless cart.email == ""
raise unless cart.phone_number == 0
raise unless cart.contact_info == nil

cart.email = "foo@example.com"
raise unless cart.email == "foo@example.com"
raise unless cart.phone_number == 0
raise unless cart.contact_info == :email

# Setting phone_number clears email.
cart.phone_number = 5552456789
raise unless cart.email == ""
raise unless cart.contact_info == :phone_number

# Setting phone_number to nil clears the oneof.
cart.phone_number = nil
raise unless cart.contact_info == nil
```

PR #932, opened 2022-05-13
Always run tests with `--no-doc` unless specified otherwise

Morriar:
### Motivation

Following https://github.com/Shopify/tapioca/pull/920, running the test suite increased from 10 to 40 minutes because we spend a lot of time parsing and generating the documentation in the gem RBIs.

### Implementation

Always pass `--no-doc` unless `--doc` is passed explicitly.

### Tests

See automated tests.


Morriar:
As shown [here](https://github.com/Shopify/tapioca/actions/runs/2321951584), we're back to ~10min 🎉 

image

KaanOzkan (approved):

PR #933, opened 2022-05-13
Handle gracefully if we get unparsable JSON

paracycle:
### Motivation

We seem to sometimes not read enough bytes from the Sorbet output, which ends up causing us to get an error from the JSON parser. While we investigate and fix that root cause, we might as well gracefully handle JSON parsing errors.

### Implementation

Rescue JSON parsing errors

### Tests

No tests, since I don't know how to recreate this situation.


Morriar (approved):

Morriar:
Should we provide a warning or something?

paracycle:
> Should we provide a warning or something?

I mean this to be temporary to be honest, so I didn't do that.

PR #943, opened 2022-05-24
Many protubuf dsl fixes

slinkp:
### Motivation

**I will probably abandon this and pull bits of it into smaller PRs after #966 merges**

Generated RBIs differ in many ways from the ruby protobuf documentation and the code actually generated by `protoc`.
Attempting to fix as many of them as possible.

**Ignore spurious gemfile & version bumps** - will revert those

**Rebase badly needed**

- [x] All parameters to `#initialize` should actually be nilable
- [x] [Per docs](https://developers.google.com/protocol-buffers/docs/reference/ruby-generated#embedded_message): Readers for fields of `:message` type can return `nil`. Setters for fields of `:message` type can accept `nil`. Primitive field types (`string` et al) can't be assigned nor return `nil`.  (But see the previous point - you can still pass nil to `#initialize`)
- [x] There's a generated [`clear_foo` method for all fields](https://developers.google.com/protocol-buffers/docs/reference/ruby-generated#checking-presence), was missing.
- [x] `:enum` fields can accept or return Integer or Symbol [per the docs](https://developers.google.com/protocol-buffers/docs/reference/ruby-generated#enum).
- [x] TODO: there's a missing type declaration for Enum values.  I've seen these shimmed like `FOO = type_alias(T.untyped)` but maybe should be `FOO = type_alias(T.any(Integer, Symbol))` ?
- [ ] TODO: The docs say there should be a [`has_foo?` generated method for all non-repeated fields](https://developers.google.com/protocol-buffers/docs/reference/ruby-generated#fields), but it doesn't seem to actually exist 🤔  ... maybe a more recent version of `protoc` than I have?
- [x] TODO: [per the docs there are module methods on Enums:](https://developers.google.com/protocol-buffers/docs/reference/ruby-generated#enum) `#lookup`, `#resolve`, `#descriptor` and we are missing declarations for those

### Implementation

Building on top of Ray's `make_protobuf_signature_nilable` [PR](https://github.com/Shopify/tapioca/pull/931)

The code with my changes is getting a bit hard to follow and could use some OO refactoring.
In particular, the handling of repeated and map fields is not consistent with the others.
In current state, i mainly wanted to see if I could get something working and confirm the RBI output.

### Tests

Yes, following the existing approach.  However, we are lacking tests that the generated sigs actually work as intended

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -8,7 +8,7 @@ GIT
 PATH
   remote: .
   specs:
-    tapioca (0.8.10.pre)
+    tapioca (0.8.13.pre)
       bundler (>= 1.17.3)
       parallel (>= 1.21.0)
       pry (>= 0.12.2)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -8,7 +8,7 @@ GIT
 PATH
   remote: .
   specs:
-    tapioca (0.8.15.pre)
+    tapioca (0.8.16.pre)
       bundler (>= 1.17.3)
       parallel (>= 1.21.0)
       pry (>= 0.12.2)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -8,7 +8,7 @@ GIT
 PATH
   remote: .
   specs:
-    tapioca (0.8.0.pre)
+    tapioca (0.8.1.pre)
       bundler (>= 1.17.3)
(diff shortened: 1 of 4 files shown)


slinkp:
These features all seem to exist now 🎆 🍾

PR #944, opened 2022-05-26
Fix type variable patch to respect references to late defined types

paracycle:
### Motivation

Now that Sorbet syntax for declaring type variables is via a block, code authors can very well refer from that block to types that have not been defined yet. Since the values of `fixed`, `lower` and `upper` are late evaluated, this should work normally.

However, our current attempt to retrofit the new type variable syntax was trying to early evaluate the block that was passed so that we could unify on the old syntax. However, that early evaluation breaks for the case outlined above.

### Implementation

We should instead convert the bounds passed to type variables into the block format, if a block is not passed, and always work with blocks until the point where we need to serialize the type variable. We should be pretty guaranteed that the types referred to from the block will be defined by the time we get to serialize the type variable. In any case, that is as late we can evaluate the bounds anyway.

### Tests

Added a test that breaks without this change and is fixed by it.

### DIFF
--- a/lib/tapioca/sorbet_ext/generic_name_patch.rb
+++ b/lib/tapioca/sorbet_ext/generic_name_patch.rb
@@ -21,47 +21,33 @@ module T
         Tapioca::Runtime::GenericTypeRegistry.register_type(constant, types)
       end
 
-      def type_member(variance = :invariant, fixed: nil, lower: nil, upper: nil, &blk)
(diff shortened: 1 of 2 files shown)


egiurleo (approved):
Definitely get eyes from someone else, but this change makes sense to me wrt pushing the evaluation of the block into the `serialize` method. I also find this refactor more readable!

Issue #945, opened 2022-05-26
Document default behavior for `tapioca gem`


egiurleo:
By default, running `tapioca gem` will only generate RBIs for gems that have been added or removed to a project. This is not explicitly stated in the README, which has the potential to cause some confusion.

We should make it clear that this is the default behavior of the `tapioca gem` command.

PR #954, opened 2022-05-31
Improve README documentation about `tapioca gem --all`

Morriar:
### Motivation

Closes https://github.com/Shopify/tapioca/issues/945.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,12 @@ generate RBIs from gems
 ```
 <!-- END_HELP_COMMAND_GEM -->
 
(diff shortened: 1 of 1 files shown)


vinistock (approved):

egiurleo (approved):
🎉 Yay thank you!

README.md
```diff
@@ -163,6 +163,12 @@ generate RBIs from gems
 ```
 <!-- END_HELP_COMMAND_GEM -->
 
+By default, running `tapioca gem` will only generate the RBI files for gems that have been added to or removed from the project's `Gemfile` this means that Tapioca will not regenerate the RBI files for untouched gems. However, when changing Tapioca configuration or bumping its version, it may be useful to force the regeneration of the RBI files previsouly generated. This can be done with the `--all` option:
```
paracycle, line 166: Gems added/removed or had their versions changed, to be more precise. I'll PR an addition to make that more clear.

PR #955, opened 2022-05-31
Handle multiple annotations repositories

Morriar:
### Motivation

Allow user to use multiple sources for annotations:

```sh
tapioca annotations --repo-uri URI1 URI2
```

### Implementation

1. First we pull each repo's index.
2. For each gem in the Gemfile.lock we check which indexes contain annotations files for the gem
3. Pull all the annotations files from all the repos containing one for the gem
4. Merge the annotations files into a single one

In case one of the annotation files contains a parse error, we tell so to the user and skip the gem.
In case one of annotations files can't be merged because of conflicts, we tell so to the user and skip the gem.

### Tests

See automated tests.


lib/tapioca/cli.rb
```diff
@@ -302,9 +302,9 @@ def check_shims
     end
 
     desc "annotations", "Pull gem annotations from a central RBI repository"
-    option :repo_uri, type: :string, desc: "Repository URI to pull annotations from", default: CENTRAL_REPO_ROOT_URI
+    option :repo_uri, type: :array, desc: "Repository URI to pull annotations from", default: [CENTRAL_REPO_ROOT_URI]
```
vinistock, line 305: Since it accepts an array, could we change the option name to plural (repo_uris) and also update the description?

lib/tapioca/commands/annotations.rb
```diff
@@ -60,10 +60,19 @@ def remove_expired_annotations(project_gems)
...
 
-      sig { returns(RepoIndex) }
-      def fetch_index
-        say("Retrieving index from central repository... ", [:blue, :bold])
-        content = fetch_file(CENTRAL_REPO_INDEX_PATH)
+      sig { returns(T::Hash[String, RepoIndex]) }
+      def fetch_indexes
+        repo_number = 1
+        @central_repo_root_uris.each_with_object({}) do |uri, hash|
+          hash[uri] = fetch_index(uri, repo_number: @central_repo_root_uris.size > 1 ? repo_number : nil)
```
st0012, line 67: `@central_repo_root_uris.size > 1` can probably be a local? It doesn't need to be called every time.

lib/tapioca/commands/annotations.rb
```diff
@@ -74,22 +83,33 @@ def fetch_index
...
+        end, T::Hash[String, T::Array[String]])
 
         if fetchable_gems.empty?
           say(" Nothing to do")
           exit(0)
         end
 
         say("\n")
-        fetched_gems = fetchable_gems.select { |name| fetch_annotation(name) }
+        fetched_gems = fetchable_gems.select { |gem_name, repo_uris| fetch_annotation(repo_uris, gem_name) }
```
st0012, line 101: Nitpick: the opposite order of block args and `fetch_annotation` params feels odd. Is it possible to reverse the order of `fetch_annotation` params? I think `fetch_annotation(gem_name, repo_uris)` also makes sense and reads better here.

Morriar: I kept the same order than the other methods (`fetch_index`, `fetch_local_file`, etc.) where the repo URI comes first.

lib/tapioca/cli.rb
Morriar: I renamed it as `--repos` (https://github.com/Shopify/tapioca/pull/955/commits/98663d1a602a8343f0f880aebad6a92ad101c7ad) let me know what you think.

Morriar:
I also added some documentation for the command: https://github.com/Shopify/tapioca/pull/955/commits/47539bdca305b5710cefd39436ac98818b3c15da.

KaanOzkan (commented):
Do you think allowing multiple repos will hinder contributions to `rbi-central`?

From that perspective I wasn't sure if it's a necessary feature without support for private repos.

Issue #962, opened 2022-06-03
`annotations` command should allow using private repository


Morriar:
So users can add annotations to private gems.

PR #963, opened 2022-06-03
Support authentification for private annotations repositories

Morriar:
### Motivation

Allow users to import RBI annotations from private repositories. This is useful when using private gems for example.

Closes https://github.com/Shopify/tapioca/issues/962.

### Implementation

Since we access the private repository files through http, the easiest way to go is to pass the `Authentification` header as well. This header can be provided by the user through the `--auth` option:

```shell
$ bin/tapioca annotations --sources https://raw.githubusercontent.com/$USER/$PRIVATE_REPO/$BRANCH --auth "token $TOKEN"
```

See the readme for more information.

### Next step

Once we are able to provide the authentication (this PR), the next logical step will be to read this authentication from the [netrc](https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html) file.

### Tests

See automated tests.


KaanOzkan (approved):
We aren't testing the `auth` option but I assume you manually tested it.

Morriar:
As discussed offline, I added a test using `webmock` to ensure the authorization header was passed properly 👍

PR #965, opened 2022-06-03
Fix CI

Morriar:
### Motivation

Temporary fix until https://github.com/rubyconfig/config/pull/319 is merged and released.

Fixes intermittent CI failures such as https://github.com/Shopify/tapioca/runs/6730332048?check_suite_focus=true#step:6:35.

### Implementation

Making sure `rails` is required before requiring `config`.

Depending on the order files are loaded we'll require a sub-part of Rails which will define the constant `Rails` and make loading `config` crash.

### Tests

Green CI.

### DIFF
--- a/tasks/generator_documentation.rake
+++ b/tasks/generator_documentation.rake
@@ -1,6 +1,12 @@
 # typed: true
 # frozen_string_literal: true
 
+# Temporary fix for CI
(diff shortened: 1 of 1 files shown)


egiurleo (approved):
🙏🏻 Thank you!

vinistock (approved):

RyanBrushett (approved):

PR #967, opened 2022-06-03
Refactor the interesting_ancestors_of method in the Mixins tracker

egiurleo:
### Motivation
Use `Set#compare_by_identity` to make the logic of the `interesting_ancestors_of` method easier to understand.

### Implementation
The old implementation worked as follows:
- Create `inherited_ancestors` as a new Set, and add all the object_ids of the modules in `inherited_ancestors_of(constant)`
- Iterate through the modules in `constant.ancestors` and reject any whose object_id is already in `inherited_ancestors_ids`

The new implementation is much more straightforward:
- Create `inherited_ancestors` as a new Set with `compare_by_identity`
- Add the results of `inherited_ancestors_of(constant)` to this set
- Create `ancestors` as a new Set with `compare_by_identity`
- Add the results of `ancestors_of(constant)` to this set
- Subtract `inherited_ancestors` from `ancestors` and return the result as an array

### Tests
Behavior should not change. Tests are still green.


vinistock (approved):
Nice refactor

lib/tapioca/gem/listeners/mixins.rb
```diff
@@ -71,9 +71,9 @@ def filtered_mixin?(mixin_name)
 
         sig { params(constant: Module).returns(T::Array[Module]) }
         def interesting_ancestors_of(constant)
-          inherited_ancestors_ids = Set.new(
-            inherited_ancestors_of(constant).map { |mod| object_id_of(mod) }
-          )
+          inherited_ancestors = Set.new.compare_by_identity
+          inherited_ancestors.merge(inherited_ancestors_of(constant))
```
paracycle, line 75: I think we can do all of this on a single line:

```suggestion
          inherited_ancestors = Set.new.compare_by_identity.merge(inherited_ancestors_of(constant))
```

lib/tapioca/gem/listeners/mixins.rb
```diff
@@ -91,9 +91,10 @@ def interesting_ancestors_of(constant)
           #
           # Instead, we should only drop the tail matches of the ancestors and
           # inherited ancestors, past the location of the constant itself.
-          constant.ancestors.reject do |mod|
-            inherited_ancestors_ids.include?(object_id_of(mod))
-          end
+          ancestors = Set.new.compare_by_identity
+          ancestors.merge(ancestors_of(constant))
```
paracycle, line 95: Same as above:

```suggestion
          ancestors = Set.new.compare_by_identity.merge(ancestors_of(constant))
```

paracycle (approved):
Nice. I just have a suggestion for further shortening.

egiurleo:
@paracycle beautiful! Implemented your one-liners.

PR #968, opened 2022-06-03
Support generating DSL RBI by path

jeffcarbs:
### Motivation

We often want to generate RBI for every model, worker, etc inside a namespace. Currently the only way to do that involves crawling ObjectSpace in a console or doing grep/sed magic to get the set of constants. We wanted an easier way to do this.

Another future-facing use-case is being able to run `tapioca dsl --verify some/specific/path`.
- Sidenote: At the moment, `--verify` fails unless it's called with no arguments since it thinks there's stale RBI, but that's something that can be fixed separately.

### Implementation

I thought it would be most ergonomic to be able to pass directory/filenames as args to the `tapioca dsl` but currently it's assumed that any non-flag args passed will be class names. Luckily, ruby class names and file names follow disjoint patterns:
- class names must start with a capital letter or colon
- path names typically (99.9% of path names I've ever seen) start with a lower case letter or slash.

Given that, we use that heuristic to split the passed args into either class names or paths and then treat them appropriately downstream. We then use `Tapioca::Static::SymbolLoader.symbols_from_paths` to get all the class names from the given paths.

Wanted to call out a slight difference in the error handling for paths. `tapioca dsl SomeClass SomeOtherClass` currently fails if none of those classes have processable DSL. This makes sense since you're specifically giving classes for which you want RBI generated. For paths, on the other hand you may not know if there's RBI to generate and I think that OK so we actually exist successfully if you've given paths but there's no constants to process. The one exception is if you've given use no paths that actually exist, in which case we'll exit with an error which should prevent typos like `tapioca dsl app/mode`.

### Tests

See tests

### DIFF
--- a/lib/tapioca/commands/dsl.rb
+++ b/lib/tapioca/commands/dsl.rb
@@ -10,6 +10,7 @@ module Tapioca
       sig do
         params(
           requested_constants: T::Array[String],
+          requested_paths: T::Array[Pathname],
           outpath: Pathname,
           only: T::Array[String],
           exclude: T::Array[String],
@@ -27,6 +28,7 @@ module Tapioca
       end
       def initialize(
         requested_constants:,
+        requested_paths:,
         outpath:,
         only:,
         exclude:,
@@ -42,6 +44,7 @@ module Tapioca
         app_root: "."
       )
         @requested_constants = requested_constants
+        @requested_paths = requested_paths
         @outpath = outpath
         @only = only
         @exclude = exclude
@@ -63,7 +66,7 @@ module Tapioca
       def list_compilers
         Loaders::Dsl.load_application(
           tapioca_path: @tapioca_path,
-          eager_load: @requested_constants.empty?,
+          eager_load: @requested_constants.empty? && @requested_paths.empty?,
           app_root: @app_root,
         )
 
@@ -100,6 +103,15 @@ module Tapioca
           say("Compiling DSL RBI files...")
         end
         say("")
+
+        unless @requested_paths.empty?
+          constants_from_paths = Static::SymbolLoader.symbols_from_paths(@requested_paths).to_a
+          if constants_from_paths.empty?
+            say_error("\nWarning: No constants found in: #{@requested_paths.map(&:to_s).join(", ")}", :yellow)
+          end
+
+          @requested_constants += constants_from_paths
+        end
 
         outpath = @should_verify ? Pathname.new(Dir.mktmpdir) : @outpath
         rbi_files_to_purge = existing_rbi_filenames(@requested_constants)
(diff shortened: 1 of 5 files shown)


lib/tapioca/commands/dsl.rb
```diff
@@ -222,6 +230,25 @@ def constantize_compilers(compiler_names)
...
+        name = symbol_table_json.fetch("name")
+        return [] unless kind == "CLASS_OR_MODULE" && name.fetch("kind") == "CONSTANT"
+
+        name = name.fetch("name")
+        children = symbol_table_json.fetch("children", [])
+
+        prefixed_name = name == "<root>" ? prefix : [prefix, name].compact.join("::")
+        [prefixed_name, *children.flat_map { |c| constants_from_symbol_table_json(c, prefixed_name) }].compact
+      end
+
```
Morriar, line 251: I think this logic should be moved to https://github.com/Shopify/tapioca/blob/main/lib/tapioca/static/symbol_loader.rb.

spec/tapioca/cli/dsl_spec.rb
```diff
@@ -647,6 +647,63 @@ class User; end
...
+          OUT
+
+          assert_empty_stderr(result)
+
+          assert_project_file_exist("sorbet/rbi/dsl/post.rbi")
+          refute_project_file_exist("sorbet/rbi/dsl/user.rbi")
+
+          assert_success_status(result)
+        end
+
```
Morriar, line 706: Can we add a test showing what happens when passing a wrong path `tapioca dsl path/to/nowhere.rb`? 🙏

Morriar (commented):
I like where this is going ❤️

lib/tapioca/commands/dsl.rb
jeffcarbs: Good call. I actually read through that code a bit more and realized it was going exactly what I was trying to do so we can just switch to using it directly. Will push up a commit shortly.

Morriar (commented):
There are a few conflicts with `main` and  https://github.com/Shopify/tapioca/pull/968#discussion_r891530111 hasn't been addressed.

jeffcarbs:
@paracycle @Morriar - I'm picking this PR back up. I rebased from the latest `main` branch and updated some things based on the conversation above.

Regarding whether to exit with an error/success, I think I wound up a nice middle-ground:
- if you gave us paths that actually exist and there just happen to be no processable constants, that will exit successfully
  - This is necessary to support running this as a pre-commit hook which would just pass the paths to any files that have changed. It's possible/likely that the files given just happen to not have any constants to process and that shouldn't be an error.
- if you didn't give us at least one valid path and there are no processable constants, that will exit with error.
  - This should hopefully be enough to catch typos where you gave us a non-existent path (e.g. a typo).

paracycle (approved):

PR #972, opened 2022-06-06
Fix CI in a better way

Morriar:
### Motivation

The upstream fixes about `Railtie` have been merged:
* https://github.com/rubyconfig/config/pull/319
* https://github.com/byroot/frozen_record/pull/52

We can now revert the temporary fix from #965.

### Tests

CI is green 💚


Morriar:
> We may want to stop pointing at the branches when the gems are released. Should we add a todo?

Since it's only development dependencies I'm not too worried.

PR #976, opened 2022-06-09
Support for activerecord-typedstore v1.5.0

etiennebarrie:
### Motivation

The internals changed in https://github.com/byroot/activerecord-typedstore/pull/91 and `store.accessors` is now a Hash instead of an Array. The value is the name of the method for which we need to compile the RBI.

### Implementation

I implemented it in a way that should be compatible with the previous version.

Instead of `flat_map(&:accessors).empty?`, which used to give `[]` but now gives `[{}]` if a store has no field, I use `all?`, and I use the fact that Hash objects yield two values to `each` to preserve the previous behavior.

### Tests

I made the changes first without the gem bump, to show it's still the same behavior, then the gem bump and the additional test for the new feature.

Closes #949

### DIFF
--- a/spec/tapioca/cli/dsl_spec.rb
+++ b/spec/tapioca/cli/dsl_spec.rb
@@ -41,6 +41,13 @@ module Tapioca
           @project.require_real_gem("smart_properties", "1.15.0")
           @project.require_real_gem("sidekiq", "6.2.1")
           @project.bundle_install
+          @gemfile = @project.read("Gemfile")
+          @gemfile_lock = @project.read("Gemfile.lock")
+        end
+
+        before do
+          @project.write("Gemfile", @gemfile)
+          @project.write("Gemfile.lock", @gemfile_lock)
         end
 
         after do


lib/tapioca/dsl/compilers/active_record_typed_store.rb
```diff
@@ -95,17 +95,18 @@ class ActiveRecordTypedStore < Compiler
...
           stores = constant.typed_stores
-          return if stores.values.flat_map(&:accessors).empty?
+          return if stores.values.all? { |store| store.accessors.empty? }
 
           root.create_path(constant) do |model|
             stores.values.each do |store_data|
-              store_data.accessors.each do |accessor|
-                field = store_data.fields[accessor]
+              store_data.accessors.each do |accessor, name|
+                field = store_data.fields.fetch(accessor)
```
Morriar, line 103: Does this change means we're not going to be retro-compatible with `activerecord-typedstore < 1.5.0`?

etiennebarrie: But this change here going from `fields[accessor]` to `fields.fetch(accessor)` is not strictly necessary.

paracycle (approved):
I like this change, thank you!

paracycle:
> Instead of just fixing the single test (I could save and restore Gemfile/Gemfile.lock just in the new test), I went ahead and made the setup a bit more reproducible for all tests.

I like this approach and I think it is fine having it inside this PR. I think we can extract that to the general `@project` setup/teardown, so that we don't have to worry about this kind of cross-test leaking.

> I also found a flakiness issue where one gem was added but not bundled installed, so in some cases it caused an additional Resolving dependencies... line coming from Bundler. Let me know if you'd rather have those in a separate PR.

This is also a good fix and I think we can make this better later as well, e.g. by running `bundle install`, if needed, inside the `@project.tapioca` call itself.

PR #979, opened 2022-06-13
Add `--environment` flag to `dsl`/`gem` commands

### LINKED ISSUE #657: Best practice RAILS_ENV use for --verify
Running --verify on my CI failed because RBIs were not there in RAILS_ENV=test, which of course my CI runs on. And of course locally, RAILS_ENV is dev.

In dev environment, we have unique gems loaded to help our work (Gemfile group :development), and custom routes (e.g. sidekiq web). These don't need to be included in the DSLs, but are by default.

There are several options here -- what's considered best practice?
1. Always generate and verify tapioca in RAILS_ENV=test
2. Use some other rails env (production?) for tapioca -- though this will have database connection issues
2. Create a wrapper for tapioca generation that ignores the development gems. But how to skip routes?
3. Something else?

paracycle:
### Motivation

Fixes #657 

Tapioca is primarily a development tool, and thus, should be used to load applications/gems in `development` mode by default. Up to now, however, we assumed that we could just be not opinionated about this, which led to mismatch between the output of the commands locally and on CI system (which usually run in `test` environment).

By adding an `--environment` flag that defaults to `development`, we can allow users to use the sensible default, but also to override it.

### Implementation

Add the flag to `dsl` and `gem` commands and set environment variables for `RAILS_ENV` and `RACK_ENV` as early as possible with that value.

### Tests

Added CLI tests for the default value and supplied value of the environment flag.


lib/tapioca/cli.rb
```diff
@@ -321,6 +335,10 @@ def __print_version
       def self.exit_on_failure?
         true
       end
+
+      def set_environment
```
Morriar, line 339: Should we move this to [cli_helper](https://github.com/Shopify/tapioca/blob/main/lib/tapioca/helpers/cli_helper.rb) so we can attach a signature to it?

Or maybe this should be called from inside the command and thus defined inside the base [command](https://github.com/Shopify/tapioca/blob/main/lib/tapioca/commands/command.rb)?

spec/tapioca/cli/dsl_spec.rb
```diff
@@ -1590,6 +1590,102 @@ def bar(&block); end
           refute_success_status(result)
         end
       end
+
+      describe "environment" do
+        before(:all) do
+          @project.tapioca("init")
+
+          @project.write("lib/post.rb", <<~RB)
+            require "smart_properties"
```
Morriar, line 1599: I wasn't sure why we needed `smart_properties` in this test? I guess nothing is happening is we don't have a real DSL to trigger RBI generation?

Morriar (approved):

spec/tapioca/cli/dsl_spec.rb
paracycle: Yes, that's the reason, I wanted to make sure that the RBIs generated were different if different environments defined constants differently. Alternatively I could have used a custom DSL compiler, but Smart Properties is just easier to use.

lib/tapioca/cli.rb
```diff
@@ -318,9 +333,15 @@ def __print_version
     end
 
     no_commands do
+      sig { returns(T::Boolean) }
```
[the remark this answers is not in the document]
paracycle: Ah ok, the runtime is failing because of this: https://github.com/sorbet/sorbet/issues/5786 

I'll move it to `CliHelper` instead.

lib/tapioca/helpers/env_helper.rb
```diff
@@ -0,0 +1,16 @@
+# typed: true
+# frozen_string_literal: true
+
+module Tapioca
+  module EnvHelper
+    extend T::Sig
+    extend T::Helpers
+
+    requires_ancestor { Thor }
```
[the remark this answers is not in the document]
Morriar: Urgh alright, let's keep a note that we need to fix this as well 😱

Issue #993, opened 2022-06-15
Improper loading of Rails engines 


wildmaples:
# Issue 

When `tapioca gems` command is run, rails engines are loaded in the [`Tapioca::Runtime::Loader#load_rails_engines` method](https://github.com/Shopify/tapioca/blob/main/lib/tapioca/runtime/loader.rb#L103), where engines are sorted then required/loaded. 

Any [`LoadError` exceptions are rescued](https://github.com/Shopify/tapioca/blob/main/lib/tapioca/runtime/loader.rb#L110-L112) and files are saved to be [reloaded again](https://github.com/Shopify/tapioca/blob/main/lib/tapioca/runtime/loader.rb#L115-L121). 

This causes issues because if the files are loaded incorrectly both times, they are never loaded. This cascades down to `uninitialized constant (NameError)` later when the unloaded constants are called, which is something we're currently seeing on Shopify/shopify.  

# Questions
- Why are files from the engine being sorted and loaded this way? Is there an assumption that there is autoload in place? I don't think sorting and loading them is the correct way to do it. The arbitrary loading order seems to be the root of this issue. 
- Do we need to load files from rails engines? Rails engines are listed in the Gemfile but they are part of the Rails application so it seems like they are in the gray area of whether or not Tapioca is responsible for them.

paracycle:
> Not all engines. The Active Storage engine is not part of the Rails application, or the devise engine. I think we don't need to load engines that are inside the application directory though.

Yes, this is the correct fix, which I was going to suggest as well.

paracycle:
Btw, it is also true that our Rails engine loading code is very hacky and does not work well. There is even an issue #671 that tracks that generally. But, it is still worthwhile to exclude any application engines from the loading process.

KaanOzkan:
https://github.com/Shopify/tapioca/pull/995

PR #995, opened 2022-06-15
Filter engines in application directory so they are not loaded

wildmaples:
### Motivation

Solves: https://github.com/Shopify/tapioca/issues/993

Also fixes an internal Shopify/shopify issue 

### Implementation

As Rafael suggested in https://github.com/Shopify/tapioca/issues/993#issuecomment-1156876059, add a filter to reject engines that are part of the application directory 

### Tests

Change is part of the application loading - can this be tested?

### DIFF
--- a/lib/tapioca/runtime/loader.rb
+++ b/lib/tapioca/runtime/loader.rb
@@ -5,6 +5,7 @@ module Tapioca
   module Runtime
     class Loader
       extend(T::Sig)
+      include Tapioca::GemHelper
 
       sig do
         params(gemfile: Tapioca::Gemfile, initialize_file: T.nilable(String), require_file: T.nilable(String)).void
@@ -123,16 +124,6 @@ module Tapioca
             nil
           end
         end
-      end
-
-      sig { params(gemfile_dir: String, full_gem_path: String).returns(T::Boolean) }
-      def gem_in_app_dir?(gemfile_dir, full_gem_path)
-        !gem_in_bundle_path?(full_gem_path) && full_gem_path.start_with?(gemfile_dir)
-      end
-
-      sig { params(full_gem_path: String).returns(T::Boolean) }
-      def gem_in_bundle_path?(full_gem_path)
-        full_gem_path.start_with?(Bundler.bundle_path.to_s, Bundler.app_cache.to_s)
       end
     end
   end
--- a/lib/tapioca/runtime/loader.rb
+++ b/lib/tapioca/runtime/loader.rb
@@ -53,9 +53,13 @@ module Tapioca
         return [] unless Object.const_defined?("Rails::Engine")
 
         safe_require("active_support/core_ext/class/subclasses")
+        root = File.expand_path(".")
 
         # We can use `Class#descendants` here, since we know Rails is loaded
-        Object.const_get("Rails::Engine").descendants.reject(&:abstract_railtie?)
+        Object.const_get("Rails::Engine")
+          .descendants
+          .reject(&:abstract_railtie?)
+          .reject { |engine| File.fnmatch?("#{root}/**", engine.config.root) }
       end
 
       sig { params(path: String).void }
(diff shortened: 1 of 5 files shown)


lib/tapioca/runtime/loader.rb
```diff
@@ -53,9 +53,13 @@ def rails_engines
...
 
         safe_require("active_support/core_ext/class/subclasses")
+        root = File.expand_path(".")
 
         # We can use `Class#descendants` here, since we know Rails is loaded
-        Object.const_get("Rails::Engine").descendants.reject(&:abstract_railtie?)
+        Object.const_get("Rails::Engine")
+          .descendants
+          .reject(&:abstract_railtie?)
+          .reject { |engine| File.fnmatch?("#{root}/**", engine.config.root) }
```
paracycle, line 62: I am afraid this is not enough to check if a gem lives under the application or not. The complications are for gems that are installed under the `vendor` folder inside the application folder, for example.

`Tapioca::Gemfile::GemSpec` class has the method `gem_in_app_dir?`, for example, that checks if any gem files are inside the directory that contains the `Gemfile`, which would also be appropriate here. That method explicitly excludes cases where a given path is under the `Bundler.bundle_path` or `Bundler.app_cache` folders.

lib/tapioca/runtime/loader.rb
```diff
@@ -53,9 +53,13 @@ def rails_engines
         return [] unless Object.const_defined?("Rails::Engine")
 
         safe_require("active_support/core_ext/class/subclasses")
+        root = File.expand_path(".")
```
paracycle, line 56: As mentioned above, a better heuristic for the root folder is to either use `Rails.root` or `Bundler.default_gemfile`.

paracycle (changes requested):

wildmaples:
So I duplicated `#gem_in_app_dir?` from Gemspec in https://github.com/Shopify/tapioca/pull/995/commits/2f0d6a61a7e2775e038ec7ef2fbaa8c9c6765003 because I wanted to confirm the approach works.  

This is, however, duplicated code.`Loader` is does not have convenient access to `Gemspec` unfortunately, and I don't think it's a good idea for it to have access to the instances of `Gemspec`. Both classes have different jobs and are part of different stages in Tapioca's execution. With that, another option is to extract the logic into a utility module that can be used by both `Loader` and `Gemspec`. What do you think?

rafaelfranca:
> With that, another option is to extract the logic into a utility module that can be used by both `Loader` and `Gemspec`

I think that is a valid way to share code. If you can use composition for this it would be better in my opinion but also totally fine to use inheritance.

Morriar:
I think we can extract those methods under a new `helpers/gem_helper` module and include it in both places.

@rafaelfranca for composition are you suggesting making it a class that is instantiated in both places? It doesn't seem like those methods require any state?

Morriar (approved):
> `tapioca gem --all` runs successfully.

It's also good to check if `srb tc` finished without any unexpected errors. This may be the sign we're missing some constants that should be generated.

wildmaples:
Thanks for the review! I ran `srb tc` and there are errors so I'm gonna investigate to see if they are real changes before merging this.

KaanOzkan:
@DavidPD Since we don't support Ruby 2.6 (EOL) I think this would set a bad precedent for the repo. However, I'm confident that we won't backport anything else to `0.7.x` so if you were to point to your personal fork you wouldn't be at a disadvantage compared to sticking to tapioca `0.7.x`.

So I suggest you fork the project, backport the fix and point tapioca to your fork. This might even be the better long term solution if you're sticking to Ruby 2.6 for a while since you won't be dependent on us for other backports. In the future when you can upgrade Ruby you can point back to RubyGems releases.

I'm very sorry about the inconvenience, hopefully you can utilize latest Tapioca in no time since there have been a ton of improvements 🙂

DavidPD:
@KaanOzkan Thanks for the update. Yeah that's pretty much what we're doing short-term, and hopefully not _too_ far down the road we can get our project somewhat up-to-date.

PR #996, opened 2022-06-15
Clean `Gem::Pipeline` so it can be easily split with a superclass later

Morriar:
### Motivation

I wanted to extract a superclass from the current `Gem::Pipeline` but it proved a bit messier than what was required.

This PR applies some cleaning to make it easier.

### Tests

No behavioral change.


paracycle (approved):
I like this. One question: Is `skip_xxx?` better than `ignore_xxx?`?

rafaelfranca (approved):

PR #1007, opened 2022-06-21
Dsl::Compilers::ActiveRecordRelations: fix #find_sole_by, #sole

kivikakk:
### Motivation
The specs introduced for `sole` and `find_sole_by` in https://github.com/Shopify/tapioca/commit/5f90df17a9d2b6be03f44ad9cf605743cc17cf5a are not quite right.

See https://github.com/kivikakk/rails/blob/9e1cdf27ebaff31005f3d2bf02d97dfb116a3707/activerecord/lib/active_record/relation/finder_methods.rb#L107-L131.

### Implementation
Added logic for each of these to generated the correct signatures.

### Tests
Tests adjusted.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_relations.rb
+++ b/lib/tapioca/dsl/compilers/active_record_relations.rb
@@ -543,6 +543,23 @@ module Tapioca
                 ],
                 return_type: constant_name
               )
+            when :find_sole_by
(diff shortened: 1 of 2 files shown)


KaanOzkan (approved):

kivikakk:
Thanks so much! :blush:

PR #1010, opened 2022-06-23
Properly attribute mixin locations

egiurleo:
Part of #890.

### Motivation
The goal of this PR is to more accurately attribute mixin locations so that we only generate RBIs for mixins that occur in the current gem.

Currently, when determine where a mixin took place, we look at the stack trace at the time `extend`, `prepend`, or `include` is called. If any of the locations in the stack trace are in the current gem, then we attribute the mixin to the current gem.

This is problematic because of certain edge cases. Consider the case of `ActionController#helper_method` - it causes an `include` to be performed within the actionpack gem. However, we don't want RBIs for our gem to be generated for `actionpack` -- we want the mixin to be attributed to our gem.

Instead of examining the entire stack trace, we can find the one location within the stack trace where a mixin was required. This can help us identify the gem that is actually performing the mixin and avoid any confusion that may arise from metaprogramming edge cases.

### Implementation
1. Add the `required_from_location` to `Runtime::Reflection`, which returns the nearest location where `<top (required)>` is found.
2. Modify the `Mixin` tracker to associate all mixins with the location returned from `required_from_location`.
3. Update the `ForeignConstants` listener to use the modified tracker; remove some hacky behavior.

### Tests
All existing tests pass, and a new test is added that wouldn't pass without these changes.

### Tophatting 🎩 
I have generated gem RBIs in Shopify core with these changes.


spec/tapioca/cli/gem_spec.rb
```diff
@@ -1116,6 +1116,76 @@ module TypedParameters; end
...
+              def alert(*args, **_arg1, &block); end
+              def combined_fragment_cache_key(*args, **_arg1, &block); end
+              def content_security_policy?(*args, **_arg1, &block); end
+              def content_security_policy_nonce(*args, **_arg1, &block); end
+              def cookies(*args, **_arg1, &block); end
+              def form_authenticity_token(*args, **_arg1, &block); end
+              def notice(*args, **_arg1, &block); end
+              def protect_against_forgery?(*args, **_arg1, &block); end
+              def view_cache_dependencies(*args, **_arg1, &block); end
+            end
```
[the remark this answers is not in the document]
paracycle: This is a difference in behaviour between Ruby 2.7/3.0 and 3.1. It is related to how anonymous argument forwarding works.

lib/tapioca/runtime/reflection.rb
```diff
@@ -20,6 +20,8 @@ module Reflection
       PRIVATE_INSTANCE_METHODS_METHOD = T.let(Module.instance_method(:private_instance_methods), UnboundMethod)
       METHOD_METHOD = T.let(Kernel.instance_method(:method), UnboundMethod)
 
+      REQUIRED_FROM_LABELS = T.let(["<top (required)>", "<main>"].freeze, T::Array[String])
```
egiurleo, line 23: @Morriar How do you feel about this constant name/location?

spec/tapioca/cli/gem_spec.rb
```diff
@@ -1116,6 +1116,94 @@ module TypedParameters; end
...
+          some_engine = mock_gem("some_engine", "0.0.2") do
+            write("lib/some_engine.rb", <<~RUBY)
+              require "action_controller"
+
+              module SomeEngine
+                class SomeController < ActionController::Base
+                  # This method triggers a dynamic mixin
+                  # which should be attributed to this gem and not
+                  # actionpack, even though the real `include` happens
+                  # inside actionpack
```
Morriar, line 1129: Nitpick:

Do
we
really
need
that
many
lines
for
this
message?

I
though
the
limit
was
120
characters?

😂

Morriar (approved):

PR #1012, opened 2022-06-27
Filter out mixins that weren't performed in current gem

### LINKED ISSUE #890: Better mixin generation for gem RBIs

egiurleo:
Closes #890

### Motivation
This is the culmination of all the dynamic mixin work!

Currently, when RBIs are generated for mixins, tapioca doesn't take into account whether a mixin was performed in the current gem or in another gem. This results in strange scenarios where one gem will re-open a class to perform a mixin, and then that mixin is reflected in another gem's RBI. We should only generate RBI for mixins that are performed in the current gem.

### Implementation
This PR can be read commit-by-commit.
1. The first commit refactors the mixin tracker to remove the `mixin_locations_for` method, which is no longer used, and modify the `constants_with_mixin` method to keep track of mixin locations.
2. The second commit modifies the mixins pipeline listener to use the `mixin_locations_for` method to skip any mixins that weren't performed in the current gem.

### Tests
Modified existing tests to remove mixins that should be filtered out. Un-commented two lines in a cli test that now pass because of these changes!

### Tophatting 🎩 
TBD


lib/tapioca/gem/listeners/mixins.rb
```diff
@@ -26,26 +26,38 @@ def on_scope(event)
...
           mods
             .select do |mod|
               name = @pipeline.name_of(mod)
 
               name && !filtered_mixin?(name)
             end
             .map do |mod|
+              mixin_location = mixin_location(constant, mod, mixin_type)
+
+              # Some extend mixins are actually includes or prepend on the singleton class.
```
Morriar, line 52:
```suggestion
              # Some extend mixins are actually includes or prepends on the singleton class.
```

lib/tapioca/gem/listeners/mixins.rb
```diff
@@ -62,6 +74,40 @@ def add_mixins(tree, mods, mixin_type)
             end
         end
 
+        sig do
+          params(
+            location: String,
+          ).returns(T::Boolean)
+        end
+        def mixed_in_by_gem?(location)
```
Morriar, line 82: The method is `mixed_in_by_gem?` but works with a `location` only, which doesn't seem strictly related to mixins, should we rename it `location_in_gem?`?

Or maybe `mixed_in_by_gem?` should take a constant?

lib/tapioca/gem/listeners/mixins.rb
```diff
@@ -62,6 +74,40 @@ def add_mixins(tree, mods, mixin_type)
             end
         end
 
+        sig do
+          params(
+            location: String,
+          ).returns(T::Boolean)
+        end
```
Morriar, line 81: Let's inline signatures < 120 chars please 🙏

lib/tapioca/gem/listeners/mixins.rb
```diff
@@ -62,6 +74,40 @@ def add_mixins(tree, mods, mixin_type)
...
+            constant: Module,
+            mixin: Module,
+          ).returns(T.nilable(String))
+        end
+        def singleton_class_mixin_location(constant, mixin)
+          singleton_class = singleton_class_of(constant)
+          mixin_locations = Runtime::Trackers::Mixin.constants_with_mixin(mixin)
+
+          T.must(mixin_locations[Runtime::Trackers::Mixin::Type::Include])[singleton_class] ||
+            T.must(mixin_locations[Runtime::Trackers::Mixin::Type::Prepend])[singleton_class]
```
paracycle, line 108: Again, `dig` would be much nicer:
```suggestion
          mixin_locations.dig(Runtime::Trackers::Mixin::Type::Include, singleton_class) ||
            mixin_locations.dig(Runtime::Trackers::Mixin::Type::Prepend, singleton_class)
```

spec/tapioca/gem/pipeline_spec.rb
```diff
@@ -313,8 +313,6 @@ def hello; end
 
       object_output = template(<<~RBI)
         class Object < ::BasicObject
-          include ::Kernel
```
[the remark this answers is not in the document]
paracycle: Can we also remove the `gsub`s here: https://github.com/Shopify/tapioca/pull/1012/files#diff-98bc3f9ecbf2cc271100cfb6c25897bbfa5fc8ee53726bb0da2f67a702984facR321-R323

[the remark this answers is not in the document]
egiurleo: @paracycle For some reason that link doesn't have any highlighted lines or anything, and I can't find a gsub on the page. Can you tell me where you're seeing the gsub?

lib/tapioca/runtime/reflection.rb
```diff
@@ -168,6 +168,17 @@ def required_from_location
 
         required_location.absolute_path || ""
       end
+
+      sig { params(constant: Module).returns(T.nilable(String)) }
+      def constant_name_from_singleton_class(constant)
```
egiurleo, line 173: I moved these two methods into reflection because I had to use them in `Trackers::Mixin` to find the attached class based on a singleton class.

lib/tapioca/gem/listeners/foreign_constants.rb
```diff
@@ -53,11 +55,6 @@ def mixed_in_by_gem?(location)
           @pipeline.gem.contains_path?(location)
         end
 
-        sig { params(constant: Module).returns(T.nilable(String)) }
-        def constant_name_from_singleton_class(constant)
-          constant.to_s.match("#<Class:(.+)>")&.captures&.first
-        end
```
egiurleo, line 59: This method was moved into `Runtime::Reflection`

lib/tapioca/gem/listeners/mixins.rb
```diff
@@ -62,6 +65,23 @@ def add_mixins(tree, mods, mixin_type)
...
+        end
+        def mixed_in_by_gem?(constant, mixin, mixin_type)
+          mixin_location =
+            T.cast(
+              Runtime::Trackers::Mixin.constants_with_mixin(mixin).dig(mixin_type, constant),
+              T.nilable(String)
+            )
+
+          !!mixin_location && @pipeline.gem.contains_path?(mixin_location)
+        end
```
egiurleo, line 83: @Morriar I took your advice here and combined identifying the mixin location and determining whether it is in the gem path into one method. I did some stuff to get type checking to work, like casting the mixin location and then adding `!!` to have it always return a boolean. Let me know if there is a simpler way to resolve these issues?

lib/tapioca/runtime/trackers/mixin.rb
```diff
@@ -81,5 +79,21 @@ def extend_object(obj)
...
       super
     end
+
+    private
+
+    # Including or prepending on a singleton class is functionally equivalent to extending the
+    # attached class. Registering the mixin as an extend on the attached class ensures that
+    # this mixin can be found whether searching for an include/prepend on the singleton class
+    # or an extend on the attached class.
+    def register_extend_on_attached_class(constant)
```
egiurleo, line 89: This functionality was moved from `Listeners::Mixins` so that an include/prepend on a singleton class can also be found as an extend on the attached class.

spec/tapioca/gem/pipeline_spec.rb
```diff
@@ -313,16 +313,11 @@ def hello; end
...
-
           def hello; end
         end
       RBI
 
       compiled = compile
-        .gsub(/^\s+include ::Minitest::Expectations\s/, "")
-        .gsub(/^\s+include ::JSON::Ext::Generator::GeneratorMethods::Object\s/, "")
-        .gsub(/^\s+include ::PP::ObjectMixin\s/, "")
```
egiurleo, line 325: @paracycle Is this gsub you were talking about?

paracycle (approved):

PR #1018, opened 2022-06-28
Use default gemfile path instead of Rails root when listing engines

vinistock:
### Motivation

For projects that aren't Rails, such as Tapioca itself, `Rails.root` is `nil`, which then breaks when trying to invoke `Rails.root.to_path`.

### Implementation

Add a check for verifying that `Rails.root` is set before continuing down to list engines. This enables Tapioca to generate RBIs for itself again.

### Tests

I couldn't figure out a way of forcing `Rails.root` to return `nil`. Let me know if you have ideas.

### DIFF
--- a/lib/tapioca/runtime/loader.rb
+++ b/lib/tapioca/runtime/loader.rb
@@ -55,11 +55,12 @@ module Tapioca
 
         safe_require("active_support/core_ext/class/subclasses")
 
+        project_path = Bundler.default_gemfile.parent.expand_path
         # We can use `Class#descendants` here, since we know Rails is loaded
         Object.const_get("Rails::Engine")
           .descendants
           .reject(&:abstract_railtie?)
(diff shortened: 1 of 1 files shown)


Morriar (approved):

paracycle (changes requested):
I don't think this is the correct way to go about this, since the `Rails.root` is not even really needed except for filtering app engines.

I'd rather that we do the same thing we do for filtering out app gems, which is to use the folder of the `Gemfile` instead of something like `Rails.root`. Ref: https://github.com/Shopify/tapioca/blob/39b8a4b4ee9900bdb9a5e9f067afb98563412d82/lib/tapioca/gemfile.rb#L131-L133

RyanBrushett (approved):

paracycle (approved):

PR #1020, opened 2022-06-29
Test Tapioca own gem RBIs on CI

Morriar:
### Motivation

Allow us to catch more errors when pushing new changes.

This test is currently broken until we 1) fix Tapioca load order and 2) regenerate the RBIs.


RyanBrushett (approved):
Looks like it makes sense 👍

Morriar:
Superseded by https://github.com/Shopify/tapioca/pull/1021.

PR #1025, opened 2022-06-30
Add original gem source location to RBIs

vinistock:
### Motivation

Tapioca knows the original location of where methods, classes and modules were defined while generating the RBIs. By including this information as comments, we can enable go to gem source using the Ruby LSP quite easily.

### Implementation

The first commit adds the original method and passes it around the pipeline. The second commit creates a new listener that decorates all nodes with the original location information using the format `source://GEM_NAME-GEM_VERSION/path/to/file`.

We can use the `DocumentLink` request to connect these special URIs to the proper file system path and enable go to def.

### Tests

Added tests.


spec/tapioca/cli/gem_spec.rb
```diff
@@ -981,6 +981,7 @@ def foo(a, b, c, d, e, f, g, h); end
           @project.bundle_install
 
           result = @project.tapioca("gem foo")
+          puts result
```
Morriar, line 984: left over?

spec/tapioca/gem/pipeline_spec.rb
```diff
@@ -3898,6 +3898,12 @@ def foo; end
         end
       RB
 
+      sorbet_initialize = if RUBY_VERSION >= "3.1"
```
Morriar, line 3901: You can use erb instead:

```
        <% if ruby_version(">= 3.1") %>
          def initialize(*args, **_arg1, &blk); end
        <% else %>
          def initialize(*args, &blk); end
        <% end %>
```

See [here](https://github.com/Shopify/tapioca/blob/main/spec/tapioca/gem/pipeline_spec.rb#L162-L166) for example.

Morriar (approved):

lib/tapioca/gem/listeners/gem_location.rb
```diff
@@ -0,0 +1,50 @@
...
+
+        sig { override.params(event: ConstNodeAdded).void }
+        def on_const(event)
+          file, line = Object.const_source_location(event.symbol)
+          add_gem_location_comment(event.node, file, line)
+        end
+
+        sig { override.params(event: ScopeNodeAdded).void }
+        def on_scope(event)
+          file, line = Object.const_source_location(event.symbol)
```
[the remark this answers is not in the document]
paracycle: Sure, that would be a good addition.

spec/tapioca/gem/pipeline_spec.rb
```diff
@@ -3847,5 +3848,128 @@ module ViewHelpers::UrlHelper; end
...
+        # source://the-dep-1.1.2/lib/foo.rb:6
+        module Foo::Helper
+          # source://the-dep-1.1.2/lib/foo.rb:7
+          def helper_method; end
+        end
+
+        # source://the-dep-1.1.2/lib/foo.rb:33
+        class Quux < ::T::Struct
+          class << self
+            # source://sorbet-runtime-0.5.10125/lib/types/struct.rb:13
```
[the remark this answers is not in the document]
paracycle: On this note, can we please interpolate the version number of Sorbet runtime here. Otherwise, we will keep having to update this test every time we do a gem bump.

spec/tapioca/gem/pipeline_spec.rb
```diff
@@ -3866,5 +3867,135 @@ module ViewHelpers::UrlHelper; end
...
+          def foo; end
+        end
+
+        # @abstract It cannont be directly instantiated. Subclasses must implement the `abstract` methods below.
+        #
+        # source://the-dep-1.1.2/lib/foo.rb:28
+        class Baz
+          abstract!
+
+          # source://sorbet-runtime-#{Gem::Specification.find_by_name("sorbet-runtime").version}/lib/types/private/abstract/declare.rb:37
```
paracycle, line 3958: Nitpick: Can we turn these into local variables so that we don't keep retrieving the same version numbers over and over again. It would also make this expectation read better.

spec/tapioca/gem/pipeline_spec.rb
```diff
@@ -3866,5 +3867,135 @@ module ViewHelpers::UrlHelper; end
 
       assert_equal(output, compile)
     end
+
+    it "compile RBIs with location from gem source" do
+      add_ruby_file("foo.rb", <<~RB)
```
paracycle, line 3872: Can we break some of these constants into multiple "files" so that we can ensure that our file detection is working properly?

lib/tapioca/gem/listeners/source_location.rb
```diff
@@ -0,0 +1,54 @@
...
+
+        sig { override.params(event: ConstNodeAdded).void }
+        def on_const(event)
+          file, line = Object.const_source_location(event.symbol)
+          add_source_location_comment(event.node, file, line)
+        end
+
+        sig { override.params(event: ScopeNodeAdded).void }
+        def on_scope(event)
+          location = Runtime::Trackers::ConstantDefinition.locations_for(event.symbol)
```
egiurleo, line 20: Why do we need to use the tracker to find the location here? Might be worth writing a short comment so anyone who revisits this in the future understands.

egiurleo: Also this is a nitpick, but I found the next line confusing because this variable is called `location` -- it's an array with multiple locations, right?

lib/tapioca/gem/listeners/source_location.rb
```diff
@@ -0,0 +1,54 @@
...
+          return unless file && line
+
+          gem = @pipeline.gem
+          path = Pathname.new(file)
+          return unless File.exist?(path)
+
+          path = if path.realpath.to_s.start_with?(gem.full_gem_path)
+            "#{gem.name}-#{gem.version}/#{path.realpath.relative_path_from(gem.full_gem_path)}"
+          else
+            path.sub("#{Bundler.bundle_path}/gems/", "")
```
egiurleo, line 45: Do we ever hit this line in the test we added?

egiurleo (approved):
This looks awesome! I left a question and a couple of nitpick comments but they're not blocking.

lib/tapioca/gem/listeners/source_location.rb
vinistock: I added a comment. And yes, it's supposed to be `locations` in plural and then `location` after the `find`.

Morriar (approved):

lib/tapioca/gem/listeners/source_location.rb
egiurleo: Awesome, thanks! Just making sure I understand.

PR #1026, opened 2022-06-30
Make `tapioca init` behave like `srb init`

Morriar:
### Motivation

Initializing a project with Tapioca is basically a succession of 4 commands:

```shell
$ tapioca init
$ tapioca annotations
$ tapioca gems
$ tapioca todo
```

This PR does two things:

1. it renames the current `tapioca init` command into `tapioca configure` 
2. it makes `tapioca init` run the `configure`, `annotations`, `gems` and `todo` commands all in one

This brings us closer from the `srb init` experience.

I didn't add `tapioca dsl` in the mix has over the few projects I tried the new `tapioca init` command, all crashed when trying to generate the DSLs.


lib/tapioca/cli.rb
```diff
@@ -21,10 +21,53 @@ class Cli < Thor
       desc: "Verbose output for debugging purposes",
       default: false
 
-    desc "init", "initializes folder structure"
-    option :postrequire, type: :string, default: DEFAULT_POSTREQUIRE_FILE
+    desc "init", "get project ready for type checking"
     def init
-      command = Commands::Init.new(
+      configure_command = Commands::Configure.new(
```
rafaelfranca, line 26: Why isn't this command just calling the other methods instead of reimplementing them?

```ruby
invoke :configure
invoke :annotations
invoke :gem
invoke :todo
```

lib/tapioca/cli.rb
```diff
@@ -21,10 +21,53 @@ class Cli < Thor
...
+        rbi_formatter: rbi_formatter({ rbi_max_line_length: DEFAULT_RBI_MAX_LINE_LENGTH })
+      )
+      gems_command.execute
+
+      todo_command = Commands::Todo.new(
+        todo_file: DEFAULT_TODO_FILE,
+        file_header: true
+      )
+      todo_command.execute
+    end
```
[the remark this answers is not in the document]
rafaelfranca: I think it makes sense.

lib/tapioca/cli.rb
```diff
@@ -21,10 +21,18 @@ class Cli < Thor
...
 
-    desc "init", "initializes folder structure"
-    option :postrequire, type: :string, default: DEFAULT_POSTREQUIRE_FILE
+    desc "init", "get project ready for type checking"
     def init
-      command = Commands::Init.new(
+      invoke(:configure)
+      invoke(:annotations)
+      invoke(:gem)
+      invoke(:todo)
```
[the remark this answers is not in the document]
egiurleo: Yeah this makes me slightly nervous too. I think we should decide what we value most with this change. If we value having tapioca "just work" when a user sets it up, then running `todo` makes sense. If we value avoiding generating potentially incorrect types, then it makes sense to take it out.

Maybe we can also include a note in the "What's next" message (see comment below this one) reminding the user to look through the types generated in `todo.rbi` to check for correctness?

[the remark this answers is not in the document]
egiurleo: It just occurred to me - this is even more of a problem if we're not compiling DSLs, right? It is likely that todo will generate signatures for constants that would be generated via DSLs, which could create a bad experience.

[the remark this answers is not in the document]
Morriar: > reminding the user to look through the types generated in todo.rbi to check for correctness?

Yeah I think we should make it clear in the output that this file should be reviewed carefully as it would hide potential real error.

> todo will generate signatures for constants that would be generated via DSLs

Not sure what you mean here as `todo` only generates `module X; end` for each undefined constant X. So no real signature? Having useless constants in this file after generating DSL should be caught by `check-shims`, maybe we should mention something about this command in the what's next?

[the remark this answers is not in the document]
vinistock: Yes, exactly. I'm not sure what the best approach is. Running `todo` certainly helps the adoption, since it just makes type checking pass, but it doesn't provide very accurate typing information.

Not a blocker for the PR though, just a thought on how we'd like to have this setup.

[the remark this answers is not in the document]
egiurleo: Oh I think I imagined worse consequences for having RBI definitions for constants that don't exist in the app. If it won't practically cause any problems then I'm less worried about it!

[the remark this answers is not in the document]
Morriar: It still is a problem as it may be hiding a non-defining constant that should actually be reported as an error. This should be explained properly the message printed after `tapioca init`. Or maybe the message for `tapioca todo` should be enhanced first.

lib/tapioca/cli.rb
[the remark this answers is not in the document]
egiurleo: I've added some output that is similar to sorbet's "What's Next" message. Here's a screenshot of what it looks like in the terminal:

Screen Shot 2022-07-06 at 10 14 34 AM

lib/tapioca/cli.rb
```diff
@@ -311,5 +321,81 @@ def self.exit_on_failure?
...
+
+      OUTPUT
+
+      say("  srb tc", :cyan)
+
+      say("\n")
+      say("Documentation", :bold)
+      say("\n")
+
+      say("We recommend skimming these docs to get a feel for how to use Sorbet:")
```
egiurleo, line 371: The `say` API doesn't seem to be built for this type of longer message - the only way I could find to switch colors in one line is by calling the method multiple times. Am I missing something?

lib/tapioca/cli.rb
```diff
@@ -311,5 +321,79 @@ def self.exit_on_failure?
         true
       end
     end
+    private
```
Morriar, line 324:
```suggestion

    private
```

lib/tapioca/cli.rb
```diff
@@ -311,5 +321,79 @@ def self.exit_on_failure?
...
+      say("We recommend skimming these docs to get a feel for how to use Sorbet:")
+      say("- Gradual Type Checking: ", nil, false)
+      say("https://sorbet.org/docs/gradual", :cyan)
+      say("- Enabling Static Checks: ", nil, false)
+      say("https://sorbet.org/docs/static", :cyan)
+      say("- RBI Files: ", nil, false)
+      say("https://sorbet.org/docs/rbi", :cyan)
+      say("\n")
+
+      say("Next Steps", :bold)
```
Morriar, line 378: The items of this section could be merged after the numbered list at the beginning

lib/tapioca/cli.rb
egiurleo: Wow @Morriar thank you!!! The description for this method did not make it clear it can be used this way. I'll submit a PR to them to make it clearer.

lib/tapioca/cli.rb
```diff
@@ -311,5 +321,63 @@ def self.exit_on_failure?
         true
       end
     end
+
+    private
+
+    def print_init_next_steps
+      say(<<~OUTPUT)
+      #{set_color("This project is now set up for use with Sorbet and Tapioca", :bold)}
```
egiurleo, line 329: @Morriar 🎉🎉🎉🎉

lib/tapioca/cli.rb
```diff
@@ -311,5 +321,63 @@ def self.exit_on_failure?
...
+      Please check this folder into version control.
+
+      #{set_color("🤔 What's next", :bold)}
+
+      1. Many Ruby applications use metaprogramming DSLs to dynamically generate constants and methods.
+         To generate type definitions for any DSLs in your application, run:
+
+         #{set_color("bin/tapioca dsl", :cyan)}
+
+      2. Check whether the constants in the #{set_color("sorbet/rbi/todo.rbi", :cyan)} file actually exist in your project.
```
egiurleo, line 351: @paracycle Let me know if this is a better way to talk about todo.rbi.

lib/tapioca/cli.rb
```diff
@@ -311,5 +321,63 @@ def self.exit_on_failure?
         true
       end
     end
+
+    private
+
+    def print_init_next_steps
+      say(<<~OUTPUT)
+        #{set_color("This project is now set up for use with Sorbet and Tapioca", :bold)}
```
egiurleo, line 329: @Morriar This is awesome 🎉

lib/tapioca/cli.rb
```diff
@@ -311,5 +321,63 @@ def self.exit_on_failure?
...
+        Please check this folder into version control.
+
+        #{set_color("🤔 What's next", :bold)}
+
+        1. Many Ruby applications use metaprogramming DSLs to dynamically generate constants and methods.
+          To generate type definitions for any DSLs in your application, run:
+
+          #{set_color("bin/tapioca dsl", :cyan)}
+
+        2. Check whether the constants in the #{set_color("sorbet/rbi/todo.rbi", :cyan)} file actually exist in your project.
```
egiurleo, line 351: @paracycle Let me know if this is a better way to talk about todo.rbi

lib/tapioca/cli.rb
```diff
@@ -311,5 +321,63 @@ def self.exit_on_failure?
...
+        3. Typecheck your project:
+
+          #{set_color("bundle exec srb tc", :cyan)}
+
+          There should not be any typechecking errors.
+
+        4. Upgrade a file marked "#{set_color("# typed: false", :cyan)}" to "#{set_color("# typed: true", :cyan)}".
+          Then, run: #{set_color("bundle exec srb tc", :cyan)} and try to fix any errors.
+
+          If you are using Spoom, you may run:
```
egiurleo, line 365: @Morriar I mentioned spoom here as an alternate way of bumping files because I'm sure not everyone will know what Spoom is. Let me know if you think this is enough.

Morriar: Tapioca depends on Spoom so we may just give the command:

```md
You can use Spoom to bump files for you:

#{set_color("spoom bump --from false --to true", :cyan)}

To learn more about Spoom, visit: #{set_color("https://github.com/Shopify/spoom", :cyan)}
```

lib/tapioca/cli.rb
```diff
@@ -311,5 +321,63 @@ def self.exit_on_failure?
...
+
+          #{set_color("spoom bump --from false --to true", :cyan)}
+
+          To learn more about Spoom, visit: #{set_color("https://github.com/Shopify/spoom", :cyan)}
+
+        5. Add signatures to your methods with #{set_color("sig", :cyan)}. To learn how, read: #{set_color("https://sorbet.org/docs/sigs", :cyan)}
+
+        6. Upgrade a file marked "#{set_color("# typed: ignore", :cyan)}" to "#{set_color("# typed: false", :cyan)}".
```
Morriar, line 373: Let's remove this one, there should not be any file at `# typed: ignore` at this stage.

st0012 (approved):

PR #1029, opened 2022-07-04
Add line numbers to the ConstantDefinition tracker

vinistock:
### Motivation

Based on [this suggestion](https://github.com/Shopify/tapioca/pull/1025#discussion_r911400473).

We need the line numbers where constants were defined in order to add the [original gem source locations](https://github.com/Shopify/tapioca/pull/1025) for jump to definition. This PR saves this information in the `ConstantDefinition` tracker.

### Implementation

Added a `T::Struct` to keep track of path and line number, started populating it in the tracker and used the new method in the only occurrence of `files_for`.

### Tests

I contemplated adding a new test only for this tracker, but noticed that's not a pattern we're using in tests. Should I create a new one for this tracker?

### DIFF
--- a/lib/tapioca/runtime/trackers/constant_definition.rb
+++ b/lib/tapioca/runtime/trackers/constant_definition.rb
@@ -45,8 +45,8 @@ module Tapioca
         # or where metaprogramming was used via +eval+, etc.
         def self.files_for(klass)
           name = String === klass ? klass : name_of(klass)
-          files = @class_files[name]
-          files&.map(&:path)&.to_set || Set.new
+          files = @class_files.fetch(name, [])
+          files.map(&:path).to_set
         end
       end
     end
--- a/lib/tapioca/runtime/trackers/constant_definition.rb
+++ b/lib/tapioca/runtime/trackers/constant_definition.rb
@@ -10,6 +10,12 @@ module Tapioca
       # available in the ruby runtime without extra accounting.
       module ConstantDefinition
         extend Reflection
+        extend T::Sig
+
+        class ConstantLocation < T::Struct
+          const :lineno, Integer
+          const :path, String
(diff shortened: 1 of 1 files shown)


lib/tapioca/gem/pipeline.rb
```diff
@@ -337,7 +337,7 @@ def defined_in_gem?(constant, strict: true)
         return !strict if files.empty?
 
         files.any? do |file|
-          @gem.contains_path?(file)
+          @gem.contains_path?(file.path)
```
paracycle, line 340: Hmmm, I am not sure if this will work since `get_file_candidates` probably returns just an enumerable of files. I think it would be better to change `fules_for` to return just filenames instead.

paracycle (commented):
I am ok accepting this without a test since it is relatively hard to test it properly (since we would have to setup multiple Ruby files, etc) and there isn't that much logic in the tracker in the first place.

lib/tapioca/gem/pipeline.rb
vinistock: Changed it to continue returning a set.

lib/tapioca/runtime/trackers/constant_definition.rb
```diff
@@ -34,7 +46,7 @@ module ConstantDefinition
         def self.files_for(klass)
           name = String === klass ? klass : name_of(klass)
           files = @class_files[name]
-          files || Set.new
+          files&.map(&:path)&.to_set || Set.new
```
paracycle, line 49: Could we instead do a `.fetch(name, [])` in the line above so that this line becomes:
```suggestion
          files.map(&:path).to_set
```

KaanOzkan (approved):

PR #1038, opened 2022-07-06
DSL compiler for graphql-ruby mutations and input objects

jeffcarbs:
### Motivation

You can manually type the `resolve` method of your graphql mutations but there's nothing guaranteeing the sig you added will match the types specified in your graphql definition. It's also fairly cumbersome to type the getters of InputObjects.

### Implementation

This PR adds two DSL compilers that reflects on the graphql definition on your mutations and input objects and type the resolve method of mutations and getter methods of input objects.

Some notes:
- The return value of the resolve is typed as `T.untyped`. It seems like it should be possible to type the return value but given that graphql objects can be backed by anything (e.g. `Objects::Foo` can be back by an instance of the model `Foo`, a value object that implements the necessary methods, even just a hash, etc) it's not really feasible.
- If you write an inline sig on the resolve method, no RBI will be generated. This gives control to the consumer to write their own sig if desired.

Open questions:
- How should we handle custom argument types (e.g. custom scalars)? This PR was pulled out of an internal implementation that had just hard-coded our custom types but obviously we need some way to make it configurable or instruct folks to `prepend` a monkey-patch for `GraphqlTypeHelper#type_for`.

### Tests

See included tests.


jeffcarbs:
I can work on adding more tests and fixing the merge conflicts. Do you have thoughts on how to configure custom argument types (eg custom scalars)? We can just not for now and folks could  just monkey-patch/prepend to handle.

lib/tapioca/dsl/helpers/graphql_type_helper.rb
```diff
@@ -0,0 +1,60 @@
...
+            value_types = enum_values.map { |v| qualified_name_of(v.value.class) }.uniq
+            if value_types.size == 1
+              value_types.first
+            else
+              "T.any(#{value_types.join(", ")})"
+            end
+          elsif unwrapped_type <= GraphQL::Schema::InputObject
+            qualified_name_of(unwrapped_type)
+          else
+            raise ArgumentError, "Unknown type: #{unwrapped_type}"
```
paracycle, line 42: So far we've been choosing to fall back to T.untyped instead of raising errors, do you think that's also possible here?

lib/tapioca/dsl/helpers/graphql_type_helper.rb
```diff
@@ -0,0 +1,60 @@
...
+            enum_values = T.cast(unwrapped_type.enum_values, T::Array[GraphQL::Schema::EnumValue])
+            value_types = enum_values.map { |v| qualified_name_of(v.value.class) }.uniq
+            if value_types.size == 1
+              value_types.first
+            else
+              "T.any(#{value_types.join(", ")})"
+            end
+          elsif unwrapped_type <= GraphQL::Schema::InputObject
+            qualified_name_of(unwrapped_type)
+          else
```
paracycle, line 41: I am super wary of calling `==` and `<=` on random constants not under our control. I am sure you remember what happened the last time we had a `constant <= T::Enum` call 😂.

So can we instead swap this over to:
```suggestion
          parsed_type = case unwrapped_type
          when GraphQL::Types::Boolean.singleton_class
            "T::Boolean"
          when GraphQL::Types::Float.singleton_class
            qualified_name_of(Float)
          when GraphQL::Types::ID.singleton_class
            qualified_name_of(String)
          when GraphQL::Types::Int.singleton_class
            qualified_name_of(Integer)
          when GraphQL::Types::ISO8601Date.singleton_class
            qualified_name_of(Date)
          when GraphQL::Types::ISO8601DateTime.singleton_class
            qualified_name_of(DateTime)
          when GraphQL::Types::JSON.singleton_class
            "T::Hash[String, T.untyped]"
          when GraphQL::Types::String.singleton_class
            qualified_name_of(String)
          when GraphQL::Schema::Enum.singleton_class
            enum_values = T.cast(unwrapped_type.enum_values, T::Array[GraphQL::Schema::EnumValue])
            value_types = enum_values.map { |v| qualified_name_of(v.value.class) }.uniq
            if value_types.size == 1
              value_types.first
            else
              "T.any(#{value_types.join(", ")})"
            end
          when GraphQL::Schema::InputObject.singleton_class
            qualified_name_of(unwrapped_type)
          else
```

paracycle (commented):
Thanks, I have a couple of safety comments. I think it is fine to punt on unknown types for now. As long as we generate `T.untyped` for them, it is still better than not exposing missing methods anyway.

lib/tapioca/helpers/rbi_helper.rb
```diff
@@ -95,6 +95,15 @@ def sanitize_signature_types(sig_string)
         .gsub(".params()", "")
     end
 
+    sig { params(type: String).returns(String) }
+    def as_nilable_type(type)
```
jeffcarbs, line 99: Moved this out into the `RBIHelper` module.

lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
```diff
@@ -74,15 +75,6 @@ def do_not_generate_strong_types?(constant)
             !(constant.singleton_class < Object.const_get(:StrongTypeGeneration))
         end
 
-        sig { params(type: String).returns(String) }
-        def as_nilable_type(type)
-          if type.start_with?("T.nilable(") || type == "T.untyped"
```
jeffcarbs, line 79: Looks like a copy/paste of `as_nilable_type` from the base compiler class but was (presumably unintentionally?) not updated to handle `::T.nilable(` or `::T.untyped` like the other was.

jeffcarbs (commented):
@KaanOzkan @paracycle - this should be ready for another look. I integrated the feedback you both suggested, making the code changes and adding a test that exercised all graphql types.

One thing to call out was moving the `as_nilable_type` method to the `RBIHelpers` since it was currently defined inside the base compiler class. I left some additional comments in reference to the change.

lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
paracycle: Yeah, good catch!

jeffcarbs:
> Not sure why there are build failures

These are fixed 🙂 I had missed `gem("graphql", require: false)` in the additional gemfiles (e.g. `Gemfile-rails-6-1`), which meant that when we required the files for those compilers they would return early due to the LoadError handling:
```ruby
begin
  require "graphql"
rescue LoadError
  return
end
```

PR #1044, opened 2022-07-07
Track class new in ConstantDefinition

vinistock:
### Motivation

Based on @paracycle's work [in this commit](https://github.com/Shopify/tapioca/commit/4fecd9d0ae0aaefca4ae18f996fb8273e427a20f).

Start handling `c_return` to capture constants defined by `Constant = Class.new`. Currently, there's a bug in main where we can't generate RBIs including locations because these occurrences aren't tracked.

After this PR, generating all RBIs in Tapioca with locations works.

### Implementation

1. Started comparing classes by identity, because classes defined by `Something = Class.new` don't have names by the time we capture them
2. Changed all relevant code to pass only class instances to ConstantDefinition
3. Started handling `c_return` in ConstantDefinition to capture these occurrences
4. After 3, we start having a `StackLevelTooDeep` error that is caused by the `NamePatch`. As per @paracycle's suggestion, I inlined the implementation of `qualified_name_of`, which fixes the recursion issue
5. Finally, started passing `event.constant` instead of `event.symbol` in the source location listener and stopped trying to add comments if `location` is `nil` (which can still happen for constant aliases)

### Tests

Added a test that does not pass without handling `c_return`.


vinistock:
Applied the suggested refactors and shortener the ConstantDefinition a bit. However, there are two tests that are now failing because certain modules are no longer generated for them. I'm not 100% sure what's going on.

lib/tapioca/runtime/trackers/constant_definition.rb
```diff
@@ -17,41 +17,60 @@ class ConstantLocation < T::Struct
...
-        @class_files = {}
+        @class_files = {}.compare_by_identity
 
         # Immediately activated upon load. Observes class/module definition.
-        TracePoint.trace(:class) do |tp|
-          unless tp.self.singleton_class?
-            key = name_of(tp.self)
-            file = tp.path
-            lineno = tp.lineno
+        TracePoint.trace(:class, :c_return) do |tp|
```
egiurleo, line 23: This is entirely a suggestion, feel free to disregard - I personally find the single TracePoint with the `case` statement hard to parse. Some of the lines are pretty deeply nested, and the two cases don't really share any common functionality, aside from setting the key-value-pairs in `@class_files`.

What would you think about calling `TracePoint` twice, separating out the behavior for `class` and `c_return`, and then extracting `@class_files[key] ||= Set.new ...` into a shared method? I personally would find that much easier to read, but I don't think it's a huge deal either way.

(I'm also curious if there's any performance benefit to calling T`TracePoint` once?)

```suggestion
TracePoint.trace(:class) do |tp|
  next if tp.self.singleton_class?

  if tp.path == "(eval)"
    caller_location = T.must(caller_locations)
      .drop_while { |loc| loc.path == "(eval)" }
      .first

    next unless caller_location

    loc = ConstantLocation.new(
      path: caller_location.absolute_path || "",
      lineno: caller_location.lineno
    )
  else
    loc = build_constant_location(tp, caller_locations)
  end

  add_to_class_files(tp.self, loc)
end

TracePoint.trace(:c_return) do |tp|
  next unless tp.method_id == :new
  next unless Module === tp.return_value

  key = tp.return_value
  loc = build_constant_location(tp, caller_locations)

  add_to_class_files(key, loc)
end
```

egiurleo (approved):
Nice! I like your extension of `resolve_loc` to make it useful in more cases.

Morriar (approved):

PR #1046, opened 2022-07-08
Implement Rubocop DSL RBI compiler

sambostock:
### Motivation

When authoring a custom cop, one is likely to use the `def_node_matcher` or `def_node_search` macros, which generate methods. Sorbet will complain if the generated methods are used, as it does not know they exist, requiring the developer to write a shim.

Let's automate that.

### Implementation

I looked at a couple other DSL compilers and based my work off of them.

One tricky thing is that Rubocop expects custom cops to be defined in its namespace (i.e. `RuboCop::Cop::<department name>::<cop name>`). Therefore, we need a way to figure out which constants to generate DSLs for, and which to skip. The approach I came up with was to check if the `const_source_location` is in a gem, or within the host app. Although there may be a better way to do this.

### Tests

I've added simple tests that check a couple uses of the macros, which result in slightly different method signatures.

### DIFF
--- a/lib/tapioca/dsl/compilers/rubocop.rb
+++ b/lib/tapioca/dsl/compilers/rubocop.rb
@@ -1,0 +1,72 @@
+# typed: strict
+# frozen_string_literal: true
+
+begin
+  require "rubocop"
+rescue LoadError
+  return
+end
+
+module Tapioca
+  module Dsl
+    module Compilers
+      # TODO: documentation
+      # TODO: Should we call this Rubocop or RuboCop?
+      class Rubocop < Compiler
+        GEM_LOAD_PATH_PREFIX_PATTERN = /\A#{Regexp.union(
+          ::Gem.loaded_specs.each_value.flat_map(&:load_paths),
+        )}/
+
+        # TODO: Should this define a CONSTANT_TYPE?
+
+        def self.gather_constants
+          descendants_of(::RuboCop::Cop::Base).reject do |constant|
+            constant_name = name_of(constant)
+            next true if constant_name.nil?
+
+            # All cops (built-in, extensions, custom) are supposed to be of the
+            # form ::Rubocop::Cop::<department>::<name>, so we can't just look at
+            # the name prefix. Best we can do is look at if the constant
+            # definition path is in a gem or not.
+
+            path = const_source_location(constant_name)&.first
+            next true if path.nil?
+
+            GEM_LOAD_PATH_PREFIX_PATTERN.match?(path)
+          end
+        end
+
+        def decorate
+          return unless used_macros?
+
+          root.create_path(constant) do |cop_klass|
+            node_matchers.each do |name|
+              create_method_from_def(cop_klass, constant.instance_method(name))
+            end
+
+            node_searches.each do |name|
+              create_method_from_def(cop_klass, constant.instance_method(name))
+            end
+          end
+        end
+
+        private
+
+        def used_macros?
+          return true unless node_matchers.empty?
+          return true unless node_searches.empty?
+
+          false
+        end
+
+        def node_matchers
+          constant.__tapioca_node_matchers
+        end
+
+        def node_searches
+          constant.__tapioca_node_searches
+        end
+      end
+    end
+  end
+end
--- a/lib/tapioca/dsl/extensions/rubocop.rb
+++ b/lib/tapioca/dsl/extensions/rubocop.rb
@@ -1,0 +1,44 @@
+# typed: strict
+# frozen_string_literal: true
+
+begin
+  require "rubocop"
+rescue LoadError
+  return
+end
+
+module Tapioca
+  module Dsl
+    module Compilers
+      module Extensions
+        module Rubocop
+          def def_node_matcher(name, *, **defaults)
+            __tapioca_node_matchers << name
+            __tapioca_node_matchers << :"without_defaults_#{name}" unless defaults.empty?
+
+            super
+          end
+
+          def def_node_search(name, *, **defaults)
+            __tapioca_node_searches << name
+            __tapioca_node_searches << :"without_defaults_#{name}" unless defaults.empty?
+
+            super
+          end
+
+          def __tapioca_node_matchers
+            @__tapioca_node_matchers = T.let(@__tapioca_node_matchers, T.nilable(T::Array[Symbol]))
+            @__tapioca_node_matchers ||= []
+          end
+
+          def __tapioca_node_searches
+            @__tapioca_node_searches = T.let(@__tapioca_node_searches, T.nilable(T::Array[Symbol]))
+            @__tapioca_node_searches ||= []
+          end
+
+          ::RuboCop::Cop::Base.singleton_class.prepend(self)
+        end
+      end
+    end
+  end
+end
--- a/spec/tapioca/dsl/compilers/rubocop_spec.rb
+++ b/spec/tapioca/dsl/compilers/rubocop_spec.rb
@@ -1,0 +1,110 @@
+# typed: strict
+# frozen_string_literal: true
+
+require "spec_helper"
+require "rubocop"
+require "rubocop-sorbet"
+
+module Tapioca
+  module Dsl
+    module Compilers
+      class RubocopSpec < ::DslSpec
+        describe "Tapioca::Dsl::Compilers::Rubocop" do
+          def before_setup
+            require "tapioca/dsl/extensions/rubocop"
+            super
+          end
+
+          describe "initialize" do
+            it "gathers no constants if there are no classes that inherit from Rubocop::Cop::Base" do
+              assert_empty(gathered_constants)
+            end
+
+            it "gathers only cops defined in the app" do
+              add_ruby_file("content.rb", <<~RUBY)
+                class MyCop < ::RuboCop::Cop::Base
+                end
+
+                class MyLegacyCop < ::RuboCop::Cop::Cop
+                end
+
+                module ::RuboCop
+                  module Cop
+                    module MyApp
+                      class MyNamespacedCop < Base
+                      end
+                    end
+                  end
+                end
+              RUBY
+
+              assert_equal(["MyCop", "MyLegacyCop", "RuboCop::Cop::MyApp::MyNamespacedCop"], gathered_constants)
+            end
+          end
+
+          describe "decorate" do
+            it "generates empty RBI when no DSL used" do
+              add_ruby_file("content.rb", <<~RUBY)
+                class MyCop < ::RuboCop::Cop::Base
+                end
+              RUBY
+
+              expected = <<~RBI
+                # typed: strong
+              RBI
+
+              assert_equal(expected, rbi_for(:MyCop))
+            end
+
+            it "generates correct RBI file" do
+              add_ruby_file("content.rb", <<~RUBY)
+                class MyCop < ::RuboCop::Cop::Base
+                  def_node_matcher :some_matcher, "(...)"
+                  def_node_matcher :some_matcher_with_params, "(%1 %two ...)"
+                  def_node_matcher :some_matcher_with_params_and_defaults, "(%1 %two ...)", two: :default
+                  def_node_search :some_search, "(...)"
+                  def_node_search :some_search_with_params, "(%1 %two ...)"
+                  def_node_search :some_search_with_params_and_defaults, "(%1 %two ...)", two: :default
+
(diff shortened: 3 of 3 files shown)


lib/tapioca/dsl/compilers/rubocop.rb
```diff
@@ -0,0 +1,77 @@
...
+
+module Tapioca
+  module Dsl
+    module Compilers
+      # TODO: documentation
+      # TODO: Should we call this Rubocop or RuboCop?
+      class Rubocop < Compiler
+        GEM_LOAD_PATH_PREFIX_PATTERN = %r{\A#{Regexp.union(
+          ::Gem.loaded_specs.each_value.flat_map(&:load_paths),
+        )}}
```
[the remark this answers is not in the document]
sambostock: I've stopped trying to filter out constants from gems, so this is no longer applicable.

spec/tapioca/dsl/compilers/rubocop_spec.rb
```diff
@@ -0,0 +1,149 @@
...
+require "rubocop"
+require "rubocop-sorbet"
+
+module Tapioca
+  module Dsl
+    module Compilers
+      class RuboCopSpec < ::DslSpec
+        EXISTING_CONSTANTS = Runtime::Reflection
+          .descendants_of(::RuboCop::Cop::Base)
+          .filter_map { |constant| Runtime::Reflection.name_of(constant) }
```
sambostock, line 14: So as to be able to focus on just the constants we introduce in our tests, we need to compile a list of all the constants we'll gather from gems **before we define a bunch in our tests**.

spec/tapioca/dsl/compilers/rubocop_spec.rb
```diff
@@ -0,0 +1,149 @@
...
+
+        class << self
+          def target_class_file
+            # Against convention, RuboCop uses "rubocop" in its file names, so we do too.
+            super.gsub("rubo_cop", "rubocop")
+          end
```
sambostock, line 20: Originally, I'd renamed the files to `rubo_cop.rb` too, due to the loading not working, but I like this better, as it allows us to keep both the constant names and file names matching RuboCop's convention (which was the whole point of renaming the constants).

sambostock:
I've addressed feedback, written docs, and tidied up the commits.

One thing I considered was seeing if we could specify that the first parameter to the methods should be a `::RuboCop::AST::Node`. However, the methods both default to using `self` if no node is passed in, as well as implement a [guard clause](https://github.com/rubocop/rubocop-ast/blob/e487124ee4a5160fd85d2a47ee4adde6ebcd0775/lib/rubocop/ast/node_pattern/compiler/node_pattern_subcompiler.rb#L129-L131), so technically anything is allowed to be passed in, so for simplicity I stuck with the generated sig whose first param is `T.untyped`.

vinistock (approved):

KaanOzkan (approved):
🎉

lib/tapioca/dsl/compilers/rubocop.rb
```diff
@@ -0,0 +1,83 @@
...
+
+        private
+
+        sig { returns(T::Boolean) }
+        def used_macros?
+          return true unless node_matchers.empty?
+          return true unless node_searches.empty?
+
+          false
+        end
```
Morriar, line 69:
```suggestion
        def used_macros?
          node_matchers.any? || node_searches.any?
        end
```

lib/tapioca/dsl/compilers/rubocop.rb
```diff
@@ -0,0 +1,83 @@
...
+          extend T::Sig
+          sig { override.returns(T::Enumerable[Class]) }
+          def gather_constants
+            descendants_of(::RuboCop::Cop::Base).select { |constant| name_of(constant) }
+          end
+        end
+
+        sig { override.void }
+        def decorate
+          return unless used_macros?
```
Morriar, line 48: I think you can simplify all of this with one collection:

```rb
methods = node_macros
return if methods.empty?

root.create_path(constant) do |cop_klass|
  methods.each do |name|
    create_method_from_def(cop_klass, constant.instance_method(name))
  end
end
```

Since it doesn't look like we have special behavior depending on the kind.

lib/tapioca/commands/dsl.rb
```diff
@@ -363,7 +363,8 @@ def underscore(class_name)
 
       sig { params(constant: String).returns(String) }
       def rbi_filename_for(constant)
-        underscore(constant) + ".rbi"
+        # TODO: Figure out a better way than this.
+        underscore(constant).gsub!("rubo_cop", "rubocop") + ".rbi"
```
Morriar, line 367: This is definitely not ideal.

Since it's a generated file, why do we care about the name having an underscore or not? Can we let the file be `rubo_cop.rbi`?

lib/tapioca/commands/dsl.rb
```diff
@@ -363,7 +363,8 @@ def underscore(class_name)
 
       sig { params(constant: String).returns(String) }
       def rbi_filename_for(constant)
-        underscore(constant) + ".rbi"
+        # TODO: Figure out a better way than this.
+        underscore(constant).gsub("rubo_cop", "rubocop") + ".rbi"
```
vinistock, line 367: Apologies for setting you on the wrong path, but let's switch the name of the compiler to `Rubocop` instead.

sambostock: Sorry these are two separate things (I originally erroneously thought they were related):

- **`Tapioca::Commands::DSL#rbi_filename_for` controls the output file name for the generated RBIs**, which are based on the name of the constant they were generated for. This is hard to override for just constants with `RuboCop` in their name (and we don't control those constant names), and probably not worth doing (as @Morriar [suggested](https://github.com/Shopify/tapioca/pull/1046#discussion_r1008180704)).
- **`DslSpec#target_class_file` controls which compiler file is required based on the name of the spec file.** Fortunately, that module is included in the specs, so we can just override the method in the spec to say that we want to `require` `"tapioca/dsl/compilers/rubocop.rb"` instead of `"tapioca/dsl/compilers/rubo_cop.rb"`, which makes it trivial for Tapioca to **internally** match both the `RuboCop` constant name and `rubocop` file name conventions the RuboCop gem follows.

vinistock: Oh, I see what you mean. It's not related to the DSL compiler class. That's unfortunate for sure, but not sure if there's a better way of doing this other than treating rubocop as a special case

sambostock: Yeah, exactly, and if we were to allow for that (perhaps through a configurable inflector), it would probably be a change that doesn't belong in this PR.

lib/tapioca/dsl/compilers/rubocop.rb
sambostock: After consolidating the method collections, this is no longer relevant.

lib/tapioca/dsl/extensions/rubocop.rb
```diff
@@ -0,0 +1,46 @@
...
+            __tapioca_node_methods << name
+            __tapioca_node_methods << :"without_defaults_#{name}" unless defaults.empty?
+
+            super
+          end
+
+          sig { returns(T::Array[MethodName]) }
+          def __tapioca_node_methods
+            @__tapioca_node_methods = T.let(@__tapioca_node_methods, T.nilable(T::Array[MethodName]))
+            @__tapioca_node_methods ||= []
```
Morriar, line 38: I think Sorbet supports this as a one-liner now:

```suggestion
            @__tapioca_node_methods ||= T.let([], T.nilable(T::Array[MethodName]))
```

spec/tapioca/dsl/compilers/rubocop_spec.rb
Morriar: Do we still need this since we don't rename the file in the DSL command?

sambostock: This allows us to have

- `lib/tapioca/dsl/compilers/rubocop.rb`
- `lib/tapioca/dsl/extensions/rubocop.rb`
- `spec/tapioca/dsl/compilers/rubocop_spec.rb`

instead of


- `lib/tapioca/dsl/compilers/rubo_cop.rb`
- `lib/tapioca/dsl/extensions/rubo_cop.rb`
- `spec/tapioca/dsl/compilers/rubo_cop_spec.rb`

which I think is preferable, especially given it's really just limited to telling the spec file how to require the correct compiler.

lib/tapioca/dsl/extensions/rubocop.rb
Morriar: Great call! The feature was introduced in Sorbet v.0.5.10207 and we require only v0.5.9204.

So let's keep this construct 👍

spec/tapioca/dsl/compilers/rubocop_spec.rb
Morriar: I would rather rename the constant `Rubocop` or let the file be named `rubo_cop` rather than play with this.

It seems all the files under `dsl/compilers` follow the same convention. `class CompilerName` -> `compiler_name.rb`. I'm not sure why we should deviate from this.

sambostock: 🤷 I was trying to match Rubocop's own naming convention (e.g. `RuboCop::Version` is defined in `lib/rubocop/version.rb`, not `lib/rubo_cop/version.rb`), but given the can of worms it would be to do this for the RBI files, maybe it's not worth doing for the compiler files either. It just feels weird seeing `rubo_cop` TBH, but I'm glad to deal with it if you think it's not worth the extra method in the test file.

lib/tapioca/dsl/extensions/rubocop.rb
sambostock: Perhaps worth asking: do we run CI against `v0.5.9204`? If we don't, perhaps it's worth adding that (in a separate PR) to ensure that we detect if we ever introduce a dependency on features from a newer version. Had I used the new notation, we might have unknowingly broken compatibility.

sambostock: Haha, [according to CI](https://github.com/Shopify/tapioca/actions/runs/3364568062/jobs/5579066013), that version of `sorbet-static-and-runtime` (apparently) doesn't even exist...

paracycle: This should be fine to change now, since we force a newer version of Sorbet.

sambostock:
I've pushed some updates, mainly consisting of:
- Adding a `extenders_of` helper method (not sure about this name).
- Extending gathered constants to include all modules which extend `RuboCop::AST::NodePattern::Macros`, rather than descendants of `RuboCop::Cop::Base`.
- Updating the lazy compiler definition approach from using rescuing `LoadError` to checking if constants are `defined?`.

I think it should be good to go, but given the changes it might be worth another look.

spec/tapioca/dsl/compilers/rubocop_spec.rb
sambostock: Okay, so circling back to this, https://github.com/Shopify/tapioca/commit/a80c4ecbd1215546c5c3e6fd470c55d30303a4ef makes the changes necessary to avoid this workaround, but my gut feeling is that this leads to the code being more confusing, because now we're dealing with a mix of constants that include `RuboCop` and constants that include `Rubocop`, and it's not obvious why they differ.

Likewise, we could go for the `rubo_cop` file naming, but then we'd have a mid of those files and ones like `.rubocop.yml`, again without an obvious reason why.

Therefore, I think it's overall more straightforward to keep the `RuboCop` naming to match upstream, and be consistent across our constants and theirs, and have this one workaround method with the comment documenting why it's there. I think this would be in line with the "principle of least surprise", in that this one method makes the rest of the app less surprising because things _just work_.

vinistock:
> Extending gathered constants to include all modules which extend RuboCop::AST::NodePattern::Macros, rather than descendants of RuboCop::Cop::Base.

What's the advantage of this? Wouldn't all cops inherit from `RuboCop::Cop::Base`?

Also, are all cops we want to generate RBIs for required when Tapioca loads the application? For example, extensions like `rubocop-rails` are required through the `.rubocop.yml` configuration file and you can do the same thing for custom cops defined inside the application.

Just trying to understand if custom cops will need to be manually required in the `require.rb` file.

PR #1047, opened 2022-07-08
Remove `T.unsafe` from extension method call

sambostock:
### Motivation

Sorbet doesn't know that `::FrozenRecord::Base` will have `::Tapioca::Dsl::Compilers::Extensions::FrozenRecord` prepended to its `singleton_class`. Therefore, it does not allow us to call the method provided by that module.

`T.unsafe` is used to allow the call, which can be improved.

### Implementation

If we tell Sorbet that `ConstantType` will be the union of both types, it allows us to call the method.

I came up with this approach in #1046, which also makes use of an extension. I did not want to use `T.usafe` if possible, so I came up with this approach.

### Tests

`bundle exec srb tc` is the test 😅

### DIFF
--- a/lib/tapioca/dsl/compilers/frozen_record.rb
+++ b/lib/tapioca/dsl/compilers/frozen_record.rb
@@ -65,7 +65,7 @@ module Tapioca
       class FrozenRecord < Compiler
         extend T::Sig
 
-        ConstantType = type_member { { fixed: T.class_of(::FrozenRecord::Base) } }
+        ConstantType = type_member { { fixed: T.all(T.class_of(::FrozenRecord::Base), Extensions::FrozenRecord) } }
 
         sig { override.void }
         def decorate
@@ -97,7 +97,7 @@ module Tapioca
 
         sig { params(record: RBI::Scope).void }
         def decorate_scopes(record)
-          scopes = T.unsafe(constant).__tapioca_scope_names
+          scopes = constant.__tapioca_scope_names
           return if scopes.nil?
 
           module_name = "GeneratedRelationMethods"


lib/tapioca/dsl/compilers/frozen_record.rb
```diff
@@ -65,7 +65,7 @@ module Compilers
       class FrozenRecord < Compiler
         extend T::Sig
 
-        ConstantType = type_member { { fixed: T.class_of(::FrozenRecord::Base) } }
+        ConstantType = type_member { { fixed: T.all(T.class_of(::FrozenRecord::Base), Extensions::FrozenRecord) } }
```
sambostock, line 68: Is there an alternative way to explicitly tell Sorbet that `T.class_of(::FrozenRecord::Base).is_a?(Extensions::FrozenRecord)`?

vinistock: I don't think so.

vinistock (approved):

paracycle (approved):

paracycle:
> Is there a good way to specify both, or is the idea here that it's not worth it since generics are erased at runtime anyway, so `Module` is sufficient?

So a few things here:
1. We dump all of the constants returned from all the `gather_constants` calls into a set of `Module`s anyway, so having a more precise type here does not help much for that stage.
2. More importantly, this quickly gets into co/contra-variance since we would be expecting the return type of the `gather_constants` to be substitutable for the return type of the same method on the base class, so the type variable needs to be covariant (i.e. `:out` in Sorbet), but you can only declare co/contra-variant type parameters in modules and not classes, so that ends up making the whole thing more complicated to pull off. See here for reference: https://sorbet.org/docs/generics#in-out-and-variance

PR #1056, opened 2022-07-12
Parallelize RBI parsing to speed up the `check-shims` command

Morriar:
### Motivation

Running `tapioca check-shims` can take up to 7min in Core (~13 when running on CI). The culprit is parsing the RBI files.

This PR proposes to speed up the process by parallelizing RBI parsing.

### Implementation

I reused `Executor` that we already use to parallelize RBI generation.

### Tests

No behavior change.

I tested the speed-up on Tapioca itself:

Before:

```shell
$ time bundle exec tapioca check-shims

Loading Sorbet payload...  Done (5.36s)
Loading shim RBIs from sorbet/rbi/shims...  Done (0.01s)
Loading gem RBIs from sorbet/rbi/gems...  Done (17.93s)
Looking for duplicates...  Done (0.02s)

real	0m24.291s
user	0m23.251s
sys	0m0.564s
```

After:

```shell
$ time bundle exec tapioca check-shims -w 10
Loading Sorbet payload...  Done (1.51s)
Loading shim RBIs from sorbet/rbi/shims...  Done (0.21s)
Loading gem RBIs from sorbet/rbi/gems...  Done (7.16s)
Looking for duplicates...  Done (0.03s)

real	0m9.561s
user	0m31.153s
sys	0m3.086s
```

So a > 50% improvement here.

I also tested on Core:

Before:

```shell
$ time bundle exec tapioca check-shims
Loading Sorbet payload...  Done (10.05s)
Loading shim RBIs from sorbet/rbi/shims...  Done (1.32s)
Loading gem RBIs from sorbet/rbi/gems...  Done (165.56s)
Loading dsl RBIs from sorbet/rbi/dsl...  Done (227.36s)
Loading annotation RBIs from sorbet/rbi/annotations...  Done (0.08s)
Looking for duplicates...  Done (0.5s)

real	6m52.147s
user	6m48.205s
sys	0m3.855s
```

After:

```shell
$ time bundle exec tapioca check-shims -w 10
Loading Sorbet payload...  Done (6.49s)
Loading shim RBIs from sorbet/rbi/shims...  Done (3.36s)
Loading gem RBIs from sorbet/rbi/gems...  Done (105.36s)
Loading dsl RBIs from sorbet/rbi/dsl...  Done (78.48s)
Loading annotation RBIs from sorbet/rbi/annotations...  Done (3.02s)
Looking for duplicates...  Done (0.44s)

real	3m24.315s
user	9m19.503s
sys	1m1.437s
```

So a > 50% improvement here too.

### DIFF
--- a/lib/tapioca/helpers/rbi_files_helper.rb
+++ b/lib/tapioca/helpers/rbi_files_helper.rb
@@ -15,14 +15,14 @@ module Tapioca
 
       say("Loading #{kind} RBIs from #{file}... ")
       time = Benchmark.realtime do
-        parse_and_index_file(index, file)
+        parse_and_index_files(index, [file], number_of_workers: 1)
       end
       say(" Done ", :green)
       say("(#{time.round(2)}s)")
     end
 
-    sig { params(index: RBI::Index, kind: String, dir: String).void }
-    def index_rbis(index, kind, dir)
+    sig { params(index: RBI::Index, kind: String, dir: String, number_of_workers: T.nilable(Integer)).void }
+    def index_rbis(index, kind, dir, number_of_workers:)
       return unless Dir.exist?(dir) && !Dir.empty?(dir)
 
       if kind == "payload"
@@ -32,7 +32,7 @@ module Tapioca
       end
       time = Benchmark.realtime do
         files = Dir.glob("#{dir}/**/*.rbi").sort
-        parse_and_index_files(index, files)
+        parse_and_index_files(index, files, number_of_workers: number_of_workers)
       end
       say(" Done ", :green)
       say("(#{time.round(2)}s)")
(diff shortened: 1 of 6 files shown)


lib/tapioca/helpers/rbi_files_helper.rb
```diff
@@ -250,5 +248,14 @@ def update_gem_rbis_strictnesses(errors, gem_dir)
     def gem_name_from_rbi_path(path)
       T.must(File.basename(path, ".rbi").split("@").first)
     end
+
+    sig { params(block: T.proc.void).returns(Numeric) }
+    def with_timer(&block)
```
vinistock, line 253: Could we use `Benchmark` instead? I think it does exactly the same. Also, do we actually need to measure this?

lib/tapioca/helpers/rbi_files_helper.rb
```diff
@@ -150,21 +150,19 @@ def validate_rbi_files(command:, gem_dir:, dsl_dir:, auto_strictness:, gems: [],
...
     private
 
-    sig { params(index: RBI::Index, files: T::Array[String]).void }
-    def parse_and_index_files(index, files)
-      files.each do |file|
-        parse_and_index_file(index, file)
-      end
-    end
+    sig { params(index: RBI::Index, files: T::Array[String], number_of_workers: T.nilable(Integer)).void }
+    def parse_and_index_files(index, files, number_of_workers:)
```
vinistock, line 154: Should we set the default value of `number_of_workers` to 1 to avoid having to pass it for sequential tasks?

lib/tapioca/helpers/rbi_files_helper.rb
Morriar: > Could we use Benchmark instead? 

Good idea!

> Also, do we actually need to measure this?

I was actually thinking to extend the concept to other commands so we can easily see where Tapioca is slow or slower.

lib/tapioca/helpers/rbi_files_helper.rb
Morriar: This is a private low-level command, if you don't want to deal with the details you should call the public `index_rbi` and `index_rbis` methods.

lib/tapioca/helpers/rbi_files_helper.rb
```diff
@@ -48,13 +48,16 @@ def index_rbis(index, kind, dir)
     def duplicated_nodes_from_index(index, shim_rbi_dir:, todo_rbi_file:)
       duplicates = {}
       say("Looking for duplicates... ")
-      index.keys.each do |key|
-        nodes = index[key]
-        next unless shims_or_todos_have_duplicates?(nodes, shim_rbi_dir: shim_rbi_dir, todo_rbi_file: todo_rbi_file)
+      time = Benchmark.measure do
```
vinistock, line 51: To avoid repeating the pattern `result = Benchmark.measure {}; result.real`, you can use `Benchmark.realtime` which just returns the number straight away.

vinistock (approved):

lib/tapioca/helpers/rbi_files_helper.rb
Morriar: Even better 👍

KaanOzkan (approved):

PR #1058, opened 2022-07-12
Include excluded gems in verify logic

adisonlampert:


spec/tapioca/cli/gem_spec.rb
```diff
@@ -1604,30 +1604,23 @@ class User; end
           assert_success_status(result)
         end
 
-        it "advises of removed file(s) and returns exit_status 1" do
+        it "is aware of exclude option and does not error due to removed files" do
```
vinistock, line 1607: I think the previous test has value. Can we add a new example instead of modifying the existing one?

lib/tapioca/commands/gem.rb
```diff
@@ -204,13 +204,15 @@ def compile_gem_rbi(gem)
...
-      sig { void }
-      def perform_sync_verification
+      sig { params(exclude: T::Array[String]).void }
+      def perform_sync_verification(exclude: [])
         diff = {}
 
         removed_rbis.each do |gem_name|
-          filename = existing_rbi(gem_name)
-          diff[filename] = :removed
+          unless exclude.include?(gem_name)
```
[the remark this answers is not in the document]
adisonlampert: `exclude` prevents the creation of an RBI for the excluded gem and if one exists, deletes it so my understanding is that there really shouldn't be a case where an excluded RBI ends up in the`added_rbis` array

spec/tapioca/cli/gem_spec.rb
adisonlampert: Thanks for the clarification!

st0012 (approved):

Morriar (approved):

PR #1059, opened 2022-07-12
Do not remove final newlines from annotations

egiurleo:
### Motivation
`tapioca annotations` currently removes the final newline from every annotation RBI file. This triggers linting errors in projects where linters expect every file to have a final newline.

See https://github.com/Shopify/rbi-central/issues/83

### Implementation
Don't remove final newlines when splitting up the annotation files to add the header.

### Tests
Modified existing tests to pass with this change.

### DIFF
--- a/lib/tapioca/commands/annotations.rb
+++ b/lib/tapioca/commands/annotations.rb
@@ -193,7 +193,8 @@ module Tapioca
           # Please run `#{default_command(:annotations)}` to update it.
         COMMENT
 
-        contents = content.split("\n")
+        # Split contents into newlines and ensure trailing empty lines are included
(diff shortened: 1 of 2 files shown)


lib/tapioca/commands/annotations.rb
```diff
@@ -195,7 +195,7 @@ def add_header(name, content)
 
         contents = content.split("\n")
```
paracycle, line 196: As per my other comment:
```suggestion
        # Split contents into newlines and ensure trailing empty lines are included
        contents = content.split("\n", -1) 
```

paracycle (approved):

Morriar (approved):

PR #1063, opened 2022-07-14
Fix crash on `realpath` call on non-existing paths

Morriar:
### Motivation

Currently, loading any application (gem/dsl) that has a dependency to a gem eval'ing a constant (such as `rbtrace`) will crash Tapioca:

```
Requiring all gems to prepare for compiling... bundler: failed to load command: tapioca
/gems/tapioca-0.9.1/lib/tapioca/runtime/trackers/constant_definition.rb:54:in `realpath': No such file or directory @ rb_check_realpath_internal - eval (Errno::ENOENT)
	from gems/tapioca-0.9.1/lib/tapioca/runtime/trackers/constant_definition.rb:54:in `build_constant_location'
	from gems/tapioca-0.9.1/lib/tapioca/runtime/trackers/constant_definition.rb:37:in `block in <module:ConstantDefinition>'
	from eval:2:in `<module:RBTrace>'
	from eval:1:in `/home/spin/.bundle/shopify/gems/rbtrace-0.4.14/lib/rbtrace.so'
```

The problem here is calling `realpath` with a "file" named `rb_check_realpath_internal - eval` and `realpath` expects "all components of the pathname must exist when this method is called".

### Implementation

Do not call `realpath` is the file does not exist.

### Tests

See added test.


lib/tapioca/runtime/trackers/constant_definition.rb
```diff
@@ -51,7 +51,11 @@ class ConstantLocation < T::Struct
 
         def self.build_constant_location(tp, locations)
           file = resolve_loc(caller_locations)
-          lineno = file == File.realpath(tp.path) ? tp.lineno : 0
+          lineno = 0
+
+          if File.exist?(tp.path) && file == File.realpath(tp.path)
```
st0012, line 56: We may want to check `tp.path` as well because it can also be `nil`. We had this issue in the debugger before: https://github.com/ruby/debug/issues/466

It's a very rare case though

lib/tapioca/runtime/trackers/constant_definition.rb
```diff
@@ -25,16 +25,17 @@ class ConstantLocation < T::Struct
...
 
-          if tp.path == "(eval)"
+          path = tp.path
+          if File.exist?(path)
+            loc = build_constant_location(tp, caller_locations)
+          else
             caller_location = T.must(caller_locations)
-              .drop_while { |loc| loc.path == "(eval)" }
+              .select { |loc| loc.path && File.exist?(loc.path) }
               .first
```
paracycle, line 34: Maybe `find` instead:
```
            caller_location = T.must(caller_locations)
              .find { |loc| loc.path && File.exist?(loc.path) }
```
Now that I look at this code, I have no idea why I didn't use `find` myself 😂

paracycle (approved):

PR #1064, opened 2022-07-15
Track proc bindings of existing signatures

vinistock:
### Motivation

When we generate gem RBIs, we're not carrying over proc bindings from existing signatures. The reason is because `sorbet-runtime` doesn't actually keep track of the bind and just discards it.

This ends up creating some frustration when generating gem RBIs as proc bindings are ignored.

### Implementation

Add a new Sorbet patch to keep track of the bind and print the `T::Types::Proc` name including it.

### Tests

Added an example in the pipeline spec.

### DIFF
--- a/spec/tapioca/gem/pipeline_spec.rb
+++ b/spec/tapioca/gem/pipeline_spec.rb
@@ -3868,6 +3868,26 @@ class Tapioca::Gem::PipelineSpec < Minitest::HooksSpec
       assert_equal(output, compile)
     end
 
+    it "compiles proc bindings" do
+      add_ruby_file("foo.rb", <<~RUBY)
+        class Foo
+          extend T::Sig
+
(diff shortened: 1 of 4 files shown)


paracycle (approved):
Perfect, thank you!

spec/tapioca/gem/pipeline_spec.rb
```diff
@@ -3868,6 +3868,26 @@ module ViewHelpers::UrlHelper; end
       assert_equal(output, compile)
     end
 
+    it "compiles proc bindings" do
+      add_ruby_file("foo.rb", <<~RUBY)
+        class Foo
+          extend T::Sig
+
+          sig { params(block: T.proc.bind(String).void).void }
```
Morriar, line 3876: Can we have a test with `T.proc.params`? 🙏

Morriar (approved):

PR #1066, opened 2022-07-15
Rework method/parameter name validation to use the compiler

paracycle:
### Motivation

Our current way of doing method name and parameter name checks are a bit hacky and [that's been showing through](https://github.com/Shopify/tapioca/pull/1045).

### Implementation

A better way to do this is to use the compiler to do the name checks, similar to [how Ruby does it in its codebase](https://github.com/ruby/ruby/blob/241dced625f9ba8a4071954579778a0940e75179/ext/rubyvm/lib/forwardable/impl.rb#L3-L9). (Thanks to @tenderlove for showing me this method name validation check)

This method relies on compiling a method definition with the given name and then checking if the compiled code actually defines a method with that name.

### Tests

Only updated the error message from a test to make it explicit which case is failing.

### DIFF
--- a/lib/tapioca/helpers/rbi_helper.rb
+++ b/lib/tapioca/helpers/rbi_helper.rb
@@ -93,15 +93,36 @@ module Tapioca
 
     sig { params(name: String).returns(T::Boolean) }
     def valid_method_name?(name)
-      name == "==" || !(
-        name.to_sym.inspect.start_with?(':"', ":@", ":$") ||
-        name.delete_suffix("=").to_sym.inspect.start_with?(':"', ":@", ":$")
-      )
+      # try to parse a method definition with this name
+      iseq = RubyVM::InstructionSequence.compile("def #{name}; end", nil, nil, 0, false)
+      # pull out the first operation in the instruction sequence and its first argument
+      op, arg, _data = iseq.to_a.dig(-1, 0)
+      # make sure that the operation is a method definition and the method that was
+      # defined has the expected name, for example, for `def !foo; end` we don't get
+      # a syntax error but instead get a method defined as `"foo"`
+      op == :definemethod && arg == name.to_sym
+    rescue SyntaxError
+      false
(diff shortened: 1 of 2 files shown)


lib/tapioca/helpers/rbi_helper.rb
```diff
@@ -93,15 +93,32 @@ def self.serialize_type_variable(type, variance, fixed, upper, lower)
...
+      # a syntax error but instead get a method defined as `"foo"`
+      op == :definemethod && arg == name.to_sym
+    rescue SyntaxError
+      false
     end
 
     sig { params(name: String).returns(T::Boolean) }
     def valid_parameter_name?(name)
-      /^([[:lower:]]|_|[^[[:ascii:]]])([[:alnum:]]|_|[^[[:ascii:]]])*$/.match?(name)
+      sentinel_method_name = :sentinel_method_name
```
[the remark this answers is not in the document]
Morriar: Interesting!

It's definitely better than the symbol hack and, as you said, at least we don't have to bother with it anymore since now it's handled by the Ruby VM.

lib/tapioca/helpers/rbi_helper.rb
```diff
@@ -93,15 +93,32 @@ def self.serialize_type_variable(type, variance, fixed, upper, lower)
...
+      # try to parse a method definition with this name as the name of kwarg
+      iseq = RubyVM::InstructionSequence.compile("def #{sentinel_method_name}(#{name}:); end", nil, nil, 0, false)
```
Morriar, line 112: Why does this need to be a `kwarg` rather than a positional?

paracycle: Hah, sorry, maybe I should have commented about that. The reason is that if we do a positional, then `&` ends up being a valid parameter name, for one thing.

Morriar (approved):

PR #1049, opened 2022-07-08
Add table of contents to README 

wildmaples:
### Motivation

* Add table of contents to README 
* Link to `Migrating to Tapioca` on the README 

### Implementation
N/A

### Tests
N/A

PR #1068, opened 2022-07-19
Auto-generate Table of Contents

egiurleo:
Based off of @wildmaples's work in #1049 

### Motivation
The Tapioca README is getting kind of long, and a table of contents would make it easier to navigate without scrolling.

### Implementation
Use the [TOC Generator](https://github.com/marketplace/actions/toc-generator) GitHub action to push a commit updating the table of contents every time the README is changed.

Here's what the auto-generated table of contents looks like:

Screen Shot 2022-07-19 at 12 32 44 PM


Morriar (commented):
I think it would make more sense to do this locally from `bin/readme`.

PR #1074, opened 2022-07-22
Don't crash on rails < 7

fsateler:
The method `subtype` was made public in rails 7, but it did exist previously so we can use it via `send`.

### Motivation

With rails < 7, tapioca dsl crashes with: `lib/tapioca/dsl/helpers/active_record_column_type_helper.rb:122:in 'enum_setter_type': private method 'subtype' called for #<ActiveRecord::Enum::EnumType:0x000055c8a8aefb70> (NoMethodError)`

### Implementation

subtype was not public in rails 6, but it did exist since at least 5.0 https://github.com/rails/rails/blob/v5.0.0/activerecord/lib/active_record/enum.rb#L143

### Tests

No idea how to add a test for this.

### DIFF
--- a/lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
+++ b/lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
@@ -119,7 +119,8 @@ module Tapioca
 
         sig { params(column_type: ActiveRecord::Enum::EnumType).returns(String) }
         def enum_setter_type(column_type)
-          case column_type.subtype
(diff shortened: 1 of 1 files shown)


paracycle (approved):

rafaelfranca:
I don't see a failure when I run tapioca against Rails 6.1 before this patch. Is this helper being tested?

rafaelfranca:
Ah, found it. It is because the enum definition doesn't work in Rails 6.1, so I had to change that before seeing this failure.

PR #1079, opened 2022-07-28
Generate enum module definitions

### LINKED ISSUE #604: protobuf dsl compiler does not create type definitions for enums
Method signatures which return an enum are generated correctly, but no definitions of the enums themselves are generated.

There is an example of this in the test suite already https://github.com/Shopify/tapioca/blob/afcecf2da3e3b0288978b1acc02f37ac5a4650a7/spec/tapioca/compilers/dsl/protobuf_spec.rb#L151

No rbi is generated for `Cart::VALUE_TYPE` which will result in type errors (`Unable to resolve constant`)

georgebrock:
### Motivation

Fixes #604 

### Implementation

This PR is an update to #607 with a few extra commits to make the following changes:

- The return type of all setter methods is set to `void` because Ruby ignores return values from setter methods.
- The return type of enum value getter methods is changed to `T.any(Symbol, Integer)` to account for how Protobuf handles invalid enum values.

### Tests

Includes updated tests 👍

### DIFF
--- a/lib/tapioca/compilers/dsl/protobuf.rb
+++ b/lib/tapioca/compilers/dsl/protobuf.rb
@@ -64,7 +64,6 @@ module Tapioca
         class Field < T::Struct
           prop :name, String
           prop :type, String
-          prop :setter_type, String
           prop :init_type, String
           prop :default, String
         end
@@ -167,7 +166,6 @@ module Tapioca
               Field.new(
                 name: descriptor.name,
                 type: type,
-                setter_type: type,
                 init_type: "T.nilable(T.any(#{type}, T::Hash[#{key_type}, #{value_type}]))",
                 default: "Google::Protobuf::Map.new(#{default_args.join(", ")})"
               )
@@ -181,7 +179,6 @@ module Tapioca
               Field.new(
                 name: descriptor.name,
                 type: type,
-                setter_type: type,
                 init_type: "T.nilable(T.any(#{type}, T::Array[#{elem_type}]))",
                 default: "Google::Protobuf::RepeatedField.new(#{default_args.join(", ")})"
               )
@@ -192,7 +189,6 @@ module Tapioca
             Field.new(
               name: descriptor.name,
               type: type,
-              setter_type: type,
               init_type: "T.nilable(#{type})",
               default: "nil"
             )
@@ -215,7 +211,7 @@ module Tapioca
 
           klass.create_method(
             "#{field.name}=",
-            parameters: [create_param("value", type: field.setter_type)],
+            parameters: [create_param("value", type: field.type)],
             return_type: "void"
           )
 
--- a/lib/tapioca/compilers/dsl/protobuf.rb
+++ b/lib/tapioca/compilers/dsl/protobuf.rb
@@ -167,7 +167,7 @@ module Tapioca
                 name: descriptor.name,
                 type: type,
                 setter_type: type,
-                init_type: "T.any(#{type}, T::Hash[#{key_type}, #{value_type}])",
+                init_type: "T.nilable(T.any(#{type}, T::Hash[#{key_type}, #{value_type}]))",
                 default: "Google::Protobuf::Map.new(#{default_args.join(", ")})"
               )
             else
@@ -181,7 +181,7 @@ module Tapioca
                 name: descriptor.name,
                 type: type,
                 setter_type: type,
-                init_type: "T.any(#{type}, T::Array[#{elem_type}])",
+                init_type: "T.nilable(T.any(#{type}, T::Array[#{elem_type}]))",
                 default: "Google::Protobuf::RepeatedField.new(#{default_args.join(", ")})"
               )
             end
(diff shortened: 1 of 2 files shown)


KaanOzkan (commented):
@georgebrock Can you complete the CLA step?

paracycle:
> * The return type of all setter methods is set to `void` because Ruby ignores return values from setter methods.

I don't think this is quite correct. In Ruby the setter always returns the value that was passed in, regardless of what the setter method returns:
```
[1] pry(main)> def a=(value)
[1] pry(main)*   :foo
[1] pry(main)* end
=> :a=
[2] pry(main)> a = 42
=> 42
```

Moreover, the return value is NOT ignored, since it is essential for doing chained sets like `a = b = c = 1` (which roughly translates to `send(:a=, send(:b=, send(:c=, 1)))`).

georgebrock:
> I don't think this is quite correct. In Ruby the setter always returns the value that was passed in, regardless of what the setter method returns

There's a subtle but important difference between the value of the assignment expression and the return value of the setter method. Here's a (slightly modified[^1]) version of the example you gave:

```
[1] pry(main)> def a=(value)
[1] pry(main)*   :foo
[1] pry(main)* end
=> :a=
[2] pry(main)> self.a = 42
=> 42
```

What we're seeing as the resulting value in pry is the result of Ruby interpreting the assignment expression `self.a = 42`, and while the `#a=` method is called as part of this expression its return value isn't what determines the value of the expression.

We can prove that the _return value_ is `:foo` if we use `#send` or `Method#call` to invoke the method directly[^1] instead of implicitly invoking it as part of an assignment expression:

```
[3] pry(main)> self.send(:a=, 42)
=> :foo
[4] pry(main)> self.method(:a=).call(42)
=> :foo
```

You gave an example of chained assignments being equivalent to composed calls to `#send`, but we can see from the behaviour of `#call` that isn't the case.

<details>
<summary>We can write that example out in full to see the difference …</summary>

``` ruby
def a=(v)
  puts "#a= called with #{v}"
  return :foo
end

def b=(v)
  puts "#b= called with #{v}"
  return :bar
end

def c=(v)
  puts "#c= called with #{v}"
  return :baz
end
```

```
[4] pry(main)> self.a = self.b = self.c = 42
#c= called with 42
#b= called with 42
#a= called with 42
=> 42
[5] pry(main)> self.send(:a=, self.send(:b=, self.send(:c=, 42)))
#c= called with 42
#b= called with baz
#a= called with bar
=> :foo
```

</details>

Even if we use syntax that looks more like a method call than a typical assignment, Ruby still interprets it as an assignment expression:

```
[5] pry(main)> self.a=(42)
=> 42
```

<details>
<summary>If we want to get really in the weeds …</summary>

… we can prove that this last example really is interpreted as an assignment expression instead of a method call by running it through the Ruby parser and inspecting the AST:

```
[6] pry(main)> require 'ripper'
[7] pry(main)> require 'pp'
[8] pry(main)> pp Ripper.sexp('self.a=(42)')
[:program,
 [[:assign,
   [:field,
    [:var_ref, [:@kw, "self", [1, 0]]],
    [:@period, ".", [1, 4]],
    [:@ident, "a", [1, 5]]],
   [:paren, [[:@int, "42", [1, 8]]]]]]]
```

Compared to the AST of a method call to a non-setter method:

```
[9] pry(main)> pp Ripper.sexp('self.a(42)')
[:program,
 [[:method_add_arg,
   [:call,
    [:var_ref, [:@kw, "self", [1, 0]]],
    [:@period, ".", [1, 4]],
    [:@ident, "a", [1, 5]]],
   [:arg_paren, [:args_add_block, [[:@int, "42", [1, 7]]], false]]]]]
```

</details>


There were more detailed examples of how all of this applies to Sorbet in my comment here: https://github.com/Shopify/tapioca/pull/607#issuecomment-1196370225

[^1]: There's a minor bug in the original example. The second expression (`a = 42`) is a local variable assignment, which doesn't make use of the `#a=` method at all. I've changed it to an explicit `self.a = 42` to ensure that the method is called.

[^2]: The behaviour of `#send` and `Method#call` don't really matter in the context of this PR, since Sorbet considers the their return values to be `T.untyped`, but it helps illustrate the difference between the return value and the assignment expression.

lib/tapioca/compilers/dsl/protobuf.rb
```diff
@@ -153,7 +166,7 @@ def field_of(descriptor)
               Field.new(
                 name: descriptor.name,
                 type: type,
-                init_type: "T.any(#{type}, T::Hash[#{key_type}, #{value_type}])",
+                init_type: "T.nilable(T.any(#{type}, T::Hash[#{key_type}, #{value_type}]))",
```
paracycle, line 169: These `init_type` changes also make sense, but seem like distinct from the change this PR is trying to make. Could we make these changes in a separate PR please?

georgebrock:
CLA is now sorted.

I think I've worked through all of the feedback here, so this is probably ready for another look.

There were a couple of suggestions about splitting this out into multiple PRs, and I definitely could do that if it's still deemed necessary. It's complicated a little by the fact that I've taken over work on an existing PR here, and tried not to touch the original author's commits.

paracycle (approved):
This looks much better, thank you! I think there might be a missed expectation in the new enum module test, when that is fixed, I think this is good to ship.

Thank you 🙏

georgebrock:
@paracycle Thanks!

I pushed a commit to fix the failing test, and another to fix a small bug that I found in the process.

PR #1082, opened 2022-07-28
Disable TracePoints when running the check-shims command

rafaelfranca:
When profiling why this command was taking so much time, I found out that in the specific application I was working on 37% of the time was spent in the `TracePoint#method_id` method.

Since we don't need this TracePoint enabled in this command, this patch disables it.

## Profiling before
Screen Shot 2022-07-28 at 2 28 31 PM

## Profiling after
Screen Shot 2022-07-28 at 2 29 06 PM

Total time went down from 3:50min to 1:20min

### DIFF
--- a/lib/tapioca.rb
+++ b/lib/tapioca.rb
@@ -6,6 +6,18 @@ require "sorbet-runtime"
 module Tapioca
   extend T::Sig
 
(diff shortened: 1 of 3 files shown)


vinistock (approved):

rafaelfranca:
https://www.speedscope.app/

PR #1084, opened 2022-07-29
Upgrade all gems

rafaelfranca:
I'm changing CI to not use the Gemfile.lock anymore (as it should be), so we can test different versions of Rails on CI.
That means that new versions of Sorbet are used on CI as soon they are released.

The new version of Sorbet doesn't work with our code, so I'm upgrading all gems so can fix the issues with Sorbet.


paracycle (approved):
I am 👍 with testing without `Gemfile.lock` but I feel like we will end up having a lot of pain that way due to random Sorbet changes.

rafaelfranca:
> I am 👍 with testing without `Gemfile.lock` but I feel like we will end up having a lot of pain that way due to random Sorbet changes.

Yeah. I can add multiple Gemfiles if that is a concern, but that will make development harder.

paracycle:
> ... that will make development harder.

Honestly, as long as mainline development against latest Ruby/Rails is not much harder, then I am ok with that cost. CI can catch errors with other Ruby/Rails versions and reproducing those might end up being a little harder. That's not a huge concern for me.

rafaelfranca:
Sounds, good. I'll try that direction first then.

Issue #1080, opened 2022-07-28
bin/tapioca dsl undefined method `config' for nil:NilClass (NoMethodError)


ccoffey:
Version of Tapioca that I am using:
```
bin/tapioca --version                                                  
Tapioca v0.9.2
```

Version of Ruby that I am using:
```
cat .ruby-version 
2.7.5
```

Version of Rails that I am using:
```
grep "'rails'" Gemfile
gem 'rails', '7.0.3.1'
```

The error:

```
bin/tapioca dsl
Loading Rails application... Traceback (most recent call last):
	33: from bin/tapioca:29:in `<main>'
	32: from bin/tapioca:29:in `load'
	31: from /home/cathal/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/tapioca-0.9.2/exe/tapioca:23:in `<top (required)>'
	30: from /home/cathal/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/thor-1.2.1/lib/thor/base.rb:485:in `start'
	29: from /home/cathal/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/thor-1.2.1/lib/thor.rb:392:in `dispatch'
	28: from /home/cathal/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/thor-1.2.1/lib/thor/invocation.rb:127:in `invoke_command'
	27: from /home/cathal/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/thor-1.2.1/lib/thor/command.rb:27:in `run'
	26: from /home/cathal/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/tapioca-0.9.2/lib/tapioca/cli.rb:144:in `dsl'
	25: from /home/cathal/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/sorbet-runtime-0.5.10219/lib/types/private/methods/_methods.rb:272:in `block in _on_method_added'
	24: from /home/cathal/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/sorbet-runtime-0.5.10219/lib/types/private/methods/_methods.rb:272:in `bind_call'
	23: from /home/cathal/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/tapioca-0.9.2/lib/tapioca.rb:17:in `silence_warnings'
[...]

PR #1087, opened 2022-07-29
Report Rails load errors and continue generating

KaanOzkan:
### Motivation

Resolves https://github.com/Shopify/tapioca/issues/1080
### Implementation

Rescue exceptions thrown during eager loading of the rails application

### Tests

Added a test

### DIFF
--- a/lib/tapioca/loaders/loader.rb
+++ b/lib/tapioca/loaders/loader.rb
@@ -40,12 +40,17 @@ module Tapioca
         silence_deprecations
 
         if environment_load
-          safe_require("./config/environment")
+          require "./config/environment"
         else
-          safe_require("./config/application")
+          require "./config/application"
         end
 
         eager_load_rails_app if eager_load
+      rescue LoadError, StandardError => e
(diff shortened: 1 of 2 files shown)


lib/tapioca/loaders/loader.rb
```diff
@@ -40,12 +40,17 @@ def load_rails_application(environment_load: false, eager_load: false)
...
-          safe_require("./config/environment")
+          require "./config/environment"
         else
-          safe_require("./config/application")
+          require "./config/application"
         end
 
         eager_load_rails_app if eager_load
+      rescue LoadError => e
+        say("Tapioca attempted to load the Rails application after encountering a `config/application.rb` file," \
```
rafaelfranca, line 50: Would not this only happen when the file isn't found? I think that would be the only cause for a `LoadError`.

Morriar: What do you think about catching all kinds of errors and showing them in the output rather than just `LoadError`?

paracycle (approved):

PR #1088, opened 2022-07-29
Test with supported versions of Rails

rafaelfranca:
Some of our DSL generators depend on Rails code, which might change between versions.

Since Rails is an important dependency and people might be still in an old version. It is better to test against all of them.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,30 @@ env:
   SRB_SKIP_GEM_RBIS: true
 
 jobs:
-  build:
+  linters:
+    runs-on: ubuntu-latest
+    name: Linters
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: "3.1"
+          bundler-cache: true
+      - name: Run type check
+        run: bin/typecheck
+      - name: Lint Ruby files
+        run: bin/style
+      - name: Verify documentation
+        run: bin/docs
+      - name: Verify README
+        run: bin/readme
+      - name: Verify gem RBIs are up-to-date
+        run: bundle exec exe/tapioca gem --verify
+      - name: Verify duplicates in shims
+        run: bundle exec exe/tapioca check-shims
+
+  tests:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -30,7 +46,19 @@ jobs:
...
         run: bin/test
+        continue-on-error: ${{ matrix.experimental }}
       - name: Verify gem RBIs are up-to-date
         run: bundle exec exe/tapioca gem --verify
+        if: ${{ matrix.gemfile == 'Gemfile' }}
       - name: Verify duplicates in shims
         run: bundle exec exe/tapioca check-shims
+
+  buildall:
+    if: ${{ always() && github.event.pull_request }}
```
paracycle, line 57: Out of curiosity: why do we need `always()` here? Would this not work the same way if it was `${{ github.event.pull_request }}` only?

rafaelfranca: Without it, If the build step fails, this will not run, so we will not have the failing status.

paracycle: Ah, got it, thanks.

rafaelfranca:
Yes. I was thinking on doing that. I'm already skipping the rbi checks for all other Gemfiles. Let me see if I can make that work.

rafaelfranca:
Should we do

Screen Shot 2022-07-29 at 5 05 56 PM

or

Screen Shot 2022-07-29 at 5 06 18 PM

paracycle:
> Should we do
> 
> Screen Shot 2022-07-29 at 5 05 56 PM
> 
> or Screen Shot 2022-07-29 at 5 06 34 PM

I was imagining linters running before tests, but parallel also works I guess. Thanks for doing this 🙏

rafaelfranca:
I'll go with parallel then since it would be faster.

paracycle (approved):
Thank you, this is great. Regardless of how the pipeline works, this is much better than what we already had.

Do you know why the `Build (matrix)` didn't report a status, though?

rafaelfranca:
> Do you know why the `Build (matrix)` didn't report a status, though?

It doesn't but only when it starts (you not seeing because I rebased many times). Not sure how to make it always report.

You can see in this commit https://github.com/Shopify/tapioca/pull/1088/commits/607daabc3951097af454f53312b508b53c0ea7f0.

After I merge I'll make it required, so it will always show here, but as pending.

.github/workflows/ci.yml
```diff
@@ -28,9 +24,51 @@ jobs:
...
+          - gemfile: gemfiles/Gemfile-rails-main
+            ruby: "2.7"
+            experimental: true
+          - gemfile: gemfiles/Gemfile-rails-main
+            ruby: "3.0"
+            experimental: true
+          - gemfile: gemfiles/Gemfile-rails-main
+            ruby: "3.1"
+            experimental: true
+    name: Ruby ${{ matrix.ruby }} - Gemfile ${{ matrix.gemfile }}
```
paracycle, line 52: One nitpick: Do we really need to write `Gemfile` here, since the gemfile itself has that in its name
```suggestion
    name: Ruby ${{ matrix.ruby }} - ${{ matrix.gemfile }}
```

PR #1091, opened 2022-08-01
Optimize and resolve TODO in SymbolTableParser

dduugg:
### Motivation

Resolves the todo introduced in https://github.com/Shopify/tapioca/commit/d18d7d00dfb3c3adf8f2aafd7606cc119158a5dd
cc @paracycle 

Also adds some adjacent perf improvements:
- avoid creating unused regex capture groups by switching `=~` to `.match?`
- extract literal collection in loop to a constant

(FWIW i think that both of these would be flagged by rubocop-performance, if it were in use)

### Implementation

I think this is self-explanatory, but happy to elaborate as needed

### Tests

Relying on existing tests 🙈

### DIFF
--- a/lib/tapioca/static/symbol_table_parser.rb
+++ b/lib/tapioca/static/symbol_table_parser.rb
@@ -6,6 +6,8 @@ module Tapioca
     class SymbolTableParser
       extend T::Sig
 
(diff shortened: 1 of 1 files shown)


paracycle (approved):
🙏 Thanks for the cleanup for this section of the codebase.

PR #1094, opened 2022-08-02
Use a custom URI class to handle source location comments

vinistock:
### Motivation

We are currently relying on regexes and string substitutions to find the right paths for gems. This is not reliable and has proven to not find the right paths.

### Implementation

This PR introduces a custom URI object to handle the `source://` scheme. This makes it easier to validate the consistency of the URI and we can use a similar object on the Ruby LSP to do the inverse operation of parsing it.

It also improves the detection of default gem paths, by using `Gemfile::GemSpec`, which already has the mechanisms to figure out these paths for us.

The format of source URIs is now:
```
source://tapioca/1.2.3/lib/foo.rb#13
^ scheme
         ^ gem name
                 ^ gem version (optional)
                       ^ path
                                  ^ line number
```

### Tests

Updated existing tests.


st0012 (approved):
I think it may be too much effort but are we able to have a test for the `BUNDLER_HOME` scenario?

paracycle (commented):
I am not sure if I understand why we are doing more breakdowns here. Isn't it the case that if I have `foo-1.2.3` as a dependency, that it should be referenced in the RBI file as `foo-1.2.3`? Do we care if I suddenly change my Bundler setup so that it installs `foo` to the bundle path? If so, why do we care?

vinistock:
@paracycle while testing this on Core, I realized that we have a mix of both. Some gems are installed under `Bundler.home` and others are installed under `Bundle.bundle_path`.

If this information is not present somehow in the source location comments, the Ruby LSP has no way of figuring out where the appropriate path is.

I have no clue why we have gems on both locations though. If you have more context on why this could be, let me know.

paracycle:
> If this information is not present somehow in the source location comments, the Ruby LSP has no way of figuring out where the appropriate path is.

LSP can search in both locations?

vinistock:
Is that better though? That means that instead of processing that ahead of time in Tapioca, we'll need to search on both when serving the request.

lib/tapioca/helpers/source_uri.rb
```diff
@@ -0,0 +1,42 @@
+# typed: true
+# frozen_string_literal: true
+
+require "uri/file"
+
+module URI
```
Morriar, line 6: Shouldn't this be under the `Tapioca` namespace?

lib/tapioca/helpers/source_uri.rb
```diff
@@ -0,0 +1,42 @@
...
+
+      @gem_version, @path = v.delete_prefix("/").split("/", 2)
+    end
+
+    sig { returns(String) }
+    def to_s
+      "source://#{gem_name}/#{gem_version}/#{path}##{line_number}"
+    end
+
+    if RUBY_VERSION >= "3.1"
```
paracycle, line 36: Can we do this like how `URI::GlobalId` does it: https://github.com/rails/globalid/blob/main/lib/global_id/uri/gid.rb#L176 (basically by using `respond_to?` instead of doing version checks)

lib/tapioca/gem/listeners/source_location.rb
```diff
@@ -41,25 +55,31 @@ def on_method(event)
...
+            "#{gem.name}//#{path.realpath.relative_path_from(RbConfig::CONFIG["rubylibdir"])}"
           else
-            path.sub("#{Bundler.bundle_path}/gems/", "").to_s
+            "#{gem.name}//#{path.realpath.relative_path_from(gem.full_gem_path)}"
           end
 
-          # Strip out the RUBY_ROOT prefix, which is different for each user
-          path = path.sub(RbConfig::CONFIG["rubylibdir"], "RUBY_ROOT")
-
+          uri = URI("source://#{path}##{line}")
```
paracycle, line 78: Can we instead use `URI.build` and pass in all the components explicitly? I feel like we are still treating URI building as string building which isn't the best thing to do.

lib/tapioca/gem/listeners/source_location.rb
```diff
@@ -41,25 +55,31 @@ def on_method(event)
...
+          # If the definition source location is not from the current gem, look for it in dependencies
+          gem = @gem_specs[realpath] unless realpath.start_with?(gem.full_gem_path)
+          return if gem.nil?
+
+          path = if gem.default_gem?
+            "#{gem.name}//#{path.realpath.relative_path_from(RbConfig::CONFIG["rubylibdir"])}"
           else
-            path.sub("#{Bundler.bundle_path}/gems/", "").to_s
+            "#{gem.name}//#{path.realpath.relative_path_from(gem.full_gem_path)}"
           end
```
paracycle, line 76: I think it would be best to encapsulate the logic of finding the relative path of a given file path in the `GemSpec` class, so that we don't have to know about libdirs and default gems, etc here.

Something like:
```
path = gem.relative_path_for(file)
```

lib/tapioca/gem/listeners/source_location.rb
vinistock: You can see what I mean [here](https://github.com/Shopify/ruby-lsp/pull/237).

paracycle: Right, I see what you are doing there, and honestly I don't want to delay this PR longer, but I also feel like it would be so much more efficient and less code if we did almost the exact opposite of what we are doing here in the client.

Basically, the client can:
1. Using `Gem::Specification.default_stubs` and `.stubs`, build a hash with keys as `[gem_name, gem_version]` and values as a hash from relative file names to full file names.
2. When we are asked handle a source: URI, we parse it and grab the `gem_name`, which is guaranteed to always exist.
3. We then parse the `gem_version`, if it doesn't exist, then we parse it from the file name. We expect this to always end up with a gem version number, if not, we bail and show an error or something.
4. Now that we have a `[gem_name, gem_version]` we look it up in our table. If the gem name and version is also loaded, that gives us another hash table. If the gem name/version is not found, we fail gracefully.
5. Now we are left with a hashtable of relative file names to full file names, and we use the `path` of the URI to look up the full file path.
6. We open that file on the `line_number` from the URI.

This feels much easier to implement and way more robust than looking up files in various magical folders, and trying to guess the version of the gem, etc. WDYT?

vinistock: I agree with this approach. But if we look up in `default_stubs + stubs`, would we ever need to try to figure out the version from the file name?

Also, given this strategy, do you still wish to include the versions in the comments?

paracycle: I think there is still value and that is in cases where the RBI file has gone stale. Many projects are behind on updating this gem RBI files and we should be resilient to that. So including the version (when it can't be inferred from the filename) is a good thing for graceful handling of version mismatches. Basically, a gem's installed version could be drastically different from the RBI gem version and file paths/line numbers could have changed a lot in between those versions as well.

Specifically, if we change my suggestion above to be have an extra layer of lookup, with initial layer begin gem name and second layer being gem version and last layer being relative file path, then we can fail in this way:

1. If the lookup on gem name fails then LSP client can warn saying that the gem dependency has been removed and prompt to update RBI files.
2. If the gem version lookup fails, then we can prompt saying the gem version is different and file/line might have changed and ask if the user still wants to continue.
3. If the relative file path has changed, communicate that tot hw user and maybe again prompt for RBI generation again.

dirceu:
Code looks good to me. Questions for my own understanding:

1. Is ruby-lsp the only client that currently consumes those source location comments?
2. Given that Tapioca usually doesn't regenerate RBIs unless you change the version of a gem, does that mean that ruby-lsp will have to support multiple formats of source location comments?

paracycle (approved):

PR #1099, opened 2022-08-05
Don't skip over abstract FrozenRecord classes

amomchilov:
### Motivation

On `shop-server`, we have an `abstract!` base class that derives from `FrozenRecord::Base`.

Currently, this class doesn't have an RBI file generated, because the FrozenRecord Tapioca compiler is explicitly skipping over abstract classes.

I don't think there's a reason *not* to include these, so let's just do it.

### Tests

There wasn't a test to check that `abstract` classes are skipped. If there was one, I would have deleted it. I don't think there's a need to add a test to verify that's we're not not including abstract files, but LMK if you think I should add one anyway :)

### DIFF
--- a/lib/tapioca/dsl/compilers/frozen_record.rb
+++ b/lib/tapioca/dsl/compilers/frozen_record.rb
@@ -90,7 +90,7 @@ module Tapioca
 
         sig { override.returns(T::Enumerable[Module]) }
         def self.gather_constants
-          descendants_of(::FrozenRecord::Base).reject(&:abstract_class?)
+          descendants_of(::FrozenRecord::Base)
         end
 
         private


vinistock (approved):

Morriar (commented):
> LMK if you think I should add one

I think you can add a test under https://github.com/Shopify/tapioca/blob/main/spec/tapioca/dsl/compilers/frozen_record_spec.rb showing that we do not skip abstract classes anymore. At least to show that we do not skip them intentionally.

amomchilov:
We actually moved away from having that abstract base class. There were other edge-cases in FrozenRecord that made it more painful than it was worth. I won't have time to investigate this before BFCM, so perhaps we could just close this issue

PR #1101, opened 2022-08-08
Revert "Make ActiveRecord#find return value typed"

paracycle:
Reverts Shopify/tapioca#1089

The PR was merged, but the implementation is flawed and will not work in real applications. 

Union types as return values don't work the way folks imagine them to work:
```ruby
user = User.find(1)

user.name # Method name does not exist on T::Array[User] component of T.any(T::Array[User], User)
```

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_relations.rb
+++ b/lib/tapioca/dsl/compilers/active_record_relations.rb
@@ -522,7 +522,7 @@ module Tapioca
                 parameters: [
                   create_rest_param("args", type: "T.untyped"),
                 ],
-                return_type: "T.any(#{constant_name}, T::Array[#{constant_name}])"
+                return_type: "T.untyped"
               )
             when :find_by
               create_common_method(
--- a/spec/tapioca/dsl/compilers/active_record_relations_spec.rb
+++ b/spec/tapioca/dsl/compilers/active_record_relations_spec.rb
@@ -93,7 +93,7 @@ module Tapioca
                     sig { returns(::Post) }
                     def fifth!; end
 
-                    sig { params(args: T.untyped).returns(T.any(::Post, T::Array[::Post])) }
+                    sig { params(args: T.untyped).returns(T.untyped) }
                     def find(*args); end
 
                     sig { params(args: T.untyped).returns(T.nilable(::Post)) }


vinistock:
I know this is not the point of this PR, but I wanted to call attention to the fact that Sorbet has started accepting overloaded method signatures ([example](https://github.com/sorbet/sorbet/blob/master/rbi/core/kernel.rbi#L726-L733)).

Could we (in a follow up PR) change this generator to add two signatures? One that accepts and returns an array and another one that accepts a string or integer and returns a single record?

paracycle:
> I know this is not the point of this PR, but I wanted to call attention to the fact that Sorbet has started accepting overloaded method signatures ([example](https://github.com/sorbet/sorbet/blob/master/rbi/core/kernel.rbi#L726-L733)).
> 
> Could we (in a follow up PR) change this generator to add two signatures? One that accepts and returns an array and another one that accepts a string or integer and returns a single record?

Sorbet has always accepted overloaded signatures in the payload since the beginning. But that is only guaranteed to work in `# typed: __STDLIB_INTERNAL` files which are supposed to be files in the payload.

[I'd made a case](https://sorbet-ruby.slack.com/archives/C02VD06EQ3U/p1657580889228799?thread_ts=1657567007.097839&cid=C02VD06EQ3U) for making that possible in all RBI files, which was met with a [hesitantly warm reception](https://sorbet-ruby.slack.com/archives/C02VD06EQ3U/p1657580912035449?thread_ts=1657567007.097839&cid=C02VD06EQ3U), but there are [edge cases to consider](https://sorbet-ruby.slack.com/archives/C02VD06EQ3U/p1657580970731379?thread_ts=1657567007.097839&cid=C02VD06EQ3U).

andredriem:
> > I know this is not the point of this PR, but I wanted to call attention to the fact that Sorbet has started accepting overloaded method signatures ([example](https://github.com/sorbet/sorbet/blob/master/rbi/core/kernel.rbi#L726-L733)).
> > Could we (in a follow up PR) change this generator to add two signatures? One that accepts and returns an array and another one that accepts a string or integer and returns a single record?
> 
> Sorbet has always accepted overloaded signatures in the payload since the beginning. But that is only guaranteed to work in `# typed: __STDLIB_INTERNAL` files which are supposed to be files in the payload.
> 
> [I'd made a case](https://sorbet-ruby.slack.com/archives/C02VD06EQ3U/p1657580889228799?thread_ts=1657567007.097839&cid=C02VD06EQ3U) for making that possible in all RBI files, which was met with a [hesitantly warm reception](https://sorbet-ruby.slack.com/archives/C02VD06EQ3U/p1657580912035449?thread_ts=1657567007.097839&cid=C02VD06EQ3U), but there are [edge cases to consider](https://sorbet-ruby.slack.com/archives/C02VD06EQ3U/p1657580970731379?thread_ts=1657567007.097839&cid=C02VD06EQ3U).

Those links are private now. Is there any news regarding the overloaded signatures?

Morriar:
> Is there any news regarding the overloaded signatures?

Overloads are now allowed in RBI files as an experimental feature: https://sorbet.org/docs/overloads.

PR #1110, opened 2022-08-10
Fix failures against Rails `main`

paracycle:
### Motivation

Our builds against Rails `main` have been failing, so this PR attempts to address all the changes since Rails 7.

### Implementation

Each commit explains what was changed with a link to upstream change that made it necessary.

### Tests

No additional tests

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_fixtures.rb
+++ b/lib/tapioca/dsl/compilers/active_record_fixtures.rb
@@ -42,13 +42,12 @@ module Tapioca
 
         sig { override.void }
         def decorate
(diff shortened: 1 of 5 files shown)


vinistock (approved):

rafaelfranca:
CI for all other Rails versions is now broken.

paracycle:
@rafaelfranca Thanks, I missed that the block adding `without` was for `>= 7.0`. The `with` method is only added for `>= 7.1.alpha`, so I updated that.

rafaelfranca (approved):

PR #1112, opened 2022-08-15
Show full result when failing to match output in tests

dirceu:
We seem to have a [flaky test in `main`](https://github.com/Shopify/tapioca/runs/7813457788?check_suite_focus=true#step:4:650); the test failure only tells us that the output is empty, but it doesn't show anything about stderr or about the overall success.

This PR introduces and uses two new assertions that might make it easier to debug such test failures:

* `assert_stdout_includes`
* `assert_stderr_includes`

If I manually setup a failure on that flaky test, this is what we get now:

```
cli
  gem
    generate
      it must not include `rbi` definitions into `tapioca` RBI        FAIL (1.61s)
        ########## STDOUT ##########
        Requiring all gems to prepare for compiling...  Done
        
        Removing RBI files of gems that have been removed:
        
          Nothing to do.
        
          Compiled tapioca
              create  sorbet/rbi/gems/tapioca@0.9.2.rbi
        
        Checking generated RBI files...  Done
        
          Changed strictness of sorbet/rbi/gems/tapioca@0.9.2.rbi to `typed: false` (conflicting with DSL files)
        
        All operations performed in working directory.
        Please review changes and commit them.
        
        ########## STDERR ##########
        <empty>
        ########## STATUS: true ##########
```

### DIFF
--- a/spec/spec_with_project.rb
+++ b/spec/spec_with_project.rb
@@ -127,6 +127,16 @@ module Tapioca
       refute(result.status)
     end
 
+    sig { params(result: MockProject::ExecResult, snippet: String).void }
+    def assert_stdout_includes(result, snippet)
+      assert_includes(result.out, snippet, result.to_s)
+    end
+
+    sig { params(result: MockProject::ExecResult, snippet: String).void }
+    def assert_stderr_includes(result, snippet)
+      assert_includes(result.err, snippet, result.to_s)
+    end
+
     private
 
     sig { returns(String) }
--- a/spec/tapioca/cli/annotations_spec.rb
+++ b/spec/tapioca/cli/annotations_spec.rb
@@ -40,8 +40,8 @@ module Tapioca
 
         result = @project.tapioca("annotations --sources #{repo.path}")
 
-        assert_includes(result.out, "remove  sorbet/rbi/annotations/bar.rbi")
-        assert_includes(result.out, "remove  sorbet/rbi/annotations/foo.rbi")
+        assert_stdout_includes(result, "remove  sorbet/rbi/annotations/bar.rbi")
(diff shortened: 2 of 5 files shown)


spec/spec_with_project.rb
```diff
@@ -127,6 +127,16 @@ def refute_success_status(result)
       refute(result.status)
     end
 
+    sig { params(result: MockProject::ExecResult, snippet: String).void }
+    def assert_output_includes(result, snippet)
```
paracycle, line 131: We'd been using `stderr` and `stdout` in assertion names, for example, `assert_empty_stderr` and `assert_empty_stdout` above, so it would be much nicer if these were also `assert_stdout_includes` and `assert_stderr_includes`, as well.

spec/spec_with_project.rb
```diff
@@ -127,6 +127,16 @@ def refute_success_status(result)
...
+      assert_includes(result.out, snippet, result.to_s)
```
paracycle, line 132: Does this still print the original failure message that includes the diff as well? If not, it would be great to have that since it is also helpful to know what was being expected for match.

dirceu: It does still print the original failure message 👍 

```
Expected "Requiring all gems to prepare for compiling...  Done\n\nRemoving RBI files of gems that have been removed:\n\n  Nothing to do.\n\n  Compiled tapioca\n      create  sorbet/rbi/gems/tapioca@0.9.2.rbi\n\nChecking generated RBI files...  Done\n\n  Changed strictness of sorbet/rbi/gems/tapioca@0.9.2.rbi to `typed: false` (conflicting with DSL files)\n\nAll operations performed in working directory.\nPlease review changes and commit them.\n" to include "Compiled tapioca1\n".```

paracycle (approved):
🙏

PR #1113, opened 2022-08-15
Add flag to list compilers for DSL command

paracycle:
### Motivation

Since we are now loading all compilers from any loaded gem that exports a DSL compiler, the full set of DSL compilers in play might be a little confusing.

### Implementation

Thus, this PR adds a way to ask Tapioca to list the set of DSL compilers and get a tabular output that shows them as well as their enabled/disabled status.

### Tests

Added CLI tests for the feature.

### DIFF
--- a/lib/tapioca/cli.rb
+++ b/lib/tapioca/cli.rb
@@ -115,6 +115,11 @@ module Tapioca
       type: :string,
       desc: "The Rack/Rails environment to use when generating RBIs",
       default: DEFAULT_ENVIRONMENT
+    option :list_compilers,
+      aliases: ["-l"],
+      type: :boolean,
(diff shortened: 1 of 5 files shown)


vinistock (approved):

lib/tapioca/cli.rb
```diff
@@ -133,7 +138,11 @@ def dsl(*constants)
       )
 
       Tapioca.silence_warnings do
-        command.execute
+        if options[:list_compilers]
```
Morriar, line 141: I'm surprised this needs to be inside the `silence_warnings` block.

paracycle: Honestly, never even paid attention to that. It probably can be outside that block indeed.

PR #1134, opened 2022-08-24
Bump yard-sorbet

dduugg:
### Motivation

yard-sorbet v0.7.0 was just released, which includes a typo fix that will break an existing test. I'm upgrading it and fixing the test in this PR, rather than wait for dependabot to flag the change.

### Implementation

N/A

### Tests

N/A

cc @KaanOzkan

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -337,7 +337,7 @@ GEM
       nokogiri (~> 1.8)
     yard (0.9.28)
       webrick (~> 1.7.0)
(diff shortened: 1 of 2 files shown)


Morriar:
You'll need to update the RBIs for gems using `bundle exec tapioca gems`.

dduugg:
> You'll need to update the RBIs for gems using `bundle exec tapioca gems`.

Thanks, done!

paracycle (approved):

PR #1137, opened 2022-08-29
Revert "Bump sorbet-static-and-runtime from 0.5.10326 to 0.5.10370"

KaanOzkan:
Reverts Shopify/tapioca#1135 as we lose platform specific versions


KaanOzkan:
>Should we temporarily disable Rubocop on this gem until the problem is fixed upstream? We maybe accidentally merge the next bump and find ourselves in the same situation again.

I assume you mean dependabot. If we were able to close it company wide I'd definitely be in favour of that. For our projects I'm responsible of bumps this week and I can be vigilant since problem is assumed to be fixed by the end of the week. However, I'm not against skipping it for Tapioca.

PR #1145, opened 2022-09-01
Type the rails find method

richard-thatch:
### Motivation

For find_by, we have no problem inferring that it will return
T.nilable(constant_name), however we eject from trying to type find. Per
https://github.com/rails/rails/blob/33100212244efbd1adf21587e1e37090fdd3591d/activerecord/lib/active_record/relation/finder_methods.rb#L13-L19
(current main revision at time of authoriship)

We should be able to provide something a bit better, even if developers
have to T.cast when they know they're not getting an array back.

I have _zero_ idea if there was a specific reason for T.untyped in this
regard, but I thought I'd contribute this as a suggestion if this is
tenable, or a question if it is not so I can better understand.

### Implementation
Instead of having ActiveRecord::Base::find return T.untyped, have it return the constant it belongs to similar to find_by, or an array of those constants per the documentation and code on rails/rails

### Tests
I added the test I thought relevant, the tests take a while so I'll check branch CI for anything else I've missed.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_relations.rb
+++ b/lib/tapioca/dsl/compilers/active_record_relations.rb
@@ -526,7 +526,7 @@ module Tapioca
                 parameters: [
                   create_rest_param("args", type: "T.untyped"),
                 ],
-                return_type: "T.untyped"
+                return_type: "T.any(#{constant_name}, T::Enumerable[#{constant_name}])"
               )
             when :find_by
               create_common_method(
--- a/spec/tapioca/dsl/compilers/active_record_relations_spec.rb
+++ b/spec/tapioca/dsl/compilers/active_record_relations_spec.rb
@@ -93,7 +93,7 @@ module Tapioca
                     sig { returns(::Post) }
                     def fifth!; end
 
-                    sig { params(args: T.untyped).returns(T.untyped) }
+                    sig { params(args: T.untyped).returns(T.any(::Post, T::Enumerable[::Post])) }
                     def find(*args); end
 
                     sig { params(args: T.untyped).returns(T.nilable(::Post)) }


richard-thatch:
Thanks Rafael!

rafaelfranca:
Thank you for the pull request. I wonder if we have a better way to document this decision. Maybe we should document in the manual for this generator and expose the manual in the generated rbi?

richard-thatch:
I did read the manual, and I did do a quick search before asking and looking (sorry I didn't find it haha). Could also be an inline comment on the generator itself where the method is defined.

I think because it's so odd, including it in the RBI isn't such a bad idea! But I don't know if that's a current practice.

richard-thatch:
Just to avoid necro-ing the other thread really quick. I did expect this to introduce a type error when using Model.find(1). The  PR you linked noted that this was not ideal (I agree, re the future overloaded types stuff), but I was prepared to have to do the following:

Original [sorbet run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20User%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7B%20params%28args%3A%20T.untyped%29.returns%28T.any%28User%2C%20T%3A%3AArray%5BUser%5D%29%29%7D%0A%20%20def%20self.find%28*args%29%0A%20%20%20%20T.unsafe%28nil%29%0A%20%20end%0A%0A%20%20def%20name%0A%20%20end%0Aend%0A%0Auser%20%3D%20User.find%281%29%0A%0Auser.name%20%23%20Method%20name%20does%20not%20exist%20on%20T%3A%3AArray%5BUser%5D%20component%20of%20T.any%28T%3A%3AArray%5BUser%5D%2C%20User%29%0A)

What I would expect to have to do: [sorbet run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20User%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7B%20params%28args%3A%20T.untyped%29.returns%28T.any%28User%2C%20T%3A%3AArray%5BUser%5D%29%29%7D%0A%20%20def%20self.find%28*args%29%0A%20%20%20%20T.unsafe%28nil%29%0A%20%20end%0A%0A%20%20def%20name%0A%20%20end%0Aend%0A%0Auser%20%3D%20T.cast%28User.find%281%29%2C%20User%29%0A%0Auser.name%20%23%20Method%20name%20does%20not%20exist%20on%20T%3A%3AArray%5BUser%5D%20component%20of%20T.any%28T%3A%3AArray%5BUser%5D%2C%20User%29%0A)

I know that this isn't ideal, but we have to do exactly this for ActionControllerParameters, as it can return three different types. Is the decision here that this extra friction isn't worth having the value typed?

I guess in this case we can still cast T.untyped to Model to get our typing? Maybe that's the message to be put in the manual if we don't want to force the cast? E.g. (reductively paraphrasing here) "If you need a strong type from #find, you can cast from untyped like so.."

```ruby
user = T.cast(User.find(1), User)
```

PR #1148, opened 2022-09-06
Handle constants with nil value

vinistock:
### Motivation

I noticed that Tapioca was not generating RBIs for constants that have their value set to nil. E.g.:
```ruby
FOO = nil # => ignored by Tapioca
```
The reason is because inside `constantize`, applying `const_get` will return `nil` and then we skip that constant. I'm not sure if that was intentional, so I appreciate feedback on the implementation.

### Implementation

Basically, moved part of the implementation of `constantize` to only skip constants that are not defined and include the ones that have their value set to `nil`.

### Tests

Added an example.

### DIFF
--- a/lib/tapioca/gem/pipeline.rb
+++ b/lib/tapioca/gem/pipeline.rb
@@ -162,7 +162,7 @@ module Tapioca
         return if symbol_in_payload?(symbol) && !@bootstrap_symbols.include?(symbol)
 
         constant = constantize(symbol)
-        push_constant(symbol, constant) if constant
+        push_constant(symbol, constant) if Runtime::Reflection.constant_defined?(constant)
       end
 
       sig { params(event: Gem::ConstantFound).void.checked(:never) }
@@ -260,6 +260,7 @@ module Tapioca
         return if klass_name&.start_with?("T::Types::", "T::Private::")
 
         type_name = klass_name || "T.untyped"
+        type_name = "T.untyped" if type_name == "NilClass"
         node = RBI::Const.new(name, "T.let(T.unsafe(nil), #{type_name})")
         push_const(name, klass, node)
         @root << node
@@ -321,7 +322,7 @@ module Tapioca
           next unless superclass_name
 
           resolved_superclass = constantize(superclass_name)
-          next unless Module === resolved_superclass
+          next unless Module === resolved_superclass && Runtime::Reflection.constant_defined?(resolved_superclass)
           next if name_of(resolved_superclass) == constant_name
 
           # We found a suitable superclass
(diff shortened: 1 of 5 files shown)


paracycle (commented):
Can we instead make `constantize` return a sentinel value (i.e. a constant with the value `Module.new.freeze`) instead of `nil` to disambiguate actual `nil` valued constants? The `constantize` method used to do various relative or non-inherited lookup and I am scared that `const_get` behaviour might not match exactly.

lib/tapioca/gem/pipeline.rb
```diff
@@ -161,8 +161,10 @@ def on_symbol(event)
         symbol = event.symbol.delete_prefix("::")
         return if symbol_in_payload?(symbol) && !@bootstrap_symbols.include?(symbol)
 
-        constant = constantize(symbol)
-        push_constant(symbol, constant) if constant
```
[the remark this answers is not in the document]
vinistock: To be honest, I don't know why we use `const_get` instead of `const_defined?`. @paracycle any ideas on this?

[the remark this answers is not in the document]
paracycle: We don't need `const_defined?` inside the `constantize` method because the expectation is that the operation will succeed and it not succeeding is an exception.

If we did a `const_defined?` followed by a `const_get` every time, that would slow down the common path (since it might end up making the const lookup twice).

We just need to handle the exception more gracefuly.

paracycle (approved):

lib/tapioca/runtime/reflection.rb
```diff
@@ -19,6 +19,7 @@ module Reflection
       PROTECTED_INSTANCE_METHODS_METHOD = T.let(Module.instance_method(:protected_instance_methods), UnboundMethod)
       PRIVATE_INSTANCE_METHODS_METHOD = T.let(Module.instance_method(:private_instance_methods), UnboundMethod)
       METHOD_METHOD = T.let(Kernel.instance_method(:method), UnboundMethod)
+      UNDEFINED_CONSTANT = T.let(Module.new.freeze, Module)
```
Morriar, line 22: What do you think about providing an helper method to use this like `Runtime::Reflection.is_constant_undefined?`?

vinistock: What would the method do? Wouldn't it end up being the same as `constantize`?

vinistock: I tried putting this method together, but it actually fails in Sorbet runtime validations
```ruby
sig { params(constant: BasicObject).returns(T::Boolean) }
def constant_defined?(constant)
  !UNDEFINED_CONSTANT.eql?(constant)
end
```
We have a test in the pipeline spec that verifies we can process `BasicObject` and it fails with
```
NoMethodError: undefined method `is_a?' for #<BasicObjectTest::B:0x00000001184d5cb0>
  obj.is_a?(@raw_type)
                 ^^^^^^
  sorbet-runtime-0.5.10326/lib/types/types/simple.rb:24:in `valid?'
```
Any ideas on how to structure this that wouldn't break the simple validator?

vinistock: Done.

Morriar (approved):

PR #1152, opened 2022-09-08
Prevent failure when using generic interfaces in constant assignments.

amanda-mitchell:
### Motivation

I found a bug!

Tapioca is unable to compile code that uses constant assignments to generic interfaces because of a monkeypatch into Sorbet that breaks the runtime type checker.

### Implementation

The aforementioned monkeypatch creates copies of generic classes and modules in order to store information about the generic parameters that were used. (NB: I suspect that the need for the monkeypatch will go away if/when Sorbet adds runtime type checking for generics, but I digress)

I augmented this so that the created type also stores an explicit reference to the original type. I then modified the behavior of Utils#coerce (which is also monkeypatched already) so that it checks whether we're dealing with one of these copies and instead returns the value for the original type.

### Tests

I added a failing test and then fixed the behavior so it would pass.


spec/tapioca/gem/pipeline_spec.rb
```diff
@@ -3271,6 +3271,51 @@ def non_abstract_children; end
...
+
+        class Concrete
+          extend T::Generic
+
+          include GenericInterface
+          Parameter = type_member
+        end
+
+        GENERIC_CONSTANT = T.let(
+          Concrete[Numeric].new,
```
rafaelfranca, line 3292:
```suggestion
          Concrete.new,
```

The `#[]` should not be used at runtime. What is funny is that sorbet even show an error when you try to do that in some kind of expressions like `if`.

rafaelfranca: But given that, is this what causes the error you are trying to fix?

rafaelfranca:
Thank you for the pull request. The implementation makes sense, but I'm failing to understand what was broken. Did Tapioca straight failed before this PR or did the type generated for the generic constant was wrong?

rafaelfranca:
Got it. Makes sense. CI is broken though. It seems one of the generics is now using `T.untyped` when before was using `Numeric`. Can you take a look please?

amanda-mitchell:
Okay, take two: rather than returning a `T::Type::Base` for the underlying type, I made a new subclass of `T::Type::Simple` that accepts an underlying type and just overrides `T::Type::Base#is_valid?`, which was the only problematic behavior.

lib/tapioca/sorbet_ext/generic_name_patch.rb
```diff
@@ -16,6 +16,7 @@ def [](*types)
         # `register_type` method builds and returns an instantiated clone of the generic type
         # so, we just return that from this method as well.
         Tapioca::Runtime::GenericTypeRegistry.register_type(constant, types)
+        Tapioca::Runtime::GenericTypeRegistry.register_type(constant, types)
```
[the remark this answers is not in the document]
amanda-mitchell: Oops…not sure how that got in there. Will remove.

PR #1146, opened 2022-09-02
map anonymous GraphQL InputObjects to T.untyped

ryanquanz:
### Motivation

When run `tapioca dsl` against a GraphQLType that used an anonymous class inheriting from InputObject, I was seeing errors like these:

(when sorbet assertions are being enforced)
```
TypeError: Passed `nil` into T.must
            /home/spin/.bundle/tapioca/gems/sorbet-runtime-0.5.10326/lib/types/_types.rb:218:in `must'
            /home/spin/src/github.com/Shopify/tapioca/lib/tapioca/dsl/helpers/graphql_type_helper.rb:47:in `type_for'
            /home/spin/.bundle/tapioca/gems/sorbet-runtime-0.5.10326/lib/types/private/methods/call_validation_2_7.rb:106:in `bind_call'
            /home/spin/.bundle/tapioca/gems/sorbet-runtime-0.5.10326/lib/types/private/methods/call_validation_2_7.rb:106:in `block in create_validator_method_fast1'
            /home/spin/src/github.com/Shopify/tapioca/lib/tapioca/dsl/compilers/graphql_input_object.rb:55:in `block (2 levels) in decorate'
            /home/spin/src/github.com/Shopify/tapioca/lib/tapioca/dsl/compilers/graphql_input_object.rb:53:in `each'
            /home/spin/src/github.com/Shopify/tapioca/lib/tapioca/dsl/compilers/graphql_input_object.rb:53:in `block in decorate'
            /home/spin/src/github.com/Shopify/tapioca/lib/tapioca/rbi_ext/model.rb:40:in `block in create_class'
            <internal:kernel>:90:in `tap'
            /home/spin/src/github.com/Shopify/tapioca/lib/tapioca/rbi_ext/model.rb:39:in `create_class'
            /home/spin/.bundle/tapioca/gems/sorbet-runtime-0.5.10326/lib/types/private/methods/call_validation.rb:157:in `bind_call'
            /home/spin/.bundle/tapioca/gems/sorbet-runtime-0.5.10326/lib/types/private/methods/call_validation.rb:157:in `validate_call'
            /home/spin/.bundle/tapioca/gems/sorbet-runtime-0.5.10326/lib/types/private/methods/_methods.rb:270:in `block in _on_method_added'
            /home/spin/src/github.com/Shopify/tapioca/lib/tapioca/rbi_ext/model.rb:16:in `create_path'
[...]

PR #1153, opened 2022-09-09
Handle anonymous GraphQL input classes

vinistock:
### Motivation

When defining GraphQL input objects from a schema, anonymous classes are generated to define them in Ruby. This means that invoking `qualified_name_of` returns `nil` and we end up breaking in `rbi_helper` when trying to turn it into a nilable type.

For example, loading this GraphQL schema
```graphql
input CommentInput {
  body: CommentBodyInput
}

input CommentBodyInput {
  content: String!
}
```

### Implementation

Started returning `T.untyped` if the GraphQL input object's name is nil. Not sure if we can do better than this.

### Tests

I tried writing tests by adding a content file with the schema and then loading it, but marshal loading for our isolation test runner was failing when loading the anonymous class. If you have a better idea on how to test this, please let me know.

### DIFF
--- a/lib/tapioca/dsl/helpers/graphql_type_helper.rb
+++ b/lib/tapioca/dsl/helpers/graphql_type_helper.rb
@@ -39,7 +39,7 @@ module Tapioca
               "T.any(#{value_types.join(", ")})"
             end
           when GraphQL::Schema::InputObject.singleton_class
-            qualified_name_of(unwrapped_type)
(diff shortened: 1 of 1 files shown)


paracycle (commented):
Isn't this a duplicate of https://github.com/Shopify/tapioca/pull/1146?

vinistock:
@paracycle it is indeed! I totally missed that PR. Closing in favour of #1146.

PR #1162, opened 2022-09-13
Make sure we can type GraphQL object with anonymous input fields

rafaelfranca:
### Motivation
When the class is anonymous, `qualified_name_of` returns `nil`. This
means that `type_for` would raise an exception, causing the rbi generator
to fail.

### Implementation
Return `T.untyped` for the anonymous case

### Other info
Those commits are already applied to master (I used VSCode to sync the code and it end up pushing the changes), so this is only for the 0-10-stable branch.

### DIFF
--- a/lib/tapioca/dsl/compilers/graphql_input_object.rb
+++ b/lib/tapioca/dsl/compilers/graphql_input_object.rb
@@ -47,12 +47,10 @@ module Tapioca
           arguments = constant.all_argument_definitions
           return if arguments.empty?
 
-          graphql_type_helper = Helpers::GraphqlTypeHelper.new
-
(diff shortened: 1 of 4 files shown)


KaanOzkan (approved):
First commit is functionally equivalent to this PR https://github.com/Shopify/tapioca/pull/1146/. I don't have strong opinions between the `type_for_constant` helper method and checking if `parsed_type` is nil at the end of the switch.

sambostock (approved):

rafaelfranca:
That PR wan't handling the case when the one of values of the enum was anonymous. Not that I think it should happen, but better to be safe.

PR #1164, opened 2022-09-14
Handle `all_argument_definitions` not being defined in `graphql` `< 1.13`

sambostock:
### Motivation

We use [`GraphQL::Schema::Member::HasArguments.all_argument_definitions`](https://github.com/rmosolgo/graphql-ruby/blob/28bb4362c19c6ca6b2fab35edb48bf2627cd9b3f/lib/graphql/schema/member/has_arguments.rb#L141) in the DSL compiler for GraphQL mutations. This method was [introduced in `graphql@1.13`](https://github.com/rmosolgo/graphql-ruby/commit/6e5fe3731d81b5d22d869e3de51ded9a7d831b40), which means our implementation breaks in versions prior to that.

Therefore, we either need a workaround, or to restrict the versions we're compatible with.

### Implementation

This first pass tries to switch to the `arguments` method, which returns a `Hash` whose `values` we can extract, which hopefully has the same semantics.

### Tests

Currently seeing if existing tests pass.

### After Merging
- [ ] Backport to `0-10-stable`, which introduced GraphQL DSL compilation

### DIFF
--- a/lib/tapioca/dsl/compilers/graphql_input_object.rb
+++ b/lib/tapioca/dsl/compilers/graphql_input_object.rb
@@ -44,7 +44,7 @@ module Tapioca
 
         sig { override.void }
         def decorate
-          arguments = constant.all_argument_definitions
+          arguments = constant.arguments.values
           return if arguments.empty?
 
           root.create_path(constant) do |input_object|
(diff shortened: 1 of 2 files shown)


sambostock:
That makes sense, @rafaelfranca, I can add that. Do you think we should
- still support that older version (since this PR provides a fix that allows us to support it), or
- drop support for anything older than `graphql@1.13` (merging `gem "graphql", ">= 1.13"` in a separate PR, and closing this one)?

rafaelfranca:
I think we should do the latter.

Issue #1149, opened 2022-09-08
tapioca 0.10: tapioca dsl hangs when Rails app uses EventedFileUpdateChecker


bmulholland:
tapioca 0.9.4 would finish and exit no problem. When I upgrade to 0.10 `tapioca dsl` completes, but hangs until killed. This happens only when the number of processes for `Parallel` is >1, line ~31 of `Tapioca::Executor`, and it only happens when the rails app is initialized with `config.file_watcher = ActiveSupport::EventedFileUpdateChecker`

One workaround is to use the default Rails file watcher, removing that config. Another is to set workers to 1 with `--workers 1`, though this is incompatible with `--verify` in e.g. git hooks.

I tried tracing it and the exe/tapioca file gets to the end, but the process never closes. A simple repro is to change tapioca's DSL `execute` to:

```ruby
        Loaders::Dsl.load_application(
          tapioca_path: @tapioca_path,
          eager_load: @requested_constants.empty?
        )

        Parallel.map([1,2,3], { in_processes: 2 }){}
        say("done")
        return
```

MBP M1 Pro

Here's the last of the output:

```
Done

Checking generated RBI files...  Done
  No errors found

All operations performed in working directory.
Please review changes and commit them.
^C
```

dirceu:
@bmulholland thank you very much for the investigation! Can you please check if this works for you? https://github.com/Shopify/tapioca/pull/1179

PR #1179, opened 2022-09-20
Fix issue with Tapioca DSL hanging

dirceu:
I was able to reproduce this on an existing application and this seems to reliably fix the problem by preventing ruby/debug from attaching to child processes.

Fixes https://github.com/Shopify/tapioca/issues/1149.

### DIFF
--- a/lib/tapioca/helpers/env_helper.rb
+++ b/lib/tapioca/helpers/env_helper.rb
@@ -11,6 +11,7 @@ module Tapioca
     sig { params(options: T::Hash[Symbol, T.untyped]).void }
     def set_environment(options) # rubocop:disable Naming/AccessorMethodName
       ENV["RAILS_ENV"] = ENV["RACK_ENV"] = options[:environment]
+      ENV["RUBY_DEBUG_FORK_MODE"] = "parent"
     end
   end
 end


vinistock (approved):

dirceu:
I can reproduce manually but I'm having a hard time finding the exact combination of factors in a fake Rails app :/

```diff
diff --git a/spec/tapioca/cli/dsl_spec.rb b/spec/tapioca/cli/dsl_spec.rb
index 649ecb71..412373d1 100644
--- a/spec/tapioca/cli/dsl_spec.rb
+++ b/spec/tapioca/cli/dsl_spec.rb
@@ -1321,6 +1321,40 @@ module Tapioca
           assert_success_status(result)
         end
 
+        # prevents regressions on https://github.com/Shopify/tapioca/issues/1149
+        it "does nothing and returns exit status 0 when using ruby/debug" do
+          @project.write("config/application.rb", <<~RB)
+            require "rails"
+
+            class Application < Rails::Application
+              attr_reader :config
+
+              def initialize
+                super
+                root = Pathname.new("#{@project.path}")
+                @config = Rails::Application::Configuration.new(root)
+                @config.file_watcher = ActiveSupport::EventedFileUpdateChecker
+              end
+            end
+          RB
+          @project.write("config/environment.rb", <<~RB)
+            require_relative "application.rb"
+          RB
+          @project.require_real_gem("listen")
+          @project.require_real_gem("rails")
+          @project.require_real_gem("debug", "1.6.2")
+          @project.bundle_install
+
+          @project.tapioca("dsl")
+          result = @project.tapioca("dsl --verify --workers 2")
+
+          assert_stdout_includes(result, <<~OUT)
+            Nothing to do, all RBIs are up-to-date.
+          OUT
+          assert_empty_stderr(result)
+          assert_success_status(result)
+        end
+
         it "advises of removed file(s) and returns exit status 1 when files are excluded" do
           @project.tapioca("dsl")
           result = @project.tapioca("dsl --verify --exclude SmartProperties")
```

st0012:
But #1149's reproduction script doesn't seem to include `debug`?

bmulholland:
FWIW my repro (in #1149) _did_ seem to require `debug` to be present -- though in weird ways I couldn't quite understand. Like @dirceu, I could only repro this in a rails environment (i.e. with `rails runner`, not with `ruby xx.rb`)

I don't understand the parts involved, so I only have a light understanding of what was said over at https://github.com/grosser/parallel/issues/329, but I suspect the issue may actually be in `EventedFileUpdateChecker`.

Anyway, I confirm that this PR resolves the hang issue for me (as does dropping `EventedFileUpdateChecker`, which we've also done).

PR #1188, opened 2022-09-28
Various performance optimizations for trackers

vinistock:
### Motivation

With the new foreign constant mixin attributions, generating DSLs has been slowed quite significantly. Part of the reason is because we eagerly track those, even when not using them.

### Implementation

There are four changes in this PR
- We were invoking `caller_locations` twice in `ConstantDefinition`, despite passing the value through and argument. I removed the second invocation and used the `locations` argument, since invoking `caller_locations` is expensive
- Started lazily resolving attached_class extends on singleton classes only when requesting a specific mixin. This makes us avoid invoking `ObjectSpace.each_object`, which is super expensive
- Reuse the `return_value` result by moving a variable up
- Started using `File.identical?(a, b)` for file comparisons. This is much faster than computing `File.realpath`

### Tests


dirceu (approved):

st0012 (approved):
🚀

PR #1195, opened 2022-09-29
Add release workflow

vinistock:
### Motivation

Add release workflow that automatically creates the GitHub release when a tag starting with `v` is pushed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,27 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - v*
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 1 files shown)


dirceu (approved):

KaanOzkan (approved):

Morriar:
> I'm not sure that's possible

From the link I posted in the comment:

> You can use special characters in path, branch, and tag filters.
> 
> `*`: Matches zero or more characters, but does not match the `/` character.
> `**`: Matches zero or more of any character.
> `?`: Matches zero or one of the preceding character.
> `+`: Matches one or more of the preceding character.
> `[]` Matches one character listed in the brackets or included in ranges.

PR #1197, opened 2022-09-30
Backport "Use correct type for `GraphQL::Types::ISO8601DateTime`"

sambostock:
This backports #1196 to `0-10-stable`, as `0.10` introduced the bug.


sambostock:
Oh, are we not releasing 0.10.x anymore?

PR #1201, opened 2022-10-03
Add explanation in readme what ConstantType is for

bdewater:
It's listed in the example and not too hard to deduce by copy-pasting the example and tweaking until it works (or have `srb tc` complain), but it is never actually explained. 

cc @paracycle from our Slack convo


bdewater:
I signed the CLA, that step should be green once re-run.

egiurleo (approved):

egiurleo:
@bdewater Would you mind rebasing? I just merged a fix for the failing CI tests and I think CI got stuck last night.

PR #1211, opened 2022-10-05
Update RBI gem and fix tests

egiurleo:
### Motivation
There was a change in the RBI gem (https://github.com/Shopify/rbi/pull/140) to stop sorting `prop` and `const` attributes on T::Struct objects. This change has broken a test in the tapioca test suite that assumed such fields would be sorted alphabetically by name. See #1201 for the test failing in CI.

### Implementation
I updated the RBI gem to the latest version and fixed the tests in question.

### Tests
No new tests, existing tests pass CI.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -22,7 +22,7 @@ PATH
       netrc (>= 0.11.0)
       parallel (>= 1.21.0)
       pry (>= 0.12.2)
-      rbi (~> 0.0.0, >= 0.0.14)
+      rbi (~> 0.0.0, >= 0.0.16)
       sorbet-static-and-runtime (>= 0.5.9204)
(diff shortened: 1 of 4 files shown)


Gemfile.lock
```diff
@@ -251,7 +251,7 @@ GEM
       zeitwerk (~> 2.5)
     rainbow (3.1.1)
     rake (13.0.6)
-    rbi (0.0.15)
+    rbi (0.0.16)
```
paracycle, line 254: Let's update the required version in the gemspec file as well: https://github.com/Shopify/tapioca/blob/emily/tstruct-field-tests/tapioca.gemspec#L30

paracycle (approved):

PR #1213, opened 2022-10-06
Fix code making additions to only the singleton class of a constant in the payload

### LINKED ISSUE #1212: Tapioca generating empty RBI file for some gems
We're observing some strange behaviour from `tapioca` where it generates empty RBIs for some of our gems.
The surprising part is that if I run `srb rbi gems` instead (which I understand is deprecated in favour of `tapioca`), these RBI files are correctly generated 🤔.

As an example, we have `neatjson` in our Gemfile, which adds a `neat_generate` method to `JSON` (see [code](https://github.com/Phrogz/NeatJSON/blob/master/lib/neatjson.rb#L38)), but the following file gets generated by `tapioca`:

```ruby
# typed: true

# DO NOT EDIT MANUALLY
# This is an autogenerated file for types exported from the `neatjson` gem.
# Please instead update this file by running `bin/tapioca gem neatjson`.

# THIS IS AN EMPTY RBI FILE.
# see https://github.com/Shopify/tapioca#manually-requiring-parts-of-a-gem
```

I've already tried explicitly requiring `neatjson` in `require.rb`, but that doesn't make any difference.
Again, `sorbet` somehow is able to generate this gem's RBI file correctly, so I'm not really sure what's the difference or what we could be doing wrong.

paracycle:
### Motivation

Fixes #1212 

It turns out that the reason for Tapioca not processing `neatjson` was that the gem was making a single extension to the `JSON` module which was to add a `self.neatjson` method. When Sorbet parses that file, it only gives us `<Class:JSON>` as a symbol and not `JSON`. Thus, we would end up not generating anything for the file at all.

### Implementation

Start processing `<Class:Foo>` symbols as if they were `Foo` symbols instead.

### Tests

Added a test that recreates the situation in the `neatjson` gem, which fails before the change and passes with after this PR.

### DIFF
--- a/lib/tapioca/static/symbol_table_parser.rb
+++ b/lib/tapioca/static/symbol_table_parser.rb
@@ -43,6 +43,12 @@ module Tapioca
 
           next if name.nil?
           next unless SKIP_PARSE_KINDS.include?(kind)
+
+          # Turn singleton class names to attached class names
+          if (match_data = name.match(/<Class:(.*)>/))
+            name = match_data[1]
+          end
+
           next if name.match?(/[<>()$]/)
           next if name.match?(/^[0-9]+$/)
           next if name == "T::Helpers"


spec/tapioca/gem/pipeline_spec.rb
```diff
@@ -281,6 +281,25 @@ class Bar::A::B; end
       assert_equal(output, compile)
     end
 
+    it "compiles extensions to existing classes even it only add to the singleton class" do
```
KaanOzkan, line 284:
```suggestion
    it "compiles extensions to existing classes even if it only adds to the singleton class" do
```

PR #1218, opened 2022-10-06
Silence warnings in all CLI output

paracycle:
### Motivation

We should wrap all command executions inside `Tapioca.silence_warnings` so that we can suppress warnings in the CLI output, like the ones from the parser gem that fails out 3.2-head tests.

### Tests

Expect existing tests to pass on 3.2 builds

### DIFF
--- a/lib/tapioca/cli.rb
+++ b/lib/tapioca/cli.rb
@@ -281,7 +281,10 @@ module Tapioca
         payload: options[:payload],
         number_of_workers: options[:workers],
       )
-      command.execute
+
+      Tapioca.silence_warnings do
+        command.execute
+      end
     end
 
     desc "annotations", "Pull gem RBI annotations from remote sources"
@@ -307,7 +310,10 @@ module Tapioca
         netrc_file: netrc_file(options),
         typed_overrides: options[:typed_overrides],
       )
-      command.execute
+
+      Tapioca.silence_warnings do
+        command.execute
+      end
     end
 
     map ["--version", "-v"] => :__print_version


sambostock (approved):

vinistock (approved):
Would adding this to the gem command get rid of those `You need to install dalli` warnings? And if so, could we add it as well?

Morriar (commented):
> We should wrap all command executions inside `Tapioca.silence_warnings`

Agreed. Would it make sense to move it inside `Commnand:execute` then?

I think we can do something like this:

```rb
class Command
  sig { void }
  def execute
    Tapioca.silence_warnings do
      internal_execute
    end
  end

  private

  sig { abstract.void }
  def internal_execute; end
end

class SubCommand < Command
  private

  def internal_execute
    # real stuff
  end
end
```

paracycle:
> Why were the warnings causing CI to fail?

Because they warnings get printed into stderr and leak into the output that we verify in CLI tests.

paracycle:
> Agreed. Would it make sense to move it inside `Commnand:execute` then?

That's what I wanted to do first, but that's a little tricky because for some commands we use more than `execute` to run things, so I need to make a refactor first.

dogweather:
> > Would adding this to the gem command get rid of those `You need to install dalli` warnings? And if so, could we add it as well?
> 
> No. The `gem` command already has this in place (actually, that's why the `silence_warnings` exist), but since Rails/ActiveSupport (or whatever) does not print those using `warn`, it doesn't help.

What's the source of the `You don't have dalli installed in your application. Please add it to your Gemfile and run bundle install` warning? I don't use memcached in my Rails app yet I get these messages.

dogweather:
> > What's the source of the `You don't have dalli installed...
> 
> ...printed _because_ you are not using memcached in your Rails app. Tapioca loads this file, it hits the `require "dalli"` line and raises the `LoadError` exception which then prints that to the `$stderr`. Since Active Support is not using `warn` for printing those errors, the current silencer is not able to do much.

Ah - thanks! I'll look into writing a PR for Rails.

----

This is interesting to me, philosophically. So `mem_cache_store.rb` works by assuming, if it's loaded, then it's intended to be used — configuration by convention, not by config files. And so — according to this design — when it's loaded, it should try to load its requirements, like `dalli`.

The problem is, introspective tools might load files like this as part of their work.

paracycle:
> The problem is, introspective tools might load files like this as part of their work.

Yes, exactly. IMO, it would be so much better if these warnings were being printed via `Kernel#warn`, which introspective tools could suppress if they wanted to, instead of printed to stderr directly.

I am sure there is a good historical reason for why Rails chose to systematically use `$stderr.puts` for these cases, but I feel like it would be better to switch to `Kernel#warn` at this point.

paracycle:
> > > What's the source of the `You don't have dalli installed...
> > 
> > 
> > ...printed _because_ you are not using memcached in your Rails app. Tapioca loads this file, it hits the `require "dalli"` line and raises the `LoadError` exception which then prints that to the `$stderr`. Since Active Support is not using `warn` for printing those errors, the current silencer is not able to do much.
> 
> Ah - thanks! I'll look into writing a PR for Rails.
> 
> This is interesting to me, philosophically. So `mem_cache_store.rb` works by assuming, if it's loaded, then it's intended to be used — configuration by convention, not by config files. And so — according to this design — when it's loaded, it should try to load its requirements, like `dalli`.
> 
> The problem is, introspective tools might load files like this as part of their work.

@dogweather I finally opened a PR for this on Rails https://github.com/rails/rails/pull/48777 🤞

Issue #1182, opened 2022-09-22
Allow generating RBIs in Rails engines


vinistock:
In a Rails engine, the Rails application is typically defined inside `test/dummy`. However, it's a real application and we could load it to inspect the defined models and controllers.

Currently, it does not work because we [return early](https://github.com/Shopify/tapioca/blob/4829c608cb8ea395b5e17b789cd2b4b829def0f9/lib/tapioca/loaders/loader.rb#L38) if `config/application.rb` is not defined in the current directory.

If we could somehow allow this to also point to different paths, like `test/dummy/config/application.rb`, DSL generation for engines would just work.

PR #1220, opened 2022-10-07
Add option for generating RBIs for Rails engines

andyw8:
(Paired with @vinistock) 

### Motivation

Closes https://github.com/Shopify/tapioca/issues/1182

### Implementation

An `--app_root` option was added to the `tapioca dsl` command. This can be used to specify the path to the dummy app for an engine.

There were several places in the code which assumed the app is always at the root of the project. These were updated to take the `app_root` option into consideration.

### Tests

The test approach is styled after the existing RBI generation tests. We create files for the app as normal, but then move the necessary files into the `test/dummy` path before running the `dsl` commmand.


spec/tapioca/cli/dsl_spec.rb
```diff
@@ -343,6 +343,62 @@ def title=(title); end
...
+                def title; end
+
+                sig { params(title: T.nilable(::String)).returns(T.nilable(::String)) }
+                def title=(title); end
+              end
+            end
+          RBI
+
+          # Restore directory structure so to not impact other tests
+          FileUtils.mv(engine_path + "/config", @project.path)
```
vinistock, line 397: Can we put this inside an `ensure` block? It makes sure that we restore back even if something fails.

andyw8: Good call, I've pushed an update.

spec/tapioca/cli/dsl_spec.rb
```diff
@@ -343,6 +343,66 @@ def title=(title); end
...
+              end
+            end
+          RB
+
+          engine_path = @project.path + "/test/dummy"
+
+          begin
+            FileUtils.mkdir_p(engine_path)
+            FileUtils.mv(@project.path + "/config", engine_path)
+            FileUtils.mv(@project.path + "/lib", engine_path)
```
KaanOzkan, line 372: I'd prefer writing the files to `/test/dummy` directly on line 347 but not a blocker.

KaanOzkan (approved):

spec/tapioca/cli/dsl_spec.rb
andyw8: Agreed, unfortunately the top-level `before(:all)` sets the project path so we would need to restructure the tests first.

KaanOzkan: I'm a bit confused. I was only referring to the `lib` files. More specifically I was thinking of changing the 2 writes to:
`@project.write("test/dummy/lib/post.rb", <<~RB)` & `@project.write("test/dummy/lib/comment.rb", <<~RB)`

and removing the line `FileUtils.mv(@project.path + "/lib", engine_path)`.

We don't have to do it, I'm just trying to understand the project path blocker 🙂

PR #1225, opened 2022-10-12
Improve DSL test for custom `app_root`

andyw8:
### Motivation

Address feedback in https://github.com/Shopify/tapioca/pull/1220/files#r993606253

### Implementation

Write the model files directly rather than having to move them.

### Tests

n/a


KaanOzkan (commented):
I don't know why the error is happening despite the directory being purged same as before. We can keep the existing solution if it requires a long investigation.

andyw8:
Fix in progress for CI: https://github.com/Shopify/tapioca/pull/1226

vinistock (approved):

PR #1226, opened 2022-10-12
Update Sorbet and fix failing test case on `main`

andyw8:
### Motivation

Due to https://github.com/sorbet/sorbet/pull/6426, the Sorbet output changed, causing a test failure in `main`.

### Implementation

* Ran `bundle update sorbet`
* Adjusted the test based on the output from Sorbet, with the assumption that the output is correct.
* Ran `bundle exec exe/tapioca gem` and committed the RBI changes.

### Tests

n/a


sorbet/rbi/gems/mini_portile2@2.8.0.rbi
```diff
@@ -1,8 +0,0 @@
-# typed: true
```
andyw8, line 1: Removed by `bundle exec exe/tapioca gem`.

andyw8:
@paracycle I saw that you added Ruby 3.2 to the build in https://github.com/Shopify/tapioca/pull/1214 as 'experimental'. If a Ruby 3.2 job fails, is the intention that the 'overall' build should still pass?

andyw8:
Resolved by #1227.

paracycle:
> @paracycle I saw that you added Ruby 3.2 to the build in #1214 as 'experimental'. If a Ruby 3.2 job fails, is the intention that the 'overall' build should still pass?

I see this is closed, but yes, that's the general idea. The Ruby 3.2 job is to test Ruby 3.2 more than Tapioca itself. For example, last year it had caught this: https://bugs.ruby-lang.org/issues/18405

PR #1238, opened 2022-10-20
ci: regenerate Gemfile.lock when running on ruby head

flavorjones:
### Motivation

Experimental Ruby 3.2 builds are failing because Nokogiri's precompiled native platform gem only supports up to Ruby 3.1.

Example failure: https://github.com/Shopify/tapioca/actions/runs/3283616191/jobs/5408560879#step:3:41

### Implementation

CI should regenerate the lockfile when running on an unreleased version of Ruby.

### Tests

Ruby 3.2 should run as expected.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,8 @@ jobs:
       BUNDLE_GEMFILE: ${{ matrix.gemfile }}
     steps:
       - uses: actions/checkout@v3.1.0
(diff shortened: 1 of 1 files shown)


st0012:
I'm merging this so at least the CI fails at the `Run tests` step and doesn't mark the build as failures.
And then I'll take a look in fixing the error @KaanOzkan mentioned.

PR #1246, opened 2022-10-25
Update gems and their RBI files

dirceu:
Closes #1242
Closes #1243
Closes #1244
Closes #1245


dirceu:
@vinistock I reverted that and it caused failures in some of the matrix's pipelines; I removed the revert commit and other ones are failing. I'm a bit lost here as to why the `it compiles signatures and structs in source files` passes on Gemfile-rails-main but not on Gemfile-rails-6-1, for example. I wonder if there's some caching issue at play here. Any tips?

vinistock:
The reason is probably because we don't have a `Gemfile.lock` for the extra gemfiles, so you never know for sure which version is going to be installed.

One option we can take is check the Sorbet version in the test and then change the generated RBI based on it. Something like
```ruby
expected_foo_signature = if sorbet_version < 123123 && sorbet_version > 123123
  "def foo=(arg0); end"
else
  <<~RUBY
    sig { ... }
    def foo=(foo); end
  RUBY
end
```

st0012 (approved):

vinistock (approved):

PR #1250, opened 2022-10-28
Make scopes available for abstract classes

Teots:
### Motivation
At the moment sorbet can't figure out that classes, that are marked as abstract, have access to scopes. The following example illustrates this:

```ruby
class A < ApplicationRecord
  extend T::Sig
  abstract_class = true

  scope :deleted, -> { where.not(deleted_at: nil) }

  class << self
    extend T::Sig

    sig { returns(Integer) }
    def count_deleted
        # Method `deleted` does not exist on `T.class_of(A)`
        deleted.count
    end
  end
end
```

Since those scopes can be used, I think it makes sense to generate the rbi files accordingly.

### Implementation
I removed the filter for abstract classes in `ActiveRecordScope` and did the same thing in `ActiveRecordRelations` to make the generated relation methods work.

### Tests
Tests for `ActiveRecordScope` have been extended.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_relations.rb
+++ b/lib/tapioca/dsl/compilers/active_record_relations.rb
@@ -175,7 +175,7 @@ module Tapioca
 
           sig { override.returns(T::Enumerable[Module]) }
           def gather_constants
-            ActiveRecord::Base.descendants.reject(&:abstract_class?)
+            descendants_of(ActiveRecord::Base).reject(&:abstract_class?)
           end
         end
 
--- a/lib/tapioca/dsl/compilers/active_record_scope.rb
+++ b/lib/tapioca/dsl/compilers/active_record_scope.rb
@@ -83,7 +83,7 @@ module Tapioca
         class << self
           sig { override.returns(T::Enumerable[Module]) }
           def gather_constants
-            descendants_of(::ActiveRecord::Base).reject(&:abstract_class?)
+            descendants_of(::ActiveRecord::Base)
           end
         end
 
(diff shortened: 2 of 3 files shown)


paracycle:
> did the same thing in `ActiveRecordRelations` to make the generated relation methods work.

Can't see this part in the PR. Is that missing or did you decide against this?

Teots:
@paracycle Sorry, looks like I forgot to push this commit 🙈 It's there now.

lib/tapioca/dsl/compilers/active_record_relations.rb
```diff
@@ -181,7 +181,7 @@ class << self
 
           sig { override.returns(T::Enumerable[Module]) }
           def gather_constants
-            ActiveRecord::Base.descendants.reject(&:abstract_class?)
+            ActiveRecord::Base.descendants
```
paracycle, line 184: While we are at it, can we please change this to be:
```suggestion
            descendants_of(ActiveRecord::Base)
```
as well? Thanks!

paracycle (approved):
Looks good to me! We will have to check this against Core to make sure that we don't get any unwanted behaviour before merging though.

antn:
Ah, I just ran into this too. @paracycle, did this end up having unexpected behavior, or is it maybe good to merge? 😄

paracycle:
Ok, I just had a chance to test this against Shopify Core and it seems to work. However, I removed the commit that was making AR relations to consider abstract classes as well, since that's not related to scopes and I don't think we would want folks to write relations off of abstract classes.

PR #1251, opened 2022-10-31
Add `debug` to Gemfile

st0012:
This PR also bumps Sorbet for the updated `irb` rbi (PR: https://github.com/sorbet/sorbet/pull/6509)


KaanOzkan (approved):

Gemfile
```diff
@@ -7,6 +7,7 @@ gemspec
 gem("minitest")
 gem("minitest-hooks")
 gem("minitest-reporters")
+gem("debug", require: false)
```
Morriar, line 10: Shouldn't it be in the `development` group?

st0012: It should at least be included in `test` as well?
I think it belongs to where `pry-byebug` sits, which is at the top-level atm.

vinistock (commented):
Should we just remove `pry-byebug` and use only debug? Is there a reason we need to keep it?

vinistock (approved):
I vote for removing `pry-byebug`, but don't object keeping it if other people are using it

Morriar (approved):

PR #1259, opened 2022-11-02
Drop `pry` as a runtime dependency

paracycle:
### Motivation

We've been using `pry` as a runtime dependency only because it was convenient to use its `Pry::WrappedModule` functionality to figure out which files a constant might be defined in.

However, that's actually a very minimal amount of logic that we've been relying on and the runtime dependency on a development tool has always been a bit weird. It is about time we stop doing that.

### Implementation

I mostly adapted the code from https://github.com/pry/pry/blob/master/lib/pry/wrapped_module.rb#L296-L368 into our `Reflection` class.

### Tests

No added tests, existing tests should continue to run properly.


lib/tapioca/runtime/reflection.rb
```diff
@@ -183,6 +183,61 @@ def attached_class_of(singleton_class)
...
+          PRIVATE_INSTANCE_METHODS_METHOD,
+        ]
+
+        modules.product(method_list_methods).flat_map do |mod, method_list_method|
+          method_list_method.bind_call(mod, false).map { |name| mod.instance_method(name) }
+        end
+      end
+
+      sig { params(parent: Module, name: String).returns(T.nilable(Module)) }
+      def nested_module?(parent, name)
```
st0012, line 227: nitpick: having `?` postfix but not returning a boolean feels kinda surprising?

st0012 (approved):
Nice 👍

lib/tapioca/runtime/reflection.rb
paracycle: Done!

lib/tapioca/runtime/reflection.rb
```diff
@@ -183,6 +183,61 @@ def attached_class_of(singleton_class)
...
+      private
+
+      sig { params(constant: Module).returns(T::Array[UnboundMethod]) }
+      def relevant_methods_for(constant)
+        methods = methods_for(constant).select(&:source_location)
+          .reject { |x| method_defined_by_forwardable_module?(x) }
+
+        return methods unless methods.empty?
+
+        constants_of(constant).flat_map do |const_name|
```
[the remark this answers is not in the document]
paracycle: For now, I've mostly adapted the code from upstream and have not made logic changes. I think the intention was to discover files that a constant was reopened in through looking at all the file locations of the methods defined on the constant (and its singleton class) first.

However, sometimes a constant is just a namespace and has no methods defined on it, so this fallback will try to match reopenings of the constant by looking at the file locations where subconstants are defined inside the namespace.

We probably don't want that fallback for Tapioca, but in this PR I don't want to change behaviour drastically, so I'll keep this as is for now.

PR #1260, opened 2022-11-04
Require actual version of `sorbet-static-and-runtime` as minimum

sambostock:
### Motivation

While `sorbet-static` and `sorbet-runtime` versions `0.5.9204` do exist, `0.5.9519` is the first version of `sorbet-static-and-runtime` published on RubyGems, so we cannot depend on anything lesser than that.

~We can potentially require an even newer version, as discussed in https://github.com/Shopify/tapioca/pull/1252#discussion_r1010483491, but I'm not sure which, so I've not made that change yet.~
_As per [@paracycle's comment](https://github.com/Shopify/tapioca/pull/1260#issuecomment-1304274821), I've gone ahead and further bumped the minimum version requirement to a version that allows us to get rid of our feature checks._

### Implementation

This updates the version in `tapioca.gemspec` and refreshes `Gemfile.lock` accordingly.

### Tests

All existing tests apply.


paracycle:
If we can bump the minimum version to `>= 0.5.9892` (which seems to date back to April 2022) then we can get rid of all of these feature detection workarounds as well: https://github.com/Shopify/tapioca/blob/main/lib/tapioca/helpers/sorbet_helper.rb#L23-L25

paracycle (approved):

sambostock:
@paracycle, I've added commits getting rid of a feature flag by successively bumping the version, and added a documentation comment to the now empty hash (as I imagine we want to keep it around for future use).

Alternatively, I could delete all the feature checking code, if that's not something we want to keep around.

sambostock:
I don't know about the generics syntax specifically, but trying out this branch on Core passes RBI related CI checks. It's already on Sorbet `0.5.10482`.

Issue #1140, opened 2022-08-29
Use Rails relations in typing


Overload119:
My ActiveRecord model is Person.

```rb
sig { params(blck: T.proc.params(people: Person::PrivateRelation).void).void }
def foreach_matching_person_query(&blck)
  # ...
  yield(Person.joins(:companies))
end
```

I have something like this - it throws no errors in Sorbet. 
In runtime, it throws:

```rb
NameError: uninitialized constant Person (call 'Person.connection' to establish a connection)::PrivateRelation
```

How can I properly type this?

The query itself is `Person::ActiveRecord_Relation` but this is a private class I believe. Sorbet can't read it.

G2Jose:
I ended up creating an initializer that at runtime creates these classes that tapioca RBI references. This is what it looks like: 

```ruby
# config/initializers/sorbet.rb
ActiveRecord::Base.descendants.each do |klass|
  klass.const_set('PrivateAssociationRelation', Object)
  klass.const_set('PrivateAssociationRelationWhereChain', Object)
  klass.const_set('PrivateCollectionProxy', Object)
  klass.const_set('PrivateRelation', Object)
  klass.const_set('PrivateRelationWhereChain', Object)
end
```

Would love to see if there's a better way.

rafaelfranca:
See the `CAUTION` on this document https://github.com/Shopify/tapioca/blob/main/manual/compiler_activerecordrelations.md

timfjord:
The suggestion above does fix the issue, but only partly.  
Because, it is still not clear how to properly work with class methods/scopes defined in the model itself.

Even in the example, what if we need to call a specific scope from the `Person` model
```ruby
class Person < ApplicationRecord
  # ...

  scope :active, -> { where(active: true) }
end

# ...

sig { params(blck: T.proc.params(people: ActiveRecord::Relation).void).void }
def foreach_matching_person_query(&blck)
  # ...
  yield(Person.joins(:companies))
end

foreach_matching_person_query do |people|
  puts people.active.count
end
```

Sorbet won't like this because we are trying to call it `active`, but it thinks `people` is an instance of `ActiveRecord::Relation`, so it will complain that the method doesn't exist

G2Jose:
> Sorbet won't like this because we are trying to call `active` on the `people`, but it thinks `people` is an instance of `ActiveRecord::Relation`, so it will complain that the method doesn't exist

Exactly - this is why I resorted to making these private constants public using `const_set`. 

[...]

PR #1266, opened 2022-11-08
Add documentation about using PrivateRelation

alexdunae:
### Motivation
I was migrating from sorbet-rails and couldn't find a migration path for `RelationType`.  I searched the Slack channel and came up with this suggestion.  I figured it would be helpful to others.

### Implementation
n/a

### Tests
just docs

### DIFF
--- a/manual/compiler_activerecordrelations.md
+++ b/manual/compiler_activerecordrelations.md
@@ -143,3 +143,16 @@ class Post
   end
 end
 ~~~
+
+**A note about migrating from `sorbet-rails`**: there is no direct analogue to `Post::RelationType` in Tapioca.  
+
+`sorbet-rails` actually [changes the way a Rails application loads](https://github.com/chanzuckerberg/sorbet-rails/blob/447ecbe4a2586ce8aa9055290d571fe34a55e27f/lib/sorbet-rails/railtie.rb#L26-L50) in order to make `RelationType` work at runtime.  One strategy to add types for relations is to use `WithoutRuntime`.  For example:
+
+~~~rbi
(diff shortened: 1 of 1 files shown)


manual/compiler_activerecordrelations.md
```diff
@@ -143,3 +143,16 @@ class Post
...
+`sorbet-rails` actually [changes the way a Rails application loads](https://github.com/chanzuckerberg/sorbet-rails/blob/447ecbe4a2586ce8aa9055290d571fe34a55e27f/lib/sorbet-rails/railtie.rb#L26-L50) in order to make `RelationType` work at runtime.  One strategy to add types for relations is to use `WithoutRuntime`.  For example:
+
+~~~rbi
+T::Sig::WithoutRuntime.sig { returns(Post::PrivateRelation) }
+def posts
+  Post.where(id: 123)
+end
+~~~
+
+As the name suggests, this will give you type checking during development but disable it at runtime.
```
KaanOzkan, line 158: It might be beneficial to give a bit more information here about how Tapioca defines methods on `PrivateRelation` that returns a `Post` type. For example, the `first!` method. That's why this approach gives more static typing benefits.

But I didn't use sorbet-rails and if you think this information is already assumed by an existing sorbet-rails user than it's okay to omit it.

KaanOzkan (approved):
This looks good to me but I'm not 100% sure on its placement. We can add it to https://github.com/Shopify/tapioca/wiki/Migrating-to-Tapioca as well.

fazo96:
This looks like it could be a solution for #1140

PR #1268, opened 2022-11-11
Add types for postgres hstore

DanielHeath:
### Motivation
When using an hstore column in postgres, tapioca generates "T.untyped". Both keys and values in `hstore` are strings.

### Implementation
Added a case for hstore to the existing case statement.

### Tests
No tests, but hopefully it's pretty clear what it does. Checked it on a codebase that uses hstore.

### DIFF
--- a/lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
+++ b/lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
@@ -19,33 +19,7 @@ module Tapioca
 
           column_type = @constant.attribute_types[column_name]
 
-          getter_type =
-            case column_type
-            when defined?(MoneyColumn) && MoneyColumn::ActiveRecordType
-              "::Money"
-            when ActiveRecord::Type::Integer
-              "::Integer"
-            when ActiveRecord::Type::String
-              "::String"
-            when ActiveRecord::Type::Date
-              "::Date"
-            when ActiveRecord::Type::Decimal
-              "::BigDecimal"
-            when ActiveRecord::Type::Float
-              "::Float"
-            when ActiveRecord::Type::Boolean
-              "T::Boolean"
-            when ActiveRecord::Type::DateTime, ActiveRecord::Type::Time
-              "::Time"
-            when ActiveRecord::AttributeMethods::TimeZoneConversion::TimeZoneConverter
-              "::ActiveSupport::TimeWithZone"
-            when ActiveRecord::Enum::EnumType
-              "::String"
-            when ActiveRecord::Type::Serialized
-              serialized_column_type(column_type)
-            else
-              handle_unknown_type(column_type)
-            end
+          getter_type = type_for_activerecord_value column_type
 
           column = @constant.columns_hash[column_name]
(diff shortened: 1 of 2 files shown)


lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
```diff
@@ -35,6 +35,8 @@ def type_for(column_name)
               "::Float"
             when ActiveRecord::Type::Boolean
               "T::Boolean"
+            when ActiveRecord::ConnectionAdapters::PostgreSQL::OID::Hstore
```
KaanOzkan, line 38: We need to be smarter and reference this constant only when the host application is using Postgres. Other constants on this file are loaded by default when we load ActiveRecord and don't need this treatment https://github.com/Shopify/tapioca/blob/3a2a8d150cd6501e78251c25098b530df119ac84/lib/tapioca/dsl/compilers/active_record_columns.rb#L5

Other than that I'm okay with supporting database specific columns but if there are more of them we can look into a more extensible approach. Such as having a different helper file for Postgres.

DanielHeath: I've updated this to check that `ActiveRecord::ConnectionAdapters::PostgreSQL` is defined, and also extended it to correctly type postgres Array columns.

DanielHeath:
I've signed the CLA!

lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
```diff
@@ -71,6 +45,40 @@ def type_for(column_name)
...
+            "::String"
+          when ActiveRecord::Type::Date
+            "::Date"
+          when ActiveRecord::Type::Decimal
+            "::BigDecimal"
+          when ActiveRecord::Type::Float
+            "::Float"
+          when ActiveRecord::Type::Boolean
+            "T::Boolean"
+          when defined?(ActiveRecord::ConnectionAdapters::PostgreSQL) && ActiveRecord::ConnectionAdapters::PostgreSQL::OID::Hstore
```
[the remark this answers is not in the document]
DanielHeath: Does 67f8eb9 resolve this to your satisfaction?

DanielHeath:
I don't think another round of "guess what position the maintainer prefers the Money constant to be in" would be a productive use of my time, nor would responding to it be a good use of yours.

I'm going to leave getting this merged with you; you are welcome to adjust as needed to merge, or to ignore it.

KaanOzkan:
Closing this in favour of https://github.com/Shopify/tapioca/pull/1328 since I can't edit this fork.

> I don't think another round of "guess what position the maintainer prefers the Money constant to be in" would be a productive use of my time, nor would responding to it be a good use of yours.

To be clear in my comment above I didn't ask for moving of the `Money` column. I commented on the new additions for Postgresql. Money uses a `decimal` database field under the hood so it needs to be above `ActiveRecord::Type::Decimal`.

To reduce the back and forth I'd also suggest running tests locally before pushing changes.

PR #1274, opened 2022-11-16
Add patch to Ruby version in `dev.yml`

sambostock:
### Motivation

With the patch, Dev downloads a precompiled binary, but without it seems to choose to compile from source, which seems unnecessary.

### Implementation

The config is updated to Ruby 3.1.2, the latest in the 3.1 releases.

### Tests

🎩'd locally.

### DIFF
--- a/dev.yml
+++ b/dev.yml
@@ -3,7 +3,7 @@ name: tapioca
 type: ruby
 
 up:
-  - ruby: '3.1'
+  - ruby: '3.1.2'
   - bundler
 
 env:


dev.yml
```diff
@@ -3,7 +3,7 @@ name: tapioca
 type: ruby
 
 up:
-  - ruby: '3.1'
+  - ruby: '3.1.2'
```
Morriar, line 6: > With the patch, Dev downloads a precompiled binary, but without it seems to choose to compile from source, which seems unnecessary.

Should we add this as a comment explaining why we use this version?

sambostock: [Searching for `x.y` Ruby versions in `dev.yml` across our org](https://github.com/search?q=org%3AShopify+path%3A%2Fdev.yml%24%2F+%2Fruby%3A+%5B%22%27%5D%5Cd%5C.%5Cd%5B%22%27%5D%2F&type=code) only turns up 5 matches, so I think it's safe to say this is uncommon.

PR #1277, opened 2022-11-21
Fix backwards compatibility with sorbet-runtime

dirceu:
This reintroduces the monkeypatch to `T::Utils#coerce` that was removed on #1270. See https://github.com/Shopify/tapioca/pull/1270/files#r1026916745 for more information.


vinistock (approved):

lib/tapioca/sorbet_ext/generic_name_patch.rb
```diff
@@ -92,9 +92,25 @@ def coerce_and_check_module_types(val, check_val, check_module_type)
       end
 
       class << self
-        prepend(CoercePatch)
+        prepend(PrivateCoercePatch)
       end
     end
+
+    module CoercePatch
```
paracycle, line 99: I am having trouble reading this correctly on my phone, but I think this is in the wrong namespace. There didn't use to be a Private namespace before, and this was a patch to just T::Utils. It doesn't look like that is still the case.

Moreover, I doubt we can blindly patch both modules, since for the old code the Private module never existed and for the new code the coerce method should mot be messed around with.

There needs to be some feature detection to decide what to patch.

dirceu: Before https://github.com/sorbet/sorbet/pull/6559, this is how the patch looked like:

https://github.com/Shopify/tapioca/blob/f1b8db05ab6dd48240cc9545c7379aca30f97d49/lib/tapioca/sorbet_ext/generic_name_patch.rb#L80

`T::Utils.coerce` had a "full" implementation, and we tried using our internal registry for overrides but fell back to the original method after that.

After https://github.com/sorbet/sorbet/pull/6559...

* `T::Utils::Private.coerce_and_check_module_types` was introduced with an updated implementation of what `T::Utils.coerce` used to do
* `T::Utils.coerce` became just a "wrapper" that calls `T::Utils::Private.coerce_and_check_module_types` with predefined parameters

On https://github.com/Shopify/tapioca/pull/1270 we patched `T::Utils::Private.coerce_and_check_module_types` because that would also tackle `T::Utils.coerce`, which works for newer versions of Sorbet but (as you correctly pointed out) breaks the assumption for previous versions (in which `.coerce` does **not** just call that method on the `Private` namespace).

So this _should_ be on the right level.

paracycle:
> Are there other examples of a test like this, so I use a similar setup?

It was recently removed: https://github.com/Shopify/tapioca/commit/76515f5d2dcdd30a1c536f776eee1b58dc490494#diff-bee8bd10954cc2dcd94f010d4bc1b3092af1883d0494cfda0c1b653718550b11

But, basically a gem CLI test that sets up Tapioca against both an older and a newer version of the sorbet-runtime gem (on both sides of the breaking upstream PR) to test the codepath that would be broken if we got this wrong (either) way is tested properly.

lib/tapioca/sorbet_ext/generic_name_patch.rb
dirceu: Great point, thank you! I'll add the missing test and re-trigger a review.

spec/helpers/mock_project.rb
```diff
@@ -109,19 +109,19 @@ def bundle_install(version: nil)
...
+        out, err, status = Open3.capture3(env, ["bundle", "_#{bundler_version}_", "exec", command].join(" "), opts)
         ExecResult.new(out: out, err: err, status: T.must(status.success?))
       end
     end
 
     # Run a Tapioca `command` with `bundle exec` in this project context (unbundled env)
-    sig { params(command: String).returns(ExecResult) }
-    def tapioca(command)
+    sig { params(command: String, enforce_typechecking: T::Boolean).returns(ExecResult) }
+    def tapioca(command, enforce_typechecking = false)
```
Morriar, line 124: I'm in favor of always running the tests with this at `true`. Since we have a lot of CLI-tests it will expose more problems we didn't think about when writing unit tests.

dirceu: I've made a separate branch changing the default behaviour for CLI tests (making them only "swallow" errors if you explicitly ask it to): https://github.com/Shopify/tapioca/compare/dont-swallow-errors-on-cli-tests?expand=1 We could treat those `@project.tapioca(<something>, ignore_typechecking_errors: true)` as a "shitlist" while preventing any new CLI tests from implicitly hiding errors. Does that make sense @Morriar @paracycle ?

If so, my plan would be to open a PR with that branch -> merge -> rebase this one (or alternatively I could move everything to this PR, if you prefer).

Morriar: Let's go with another PR for enabling type checking 👍 

Are the elements from the shitlist hard to fix?

vinistock (approved):

PR #1284, opened 2022-11-28
Generate clear_foo methods for protobuf compiler

slinkp:
Protobuf: Add support for `clear_*` methods

### Motivation

[Per the docs](https://developers.google.com/protocol-buffers/docs/reference/ruby-generated#checking-presence), Protobuf generates a method for every field that clears it. We should have sigs declared for those.

### Implementation

Just added to the existing `create_descriptor_method` 

### Tests

Yep

### DIFF
--- a/lib/tapioca/dsl/compilers/protobuf.rb
+++ b/lib/tapioca/dsl/compilers/protobuf.rb
@@ -284,6 +284,11 @@ module Tapioca
             return_type: "void",
           )
 
+          klass.create_method(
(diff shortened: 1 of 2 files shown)


andrewn617 (approved):
LGTM!

dirceu (approved):
Just tested this against Shopify's core monolith and it looks good! Lots of new RBI declarations and no new typechecking errors.

KaanOzkan (approved):

PR #1287, opened 2022-11-30
Avoid swallowing errors in (most) CLI tests

dirceu:
By default `exe/tapioca` will ignore many errors, including typechecking ones. This is because we want to generate as much RBI as possible, even if there are errors in the codebase.

We do want to surface more errors in the CLI tests, however, so this PR is making CLI tests throw by default, while providing an option to ignore it explicitly in cases where the updates to the test setup would not be worth the hassle.

See https://github.com/Shopify/tapioca/pull/1277#issuecomment-1326769461

### DIFF
--- a/spec/helpers/mock_project.rb
+++ b/spec/helpers/mock_project.rb
@@ -120,8 +120,8 @@ module Tapioca
     end
 
     # Run a Tapioca `command` with `bundle exec` in this project context (unbundled env)
-    sig { params(command: String, ignore_typechecking_errors: T::Boolean).returns(ExecResult) }
-    def tapioca(command, ignore_typechecking_errors: false)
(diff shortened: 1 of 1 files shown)


Morriar:
I'm not sure the PR description is completely accurate here.

What we want to do is surface type errors occuring _inside_ Tapioca as we run it in test mode.

The Tapioca executable disables sorbet-runtime for performance reasons, this has the side effect of not surfacing typing errors in our own code when we run the CLI tests.

This PR does not disable the error silencing we apply when loading the target application. Maybe this is what is confusing @paracycle here.

Morriar (approved):

paracycle (approved):
One small naming suggestion, otherwise 👍

paracycle (approved):

PR #1288, opened 2022-12-05
Handle keyword argument splats correctly

timfjord:
### Motivation

Currently, keyword argument splats aren't handled in the base compiler class so methods like the below don't work
```ruby
sig { params(arg1: Symbol, kwargs: Symbol).returns(Symbol) }
def my_methods(arg1, **kwargs) = arg1
```

### Implementation

This PR adds support for keyword argument splats by relying on the `keyrest_type` method from the signature

### Tests

There are no tests for the base class.  
Theoretically, the change could be covered in `spec/tapioca/cli/dsl_spec.rb` but I couldn't make it work so I will invest more into this

### DIFF
--- a/lib/tapioca/dsl/compiler.rb
+++ b/lib/tapioca/dsl/compiler.rb
@@ -98,10 +98,11 @@ module Tapioca
         # keyword parameters types
         signature.kwarg_types.each { |_, kwarg_type| params << kwarg_type.to_s }
 
-        # rest and keyrest parameter types
-        ["", "key"].each do |prefix|
-          params << signature.public_send("#{prefix}rest_type").to_s if signature.public_send("has_#{prefix}rest")
-        end
+        # rest parameter type
+        params << signature.rest_type.to_s if signature.has_rest
+
+        # keyrest parameter type
+        params << signature.keyrest_type.to_s if signature.has_keyrest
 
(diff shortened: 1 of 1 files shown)


lib/tapioca/dsl/compiler.rb
```diff
@@ -98,8 +98,10 @@ def parameters_types_from_signature(method_def, signature)
         # keyword parameters types
         signature.kwarg_types.each { |_, kwarg_type| params << kwarg_type.to_s }
 
-        # rest parameter type
-        params << signature.rest_type.to_s if signature.has_rest
+        # rest and keyrest parameter types
+        ["", "key"].each do |prefix|
+          params << signature.public_send("#{prefix}rest_type").to_s if signature.public_send("has_#{prefix}rest")
+        end
```
Morriar, line 104: I'm not convince with the use of `public_send` here. Wouldn't calling the real method directly make it easier to read and typecheck?

Something like this:

```rb
        # rest parameter type
        params << signature.rest_type.to_s if signature.has_rest

        # keyrest parameter type
        params << signature.keyrest_type.to_s if signature.has_keyrest
```

timfjord: Agree, removed the `public_send`

PR #1299, opened 2022-12-08
Include base controller helpers when possible

waj:
### Motivation
In a Rails project is common to have helpers defined in a base controller. Currently those helpers are declared on every `.rbi` file for descendant controllers unless those controllers define new helpers. The problem with this approach is that whenever a slight change is made on the base controller helpers, most `.rbi` files for descendants get updated making reviews of PRs annoying.

### Implementation
If the helpers module of a controller is the same helpers module of it superclass then it's guaranteed that the helpers methods are the same. So, instead of generating the helpers methods, use the helper methods module generated for the parent class.

### Tests
There was already a test with the case I described, so I updated the test to check that the subclass module just includes the parent's class module.

### DIFF
--- a/spec/tapioca/dsl/compilers/action_controller_helpers_spec.rb
+++ b/spec/tapioca/dsl/compilers/action_controller_helpers_spec.rb
@@ -160,13 +160,7 @@ module Tapioca
                   def helpers; end
 
                   module HelperMethods
-                    include ::ActionController::Base::HelperMethods
-
-                    sig { returns(T.untyped) }
-                    def current_user_name; end
-
-                    sig { params(user_id: ::Integer).void }
-                    def notify_user(user_id); end
+                    include ::BaseController::HelperMethods
                   end
 
                   class HelperProxy < ::ActionView::Base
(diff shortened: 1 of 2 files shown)


spec/tapioca/dsl/compilers/action_controller_helpers_spec.rb
```diff
@@ -160,13 +160,7 @@ class UserController
                   def helpers; end
 
                   module HelperMethods
-                    include ::ActionController::Base::HelperMethods
-
-                    sig { returns(T.untyped) }
-                    def current_user_name; end
-
-                    sig { params(user_id: ::Integer).void }
```
KaanOzkan, line 168: Removing the duplication is nice but I don't think it's worth losing the typing information in the signature.

waj: Oh... I see what you mean. I didn't notice in this case the signature for `notify_user` was different in the base class. I'll see how can I avoid this case.

PR #1302, opened 2022-12-09
Decorate ActiveRecord::SignedId finder methods

bdewater-thatch:
### Motivation
Noticed I missed typing for these. https://api.rubyonrails.org/classes/ActiveRecord/SignedId/ClassMethods.html was added in Rails 6.1.

### Implementation
Similar to existing finder methods.

### Tests
Tests added.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_relations.rb
+++ b/lib/tapioca/dsl/compilers/active_record_relations.rb
@@ -217,6 +217,7 @@ module Tapioca
           T::Array[Symbol],
         )
         FINDER_METHODS = T.let(ActiveRecord::FinderMethods.instance_methods(false), T::Array[Symbol])
+        SIGNED_FINDER_METHODS = T.let(ActiveRecord::SignedId::ClassMethods.instance_methods(false), T::Array[Symbol])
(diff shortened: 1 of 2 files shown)


lib/tapioca/dsl/compilers/active_record_relations.rb
```diff
@@ -593,6 +594,33 @@ def create_common_methods(common_relation_methods_module)
...
+              create_common_method(
+                "find_signed!",
+                common_relation_methods_module,
+                parameters: [
+                  create_param("signed_id", type: "T.untyped"),
+                  create_kw_opt_param("purpose", type: "T.untyped", default: "nil")
+                ],
+                return_type: constant_name,
+              )
+            else
```
paracycle, line 619: Should we add `signed_id_verifier` and `signed_id_verifier=` as well? The nice thing about the `case`/`when` matching is that even if those methods are defined in Rails 7.0 or something, it still works properly in Rails 6.1, since they will never get matched from the actual instance methods on the module.

bdewater-thatch: I wasn't entirely sure about that one either, but more about the fact that configuring the verifier has little to do with with AR relations. Since I had no need for setting this let alone better typing for it, I decided to let it be :)

paracycle (approved):

PR #1303, opened 2022-12-12
Do not pass `T.untyped` to `ObjectSpace::WeakMap`

LTe:
### Motivation
In https://github.com/sorbet/sorbet/pull/6610 type_member changed from `out` to `fixed: T.untyped`. From this point we should not pass explicit `T.untyped` as argument to generic type.

### Implementation
There is no need to special case for `ObjectSpace::WeakMap` anymore.

### Tests
I added specs


lib/tapioca/gem/pipeline.rb
```diff
@@ -240,10 +240,7 @@ def compile_object(name, value)
 
         klass = class_of(value)
 
-        klass_name = if klass == ObjectSpace::WeakMap
-          # WeakMap is an implicit generic with one type variable
-          "ObjectSpace::WeakMap[T.untyped]"
-        elsif T::Generic === klass
+        klass_name = if T::Generic === klass
```
Morriar, line 243: Wouldn't this may create an error depending on the version of Sorbet you're using?

I think we should register this as a feature in [SorbetHelper](https://github.com/Shopify/tapioca/blob/main/lib/tapioca/helpers/sorbet_helper.rb) and generate the correct type depending on `sorbet_supports?`

spec/tapioca/gem/pipeline_spec.rb
```diff
@@ -1953,7 +1953,7 @@ module Toto; end
       RUBY
 
       output = template(<<~RBI)
-        Foo = T.let(T.unsafe(nil), ObjectSpace::WeakMap[T.untyped])
+        Foo = T.let(T.unsafe(nil), ObjectSpace::WeakMap)
```
Morriar, line 1956: Can we add a test showing the behavior with the Sorbet version before and after the type of WeakMap was changed pease?

Morriar (commented):
Very nice!

I wonder if we could add two tests under `spec/tapioca/cli/gem_spec` where we install different versions of Sorbet and check the output.

Something like this:

```rb
        it "generates `ObjectSpace::WeakMap` as a generic type with Sorbet < X.Y.Z" do
          foo = mock_gem("foo", "0.0.1") do
            write("lib/foo.rb", <<~RB)
              Foo = ObjectSpace::WeakMap.new
            RB
          end
          @project.require_mock_gem(foo)
          @project.require_real_gem("sorbet-static", "X.Y.Z")
          @project.bundle_install
          result = @project.tapioca("gem")
          assert_project_file_equal("sorbet/rbi/gems/foo@0.0.1.rbi", <<~RBI)
             Foo = T.let(T.unsafe(nil), ObjectSpace::WeakMap[T.untyped])
          RBI
        end

        it "generates `ObjectSpace::WeakMap` as a non-generic type with Sorbet >= X.Y.Z" do
          foo = mock_gem("foo", "0.0.1") do
            write("lib/foo.rb", <<~RB)
              Foo = ObjectSpace::WeakMap.new
            RB
          end
          @project.require_mock_gem(foo)
          @project.require_real_gem("sorbet-static", "X.Y.Z")
          @project.bundle_install
          result = @project.tapioca("gem")
          assert_project_file_equal("sorbet/rbi/gems/foo@0.0.1.rbi", <<~RBI)
             Foo = T.let(T.unsafe(nil), ObjectSpace::WeakMap)
          RBI
        end

File: spec/tapioca/gem/pipeline_spec.rb (hunk 2 above)
Comment (KaanOzkan) [reply spec/tapioca/gem/pipeline_spec.rb:1958]: Let's go with @Morriar's suggestion since we don't test with different sorbet versions. This  conditional will only enter sorbet version in CI. https://github.com/Shopify/tapioca/pull/1303#pullrequestreview-1216282219

**Review by KaanOzkan — APPROVED** [review]
Thanks a lot for the PR and the prompt back and forth.

**Review by Morriar — APPROVED** [review]
Thank you for adding the other test 🙏

File: spec/tapioca/cli/gem_spec.rb [hunk 3]
```diff
@@ -1466,6 +1466,54 @@ class User; end
...
+            write("lib/foo.rb", <<~RB)
+              Foo = ObjectSpace::WeakMap.new
+            RB
+          end
+
+          @project.gemfile(@project.tapioca_gemfile)
+          @project.require_mock_gem(foo)
+          @project.require_real_gem("sorbet-static", "0.5.10585")
+          @project.bundle_install
+          @project.tapioca("gem")
```
KaanOzkan, line 1481: Test failure is caused by `sorbet-static-and-runtime` version not being satisfied during tapioca execution. Mock project depends on tapioca which brings in the Sorbet version that project uses. In the failing CI builds Sorbet version is higher than `0.5.10585` and as a result `bundle install` fails since it didn't bump it.

We can run a `bundle update` after requiring `sorbet-static-and-runtime` instead:
```
          @project.gemfile(@project.tapioca_gemfile)
          @project.require_mock_gem(foo)
          @project.bundle_install
          @project.require_real_gem("sorbet-static-and-runtime", "0.5.10585")
          @project.bundle_update("sorbet-static-and-runtime")
          @project.tapioca("gem")
```

`bundle_update` can be implemented similar to [bundle_install](https://github.com/Shopify/tapioca/blob/main/spec/helpers/mock_project.rb#L89-L109). Maybe it's okay to couple this with `require_real_gem` and trigger it within that method.

paracycle: Hmm, however the test setup should not work that way. The dependency on Tapioca is now via the `.gemspec` file, which should not limit the `sorbet-static-and-runtime` version selection. Not sure what's going wrong here.

paracycle: Ok, it is about the `mock_project` bringing in the sorbet dependency automatically. The way around it is to use the `mock_project(sorbet_dependency: false)` flag, like it was here: https://github.com/Shopify/tapioca/pull/1260/files#diff-bee8bd10954cc2dcd94f010d4bc1b3092af1883d0494cfda0c1b653718550b11L1875

paracycle: I pushed a commit to fix that

KaanOzkan: Good find, this is a much better solution.

paracycle (approved):
Thank you! The work in this PR is very much appreciated. ❤️

Issue #1306, opened 2022-12-12
Fix flakyness on CI


paracycle:
We have some recurring problems that keep popping up on CI:
1. Bundler/gemspec related error: `rubygems/specification.rb:2121:in 'method_missing': undefined method 'identifier' for #<Gem::Specification:0x0000562ed629c100> (NoMethodError)`
    - https://github.com/Shopify/tapioca/actions/runs/3650000604/jobs/6165419530#step:5:588
    - https://github.com/Shopify/tapioca/actions/runs/3643033170/jobs/6150808732#step:5:62
    - https://github.com/Shopify/tapioca/actions/runs/3643012213/jobs/6150769063#step:5:704
2. `PP::ObjectMixin` flakiness on Rails 6.1 tests: 
    - https://github.com/Shopify/tapioca/actions/runs/3633766682/jobs/6131151159#step:5:485
    - https://github.com/Shopify/tapioca/actions/runs/3633749602/jobs/6131114809#step:5:386

paracycle:
For now, both issues seem to be address to various degrees of effectiveness. Will reopen if we see regressions.

KaanOzkan:
Listing regressions here as I observe them:

- Caught one in the most recent dependabot merge https://github.com/Shopify/tapioca/actions/runs/3761036498/jobs/6392358002#step:6:553

PR #1317, opened 2022-12-15
Fix CI flakyness with respect to `Bundler`

paracycle:
### Motivation

As mentioned in #1306, we'd been observing some tests failures with message `rubygems/specification.rb:2121:in 'method_missing': undefined method 'identifier' for #<Gem::Specification:0x0000562ed629c100> (NoMethodError)`. We investigated this with @Morriar and were not able to reproduce the error locally or on Spin however hard we've tried.

What does not make sense about the whole thing is that at that point in the code where the failures always happen, there shouldn't be any instances of `Gem::Specification` anywhere, all entries should be `Bundler::StubSpecification` instances. So the error is quite unexpected.

We ultimately noticed 2 important things about our CI:
1. Our CI runs for non-main `Gemfile`s do not install a recent enough Bundler version. Since there is no lockfile associated with those gemfiles, Ruby setup does not know which Bundler version was used to create a lockfile, so just uses the Bundler that comes with Ruby. This has the possibility to make our tests non-deterministic.
2. Our main lockfile was locked to a slightly older Bundler version. This might have meant that we were running into an edge case on that version.

### Implementation

The implementation in this PR addresses the 2 points above:
1. We have a new step for every test job that installs the latest Bundler version possible. This ensures that the test runs are more deterministic, at least.
2. We've update the lockfile to be bundled with `2.3.26` instead of `2.3.23`.

### Tests

Ran the test suite 10 times on this branch and the only failures that we are observing is the other issue about `PP::ObjectMixin`, so we are pretty confident the Bundler issue is resolved with these changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,6 +56,8 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true
+      - name: Install latest Bundler version
+        run: gem install bundler
       - name: Run tests
         run: bin/test
         continue-on-error: ${{ !!matrix.experimental }}
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -393,4 +393,4 @@ DEPENDENCIES
   xpath
 
 BUNDLED WITH
-   2.3.23
+   2.3.26


egiurleo (approved):

paracycle:
I know this didn't completely fix the `Bundler` related flakyness but I went ahead and merged it anyway, since I think it improves our infrastructure regardless.

PR #1328, opened 2022-12-22
Add types for postgres hstore & arrays

KaanOzkan:
A new PR for https://github.com/Shopify/tapioca/pull/1268 because I don't have push permissions to the fork.

### Motivation
When using an hstore column in postgres, tapioca generates "T.untyped". Both keys and values in hstore are strings.

### Implementation
Added a case for hstore to the existing case statement.

### Tests
Testing this is hard because we can't use postgres columns with sqlite. Open to other testing suggestions

### DIFF
--- a/lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
+++ b/lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
@@ -19,33 +19,7 @@ module Tapioca
 
           column_type = @constant.attribute_types[column_name]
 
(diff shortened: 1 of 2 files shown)


dirceu (approved):

DanielHeath:
Hey, not really a big deal or anything, but the [v0.10.5](https://github.com/Shopify/tapioca/releases/tag/v0.10.5) release notes have attributed this work incorrectly.

PR #1329, opened 2022-12-22
Improve eager loading Rails engines when Zeitwerk is present

### LINKED ISSUE #671: Missing type declaration ::Turbo::Streams::ActionHelper for turbo-rails.
I created a brand new rails 7 project. The first thing I did was I added tapioca!
I generated all the gem & dsl rbis, fixed all I could, but this one is not obvious and I bet others will have a problem with this as well.

The problem:

```
% bin/srb tc           
sorbet/rbi/dsl/rails/conductor/action_mailbox/inbound_emails_controller.rbi:16: Unable to resolve constant ActionHelper https://srb.help/5002
    16 |    include ::Turbo::Streams::ActionHelper
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  Autocorrect: Use `-a` to autocorrect
    sorbet/rbi/dsl/rails/conductor/action_mailbox/inbound_emails_controller.rbi:16: Replace with AbstractController::Callbacks::ActionFilter
    16 |    include ::Turbo::Streams::ActionHelper
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  Autocorrect: Use `-a` to autocorrect
    sorbet/rbi/dsl/rails/conductor/action_mailbox/inbound_emails_controller.rbi:16: Replace with Turbo::DriveHelper
    16 |    include ::Turbo::Streams::ActionHelper
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    sorbet/rbi/gems/actionpack@7.0.0.rbi:178: Did you mean: AbstractController::Callbacks::ActionFilter?
     178 |class AbstractController::Callbacks::ActionFilter
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    sorbet/rbi/gems/turbo-rails@1.0.0.rbi:66: Did you mean: Turbo::DriveHelper?
    66 |module Turbo::DriveHelper
        ^^^^^^^^^^^^^^^^^^^^^^^^^
...
Errors: 12
```

When I checked turbo-rails where that module lives, I found it at https://github.com/hotwired/turbo-rails/blob/main/app/helpers/turbo/streams/action_helper.rb which does not live in the lib path, but it looks like it is autoloaded by this line here: https://github.com/hotwired/turbo-rails/blob/main/lib/turbo/engine.rb#L13

Could you provide some guidance on how I would require that action_helper.rb file in my project so that tapioca will generate the appropriate type definitions?

Thanks for your time!

egiurleo:
### Motivation
Closes #671
Closes https://github.com/Shopify/rbi-central/pull/135

RBI generation for Rails engines is currently implemented with a best effort approach. We identify every file in the application's `eager_load_path` and attempt to require those files. This does not account for dependency order, so we keep track of any files that fail in the process and then try to require them _again_ once all the other files are loaded.

This approach is not exhaustive and can sometimes miss constants defined in engines. For example, if an engine defines a constant (e.g. `Turbo::Streams`) as a namespace for subconstants, but never on its own, the current approach will never generate type annotations for that constant.

### Implementation
This implementation improves engine eager loading when Zeitwerk is present in the Rails application. Rather than relying on the best effort approach, we use the behavior already implemented in Zeitwerk to eager load every file in the application's `eager_load_path`, which is much more reliable. The resulting RBI will contain constants defined in the engine that wouldn't have been found by the previous approach.

If Zeitwerk is not present, then we fall back to the best effort approach that we were already using.

### Tests
We have added a test that passes with these changes and fails without them.


spec/tapioca/cli/gem_spec.rb
```diff
@@ -2022,6 +2022,24 @@ class << self
...
+        assert_includes(response.out, "Compiled turbo-rails")
+
+        turbo_streams_rbi = @project.read("sorbet/rbi/gems/turbo-rails@1.3.2.rbi")
+
+        # The Turbo::Streams constant is never defined on its own, only as a namespace for child constants.
+        # Because of this, tapioca cannot create an annotation for it using traditional auto-loading methods
+        # By leveraging Zeitwerk auto-loading, tapioca can discover this constant and correctly create an
+        # annotation for it.
+        assert_includes(turbo_streams_rbi, "module Turbo::Streams; end")
+      end
```
egiurleo, line 2042: @paracycle Let me know if this test looks reasonable to you. I've confirmed that undo-ing our changes breaks the test.

spec/tapioca/cli/gem_spec.rb
```diff
@@ -2022,6 +2022,25 @@ class << self
...
+        assert_includes(response.out, "Compiled turbo-rails")
+
+        turbo_streams_rbi = @project.read("sorbet/rbi/gems/turbo-rails@1.3.2.rbi")
+
+        # The Turbo::Streams constant is never defined on its own, only as a namespace for child constants.
+        # Because of this, tapioca cannot create an annotation for it using traditional auto-loading methods
+        # By leveraging Zeitwerk auto-loading, tapioca can discover this constant and correctly create an
+        # annotation for it.
+        assert_includes(turbo_streams_rbi, "module Turbo::Streams; end")
+        assert_includes(turbo_streams_rbi, "module Turbo::Streams::ActionHelper")
```
egiurleo, line 2042: @paracycle is this the only helper constant you had in mind? This is the one mentioned in #671 but I wasn't sure if there were more.

paracycle (approved):
Thank you!

lib/tapioca/loaders/loader.rb
```diff
@@ -79,15 +97,41 @@ def load_rails_engines
...
+        # to be set, so we need to create one here.
+        unless rails_application
+          Rails.application = Class.new(Rails::Application)
+        end
+
+        blk.call
+      ensure
+        unless rails_application
+          Rails.app_class = Rails.application = nil
+        end
```
KaanOzkan, line 123: A small suggestion, it's simpler for me but curious what you think.

```suggestion
        Rails.app_class = Rails.application = rails_application
```

lib/tapioca/loaders/loader.rb
```diff
@@ -79,15 +97,41 @@ def load_rails_engines
...
+
+      T::Sig::WithoutRuntime.sig { returns(T::Array[T.class_of(Rails::Engine)]) }
       def rails_engines
         return [] unless Object.const_defined?("Rails::Engine")
 
         safe_require("active_support/core_ext/class/subclasses")
 
         project_path = Bundler.default_gemfile.parent.expand_path
         # We can use `Class#descendants` here, since we know Rails is loaded
-        Object.const_get("Rails::Engine")
```
KaanOzkan, line 90: I'm curious, was there a reason these used to be `const_get`?

KaanOzkan (approved):
🎉 

I think I'll close https://github.com/Shopify/rbi-central/pull/135 since Zeitwerk is default for Rails 6+

lib/tapioca/loaders/loader.rb
egiurleo: That is a really good question and I'm not sure why, besides maybe it needed to be that way at some point and nobody ever fixed it? I'm assuming that because we already had the `Object.const_defined?("Rails::Engine")` check in place, this change could have been made earlier. @paracycle might know!

lib/tapioca/loaders/loader.rb
```diff
@@ -103,30 +145,27 @@ def safe_require(path)
       sig { void }
       def silence_deprecations
         # Stop any ActiveSupport Deprecations from being reported
-        Object.const_get("ActiveSupport::Deprecation").silenced = true
+        if defined?(ActiveSupport::Deprecation)
+          ActiveSupport::Deprecation.silenced = true
+        end
       rescue NameError
         nil
```
paracycle, line 152: I don't think we can still get a `NameError` in this method, due to the `defined?` check. So we can remove this.

lib/tapioca/loaders/loader.rb
```diff
@@ -79,15 +97,39 @@ def load_rails_engines
...
+        end
+
+        blk.call
+      ensure
+        Rails.app_class = Rails.application = rails_application
+      end
+
+      T::Sig::WithoutRuntime.sig { returns(T::Array[T.class_of(Rails::Engine)]) }
       def rails_engines
         return [] unless Object.const_defined?("Rails::Engine")
```
paracycle, line 126: I think we can make this:
```suggestion
        return [] unless defined?(Rails::Engine)
```
which reads better than `const_defined?` and the constant reference can be typechecked.

egiurleo:
Okay I think I've addressed everyone's comments. I also squashed all the little commits into one big one and gave it a good description.

PR #1331, opened 2022-12-23
Ensure that trackers stay enabled until after `gem` command is run

### LINKED ISSUE #1325: Disabled tracker in `tapioca init` generates incomplete RBI files for gems
Because the `tapioca init` CLI method runs multiple commands, the first of which disables the tracker, the tracker erroneously remains disabled through the rest of the commands, including the gem command. This causes the resulting RBIs to have incomplete data and result in type checking errors.

paracycle:
### Motivation

Fixes #1325

Since all commands other than `gem` and `dsl` commands now inherit from `CommandWithoutTracker`, as soon as we invoke the `configure` command inside `init`, we ended up disabling the trackers. This meant that by the time the `gem` command got to run, the trackers were already disabled and that we generated the RBIs with missing information.

### Implementation

The fix is to create a block in which trackers can't be disabled and using that up to and past the point where the `gem` command is run.

### Tests

No additional tests, since it is quite time consuming to test the `init` command. Ultimately, we should find a way around that.

I did top-hat this in a fresh Rails app though and it works as expected, whereas it was broken before.

### DIFF
--- a/lib/tapioca/cli.rb
+++ b/lib/tapioca/cli.rb
@@ -25,9 +25,12 @@ module Tapioca
 
     desc "init", "get project ready for type checking"
     def init
-      invoke(:configure)
-      invoke(:annotations)
-      invoke(:gem)
+      # We need to make sure that trackers stay enabled until the `gem` command is invoked
+      Runtime::Trackers.with_trackers_enabled do
+        invoke(:configure)
+        invoke(:annotations)
+        invoke(:gem)
+      end
       invoke(:todo)
 
       print_init_next_steps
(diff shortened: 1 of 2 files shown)


kddnewton (approved):
This works but is pretty dang hacky.

Another approach would be to store a boolean instance variable on the class that is like `@force_enabled`. Then it would be:

```ruby
def with_trackers_enabled(&blk)
  @force_enabled = true
ensure
  @force_enabled = false
end

def disable_all!
  @trackers.each(&:disable!) unless @force_enabled
end
```

but honestly, your way works too so I'm approving.

paracycle:
> This works but is pretty dang hacky.

Yeap, I am aware, but I didn't want to add another state to the whole tracker thing, at least for now. I think we can find a cleaner way to do this later.

PR #1336, opened 2022-12-28
Add Kredis compiler

bdewater:
### Motivation
Needed type check guarantees for the dynamically created [attribute methods](https://github.com/rails/kredis/blob/v1.3.0/lib/kredis/attributes.rb) and [enum methods](https://github.com/rails/kredis/blob/v1.3.0/lib/kredis/types/enum.rb) from Kredis.

### Implementation
Depends on the annotations from https://github.com/Shopify/rbi-central/pull/140. Used an extension similar to the existing one for FrozenRecord.

### Tests
Added tests.


spec/tapioca/dsl/compilers/kredis_spec.rb
```diff
@@ -0,0 +1,209 @@
+# typed: strict
+# frozen_string_literal: true
+
+require "spec_helper"
+
```
KaanOzkan, line 5: Looks like kredis is only available for Rails 7+. We can use the `require` logic above to skip these tests as well in case it isn't loaded.

KaanOzkan (approved):
The compiler LGTM However, this is my first time looking at kredis.

spec/tapioca/dsl/compilers/kredis_spec.rb
bdewater: I had forgot to add the gem requirement to `Gemfile-rails-6-1`, Kredis itself requires Active {Model,Support} >= 6.0.

PR #1340, opened 2023-01-03
Use latest rubygems for Rails main test on CI

st0012:
Rails 7.1 now requires rubygems 3.3.13, which is newer than most Ruby versions install by default. PR: rails/rails#46817

So if we don't upgrade the default rubygems, `bundle install` would fail.


.github/workflows/ci.yml
```diff
@@ -56,6 +56,7 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true
+          rubygems: ${{ contains(matrix.gemfile, 'rails-main') && 'latest' || 'default' }}
```
paracycle, line 59: Can't we always install `latest`?

andyw8:
Nice, TIL about that config option:

https://github.com/ruby/setup-ruby/blob/ad718faf7af4b26fef9165e2d675890b51901e6c/action.yml#L11-L16

PR #1352, opened 2023-01-11
Check-shims should not error with different superclasses

KaanOzkan:
### Motivation
While looking at https://github.com/Shopify/rbi-central/pull/143 I noticed check-shims command doesn't take the superclass information into account.

### Implementation
Compare `superclass_name` fields if we are dealing with empty `RBI::Class` nodes. If the node isn't an empty `RBI::Scope` then we continue to check the body for duplicates.

### Tests
Added.

### DIFF
--- a/lib/tapioca/helpers/rbi_files_helper.rb
+++ b/lib/tapioca/helpers/rbi_files_helper.rb
@@ -175,7 +175,15 @@ module Tapioca
       return false if shims_or_todos.empty?
 
       shims_or_todos_empty_scopes = extract_empty_scopes(shims_or_todos)
-      return true unless shims_or_todos_empty_scopes.empty?
+      shims_or_todos_empty_scopes.each do |shim_or_todo|
+        return true unless shim_or_todo.is_a?(RBI::Class)
+
+        nodes.each do |node|
+          return true unless node.is_a?(RBI::Class)
+          next if node == shim_or_todo
+          return true if node.superclass_name == shim_or_todo.superclass_name
+        end
+      end
 
       mixins = extract_mixins(shims_or_todos)
       return true unless mixins.empty?
--- a/spec/tapioca/cli/check_shims_spec.rb
+++ b/spec/tapioca/cli/check_shims_spec.rb
@@ -122,6 +122,21 @@ module Tapioca
           refute_success_status(result)
         end
 
+        it "ignores duplicates that have a different superclass" do
+          @project.write("sorbet/rbi/gems/foo@1.0.0.rbi", <<~RBI)
+            class Foo
+            end
+          RBI
+
+          @project.write("sorbet/rbi/shims/foo.rbi", <<~RBI)
+            class Foo < Object
+            end
+          RBI
+
+          result = @project.tapioca("check-shims --no-payload")
+          assert_success_status(result)
+        end
+
         it "ignores duplicates that have a signature" do
           @project.write("sorbet/rbi/gems/foo@1.0.0.rbi", <<~RBI)
             class Foo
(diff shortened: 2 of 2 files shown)


spec/tapioca/cli/check_shims_spec.rb
```diff
@@ -344,10 +359,6 @@ class Bar; end
         end
 
         it "detects duplicates from Sorbet's payload" do
-          @project.write("sorbet/rbi/shims/core/object.rbi", <<~RBI)
-            class Object; end
-          RBI
```
KaanOzkan, line 349: This used to be an error because the payload defines it as `class Object < BasicObject`

lib/tapioca/helpers/rbi_files_helper.rb
```diff
@@ -175,7 +175,15 @@ def shims_or_todos_have_duplicates?(nodes, shim_rbi_dir:, todo_rbi_file:)
       return false if shims_or_todos.empty?
 
       shims_or_todos_empty_scopes = extract_empty_scopes(shims_or_todos)
-      return true unless shims_or_todos_empty_scopes.empty?
+      shims_or_todos_empty_scopes.each do |shim_or_todo|
+        return true unless shim_or_todo.is_a?(RBI::Class)
```
paracycle, line 179: Do you mean `next` here instead? What if the next empty scope is a class?

lib/tapioca/helpers/rbi_files_helper.rb
```diff
@@ -175,7 +175,15 @@ def shims_or_todos_have_duplicates?(nodes, shim_rbi_dir:, todo_rbi_file:)
...
+
+        nodes.each do |node|
+          return true unless node.is_a?(RBI::Class)
```
paracycle, line 182: Again, do you mean `next`?

lib/tapioca/helpers/rbi_files_helper.rb
```diff
@@ -175,7 +175,15 @@ def shims_or_todos_have_duplicates?(nodes, shim_rbi_dir:, todo_rbi_file:)
...
+          next if node == shim_or_todo
+          return true if node.superclass_name == shim_or_todo.superclass_name
+        end
+      end
```
paracycle, line 186: Can this all not be:
```suggestion
      return true if shims_or_todos_empty_scopes.grep(RBI::Class).any? do |shim_or_todo|
        nodes.grep(RBI::Class).any? do |node|
          node != shim_or_todo && node.superclass_name == shim_or_todo.superclass_name
        end
      end
```
which can then be encapsulated in a method.

spec/tapioca/cli/check_shims_spec.rb
```diff
@@ -122,6 +122,21 @@ def baz; end
           refute_success_status(result)
         end
 
+        it "ignores duplicates that have a different superclass" do
```
paracycle, line 125: Not sure if this test is correct, since these two cases are defining the same `Foo` class and are, in fact, duplicate definition. For missing superclasses, we should assume `Object`, and then these would be duplicates like expected.

paracycle (commented):
I am not quite aligned on why we need this, and I fear that it will end up missing some definitions that are actually duplicates.

If classes are defined with different superclasses in different files, that should be an error, IMO.

KaanOzkan:
>I am not quite aligned on why we need this, and I fear that it will end up missing some definitions that are actually duplicates.
If classes are defined with different superclasses in different files, that should be an error, IMO.

We had an offline discussion and it made it clear that shims that'd utilize this feature won't be accurate. Biggest thing is that inheritance doesn't just makes methods available for Sorbet but also makes the child class be usable in place of the parent. This would potentially hide typechecking errors and as such shouldn't be encouraged. Ideally we'll be able to tell Sorbet to delegate methods to a certain class instead.

PR #1353, opened 2023-01-12
ignore tmp directory when initializing sorbet config

salmanasiddiqui:
### Motivation

- Potentially speed up `dev tc` on local envs which can accumulate alot of stuff in tmp
- Avoid random errors from `tmp` directory

For https://github.com/Shopify/shopify/pull/399463


Morriar (approved):
Thanks!

Issue #1351, opened 2023-01-11
Tapioca doesn't generate signatures for Active Record virtual attributes


bdewater-thatch:
These compilers exist:
- https://github.com/Shopify/tapioca/blob/main/lib/tapioca/dsl/compilers/active_model_attributes.rb
- https://github.com/Shopify/tapioca/blob/main/lib/tapioca/dsl/compilers/active_record_columns.rb

But Tapioca doesn't do anything for [Active Record attributes](https://api.rubyonrails.org/classes/ActiveRecord/Attributes/ClassMethods.html) when there's no backing database column (aka [virtual attributes](https://gorails.com/episodes/virtual-attributes-and-rails-5-attribute-api)). For example:

```ruby
create_table "members", id: :string, force: :cascade do |t|
  t.string "date_of_birth"
end

class Member < ApplicationRecord
  attribute :date_of_birth, :date
  attribute :eligible, :boolean
end
```
Tapioca correctly picks up on the fact we want a Date and not a String:
```ruby
sig { returns(T.nilable(::Date)) }
def date_of_birth; end

sig { params(value: T.nilable(::Date)).returns(T.nilable(::Date)) }
def date_of_birth=(value); end
# etc
```
but nothing for `eligible`, while these exist at runtime:
```
[6] pry(main)> Member.instance_methods.grep(/eligible/)
=> [:eligible,
 :eligible_came_from_user?,
 :eligible_previously_changed?,
# etc
```

PR #1360, opened 2023-01-17
Teach ActiveRecordColumns compiler about virtual attributes

bdewater:
### Motivation
Closes https://github.com/Shopify/tapioca/issues/1351

### Implementation
`_default_attributes` contains the contents of `columns_hash` and virtual attributes, see
- https://github.com/rails/rails/blob/v7.0.4/activerecord/lib/active_record/model_schema.rb#L486-L489
- https://github.com/rails/rails/blob/v7.0.4/activerecord/lib/active_record/attributes.rb#L288

### Tests
Test added.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_columns.rb
+++ b/lib/tapioca/dsl/compilers/active_record_columns.rb
@@ -16,7 +16,8 @@ module Tapioca
       # `Tapioca::Dsl::Compilers::ActiveRecordColumns` refines RBI files for subclasses of
       # [`ActiveRecord::Base`](https://api.rubyonrails.org/classes/ActiveRecord/Base.html).
       # This compiler is only responsible for defining the attribute methods that would be
-      # created for the columns that are defined in the Active Record model.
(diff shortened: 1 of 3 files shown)


lib/tapioca/dsl/compilers/active_record_columns.rb
```diff
@@ -109,8 +110,7 @@ def decorate
 
           root.create_path(constant) do |model|
             model.create_module(AttributeMethodsModuleName) do |mod|
-              constant.columns_hash.each_key do |column_name|
-                column_name = column_name.to_s
+              constant._default_attributes.keys.each do |column_name|
```
paracycle, line 113: Can we use `attribute_names` instead, since it does not rely on the default values of attributes and is documented since Rails 6.1 (at least): http://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods.html#method-i-attribute_names

```suggestion
              constant.attribute_names.each do |column_name|
```

paracycle (commented):
Nice find!

paracycle (approved):
Thanks

PR #1369, opened 2023-01-23
Update GraphqlInputObject compiler to skip methods from code

jeffcarbs:
### Motivation

We've come across some graphql input objects that define a method with the same name as an argument. For example:
```ruby
class SomeInput < ::GraphQL::Schema::InputObject
  argument :foo_id, String

  sig { returns(Integer) }
  def foo_id
    self[:foo_id].to_i
  end
end
```

In reality, what happens is that the graphql gem meta-programmatically defines a `foo_id` method ([source](https://github.com/rmosolgo/graphql-ruby/blob/89854fa911d8243bb216910c370dbed780c8c178/lib/graphql/schema/input_object.rb#L133-L139)) and then the user-defined `foo_id` method overrides it. If the sigs are different (like in the example above) this confuses sorbet so we should avoid defining RBI for any arguments where an explicit getter is already defined in code.

### Implementation

I updated the `GraphqlInputObject` compiler to only generate RBI for methods defined via the graphql gem.

### Tests

See added tests.

### DIFF
--- a/lib/tapioca/dsl/compilers/graphql_input_object.rb
+++ b/lib/tapioca/dsl/compilers/graphql_input_object.rb
@@ -45,13 +45,11 @@ module Tapioca
 
         sig { override.void }
         def decorate
-          graphql_gem = T.must(Gemfile.new([]).gem("graphql"))
-
           # Skip methods explicitly defined in code
           arguments = constant.all_argument_definitions.select do |argument|
-            source_location = constant.instance_method(argument.keyword.to_s).source_location&.first
-            source_location && graphql_gem.contains_path?(source_location)
+            method_defined_by_graphql?(argument.keyword.to_s)
           end
+
           return if arguments.empty?
 
           root.create_path(constant) do |input_object|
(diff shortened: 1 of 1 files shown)


lib/tapioca/dsl/compilers/graphql_input_object.rb
```diff
@@ -45,7 +45,13 @@ class GraphqlInputObject < Compiler
...
         sig { override.void }
         def decorate
-          arguments = constant.all_argument_definitions
+          graphql_gem = T.must(Gemfile.new([]).gem('graphql'))
+
+          # Skip methods explicitly defined in code
+          arguments = constant.all_argument_definitions.select do |argument|
+            source_location = constant.instance_method(argument.keyword.to_s).source_location&.first
+            source_location && graphql_gem.contains_path?(source_location)
+          end
```
jeffcarbs, line 54: Open to suggestions on a better way to do this. Alternatives considered:

1. Not relying on the internal `Gemfile` class and just using bundler, something like:
```ruby
full_gem_path = Bundler.rubygems.find_name('graphql').first.full_gem_path
source_location && source_location.start_with?(full_gem_path)
```

2. Using `Gemfile::GemSpec.spec_lookup_by_file_path`, like:
```ruby
source_location && Gemfile::GemSpec.spec_lookup_by_file_path[source_location]&.name == 'graphql'
```

Morriar: I'm not familiar with `graphql`, how would it behave with subclassing?

Something like this:

```rb
class CreateCommentInput < GraphQL::Schema::InputObject
  argument :body, String, required: true
end

class CustomCreateCommentInput < CreateCommentInput
  def body; end
end
```

Could it mess up with the order of source locations?

jeffcarbs: We basically only care about typing the methods if they come from the graphql source itself (specifically this block of code: https://github.com/rmosolgo/graphql-ruby/blob/89854fa911d8243bb216910c370dbed780c8c178/lib/graphql/schema/input_object.rb#L133-L139). If the method is defined at all in user-land we don't want to define it in RBI. In your example, if `def body` were:
- not defined at all in user-land, it should show up in RBI
- defined in `CreateCommentInput`, it should **not** show up in RBI
- defined in `CustomCreateCommentInput`, it show **not** show up in RBI

so I don't think the subclassing would have a negative effect.

jeffcarbs: Looks great, thanks for cleaning that up!

paracycle (approved):

PR #1379, opened 2023-01-31
Add warning about changing the annotation header

andyw8:
### Motivation

See https://github.com/github/linguist/pull/6143/files


st0012 (approved):

andyw8:
This may becomes redundant if we implement the `.gitattributes` approach, but I think it's worth having for now.

https://github.com/Shopify/ruby-dev-exp-issues/issues/651#issuecomment-1411094224

PR #1385, opened 2023-02-07
Stop regenerating all value constants for reopened types

### LINKED ISSUE #1326: Better subconstant tracking when a gem reopens an external constant
Currently constants defined under the root namespace can be exposed in gem RBIs if the gem is reopening `Object`. This is particularly strange looking if the host application defines constants that are then visible in RBIs.

An example of this can be seen in Tapioca's gem RBIs for `nokogiri` where we're picking up a constant defined by another gem: https://github.com/Shopify/tapioca/blob/97271d62d7d57adf8d235ea825a05d6d7cb66aff/sorbet/rbi/gems/nokogiri%401.13.10.rbi#L8

We should try to filter out constants from the RBI based on their source location. We have a WIP branch that does some of this work using [`const_source_location`](https://ruby-doc.org/core-3.1.2/Module.html#const_source_location-method) but filtering isn't perfect yet: https://github.com/Shopify/tapioca/compare/uk-const-source-location

paracycle:
### Motivation

Fixes #1326 

### Implementation

When we are generating value constants (i.e. constants that have a value that is not a module), we didn't have a good way to filter them based on where they were defined, since we couldn't ask them before.

However, since Ruby 2.7, we have access to `const_source_location` so we can check to see if a value constant has been defined by the current gem or not. So the implementation uses that.

### Tests

Updated existing tests to test for this specific case where the `bar` gem is reopening a constant defined by the `foo` gem and adds a value constant under it. The expectation is that only the subconstant added in `bar` gem is generated in `bar.rbi` and none of the other subconstants defined in `foo`.

I also tested this against Core and it works without problems (and gets rid of a lot of unnecessary constant definitions on gem RBI files).

### DIFF
--- a/lib/tapioca/gem/pipeline.rb
+++ b/lib/tapioca/gem/pipeline.rb
@@ -106,6 +106,18 @@ module Tapioca
         @payload_symbols.include?(symbol_name)
       end
 
+      sig { params(name: T.any(String, Symbol)).returns(T::Boolean) }
+      def constant_in_gem?(name)
+        return true unless Object.respond_to?(:const_source_location)
+
+        source_location, _ = Object.const_source_location(name)
+        return true unless source_location
+        # If the source location of the constant is "(eval)", all bets are off.
+        return true if source_location == "(eval)"
+
+        gem.contains_path?(source_location)
+      end
+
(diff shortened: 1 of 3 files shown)


egiurleo (approved):
Nice!

PR #1387, opened 2023-02-08
Update AASM compiler to handle custom name

jeffcarbs:
### Motivation

When trying to replace our home-rolled DSL compiler for AASM, we realized the tapioca one does not handle using a custom column name (i.e. `aasm(:status) { }`). This fixes that.

### Implementation

When you call `SomeClass.aasm` you're actually implicitly calling `SomeClass.aasm(:default)` which tells AASM to create a state machine with the name `:default`. This means that if you define your state machine like `aasm(:something)`, it's saved under the name `:something` so when you call `aasm` (again, which is really `aasm(:default)`) you're going to see an "empty" state machine with no states or events because you're looking at a totally different state machine.

Here's the source: https://github.com/aasm/aasm/blob/2e952ff2088082c9f00c95e3c1261afc826f737c/lib/aasm/aasm.rb#L27-L64

This PR updates the compiler to look at all state machines for a given constant to gather the states and events for which to define methods.

### Tests

I added a failing spec, then in the second commit wrote the code to make it pass.


paracycle (approved):
Thanks!

PR #1390, opened 2023-02-09
Use Ruby 3.2 `Class#attached_object` method

egiurleo:
### Motivation
Partially thanks to our [work on Tapioca](https://bugs.ruby-lang.org/issues/12084#Use-Case-Introspection), Ruby 3.2 introduced the [`Class#attached_object` method](https://github.com/ruby/ruby/pull/6450), which fetches the attached object of a singleton class.

We are currently doing this in tapioca by _examining all of ObjectSpace_, which is not very elegant or performant.

### Implementation
This PR changes the implementation of `attached_class_of` in `Runtime::Reflection` to just call the `attached_object` method if it's defined on the class being passed in. If this method is not defined (because we're running Ruby 3.1 or older), then we fall back to the previous implementation.

### Tests
Existing tests pass with this change. I did some manual testing locally to ensure that the two branches of the method run on different Ruby versions.


lib/tapioca/runtime/reflection.rb
```diff
@@ -174,11 +174,15 @@ def resolve_loc(locations)
         resolved_loc.absolute_path || ""
       end
 
-      sig { params(singleton_class: Module).returns(T.nilable(Module)) }
+      sig { params(singleton_class: Class).returns(T.nilable(Module)) }
       def attached_class_of(singleton_class)
-        # https://stackoverflow.com/a/36622320/98634
-        result = ObjectSpace.each_object(singleton_class).find do |klass|
-          singleton_class_of(T.cast(klass, Module)) == singleton_class
+        result = if singleton_class.respond_to?(:attached_object)
```
paracycle, line 179: Can we move the check to top level, so that we don't have to keep doing the same check over and over again?

For example, if we define:
```ruby
CLASS_ATTACHED_OBJECT = Class.method_defined?(:attached_object)
```
outside this method, then we can just check that in here:
```suggestion
        result = if CLASS_ATTACHED_OBJECT
```

An even better way of doing it would be to change the method definition based on that top-level feature check like the following:
```ruby
CLASS_ATTACHED_OBJECT = Class.method_defined?(:attached_object)

if CLASS_ATTACHED_OBJECT
  def attached_class_of(singleton_class)
    singleton_class.attached_object
  end
else
  def attached_class_of(singleton_class)
    # Old definition here...
  end
end
```
since you only do the feature detection and the check for it once ever in the runtime of the program.

I am afraid that conditional method definition confuses Sorbet, though.

lib/tapioca/runtime/reflection.rb
```diff
@@ -174,14 +175,24 @@ def resolve_loc(locations)
...
 
-      sig { params(singleton_class: Module).returns(T.nilable(Module)) }
-      def attached_class_of(singleton_class)
-        # https://stackoverflow.com/a/36622320/98634
-        result = ObjectSpace.each_object(singleton_class).find do |klass|
-          singleton_class_of(T.cast(klass, Module)) == singleton_class
+      # On Ruby 3.2 or newer, Class defines an attached_object method that returns the
+      # attached class of a singleton class without iterating ObjectSpace. On older
+      # versions of Ruby, we fall back to iterating ObjectSpace.
+      if CLASS_ATTACHED_OBJECT
```
paracycle, line 181: Nice, I guess this is working. Now the linter is unhappy and I guess you have two choices:
1. Move the common sig out of the `if` block. I think it should work just fine with Sorbet, but not sure.
2. Suppress the linting rule for this part of the codebase.

lib/tapioca/runtime/reflection.rb
```diff
@@ -174,14 +175,24 @@ def resolve_loc(locations)
...
+        sig { params(singleton_class: Class).returns(T.nilable(Module)) }
+        def attached_class_of(singleton_class)
+          T.cast(singleton_class.attached_object, Module)
```
paracycle, line 184: So this code is not always going to work. The method name was chosen as `attached_object` for the reason that while a singleton class will always be a class, the object that has that singleton class does not need to be a module or a class. It can be a simple object. For example:
```ruby
foo = "foo"
sc = foo.singleton_class
Reflection.attached_class_of(sc)
```
should return `nil` since `sc` is not attached to a class, it is attached to a string object.

The fix is:
```suggestion
          result = singleton_class.attached_object
          if Module === result
            result
          else
            nil
          end
```

lib/tapioca/runtime/reflection.rb
```diff
@@ -174,14 +175,24 @@ def resolve_loc(locations)
...
+      else
+        sig { params(singleton_class: Class).returns(T.nilable(Module)) }
+        def attached_class_of(singleton_class)
+          # https://stackoverflow.com/a/36622320/98634
+          result = ObjectSpace.each_object(singleton_class).find do |klass|
+            singleton_class_of(T.cast(klass, Module)) == singleton_class
+          end
 
-        T.cast(result, Module)
+          T.cast(result, Module)
```
paracycle, line 194: This should be:
```suggestion
          T.cast(result, T.nilable(Module))
```
since there is a chance that `find` does not find a Module that has this as its singleton class.

I guess our previous code wasn't quite correct, it just worked since we were always giving it singleton classes of modules.

lib/tapioca/runtime/reflection.rb
egiurleo: Yeah Sorbet seems fine with it! I didn't realize you could do that.

paracycle: Oh, sorry, one more thing: Seen as the condition is being checked only once, WDYT about eliminating the constant and moving the condition inside the `if`:
```suggestion
      if Class.method_defined?(:attached_object)
```

lib/tapioca/runtime/reflection.rb
```diff
@@ -174,14 +175,25 @@ def resolve_loc(locations)
...
-      def attached_class_of(singleton_class)
-        # https://stackoverflow.com/a/36622320/98634
-        result = ObjectSpace.each_object(singleton_class).find do |klass|
-          singleton_class_of(T.cast(klass, Module)) == singleton_class
+      # On Ruby 3.2 or newer, Class defines an attached_object method that returns the
+      # attached class of a singleton class without iterating ObjectSpace. On older
+      # versions of Ruby, we fall back to iterating ObjectSpace.
+
+      sig { params(singleton_class: Class).returns(T.nilable(Module)) }
+      if CLASS_ATTACHED_OBJECT # rubocop:disable Sorbet/EmptyLineAfterSig
```
egiurleo, line 183: Rubocop thinks that this if statement is an empty line after the sig. Is there something I should do to fix this in a more legitimate way? Do you think the cop should be improved to anticipate a case like this?

egiurleo:
@Morriar I attempted your suggestion in commit 475f8847b50306d57687db2d2a28005a43ae0c26. Is this what you had in mind?

P.S. I'll clean up the commits once we're happy with an implementation, so sorry that they're messy for now.

paracycle:
> Yes. @paracycle what do you think of this approach?

I love it, but let's do all the requires top-level, get rid of the (now needless) `CLASS_ATTACHED_OBJECT` constant and leave only the `include` inside the `Reflection` module.

paracycle (approved):

bitwise-aiden (approved):

Morriar (approved):

lib/tapioca/runtime/reflection.rb
```diff
@@ -1,9 +1,20 @@
...
+if Class.method_defined?(:attached_object)
+  require "tapioca/runtime/attached_class_of_32"
+else
+  require "tapioca/runtime/attached_class_of_legacy"
+end
+
 module Tapioca
   module Runtime
     module Reflection
+      include AttachedClassOf
```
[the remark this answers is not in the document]
egiurleo: > Note that Bar is missing from the ancestors of Foo.singleton_class in Ruby 2.7, which is the root of the problem.
Whoah that's totally it!

PR #1398, opened 2023-02-20
Remove byebug

andyw8:
### Motivation

Prefer to use the official Ruby `debug` gem (already present). The last release of `byebug` was over 3 years ago.


st0012 (approved):
If anyone hasn't migrated to `debug` yet, I hope [this guide](https://st0012.dev/from-byebug-to-ruby-debug) would help (if not, please let me know).

paracycle:
Honestly I was still using `pry-byebug` and am sad to see it removed. Why can't we have both `debug` and `byebug` installed in dev mode?

st0012:
We are accelerating `byebug`'s deprecation (hence `pry-byebug` too) because we found that having `byebug` loaded actually breaks `debug` in some cases ([issue](https://github.com/Shopify/ruby-dev-exp-issues/issues/800)).

And this could also happen when debugging Tapioca tests because `byebug` is required in `spec_helper`. I know we can avoid this by simply not requiring `byebug` there. But to align with the deprecation effort, it'd be great if our projects can lead the migration first. That's why we're both 1) removing `byebug` and 2) setup `debug` integration with our projects (e.g. #1399).

So I hope we can give `debug` a try for a week or two. And if the inconvenience still out weights the extra features it provides, we can collect the feedback and then add `pry-byebug` back. WDYT?

PR #1408, opened 2023-02-22
More accurately type ActiveRecordRelations calculation methods

jeffcarbs:
### Motivation

Currently the calculation methods like `sum` and `count` return `T.untyped`. For some methods we cannot know the type (e.g. `minimum`) but for others we at least know it's numeric (e.g. `sum`) and for `count` we know it's an Integer.

However, I'm assuming these are untyped because if they're used with `group` the return value is actually a hash where the keys cannot be typed but the values will be the types mentioned above. This PR updates the `ActiveRecordRelations` compiler to type these correctly.

### Implementation

I first updated the return values of the calculation methods where possible. Then I largely copied from the code around `PrivateRelationWhereChain` to add a `PrivateRelationGroupChain` module which defines the calculation methods with the hash return values.

Some callouts:
- Unlike "WhereChain", there's no such analogous `GroupChain` runtime class available so this is solely made for RBI. From looking at the source it looks like when a query uses `group` some metadata is added to it.
- Technically a `group` anywhere in the query chain produces this behavior but to avoid needing to re-type every query method inside this module we make a simplifying assumption that the calculation method is called immediately after the group (e.g. `group().count` and not `group().where().count`). The one exception is `group().having().count` which is fairly idiomatic so that gets handled without breaking the chain.

### Tests

Updated the existing specs.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_relations.rb
+++ b/lib/tapioca/dsl/compilers/active_record_relations.rb
@@ -190,6 +190,8 @@ module Tapioca
             query_methods |= ActiveRecord::SpawnMethods.instance_methods(false)
             # Remove the ones we know are private API
             query_methods -= [:arel, :build_subquery, :construct_join_dependency, :extensions, :spawn]
+            # Remove "group" which needs a custom return type for GroupChains
+            query_methods -= [:group]
             # Remove "where" which needs a custom return type for WhereChains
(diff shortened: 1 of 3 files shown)


KaanOzkan:
@jeffcarbs sorry for the wait looks like the review automation failed. Could you rebase on `main`? Thanks.

Issue #1392, opened 2023-02-10
Tapioca generating actionmailer types for included helpers


tonybruess:
My actionmailers have helpers that are included and the resulting `.rbi` file is extraneous and even produces an error for one of the methods.

```
class BaseMailer < ActionMailer::Base
end

class PaymentMailer < BaseMailer
  include LogConcern
end

module LogConcern
  sig do
    type_parameters(:block_return).
      params(
        message: String,
        blk: T.proc.returns(T.type_parameter(:block_return)),
      ).
      returns(T.type_parameter(:block_return))
  end
  def log_around(message, &blk)
  end
end
```

The erroneous RBI file:

```
class OrderMailer::PaymentMailer
  class << self
    sig do
      params(
        message: ::String,
        blk: T.proc.returns(T.type_parameter(:block_return))
      ).returns(::ActionMailer::MessageDelivery)
    end
    def log_around(message, &blk); end
  end
end
```

Funny enough, `tapioca dsl` says there are no errors:

```
Checking generated RBI files...  Done
  No errors found

All operations performed in working directory.
Please review changes and commit them.
```

Sorbet disagrees:

```
sorbet/rbi/dsl/payment_mailer.rbi:65: Unspecified type parameter https://srb.help/5004
    65 |        blk: T.proc.returns(T.type_parameter(:block_return))
                                                     ^^^^^^^^^^^^^
```

Is there a way to exclude included modules? Any advice appreciated.

paracycle:
I think what Tapioca is doing here is correct, since any public instance method added to a mailer class automatically becomes a class method that is responsible for mail delivery. You can verify that by running the following script which is based exactly on your example:
```ruby
require "bundler/inline"

gemfile(false) do
  source "https://rubygems.org"
  gem "actionmailer", "~> 7.0.0"
end

require "action_mailer"
require "minitest/autorun"

module LogConcern
  def log_around(message, &blk)
  end
end

class BaseMailer < ActionMailer::Base
end

class PaymentMailer < BaseMailer
  include LogConcern
end

class MailerTest < Minitest::Test
  def test_stuff
    assert_instance_of(ActionMailer::MessageDelivery, PaymentMailer.log_around("foo"))
  end
end
```
and you will see:
```
$ ruby test.rb 
Run options: --seed 36681

# Running:

.

Finished in 0.002450s, 408.1633 runs/s, 408.1633 assertions/s.

1 runs, 1 assertions, 0 failures, 0 errors, 0 skips
[...]

PR #1409, opened 2023-02-22
Make it possible to patch ActionMailer compiler

jeffcarbs:
### Motivation

Give a workaround for #1392 where included helper methods are typed by tapioca as if they were mailer methods. As @paracycle points out in that thread, tapioca is _technically_ doing the right thing (see [the `action_methods` implementation](https://github.com/rails/rails/blob/26e21fc854a98e8c791fa3833fb7b333b217b18f/actionpack/lib/abstract_controller/base.rb#L81-L102)) although subjectively it may not always be what the consumer might want.

One suggestion was to use `helper Foo` instead of `include Foo` but that may break if `Foo` is a "concern" and uses `included do/end` that expected to be included in a `ActionMailer::Base` since `helper Foo` includes that module into a vanilla module. For example:
```ruby
module LogConcern
  extend ActiveSupport::Concern

  included do
    before_action { puts "hey" }
  end

  def log_around(message, &blk)
  end
end

class BaseMailer < ActionMailer::Base
end

class PaymentMailer < BaseMailer
  include LogConcern # Works
end

class PaymentMailer < BaseMailer
  helper LogConcern # undefined method `before_action' for PaymentMailer
end
```

### Implementation

I think there are a few options to address:
- ❌ [Not doing] The `ActionMailer` rbi compiler gets more opinionated and explicitly only generates mailer method RBI for methods defined by the mailer itself or any non-abstract parent class (i.e. not any included module ancestors).
  - While I don’t think I’ve ever come across actual mailer methods defined via an included module, it's possible that someone somewhere depends on this functionality. 
- ❌ [Not doing] Expose some configuration to make this something the consumer can configure.
  - There's not really a precedent in tapioca currently so not sure what that would look like. For example, would it be "exclude_method_from_included_moduled: true" or would there be some regex/list of modules/methods to exclude?
- ✅ **[This PR]** Extract out a helper method in `ActionMailer` compiler to fetch the action methods so it's easier to patch with some more opinionated logic by the consumer.
  - I think this is the most viable, at least in the short-term

The selected option would make it possible to:
```ruby
module Tapioca
  module Dsl
    module Compilers
      class ActionMailer
        module ActionMethodsPatch
          def action_methods_for_constant
            # my implementation
          end
        end
        prepend ActionMethodsPatch
      end
    end
  end
end
```

Happy to change options or open to other suggestions if the maintainers have any thoughts.

### Tests

This doesn't change any behavior, but I did backfill a few new tests to lock in the current behavior of the rbi compiler.

### DIFF
--- a/lib/tapioca/dsl/compilers/action_mailer.rb
+++ b/lib/tapioca/dsl/compilers/action_mailer.rb
@@ -41,7 +41,7 @@ module Tapioca
         sig { override.void }
         def decorate
           root.create_path(constant) do |mailer|
-            constant.action_methods.to_a.each do |mailer_method|
+            action_methods_for_constant.each do |mailer_method|
               method_def = constant.instance_method(mailer_method)
               parameters = compile_method_parameters_to_rbi(method_def)
               mailer.create_method(
@@ -61,6 +61,13 @@ module Tapioca
           def gather_constants
             descendants_of(::ActionMailer::Base).reject(&:abstract?)
           end
+        end
+
+        private
+
+        sig { returns(T::Array[String]) }
+        def action_methods_for_constant
+          constant.action_methods.to_a
         end
       end
     end


paracycle (approved):

paracycle:
Thanks this makes sense to me.

PR #1411, opened 2023-02-22
protobuf: Handle AbstractMessage descendants

dirceu:
### Motivation

This fixes the issue described in https://github.com/Shopify/tapioca/issues/1410.

`AbstractMessage` is [not new](https://github.com/protocolbuffers/protobuf/pull/10281) ~and I don't quite know yet why this is causing problems _now_ (and why it didn't seem to cause problems for other services)~—ah, it looks like they only added it to the most recent minor release.

### Implementation

We changed the `gather_constants` method of the Protobuf compiler to check if `Google::Protobuf::AbstractMessage` is defined:
- if it isn't, we preserve the old behaviour
- if it is, we skip that class but gather its descendants

This has been tested with the repository used to report this bug.

### DIFF
--- a/lib/tapioca/dsl/compilers/protobuf.rb
+++ b/lib/tapioca/dsl/compilers/protobuf.rb
@@ -145,7 +145,9 @@ module Tapioca
                   klass.create_method("initialize", parameters: [kwargs_parameter], return_type: "void")
                 end
               else
-                raise TypeError, "Unexpected descriptor class `#{descriptor.class.name}` for constant `#{constant}`"
+                add_error(<<~MSG.strip)
+                  Unexpected descriptor class `#{descriptor.class.name}` for `#{constant}`
+                MSG
               end
             end
           end
--- a/spec/tapioca/dsl/compilers/protobuf_spec.rb
+++ b/spec/tapioca/dsl/compilers/protobuf_spec.rb
@@ -501,6 +501,23 @@ module Tapioca
                 def contact_info; end
               RBI
             end
+
+            it "shows an error for an unexpected descriptor class" do
+              expect_dsl_compiler_errors!
+
+              add_ruby_file("protobuf.rb", <<~RUBY)
+                Cart = Class.new(::Google::Protobuf.const_get(:AbstractMessage))
+              RUBY
+
+              rbi_output = rbi_for(:Cart)
+
+              assert_equal(<<~RBI, rbi_output)
+                # typed: strong
+
+                class Cart; end
+              RBI
+              assert_equal(["Unexpected descriptor class `NilClass` for `Cart`"], generated_errors)
+            end
           end
         end
       end


lib/tapioca/dsl/compilers/protobuf.rb
```diff
@@ -91,6 +91,7 @@ def decorate
               create_type_members(klass, "Key", "Value")
             else
               descriptor = T.unsafe(constant).descriptor
+              next if descriptor.nil?
```
vinistock, line 94: NIT: this block is not a loop, so should we use `break` instead? Or alternatively, add `nil` to the `case` statement below and do nothing.

vinistock (commented):
Is it easy/possible to add a test for this so that we don't regress? I'm fine shipping the fix and figuring the test later if it's complicated.

lib/tapioca/dsl/compilers/protobuf.rb
rafaelfranca: I'd change the `gather_constants` to get subclasses of `Google::Protobuf::AbstractMessage` if it is defined instead of checking if the descriptor is `nil`. We should not generate the rbi for that class since it is already in the gem rbi.

paracycle: Yes, what @rafaelfranca said is the right call. If there is a new superclass for all protobuf classes, then all we need to do is to gather its descendants.

dirceu: Is there a "clean" way to get all descendants in this case? I can't do the following because `::Google::Protobuf::AbstractMessage` is a private constant:

```ruby
abstract_message_descendants = descendants_of(::Google::Protobuf::AbstractMessage)
```

dirceu: Thanks for the feedback! I've updated the DSL compiler to grab descendants of `::Google::Protobuf::AbstractMessage`, but ignore the constant itself (as its RBI is already in the gem RBI).

Not quite sure about how to test that yet. 👀

paracycle: You just need to add:
```ruby
              refute_includes(gather_constants, "Google::Protobuf::AbstractMessage")
```
inside the `gather_constants` tests.

lib/tapioca/dsl/compilers/protobuf.rb
```diff
@@ -145,7 +145,7 @@ def decorate
                   klass.create_method("initialize", parameters: [kwargs_parameter], return_type: "void")
                 end
               else
-                raise TypeError, "Unexpected descriptor class: #{descriptor.class.name}"
+                raise TypeError, "Unexpected descriptor class `#{descriptor.class.name}` for constant `#{constant}`"
```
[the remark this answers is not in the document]
paracycle: Yeah, after writing that, I was thinking about this, and I think that might be nice. We should also catch exceptions at the DSL pipeline layer, if any are raised, and turn them into displayed errors instead. Like I said above, in most cases, there is nothing the user can do with an exception, it just makes the experience worse. In general, some DSL RBI files, maybe with missing info, is still better than no DSL RBI files.

paracycle (approved):

PR #1412, opened 2023-02-23
Fix eager loading for DSL

jeffcarbs:
### Motivation

It appears that #1072 introduced a bug where the `eager_load:` option was not fully threaded through to the `Tapioca::Loaders::Dsl.new` so it was always falling back to the default value of `true`.

### Implementation

Pass the given argument to `Tapioca::Loaders::Dsl.new`

### Tests

There's no indication eager loading is/isn't used so there's not a straightforward way to test this specifically, but it's a one-line change that's pretty obvious.


paracycle (approved):
Ooops, thanks!

PR #1413, opened 2023-02-23
Fix DSL RBI missing PrivateRelation on abstract_class ActiveRecord

aericson:
### Motivation

In the current tapioca 0.11 scoped methods on abstract classes will end up with a return type of `PrivateRelation` and `PrivateAssociationRelation`. However, those will not exist in the DSL for those ActiveRecords as they are abstract classes.

This PR forces abstract classes to use `T.untyped` as the return type.

Please see the related PR for more context: https://github.com/Shopify/tapioca/pull/1250

### Implementation

Another option I looked into was to add `PrivateRelation` and `PrivateAssociationRelation` to the DSL of the model but I'm not sure that makes sense. I think this fix is a good way forward to at least for a patch  semver release.

### Tests

I've fixed the existing test. If my reasoning is correct then the test was asserting with the wrong return type.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_scope.rb
+++ b/lib/tapioca/dsl/compilers/active_record_scope.rb
@@ -54,30 +54,27 @@ module Tapioca
 
           return if method_names.empty?
 
-          root.create_path(constant) do |model|
-            relations_enabled = compiler_enabled?("ActiveRecordRelations") && !constant.abstract_class?
+          relation_methods_module = model.create_module(RelationMethodsModuleName)
+          create_relation_class
 
-            relation_methods_module = model.create_module(RelationMethodsModuleName)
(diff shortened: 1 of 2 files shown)


bmulholland:
Reporting in that I hit this issue with an abstract ApplicationRecord that has scopes on it, and using this branch fixed the issue for me. Thank you!

KaanOzkan (approved):

aericson:
> but I suspect we can do something better as well.

@paracycle giving it a try while I wait, added the following commit https://github.com/Shopify/tapioca/pull/1413/commits/825381c00c0962d8254d3617938d53bce5d2cd3e.
But happy to go in another direction after you check the runtime behaviour.

paracycle:
Sorry, it took me a while to get my mind around the solution space here and I think it is best to revert the original PR. I opened #1431 and tried to explain my reasoning for it.

Thanks for trying to fix this forward @aericson, I appreciate it.

PR #1414, opened 2023-02-24
Avoid raising when constantizing from paths

jeffcarbs:
### Motivation

#968 added support for generating DSL RBI via a path like `bin/tapioca dsl app/models`. The way it works is using `Static::SymbolLoader.symbols_from_paths` to find all constants defined in those paths and then acting as if you called `bin/tapioca dsl <those constants>`.

When testing this out for the first time, I realized that sometimes `symbols_from_paths` found constants that weren't actually loaded (or loadable) and would then raise an exception. Two common examples in our codebase:
1. A constant defined inside an eigenclass, for example
```ruby
class Foo
  class << self
    BAR = nil
  end
end
```
In this example, `Foo::BAR` is a constant returned from `symbols_from_paths` but you can't actually reference it outside the eigenclass.

2. Any constants (e.g. helper modules) defined in test folders. We use packwerk so my first command was something like `bin/tapioca dsl packs/xyz` which is a folder than contains both `app` and `spec` so it found a few constants from tests and blew up trying to load them.

### Implementation

When generating DSL for specific constants like `bin/tapioca dsl Some::Constant` we raise if that constant can't be resolved. This makes sense since you're explicitly asking us to generate DSL for a specific constant.

We previously had the same behavior for constants derived from given paths, however this PR changes that behavior so we'll simply ignore constants that cannot be resolved. This makes sense to me since by doing `bin/tapioca dsl some/path` you're essentially saying "generate RBI for whatever you can in `some/path`".

### Tests

Added a test for the eigenclass case and confirmed it failed before and passed after this change. I also tested out this branch of tapioca on Gusto's codebase and it worked as expected 👍

### DIFF
--- a/lib/tapioca/commands/dsl.rb
+++ b/lib/tapioca/commands/dsl.rb
@@ -93,7 +93,7 @@ module Tapioca
       def execute
         Loaders::Dsl.load_application(
           tapioca_path: @tapioca_path,
-          eager_load: @requested_constants.empty?,
+          eager_load: @requested_constants.empty? && @requested_paths.empty?,
           app_root: @app_root,
         )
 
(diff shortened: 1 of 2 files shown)


lib/tapioca/commands/dsl.rb
```diff
@@ -104,17 +104,10 @@ def execute
         end
         say("")
 
-        unless @requested_paths.empty?
-          constants_from_paths = Static::SymbolLoader.symbols_from_paths(@requested_paths).to_a
-          if constants_from_paths.empty?
-            say_error("\nWarning: No constants found in: #{@requested_paths.map(&:to_s).join(", ")}", :yellow)
```
jeffcarbs, line 110: This warning seemed unnecessary since we already have a warning if paths are given but no constants are found.

spec/tapioca/cli/dsl_spec.rb
```diff
@@ -804,8 +804,6 @@ class Post
           OUT
 
           assert_equal(<<~ERR, result.err)
-
-            Warning: No constants found in: path/to/nowhere.rb
```
jeffcarbs, line 808: I intentionally remove this warning since we already show an error if paths are given but no constants are found.

PR #1335, opened 2022-12-27
Annotate ActiveRecord::Batches methods

bdewater:
### Motivation
I wanted typing for the yielded values from the methods in  https://github.com/rails/rails/blob/main/activerecord/lib/active_record/relation/batches.rb

### Implementation
Mostly straightforward, return types are nilable in case the block form is used, eg:
```ruby
Post.find_each { |post| do_something(post) }
#=> nil

Post.find_each
#=> #<Enumerator: ...>
```

### Tests
Added tests.

PR #1417, opened 2023-02-26
in_batches/find_in_batches/find_each cannot return nil

razvan-sv:
### Motivation
* Based on https://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-in_batches, all 3 batch methods cannot return nil.

### Implementation
* Removed the `nilable` part. Tested it in our app and it works as it should.

### Tests
* Updated the tests

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_relations.rb
+++ b/lib/tapioca/dsl/compilers/active_record_relations.rb
@@ -699,7 +699,7 @@ module Tapioca
                   *(create_kw_opt_param("order", type: "Symbol", default: ":asc") if order),
                   create_block_param("block", type: "T.nilable(T.proc.params(object: #{constant_name}).void)"),
                 ],
-                return_type: "T.nilable(T::Enumerator[#{constant_name}])",
+                return_type: "T::Enumerator[#{constant_name}]",
               )
             when :find_in_batches
               order = ActiveRecord::Batches.instance_method(:find_in_batches).parameters.include?([:key, :order])
@@ -716,7 +716,7 @@ module Tapioca
                     type: "T.nilable(T.proc.params(object: T::Array[#{constant_name}]).void)",
                   ),
                 ],
-                return_type: "T.nilable(T::Enumerator[T::Enumerator[#{constant_name}]])",
+                return_type: "T::Enumerator[T::Enumerator[#{constant_name}]]",
               )
             when :in_batches
               order = ActiveRecord::Batches.instance_method(:in_batches).parameters.include?([:key, :order])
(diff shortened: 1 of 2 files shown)


aericson:
I think the reason for the `nilable` is that if they receive a block they will return `nil`. Could be wrong though.

razvan-sv:
Indeed, it's nil in that case, and it's weird. 

This will complain because you cannot do `each_with_index` on nil

```ruby
Person.in_batches.each_with_index do |relation, batch_index|
  puts "Processing relation ##{batch_index}"
  relation.delete_all
end
```

Also this:

```ruby
Person.in_batches.update_all(awesome: true)
```

aericson:
> Indeed, it's nil in that case, and it's weird.
> 
> This will complain because you cannot do `each_with_index` on nil. And you know it's not nil :)
> 
> ```ruby
> Person.in_batches.each_with_index do |relation, batch_index|
>   puts "Processing relation ##{batch_index}"
>   relation.delete_all
> end
> ```
> 
> Also this:
> 
> ```ruby
> Person.in_batches.update_all(awesome: true)
> ```

Yes, I commented here because I also ran into this and found your PR. The change in our codebase is to make it `T.must(Person.in_batches).each_with_index`. I believe the `nilable` was intentionally added since it's a valid return for those methods.

Not having it `nilable` would make it easier for my use-case but it's not the correct annotation. This for example should not be valid:

```ruby
Person.in_batches do
  puts "batch"
end.each_with_index
```

There is no overload signatures in sorbet so it has to support all the return types. But maybe someone more experienced than I has a different opinion on it.

razvan-sv:
Probably I'll have to `T.must` everywhere :)

bdewater:
Heya 👋  @aericson is right, it got collapsed by GitHub but I mentioned it in this comment as well: https://github.com/Shopify/tapioca/pull/1335#discussion_r1082092908

I wish we could type it [like Sorbet does](https://github.com/sorbet/sorbet/blob/4a0ff8dd2e1e3f4b5f52fcb9578e3c9cabd1870f/rbi/core/array.rbi#L1182-L1189) to get something like:
```ruby
sig { params(start: T.untyped, finish: T.untyped, batch_size: Integer, error_on_ignore: T.untyped, order: Symbol).returns(T::Enumerator[::Post]) }
sig { params(start: T.untyped, finish: T.untyped, batch_size: Integer, error_on_ignore: T.untyped, order: Symbol, block: T.proc.params(object: ::Post).void).void }
def find_each(start: nil, finish: nil, batch_size: 1000, error_on_ignore: nil, order: :asc, &block); end
```

bdewater:
> Indeed, it's nil in that case, and it's weird.

This is because ultimately all these methods use `in_batches` and [`loop` here](https://github.com/rails/rails/blob/0eaa58ef3f1690538e0efc672e931824479d15c6/activerecord/lib/active_record/relation/batches.rb#L236) returns nil.

razvan-sv:
@bdewater @aericson I'll close this and try something else and see if it works. Thank you !

PR #1418, opened 2023-02-27
Only run pending migrations check if necessary

jeffcarbs:
### Motivation

A point of friction we've experienced is getting blocked from running DSL generation due to pending migration even when we're not generating RBI for ActiveRecord models (e.g. when doing something like `bin/tapioca dsl Some::Sidekiq::Worker Some::Graphql::Mutation`) . The pending migration check is valuable to ensure we don't generate ActiveRecord RBI with a stale schema, but it's irrelevant when generating RBI for things like sidekiq workers, graphql mutations, or other non-ActiveRecord things.

### Implementation

Since we want to conditionalize running the pending migrations check, we need to be at a point where we have the constants we're going to process in memory.  This means we need to pull the pending migrations check out from the `Loaders::Dsl` further down into the `Dsl::Pipeline`. We then check to see if any of the constants we're going to process subclass ActiveRecord::Base and only check for pending migrations then.

This handles both the case where you don't give specific classes (`bin/tapioca dsl`) or you give some (`bin/tapioca dsl Something`) and there are test cases for both scenarios.

A small side-benefit here is that previously even a command like `bin/tapiocs dsl --list-compilers` would fail due to pending migrations since that check happened when loading rails. Since it now only runs if necessary, that command would succeed.

### Tests

I refactored the existing pending migration test to reuse some pieces (see first commit) and then added some specific test cases for this.

### DIFF
--- a/lib/tapioca/dsl/pipeline.rb
+++ b/lib/tapioca/dsl/pipeline.rb
@@ -68,6 +68,10 @@ module Tapioca
           ERROR
         end
 
+        if defined?(::ActiveRecord::Base) && constants_to_process.any? { |c| ::ActiveRecord::Base > c }
+          abort_if_pending_migrations!
+        end
+
         result = Executor.new(
           constants_to_process,
           number_of_workers: @number_of_workers,
@@ -183,6 +187,16 @@ module Tapioca
         handler = error_handler
         handler.call(error)
         exit(1)
+      end
+
+      sig { void }
+      def abort_if_pending_migrations!
+        return unless defined?(::Rake)
+
+        Rails.application.load_tasks
+        if Rake::Task.task_defined?("db:abort_if_pending_migrations")
+          Rake::Task["db:abort_if_pending_migrations"].invoke
+        end
       end
     end
   end
--- a/lib/tapioca/loaders/dsl.rb
+++ b/lib/tapioca/loaders/dsl.rb
@@ -20,7 +20,6 @@ module Tapioca
       def load
         load_dsl_extensions
         load_application
-        abort_if_pending_migrations!
         load_dsl_compilers
       end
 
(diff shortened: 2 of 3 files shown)


lib/tapioca/dsl/pipeline.rb
```diff
@@ -184,6 +188,16 @@ def report_error(error)
         handler.call(error)
         exit(1)
       end
+
+      sig { void }
+      def abort_if_pending_migrations!
+        return unless defined?(::Rake)
```
vinistock, line 194: I think you may not even need this one now that we're further down the path.

dirceu:
@jeffcarbs thanks for reporting! I believe we found the problem; I'll ping this PR to let you know to rebase once we've fixed it.

lib/tapioca/dsl/pipeline.rb
jeffcarbs: I've removed this line as well 👍

lib/tapioca/loaders/dsl.rb
```diff
@@ -70,17 +69,6 @@ def load_application
 
         say("Done", :green)
       end
-
-      sig { void }
-      def abort_if_pending_migrations!
-        return unless File.exist?("#{@app_root}/config/application.rb")
-        return unless defined?(::Rake)
```
jeffcarbs, line 77: Removed this line as well as per https://github.com/Shopify/tapioca/pull/1418#discussion_r1119142551

jeffcarbs:
Thanks @dirceu, looks like it's working now!

@vinistock - I've put the changes back and the build is green so this should be good to go :shipit:

lib/tapioca/dsl/pipeline.rb
```diff
@@ -68,6 +68,10 @@ def run(&blk)
           ERROR
         end
 
+        if defined?(::ActiveRecord::Base) && constants_to_process.any? { |c| c < ::ActiveRecord::Base }
```
paracycle, line 71: Small comment here: We've learned the hard way that it is not safe to call methods like `<` on arbitrary constants since they might have overloaded the method (here's looking at you `XPath`). Can we please rewrite this so that the call is being made on `ActiveRecord::Base` instead?
```suggestion
        if defined?(::ActiveRecord::Base) && constants_to_process.any? { |c| ::ActiveRecord::Base > c  }
```

paracycle (commented):
I have a small comment about safety in the presence of arbitrary constants, otherwise, this is great work. Thanks for the PR!

egiurleo (approved):

PR #1423, opened 2023-02-28
Fix typing for the getter of serialized columns

wpolicarpo:
### Motivation

According to [this table](https://github.com/Shopify/tapioca/pull/1194#discussion_r983870226), serialized columns return an empty Array or Hash if their coders are either an Array or Hash respectively.

Currently tapioca assumes the getter will always returns a nilable type, which is not true:

<details>
<summary>Toggle me for an example!</summary>

```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "activerecord", "7.0.4.2"
  gem "sqlite3"
end

require "active_record"
require "minitest/autorun"
require "logger"

ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")
ActiveRecord::Base.logger = Logger.new(STDOUT)

ActiveRecord::Schema.define do
  create_table :books, force: true do |t|
    t.string :authors
  end
end

class Book < ActiveRecord::Base
  serialize :authors, Array
end

class BookTest < Minitest::Test
  def test_populated_array
    book = Book.new(authors: ["foo", "bar"])
    assert_equal(["foo", "bar"], book.authors)
  end

  def test_empty_array
    book = Book.new(authors: [])
    assert_equal([], book.authors)
  end

  def test_nil
    book = Book.new(authors: nil)
    assert_equal([], book.authors)
  end
end

```

</details>

### Implementation

This PR changes the getters return type for serialized columns where the coder is one of those exceptional classes.

### Tests

Added tests for all getters, making sure the Array and Hash ones check for a not nilable type being returned.


lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
```diff
@@ -148,6 +147,23 @@ def serialized_column_type(column_type)
...
+          when ActiveRecord::Coders::YAMLColumn
+            case column_type.coder.object_class
+            when Array.singleton_class, Hash.singleton_class
+              getter_type
+            else
+              as_nilable_type(getter_type)
+            end
+          else
+            as_nilable_type(getter_type)
+          end
```
KaanOzkan, line 165: What do you think of this to remove the outer case?

```suggestion
          return as_nilable_type(getter_type) unless column_type.is_a?(ActiveRecord::Type::Serialized) &&
            column_type.coder.is_a?(ActiveRecord::Coders::YAMLColumn)

          case column_type.coder.object_class
          when Array.singleton_class, Hash.singleton_class
            getter_type
          else
            as_nilable_type(getter_type)
          end
```

lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
```diff
@@ -148,6 +147,23 @@ def serialized_column_type(column_type)
             "T.untyped"
           end
         end
+
+        sig { params(column_type: T.untyped, getter_type: String).returns(String) }
+        def maybe_nilable_type(column_type, getter_type)
```
KaanOzkan, line 152: I don't love this method name, curious about any alternatives. For example, `as_nilable_column_type`?

wpolicarpo: Another option could be define something like this:

```ruby
sig { params(column_type: T.untyped).returns(T::Boolean) }
def not_nilable_serialized_column?(column_type)
  column_type.is_a?(ActiveRecord::Type::Serialized) &&
    column_type.coder.is_a?(ActiveRecord::Coders::YAMLColumn) &&
    [Array.singleton_class, Hash.singleton_class].include?(column_type.coder.object_class.singleton_class)
end
```

And change `type_for`:

```ruby
if column&.null
  getter_type = as_nilable_type(getter_type) unless not_nilable_serialized_column?(column_type)
  return [getter_type, as_nilable_type(setter_type)]
end
```

What do you think?

wpolicarpo: I've changed the implementation.

Issue #1345, opened 2023-01-04
YARD documentation is not idempotent with multiple constant reopenings


KaanOzkan:
This occurs when dealing with gem RBIs that reopen same constants as each other. We generate the RBIs, include yard documentation in some of the RBIs but not in all. Later when RBIs are regenerated the yard documentation is removed from some and added in others, creating a larger diff and making it harder to review.

For example running `bin/tapioca gem --all` in tapioca removed this documentation on `Object` from `pry.rbi`: https://github.com/Shopify/tapioca/commit/4de76ec524eee7a3dc05a6763c3b19c175bae60e#diff-ea313bde23ea3de4be53945fd1bd09c0a67ddd77bc7d514daba0c0acfacca606L32-L50 and moved it to `rainbow.rbi`: https://github.com/Shopify/tapioca/commit/4de76ec524eee7a3dc05a6763c3b19c175bae60e#diff-bc9be23e44cbb85f07d0912b28b6253b5c9badf1c865dc024210d32e212a3b4eR16-R33.

I'm seeing this on a larger scale with `opentelemetry-instrumentation-*` [gems](https://rubygems.org/search?query=opentelemetry-instrumentation).

Maybe we can ensure the documentation is added for every reopening or at least not move it around.

PR #1426, opened 2023-03-03
Make YARD documentation idempotent

KaanOzkan:
### Motivation
Resolves https://github.com/Shopify/tapioca/issues/1345

**Problem**: `YARD.parse` stores parsed documentation into a registry that's persisted. So depending on the parsing order of the gems documentation may change. If a previous gem has documentation for a constant/method definition it'll be included in the current gem's documentation too. However, in another run if that previous gem is parsed too late its documentation won't be available. 

### Implementation

In this PR we reset the YARD registry everytime we parse a gem's documentation to ensure only the documentation from that gem is included in the output. This means parsing order has no effect in the RBIs created and that documentation generation is idempotent.

As a result of this change duplicated RBI entries will not contain as much documentation as before. For example, when a gem defines a method on the global namespace tapioca would define `class Object` on the gem RBI. If a previously processed gem had documentation for `Object` it'd get imported into that gem, [example](https://github.com/Shopify/tapioca/blob/6d7f9a776f37afc01000379e0a1682fb999b66f6/sorbet/rbi/gems/rainbow%403.1.1.rbi#L17-L33). In some cases like the example the documentation didn't even make sense. It's coming from [Rails](https://github.com/rails/rails/blob/7cf65bc335bf1cd49cf94462ddbd8ae1adda6025/activesupport/lib/active_support/core_ext/object/duplicable.rb#L4-L19).

We initially tried to keep the documentation for each gem by parsing the documentation of all gems before any RBI generation. However, this increases the runtime since parsing isn't done in parallel anymore. 

### Tests
Included

### DIFF
--- a/lib/tapioca/gem/listeners/yard_doc.rb
+++ b/lib/tapioca/gem/listeners/yard_doc.rb
@@ -25,6 +25,7 @@ module Tapioca
 
         sig { params(pipeline: Pipeline).void }
         def initialize(pipeline)
+          YARD::Registry.clear
           super(pipeline)
           pipeline.gem.parse_yard_docs
         end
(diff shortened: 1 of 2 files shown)


paracycle (approved):
Nice catch!

egiurleo (approved):
🎉

PR #1448, opened 2023-03-21
Remove support for old Bundler API in materialize_deps

KaanOzkan:
### Motivation

This piece of code was brittle and causes an error in the latest bundler version.

### Implementation

Instead of supporting different method arity's which could keep changing in the future we can remove support for the old Bundler API. [v2.2.25](https://github.com/rubygems/rubygems/releases/tag/bundler-v2.2.25) has been out since July 2021. Users with old versions of Bundler will not be able to use future Tapioca versions.

### Tests

Locally tested using `BUNDLER_VERSION=2.4.9 bundle exec tapioca gem --verify` and bumped the version used in tapioca to trigger CI run with latest bundler.

### DIFF
--- a/lib/tapioca/gemfile.rb
+++ b/lib/tapioca/gemfile.rb
@@ -105,14 +105,13 @@ module Tapioca
     sig { returns([T::Enumerable[Spec], T::Array[String]]) }
     def materialize_deps
       deps = definition.locked_gems.dependencies.values
-      missing_specs = T::Array[String].new
-      materialized_dependencies = if definition.resolve.method(:materialize).arity == 1 # Support bundler >= v2.2.25
-        md = definition.resolve.materialize(deps)
-        missing_spec_names = md.missing_specs.map(&:name)
-        missing_specs = T.cast(md.missing_specs.map { |spec| "#{spec.name} (#{spec.version})" }, T::Array[String])
-        md.to_a.reject { |spec| missing_spec_names.include?(spec.name) }
-      else
-        definition.resolve.materialize(deps, missing_specs)
+      materialized_dependencies = definition.resolve.materialize(deps)
+      missing_spec_names = materialized_dependencies.missing_specs.map(&:name).to_set
+      missing_specs = materialized_dependencies.missing_specs.map do |spec|
+        "#{spec.name} (#{spec.version})"
+      end
+      materialized_dependencies = materialized_dependencies.to_a.reject do |spec|
+        missing_spec_names.include?(spec.name)
       end
       [materialized_dependencies, missing_specs]
     end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -9,7 +9,7 @@ PATH
   remote: .
   specs:
     tapioca (0.11.2)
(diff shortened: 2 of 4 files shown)


lib/tapioca/gemfile.rb
```diff
@@ -105,15 +105,15 @@ def load_dependencies
...
-        definition.resolve.materialize(deps, missing_specs)
-      end
+      materialized_dependencies = definition.resolve.materialize(deps)
+      missing_spec_names = materialized_dependencies.missing_specs.map(&:name)
+      missing_specs = T.cast(
+        materialized_dependencies.missing_specs.map do |spec|
+          "#{spec.name} (#{spec.version})"
+        end,
+        T::Array[String],
+      )
```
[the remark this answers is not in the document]
paracycle: Not sure if it is still necessary, but it would be great if we could remove it.

Morriar (approved):

lib/tapioca/gemfile.rb
```diff
@@ -105,15 +105,12 @@ def load_dependencies
...
-        missing_specs = T.cast(md.missing_specs.map { |spec| "#{spec.name} (#{spec.version})" }, T::Array[String])
-        md.to_a.reject { |spec| missing_spec_names.include?(spec.name) }
-      else
-        definition.resolve.materialize(deps, missing_specs)
+      materialized_dependencies = definition.resolve.materialize(deps)
+      missing_spec_names = materialized_dependencies.missing_specs.map(&:name)
+      missing_specs = materialized_dependencies.missing_specs.map do |spec|
+        "#{spec.name} (#{spec.version})"
       end
+      materialized_dependencies.to_a.reject { |spec| missing_spec_names.include?(spec.name) }
```
paracycle, line 113: Ah, hold on, this code is not correct. It should be:
```suggestion
      materialized_dependencies.to_a.reject! { |spec| missing_spec_names.include?(spec.name) }
```

paracycle (changes requested):

lib/tapioca/gemfile.rb
```diff
@@ -105,14 +105,13 @@ def load_dependencies
...
-      missing_specs = T::Array[String].new
-      materialized_dependencies = if definition.resolve.method(:materialize).arity == 1 # Support bundler >= v2.2.25
-        md = definition.resolve.materialize(deps)
-        missing_spec_names = md.missing_specs.map(&:name)
-        missing_specs = T.cast(md.missing_specs.map { |spec| "#{spec.name} (#{spec.version})" }, T::Array[String])
-        md.to_a.reject { |spec| missing_spec_names.include?(spec.name) }
-      else
-        definition.resolve.materialize(deps, missing_specs)
+      materialized_dependencies = definition.resolve.materialize(deps)
+      missing_spec_names = materialized_dependencies.missing_specs.map(&:name)
```
paracycle, line 109: Nitpick: Can we make this a set, so that the `include?` operation below does not end up being quadratic:
```suggestion
      missing_spec_names = materialized_dependencies.missing_specs.map(&:name).to_set
```

PR #1459, opened 2023-03-30
Bump Spoom to v1.2.0

Morriar:
This requires two changes:

1. `Spoom.srb` was removed upstream, instead we run the `srb` command inside a `Spoom::Context`.
2. To solve 1. I initialized the context as a constant in `sorbet_helper` which is required in a specific way when running the DSL tests. This means we need to require Spoom earlier in this case.

### DIFF
--- a/lib/tapioca/dsl.rb
+++ b/lib/tapioca/dsl.rb
@@ -2,6 +2,7 @@
 # frozen_string_literal: true
 
 require "rbi"
+require "spoom"
 
 require "tapioca"
 require "tapioca/runtime/reflection"


tapioca.gemspec
```diff
@@ -28,7 +28,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency("parallel", ">= 1.21.0")
   spec.add_dependency("rbi", "~> 0.0.0", ">= 0.0.16")
   spec.add_dependency("sorbet-static-and-runtime", ">= 0.5.10187")
-  spec.add_dependency("spoom", "~> 1.1.0", ">= 1.1.11")
+  spec.add_dependency("spoom", "~> 1.2.0")
```
paracycle, line 31: Can we still leave the patch version lower-bound, so that it is helpful for future bumps?
```suggestion
  spec.add_dependency("spoom", "~> 1.2.0", ">= 1.2.0")
```

paracycle (approved):

PR #1460, opened 2023-03-30
Add TapiocaWeakFindEach compiler configuration option

tonybruess:
### Motivation
Users are confused by the return type (https://github.com/Shopify/tapioca/pull/1417) for `find_each` etc and it would be nice to configure the return type as nilable or not.

I spoke with Ufuk on Slack and he said the team is open to more configurability in the base compilers.

### Implementation
I originally looked into adding a configuration option to the CLI / YAML file but the compilers architecture is not designed to support such configuration. There is precedent for using specially named modules to allow users to configure the DSL compilers: https://github.com/shopify/tapioca/blob/main/lib/tapioca/dsl/helpers/active_record_column_type_helper.rb#L84

While originally I thought YAML was a better configuration option, I actually think using specially named modules to configure the compiler will provide users with much more flexibility.

### Tests
I want feedback on the change before I invest time in writing tests. I am happy to write specs if the response is positive.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_relations.rb
+++ b/lib/tapioca/dsl/compilers/active_record_relations.rb
@@ -689,6 +689,7 @@ module Tapioca
             case method_name
             when :find_each
               order = ActiveRecord::Batches.instance_method(:find_each).parameters.include?([:key, :order])
+              return_type = "T::Enumerator[#{constant_name}]"
               create_common_method(
                 "find_each",
                 parameters: [
@@ -699,10 +700,11 @@ module Tapioca
                   *(create_kw_opt_param("order", type: "Symbol", default: ":asc") if order),
                   create_block_param("block", type: "T.nilable(T.proc.params(object: #{constant_name}).void)"),
                 ],
-                return_type: "T.nilable(T::Enumerator[#{constant_name}])",
+                return_type: weak_find_each?(constant_name) ? as_nilable_type(return_type) : return_type,
               )
             when :find_in_batches
               order = ActiveRecord::Batches.instance_method(:find_in_batches).parameters.include?([:key, :order])
+              return_type = "T::Enumerator[T::Enumerator[#{constant_name}]]"
               create_common_method(
                 "find_in_batches",
                 parameters: [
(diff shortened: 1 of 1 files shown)


tonybruess:
Hi team, can I get some feedback on this PR please? It's been patiently sitting here for 3 weeks now :)

lib/tapioca/dsl/compilers/active_record_relations.rb
```diff
@@ -699,10 +700,11 @@ def create_common_methods
                   *(create_kw_opt_param("order", type: "Symbol", default: ":asc") if order),
                   create_block_param("block", type: "T.nilable(T.proc.params(object: #{constant_name}).void)"),
                 ],
-                return_type: "T.nilable(T::Enumerator[#{constant_name}])",
+                return_type: weak_find_each?(constant_name) ? as_nilable_type(return_type) : return_type,
```
andyw8, line 703: It seems this is changing the default behaviour - can we flip it to be the other way?

lib/tapioca/dsl/compilers/active_record_relations.rb
```diff
@@ -813,6 +816,12 @@ def create_relation_method(
             return_type: association_return_type,
           )
         end
+
+        sig { params(constant: Module).returns(T::Boolean) }
+        def weak_find_each?(constant)
+          Object.const_defined?(:TapiocaWeakFindEach) &&
```
andyw8, line 822: Let's ensure this is documented. I don't see any current place for `StrongTypeGeneration` so I'd suggest a new section in the README, e.g. `Configuration Constants`.

andyw8 (commented):
Thanks @tonybruess, the team are generally in favour of making this change. Please see the suggested changes above. We would also expect there to be tests.

tonybruess:
@andyw8 Thanks so much for the feedback. I'll get this cleaned up with the appropriate changes, documentation and tests.

Are you/the team sure we don't want to prefix or namespace these configuration classes? It feels like it would be nice if maybe they lived in a namespace e.g. `TapiocaConfig::WeakFindEach` or something.

lib/tapioca/dsl/compilers/active_record_relations.rb
rafaelfranca: I don't think we should be using this approach for this. Let's make sure we can have real configuration without having to define modules in application and include them in our classes.

PR #1468, opened 2023-04-06
feat: consistently define `fetch_multi_by`

cribbles:
### Motivation

Following https://github.com/Shopify/identity_cache/pull/534, IdentityCache will define `fetch_multi_by` for `cache_index` invocations regardless of the number of keys included in the index. (Previously only single-index keys were supported.)

We should lift the `field_length` restriction accordingly.

### Implementation
Removes the `field_length == 1` restriction (as well as the `field_length` variable designation, which is now useless).

### Tests
~Nothing changed.~

Updated a unit test implicitly testing this behavior in https://github.com/Shopify/tapioca/pull/1468/commits/64983d4af1eaaf4b383e1506c617e0cfbac0bbf9.

### DIFF
--- a/lib/tapioca/dsl/compilers/identity_cache.rb
+++ b/lib/tapioca/dsl/compilers/identity_cache.rb
@@ -172,7 +172,6 @@ module Tapioca
           ).void
         end
         def create_index_fetch_by_methods(field, klass)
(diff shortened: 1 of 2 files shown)


paracycle (approved):
Thank you!

PR #1521, opened 2023-05-30
Update minimum supported Sorbet version

andyw8:
### Motivation

As per internal discussion, we want to limit support for older versions of Sorbet.

This relates to https://github.com/Shopify/tapioca/pull/1511. We aren't updating to as recent a version as mentioned there, but this helps to move us in the direction of only supporting newer versions.

### Implementation

Update to a minimum Sorbet version from ~6 months ago.

If we go ahead with this, we should set up some kind of reminder/automation to periodically update the minimum supported version.

### Tests

n/a

### DIFF
--- a/lib/tapioca/helpers/sorbet_helper.rb
+++ b/lib/tapioca/helpers/sorbet_helper.rb
@@ -24,7 +24,6 @@ module Tapioca
     FEATURE_REQUIREMENTS = T.let(
       {
         # feature_name: ::Gem::Requirement.new(">= ___"), # https://github.com/sorbet/sorbet/pull/___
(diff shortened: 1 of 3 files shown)


lib/tapioca/helpers/sorbet_helper.rb
```diff
@@ -24,7 +24,6 @@ module SorbetHelper
     FEATURE_REQUIREMENTS = T.let(
       {
         # feature_name: ::Gem::Requirement.new(">= ___"), # https://github.com/sorbet/sorbet/pull/___
-        non_generic_weak_map: ::Gem::Requirement.new(">= 0.5.10587"), # https://github.com/sorbet/sorbet/pull/6610
```
paracycle, line 27: If we are bumping to `>= 0.5.10566`, we can't remove this feature check since it requires `>= 0.5.10587`.

PR #1525, opened 2023-06-01
Exclude tapioca dependencies during gem RBI tests

KaanOzkan:


KaanOzkan:
Mean CI time seems to be around 11m with the whole pipeline being 16m https://github.com/Shopify/tapioca/actions/runs/5148501154/jobs/9270265509.

In comparison to current main having mean ~15m and the whole pipeline finishing in 24m.

Locally I saw 3 minutes of improvement.

spec/helpers/mock_project.rb
```diff
@@ -119,14 +119,37 @@ def bundle_exec(command, env = {})
...
+    end
+    def tapioca(command, enforce_typechecking: true, exclude_tapioca_dependencies: true)
       exec_command = ["tapioca", command]
       if command.start_with?(/gem/)
         exec_command << "--workers=1" unless command.match?("--workers")
         exec_command << "--no-doc" unless command.match?("--doc")
         exec_command << "--no-loc" unless command.match?("--loc")
+        if exclude_tapioca_dependencies && !command.match?("--exclude")
+          exec_command << "--exclude #{TAPIOCA_DEPENDENCIES.join(" ")}"
+        end
```
KaanOzkan, line 152: I didn't want to complicate this but ideally we'd append the arguments to `exclude` flag with the `TAPIOCA_DEPENDENCIES` and enable this speedup on tests that use `exclude` too.

spec/helpers/mock_project.rb
```diff
@@ -119,14 +119,37 @@ def bundle_exec(command, env = {})
...
-    sig { params(command: String, enforce_typechecking: T::Boolean).returns(ExecResult) }
-    def tapioca(command, enforce_typechecking: true)
+    sig do
+      params(
+        command: String,
+        enforce_typechecking: T::Boolean,
+        exclude_tapioca_dependencies: T::Boolean,
+      ).returns(ExecResult)
+    end
+    def tapioca(command, enforce_typechecking: true, exclude_tapioca_dependencies: true)
```
Morriar, line 144: What do you think about passing the list of dependencies to ignore and allowing a default value?

```suggestion
    def tapioca(command, enforce_typechecking: true, exclude: TAPIOCA_DEPENDENCIES)
```

KaanOzkan: I see the appeal but isn't this kind of strange given other flags aren't a keyword argument? I preferred the explicit intention of the boolean but don't have strong opinions.

Morriar: If we're allowing controlling the behavior of the exclude on top of the options already passed in the command string, we may as well pass the values to be used so the same approach can be reused in other tests that may need a special treatment.

Issue #1484, opened 2023-04-18
Create DSL compiler for adding signatures to controller actions


vinistock:
When trying to use strict level of typing, Sorbet will require us to add signatures to controller action methods. These signatures don't really add much value. If we can add a DSL compiler that adds the signatures to RBIs, it would allow us to have more strict typing typing in controllers.

For example
```ruby
class PostsController < ApplicationController
  # To use strict here, we need to `extend T::Sig` and add `sig { void }` to all these actions
  def index
    @posts = Post.find_posts
  end
end
```

### Notes

We need to be careful with methods marked as `helper_method`, which may be public but aren't actions. Those methods should still be annotated manually.

PR #1526, opened 2023-06-02
Compiler for controller actions

yathi:
### Motivation

This is trying to resolve https://github.com/Shopify/tapioca/issues/1484, where we want signatures for controller actions.  

### Implementation

Wrote a compiler with the help of @vinistock to create a `sig { void }` for all public controller actions.
  
Things to look out for: Is there something in Shopify core or another big project that might break because of this?  

### Tests

There are some tests added and dsl docs generated.

### DIFF
--- a/lib/tapioca/dsl/compilers/controller_actions.rb
+++ b/lib/tapioca/dsl/compilers/controller_actions.rb
@@ -1,0 +1,75 @@
+# typed: strict
+# frozen_string_literal: true
+
+begin
+  require "action_controller"
+rescue LoadError
+  return
+end
+
+module Tapioca
+  module Dsl
+    module Compilers
+      # `Tapioca::Dsl::Compilers::ControllerActions` generates RBI files for classes that include
+      # [`ActionController::Base`](https://api.rubyonrails.org/classes/ActionController/Base.html).
+      #
+      # For example, with the following class that includes `ActionController::Base`:
+      #
+      # ~~~rb
+      # class PostsController < ApplicationController
+      #  def index
+      #   # ...
+      #  end
+      #
+      #  def create
+      #    # ...
+      #  end
+      # end
+      # ~~~
+      #
+      # this compiler will produce the RBI file `posts_controller.rbi` with the following content:
+      #
+      # ~~~rbi
+      # # posts_controller.rbi
+      # # typed: strong
+      # class PostsController
+      #   sig { void }
+      #   def index; end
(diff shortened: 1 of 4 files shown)


lib/tapioca/dsl/compilers/controller_actions.rb
```diff
@@ -0,0 +1,75 @@
...
+      # end
+      # ~~~
+      class ControllerActions < Compiler
+        extend T::Sig
+
+        ConstantType = type_member { { fixed: T.class_of(::ActionController::Base) } }
+
+        sig { override.void }
+        def decorate
+          public_methods = constant.instance_methods(false)
```
KaanOzkan, line 50: This returns `protected` methods too and [I don't think we want that](https://guides.rubyonrails.org/action_controller_overview.html#:~:text=Only%20public%20methods%20are%20callable%20as%20actions.%20It%20is%20a%20best%20practice%20to%20lower%20the%20visibility%20of%20methods%20(with%20private%20or%20protected)%20which%20are%20not%20intended%20to%20be%20actions%2C%20like%20auxiliary%20methods%20or%20filters.).

KaanOzkan: Can we also add a test with `protected` methods?

lib/tapioca/dsl/compilers/controller_actions.rb
```diff
@@ -0,0 +1,75 @@
...
+      # end
+      # ~~~
+      class ControllerActions < Compiler
+        extend T::Sig
+
+        ConstantType = type_member { { fixed: T.class_of(::ActionController::Base) } }
+
+        sig { override.void }
+        def decorate
+          public_methods = constant.public_instance_methods(false)
```
paracycle, line 50: Actually, I think we want to use `action_methods` instead, since it is what Rails uses internally: https://api.rubyonrails.org/v7.0.0/classes/AbstractController/Base.html#method-c-action_methods

yathi:
Fixed the linter issue. Maybe needs approval again?

vinistock:
We had some internal discussions and proposing this compiler has actually been a mistake in my understanding. Since controller actions are discoverable statically and you can just add signatures to them, creating a DSL compiler just for the convenience is not the intended use case. DSL compilers should instead be focused on code that isn't discoverable statically.

Sorry for the wrong suggestion. Thank you for exploring it!

PR #1529, opened 2023-06-09
Limit `AutoRequireHook` effect to be localized to loading the bundle

### LINKED ISSUE #1528: The `AutoRequireHook` patch should only apply to `tapioca gem` command
Currently, the [patch](https://github.com/Shopify/tapioca/blob/main/lib/tapioca/gemfile.rb#L15-L18) is applied globally, which causes commands like `tapioca dsl` not respecting the `require: false` definition in Gemfile when loading Rails applications. 

And because this is not intentional, the [gem exclusion list](https://github.com/Shopify/tapioca/blob/main/lib/tapioca/gemfile.rb#L154-L164) for `tapioca gem` is not applied in such situations and causes gems like `debug` to be loaded during DSL generation and introducing unwanted side-effects.

paracycle:
### Motivation

Fix #1528 

### Implementation

Move `AutoRequireHook` to a separate file, and make it enabled only while executing a block passed to a singleton method on the module named `override_require_false`. The `AutoRequireHook` still always gets prepended to `Bundler::Dependency` but its behaviour in `autorequire` is to return very early if the module is not enabled.

### Tests

Added a test that fails DSL compilation for a mock gem that is marked `require: false` and raises from its default require. The test passes after the fix.

### DIFF
--- a/lib/tapioca/bundler_ext/auto_require_hook.rb
+++ b/lib/tapioca/bundler_ext/auto_require_hook.rb
@@ -13,22 +13,41 @@ module Tapioca
       requires_ancestor { ::Bundler::Dependency }
 
       @exclude = T.let([], T::Array[String])
+      @enabled = T.let(false, T::Boolean)
 
       class << self
         extend T::Sig
 
-        sig { params(exclude: T::Array[String]).returns(T::Array[String]) }
-        attr_writer :exclude
-
         sig { params(name: T.untyped).returns(T::Boolean) }
         def excluded?(name)
           @exclude.include?(name)
+        end
+
+        def enabled?
+          @enabled
+        end
+
+        sig do
+          type_parameters(:Result).params(
+            exclude: T::Array[String],
+            blk: T.proc.returns(T.type_parameter(:Result)),
+          ).returns(T.type_parameter(:Result))
+        end
+        def override_require_false(exclude:, &blk)
+          @enabled = true
+          @exclude = exclude
+          blk.call
+        ensure
+          @enabled = false
         end
       end
 
       sig { returns(T.untyped).checked(:never) }
       def autorequire
         value = super
+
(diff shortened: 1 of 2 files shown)


lib/tapioca/bundler_ext/auto_require_hook.rb
```diff
@@ -0,0 +1,69 @@
...
+        def override_require_false(exclude:, &blk)
+          @enabled = true
+          @exclude = exclude
+          blk.call
+        ensure
+          @enabled = false
+        end
+      end
+
+      sig { returns(T.untyped).checked(:never) }
```
st0012, line 45: TIL: `checked`

lib/tapioca/gemfile.rb
```diff
@@ -67,11 +25,14 @@ def autorequire
...
     def initialize(exclude)
-      AutoRequireHook.exclude = exclude
       @gemfile = T.let(File.new(Bundler.default_gemfile), File)
       @lockfile = T.let(File.new(Bundler.default_lockfile), File)
       @definition = T.let(Bundler::Dsl.evaluate(gemfile, lockfile, {}), Bundler::Definition)
-      dependencies, missing_specs = load_dependencies
+
+      dependencies, missing_specs = BundlerExt::AutoRequireHook.override_require_false(exclude: exclude) do
+        load_dependencies
+      end
```
Morriar, line 34: What do you think about moving the `override_require_false` inside `load_dependencies`?

Morriar (approved):

lib/tapioca/gemfile.rb
paracycle: Ah good call! It also turns out that the call should be made inside `require_bundle` call anyway. 🙈

Morriar (approved):

PR #1548, opened 2023-06-29
Add support for `--include-dependencies` flag

LTe:
### Motivation

In my project we have huge Gemfile with a lot of gems. We tried to type some of the files, but we do not need all of the gems to be generated. We tried to use `tapioca gems [name_of_the_gem]` but it was generating only this gem (without dependencies). We tried `exclude` but list was huge.

I think we should have `--include-dependencies` flag that generate RBIs for dependencies as well

### Implementation
I added new flag for CLI that will pick gems and dependencies from the list.

### Tests
I added integration tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,6 +171,7 @@ Options:
   --post, -a,  [--postrequire=file]                                   # A file to be required after Bundler.require is called
                                                                       # Default: sorbet/tapioca/require.rb
   -x,          [--exclude=gem [gem ...]]                              # Exclude the given gem(s) from RBI generation
+               [--include-dependencies], [--no-include-dependencies]  # Generate RBI files for dependencies of the given gem(s)
   --typed, -t, [--typed-overrides=gem:level [gem:level ...]]          # Override for typed sigils for generated gem RBIs
                                                                       # Default: {"activesupport"=>"false"}
                [--verify], [--no-verify]                              # Verify RBIs are up-to-date
@@ -199,7 +200,7 @@ generate RBIs from gems
 ```
 <!-- END_HELP_COMMAND_GEM -->
 
-By default, running `tapioca gem` will only generate the RBI files for gems that have been added to or removed from the project's `Gemfile` this means that Tapioca will not regenerate the RBI files for untouched gems. However, when changing Tapioca configuration or bumping its version, it may be useful to force the regeneration of the RBI files previously generated. This can be done with the `--all` option:
+By default, running `tapioca gem` will only generate the RBI files for gems that have been added to or removed from the project's `Gemfile` this means that Tapioca will not regenerate the RBI files for untouched gems. If you want to force the regeneration you can supply gem names to the `tapioca gem` command. When supplying gem names if you want to generate RBI files for their dependencies as well, you can use the `--include-dependencies` option. When changing Tapioca configuration or bumping its version, it may be useful to force the regeneration of all the RBI files previously generated. This can be done with the `--all` option:
 
 ```shell
 bin/tapioca gems --all
@@ -904,6 +905,7 @@ gem:
   prerequire: ''
   postrequire: sorbet/tapioca/require.rb
   exclude: []
+  include_dependencies: false
   typed_overrides:
     activesupport: 'false'
   verify: false
--- a/lib/tapioca/cli.rb
+++ b/lib/tapioca/cli.rb
@@ -198,6 +198,10 @@ module Tapioca
       banner: "gem [gem ...]",
       desc: "Exclude the given gem(s) from RBI generation",
       default: []
+    option :include_dependencies,
+      type: :boolean,
+      desc: "Generate RBI files for dependencies of the given gem(s)",
+      default: false
     option :typed_overrides,
       aliases: ["--typed", "-t"],
       type: :hash,
@@ -251,10 +255,14 @@ module Tapioca
 
       all = options[:all]
       verify = options[:verify]
+      include_dependencies = options[:include_dependencies]
 
       raise MalformattedArgumentError, "Options '--all' and '--verify' are mutually exclusive" if all && verify
 
-      unless gems.empty?
+      if gems.empty?
+        raise MalformattedArgumentError,
+          "Option '--include-dependencies' must be provided with gems" if include_dependencies
+      else
         raise MalformattedArgumentError, "Option '--all' must be provided without any other arguments" if all
         raise MalformattedArgumentError, "Option '--verify' must be provided without any other arguments" if verify
       end
@@ -262,6 +270,7 @@ module Tapioca
       command_args = {
         gem_names: all ? [] : gems,
         exclude: options[:exclude],
+        include_dependencies: options[:include_dependencies],
         prerequire: options[:prerequire],
         postrequire: options[:postrequire],
         typed_overrides: options[:typed_overrides],
--- a/lib/tapioca/commands/abstract_gem.rb
+++ b/lib/tapioca/commands/abstract_gem.rb
@@ -13,6 +13,7 @@ module Tapioca
         params(
           gem_names: T::Array[String],
           exclude: T::Array[String],
+          include_dependencies: T::Boolean,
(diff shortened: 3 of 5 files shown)


LTe:
I think there is a couple reasons why I want this kind of feature in tapioca.

Currently my Gemfile.lock include about ~630 gems

|       | Without **only** | With **only** |
| --    | --               | --            |
| Time  | 447,81s          | 30,18s        |
| Size  | 1,4M             | 22M           |
| Count | ~560             | ~20            |
| `wc -l rbi/todo.rbi`     | 80 | 4 |

As you can see there is significant difference in performance.

Do I really need 560 generated RBIs? Most of the missing constants inside `todo.rbi` are from internal gems and this does not influence typechecking in my application.

I want this feature just for flexibility, to have control what I will generate for my application. I can of course do similar thing with external script, but I think this feature should be included in tapioca.

To be honest 560 files where `541` is not really needed can be hard to sell other team members. I think that reducing the number of files (at least on the beginning of adoption) generated by 'tapioca' may increase the number of projects that will use `tapioca` and `sorbet`.

So in summary, I want this feature in tapioca because of:

* performance
* flexibility
* in order to more easily convince the unconvinced to adopt a sorbet

KaanOzkan:
This is an interesting approach but I'm still not convinced that this is a good way of using Tapioca in the long term: Generating RBIs only for gem constants that are referenced in the application. What you want with this PR can be achieved with supplying arguments to `tapioca gem`, only difference is this PR will figure out the direct dependencies for you. I think we have to discuss as a team if that's an acceptable use case to recommend through this flag.

Biggest issue I have with this setup is the need for manual tapioca config changes every time the application starts using a new gem or starts referencing new part of a gem, it's an extra manual step to be done by the user. Knowing which gem to load can also get complex if the gem's RBIs changes conditionally based on the existence of another gem. Users will need to inspect gem's code.

Side note: why does this feature need to generate RBIs for sub-dependencies that aren't referenced by the application if you're using a `todo.rbi`? Is it because some gems have different RBIs if their sub-dependencies are loaded?

We have a similar number of gems and in practice we rarely run `bin/tapioca gem --all`, as a result performance isn't a big concern. We update individual RBIs upon version bumps and run `bin/tapioca gem --all` from a scheduled build once a week. Maybe a change in workflow would help performance concerns. We don't run `tapioca gem --verify` on CI, just `dsl --verify` and `srb tc`.

KaanOzkan:
We discussed this internally and while `--only` flag doesn't seem too beneficial for rest of the community we could add a `--include-dependencies` option for the default `tapioca gem` usage, ie. `bin/tapioca gem foo bar baz --include-dependencies`. This will support your use case and can be useful to avoid regenerating every outdated RBI using `tapioca gem` in certain situations such as dependabot PRs.

What do you think about switching the implementation to that? Biggest difference would be to iterate recursively over all dependencies and keeping track of them in a set instead of limiting to one level. We'll also have to ensure this can't be used alongside `--all`.

@paracycle I'm curious if you have opinions on this approach.

LTe:
@KaanOzkan I updated pull request and implemented `--include-dependencies` flag.

README.md
```diff
@@ -288,6 +290,20 @@ There are a few development/test environment gems that can cause RBI generation
...
 
+#### Regenerate the RBI files gem(s) explicitly
+
+In case you only want to generate RBI files for certain gems, you can use the gem command.
+
+```shell
+$ bin/tapioca gem gemA gemB
+```
+
+If you want the RBI files to be generated for dependencies as well, use the flag `--include-dependencies`:
```
KaanOzkan, line 301:
```suggestion
When supplying gem names if you want to generate RBI files for their dependencies as well, you can use the `--include-dependencies` option:
```

Something like this is better to connect it to gem names.

lib/tapioca/cli.rb
```diff
@@ -257,6 +261,7 @@ def gem(*gems)
       command_args = {
```
KaanOzkan, line 261: We need a check right above this line to raise `MalformattedArgumentError` if `--include-dependencies` is used without supplying any `gems`.

lib/tapioca/commands/abstract_gem.rb
```diff
@@ -74,15 +77,22 @@ def initialize(
       def gems_to_generate(gem_names)
         return @bundle.dependencies if gem_names.empty?
 
-        gem_names.map do |gem_name|
+        gem_candidates(gem_names).map do |gem_name|
```
KaanOzkan, line 80: No strong opinions but I prefer being explicit in the call site to have easier readability in the future.

```ruby
gem_names = gems_and_their_dependencies(gem_names) if @include_dependencies
gem_names.map do |gem_name|
  ...
end
```

lib/tapioca/commands/abstract_gem.rb
```diff
@@ -252,6 +262,26 @@ def expected_rbis
...
 
+      sig do
+        params(
+          gem_name: String,
+          resolved_dependencies: T::Array[Tapioca::Gemfile::GemSpec],
+        ).returns(T::Array[Tapioca::Gemfile::GemSpec])
+      end
+      def gem_with_dependencies(gem_name, resolved_dependencies = [])
+        gem = @bundle.gem(gem_name)
+        return [] if gem.nil?
```
KaanOzkan, line 273: I think we want to return the `gem_name` in this case so that we can raise on line 84.

lib/tapioca/commands/abstract_gem.rb
```diff
@@ -252,6 +262,26 @@ def expected_rbis
...
+
+        gem_dependencies = gem.dependencies.map { |gem_spec| @bundle.gem(gem_spec.name) }.compact
```
KaanOzkan, line 275: This method was tough to understand at first. Maybe renaming variables could help? `gem_dependencies` could be `direct_dependencies` and `resolved_dependencies` could be `gem_names`?

LTe:
@KaanOzkan I updated the code and documentation.

Do you think that this PR can be related to https://github.com/Shopify/tapioca/issues/1576?

KaanOzkan (approved):

KaanOzkan:
> Do you think that this PR can be related to https://github.com/Shopify/tapioca/issues/1576?

Good shout! I think `--include-dependencies` is a good workaround for most cases of what the issue describes. I don't know if it can solve all the occurrences in practice because RBI dependency is not same as gem dependencies. Gem `X` can  load different things based on existence of gem `Y` without depending on `Y`. But this flag would solve the example issue I described. Because of that maybe it's worth enabling this by default. I'm curious what the team thinks although we don't run into https://github.com/Shopify/tapioca/issues/1576 often and it could be resolved by a more sophisticated tracking solution. For this PR I think we can keep the default as false and re-consider after using this flag, it will also unblock you faster.

lib/tapioca/commands/abstract_gem.rb
```diff
@@ -74,6 +77,7 @@ def initialize(
       def gems_to_generate(gem_names)
         return @bundle.dependencies if gem_names.empty?
 
+        gem_names = gems_and_their_dependencies(gem_names) if @include_dependencies
```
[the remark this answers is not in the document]
LTe: I will try to simplify things

lib/tapioca/commands/abstract_gem.rb
```diff
@@ -74,14 +77,32 @@ def initialize(
...
+      end
+
+      sig do
+        params(
+          gem: Gemfile::GemSpec,
+          dependencies: T::Array[Gemfile::GemSpec],
+        ).returns(T::Array[Gemfile::GemSpec])
+      end
+      def gem_dependencies(gem, dependencies = [])
+        direct_dependencies = gem.dependencies.map { |dependency| @bundle.gem(dependency.name) }.compact
```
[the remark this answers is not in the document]
vinistock:
```suggestion
        direct_dependencies = gem.dependencies.filter_map { |dependency| @bundle.gem(dependency.name) }
```

vinistock (approved):

Issue #1563, opened 2023-07-11
Error `Kernel is already included by Tapioca::Helpers::Test::DslCompiler`


andyw8:
When upgrading Core to v0.11.7, the updated RBI fails the typecheck:

```
sorbet/rbi/gems/tapioca@0.11.7.rbi:2189: Kernel is already included by Tapioca::Helpers::Test::DslCompiler
    2189 |  requires_ancestor { Kernel }
```

Looking at `tapioca@0.11.7.rbi` I can see this is because `Tapioca::Helpers::Test::DslCompiler` includes `::Tapioca::Helpers::Test::Isolation` conditionally includes  `Tapioca::Helpers::Test::Isolation::Forking` or `Tapioca::Helpers::Test::Isolation::SubProcess`. Both of these include `Kernel`, so the `requires_ancestor` call results in a failure.

This may relate to https://github.com/Shopify/tapioca/pull/1530.

cc @paracycle

PR #1564, opened 2023-07-11
Remove extraneous `requires_ancestor` for `Kernel` on `DslCompiler`

andyw8:
In `DslCompiler`, we include `Tapioca::Helpers::Test::Isolation` which conditionally includes  `Tapioca::Helpers::Test::Isolation::Forking` or `Tapioca::Helpers::Test::Isolation::SubProcess`. Both of these include `Kernel`, so the `requires_ancestor` call results in a failure:

`Kernel is already included by Tapioca::Helpers::Test::DslCompiler`

Closes https://github.com/Shopify/tapioca/issues/1563

### Tests

No automated tests but I top-hatted against the Core PR for upgrading to v0.11.7.

### DIFF
--- a/lib/tapioca/helpers/test/isolation.rb
+++ b/lib/tapioca/helpers/test/isolation.rb
@@ -29,7 +29,9 @@ module Tapioca
 
         module Forking
           extend T::Sig
-          include Kernel
+          extend T::Helpers
+
+          requires_ancestor { Kernel }
 
           sig { params(_blk: T.untyped).returns(String) }
           def run_in_isolation(&_blk)
@@ -72,7 +74,10 @@ module Tapioca
 
         module Subprocess
           extend T::Sig
-          include Kernel
+          extend T::Helpers
+
+          requires_ancestor { Kernel }
+
           ORIG_ARGV = T.let(ARGV.dup, T::Array[T.untyped]) unless defined?(ORIG_ARGV)
 
           # Crazy H4X to get this working in windows / jruby with


KaanOzkan (commented):
I think we need to solve the root problem which is generating `include ::Kernel` when there is a `requires_ancestor { Kernel } ` call.

If that is too complex I'm okay with this PR for now to unblock users.

Morriar:
> I think we need to solve the root problem

I don't think there is actually a root cause here. Both modules _do_ include `Kernel` (see https://github.com/Shopify/tapioca/blob/main/lib/tapioca/helpers/test/isolation.rb#L32) and since https://github.com/Shopify/tapioca/pull/1530 we properly generate the include 

The interesting part is that we do not have this error locally in Tapioca, because the include is conditional so Sorbet doesn't see it. But we get it with RBIs in other projects because Tapioca finds the include through reflection. But I think it's okay.

On the other hand, I'm not sure that those two modules should indeed include `Kernel`. It seems required only to call `exit` so Maybe we should change the calls as `Kernel.exit` or replace the includes by `requires_ancestor { Kernel }` in both of them.

Morriar:
> would you be happy with this being merged @Morriar?

Not really as we would need to put it back once the issue is correctly fixed.

I took the liberty to push the change to your branch. Can you check if it fixes the RBI issue properly please?

andyw8:
If you can approve this then I will make a new Tapioca release.

PR #1565, opened 2023-07-12
Add a json_api_client DSL compiler

### LINKED ISSUE #1554: DSL generator for json_api_client
json_api_client supports schemas: https://github.com/JsonApiClient/json_api_client#schema

```rb
class User < JsonApiClient::Resource
  property :name, type: :string
  property :is_admin, type: :boolean, default: false
  property :points_accrued, type: :int, default: 0
  property :averge_points_per_day, type: :float
end
```

We should be able to infer typed `attr_accessor` from those.

lavoiesl:
### Motivation

Fixes #1554 

This could perhaps be implemented in https://github.com/JsonApiClient/json_api_client as a tapioca extension, but it's worth noting that the gem is [used in several places at Shopify](https://github.com/search?q=org%3AShopify+path%3AGemfile+json_api_client&type=code) and is the spec used by the Vault API, namely used in [iHub](https://ihub.docs.shopify.io/clients/vault).

### Implementation
The implementation was largely cargo-culted from the `smart-properties` compiler.

I paired with @Morriar

See comments for some of the design decisions.


lib/tapioca/dsl/compilers/json_api_client_resource.rb
```diff
@@ -0,0 +1,208 @@
...
+        end
+
+        private
+
+        sig { returns(T.nilable(::JsonApiClient::Schema)) }
+        def resource_schema
+          schema = constant.schema
+
+          # empty? does not exist on JsonApiClient::Schema
+          schema if schema.size > 0 # rubocop:disable Style/ZeroLengthPredicate
```
lavoiesl, line 132: Disabled rubocop rule because it was a false positive

lib/tapioca/dsl/compilers/json_api_client_resource.rb
```diff
@@ -0,0 +1,208 @@
...
+            "::Float"
+          elsif type == ::JsonApiClient::Schema::Types::Time
+            "::Time"
+          elsif type == ::JsonApiClient::Schema::Types::Decimal
+            "::BigDecimal"
+          elsif type == ::JsonApiClient::Schema::Types::Boolean
+            "T::Boolean"
+          else
+            "T.untyped"
+          end
```
lavoiesl, line 171: I couldn't find a cleaner way to do this `case`. I'm happy to consider suggestions

lib/tapioca/dsl/compilers/json_api_client_resource.rb
```diff
@@ -0,0 +1,208 @@
...
+
+        sig do
+          params(
+            mod: RBI::Scope,
+            association: JsonApiClient::Associations::BaseAssociation,
+          ).void
+        end
+        def generate_methods_for_association(mod, association)
+          # If the association is broken, it will raise a NameError when trying to access the association_class
+          klass = association.association_class
```
lavoiesl, line 188: There is a test for this

lib/tapioca/dsl/compilers/json_api_client_resource.rb
```diff
@@ -0,0 +1,208 @@
...
+
+          name, type = case association
+          when ::JsonApiClient::Associations::BelongsTo::Association
+            # id must be a string: # https://jsonapi.org/format/#document-resource-object-identification
+            [association.param.to_s, "T.nilable(::String)"]
+          when ::JsonApiClient::Associations::HasOne::Association
+            [association.attr_name.to_s, "T.nilable(#{klass})"]
+          when ::JsonApiClient::Associations::HasMany::Association
+            [association.attr_name.to_s, "T.nilable(T::Array[#{klass}])"]
```
lavoiesl, line 197: Unfortunately, all of those things can be nullable when an object is created without being loaded via the API

lib/tapioca/dsl/compilers/json_api_client_resource.rb
```diff
@@ -0,0 +1,208 @@
...
+          elsif type == ::JsonApiClient::Schema::Types::Decimal
+            "::BigDecimal"
+          elsif type == ::JsonApiClient::Schema::Types::Boolean
+            "T::Boolean"
+          else
+            "T.untyped"
+          end
+
+          if property.default.nil?
+            as_nilable_type(sorbet_type)
```
lavoiesl, line 174: Properties with default values are always initialized with their default value, even when creating a bare object.

We made the decision to make those properties non-nilable to make the usage cleaner.

That being said, it _is_ possible today to have a nil:

```rb
class User < JsonApiClient::Resource
  property :name, type: :string, default: "foo"
end

> u = User.new
=> #<User:@attributes={"type"=>"users", "name"=>"foo"}>

> u.name
=> "foo"

> u.name = 123
=> 123

> u.name # Casts to string
=> "123"

> u.name = nil

> u.name # Actually nil
=> nil
```

KaanOzkan (commented):
Don't have context on `json_api_client` but the compiler mostly LGTM.

I'm not sure if it should be a default compiler inside tapioca though 😕 I'll discuss with the team.

lib/tapioca/dsl/compilers/json_api_client_resource.rb
lavoiesl: This:
```rb
case obj
when Class
end
```

Converts to:
```rb
if obj.is_a?(Class)
```

Which isn't the same operation. Am I missing something?

KaanOzkan: Isn't the only difference inheritance checks? I don't think that's a problem in this example.

KaanOzkan (approved):

lib/tapioca/dsl/compilers/json_api_client_resource.rb
```diff
@@ -0,0 +1,208 @@
...
+        end
+
+        sig do
+          params(
+            mod: RBI::Scope,
+            association: JsonApiClient::Associations::BaseAssociation,
+          ).void
+        end
+        def generate_methods_for_association(mod, association)
+          # If the association is broken, it will raise a NameError when trying to access the association_class
```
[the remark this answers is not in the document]
paracycle: We have a way for compilers to signal errors that we use in AR associations compiler, for example: https://github.com/Shopify/tapioca/blob/bc6805edeffbaaba384111c1db29bd79b4b93795/lib/tapioca/dsl/compilers/active_record_associations.rb#L168-L176

PR #1567, opened 2023-07-12
Drop Ruby 2.7 support

andyw8:
Ruby 2.7 has been EOL since 31 Mar 2023.

### DIFF
--- a/lib/tapioca/sorbet_ext/generic_name_patch.rb
+++ b/lib/tapioca/sorbet_ext/generic_name_patch.rb
@@ -191,19 +191,6 @@ module Tapioca
     sig { returns(T.nilable(String)) }
     def name
       constant_name = super
-
-      # This is a hack to work around modules under anonymous modules not having
-      # names in 2.7: https://bugs.ruby-lang.org/issues/14895
-      #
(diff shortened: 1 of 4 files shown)


lib/tapioca/sorbet_ext/generic_name_patch.rb
```diff
@@ -191,19 +191,6 @@ def initialize(context, type, variance, fixed, lower, upper, bounds_proc)
     sig { returns(T.nilable(String)) }
     def name
       constant_name = super
-
```
andyw8, line 194: Found by grepping for `2.7`.

KaanOzkan (commented):
I believe [gemspec](https://github.com/Shopify/tapioca/blob/50027c9f85f6bbc107dbd093e3c17cae1391d348/tapioca.gemspec#L35) and [rubocop config](https://github.com/Shopify/tapioca/blob/50027c9f85f6bbc107dbd093e3c17cae1391d348/.rubocop.yml#L11) needs to be updated.

paracycle (approved):

Morriar (approved):

PR #1568, opened 2023-07-15
Add from and to parameters for _changed? generated methods

DougEdey:
### Motivation

`ActiveModel::Dirty` takes parameters for `_changed?` [generated methods](https://api.rubyonrails.org/classes/ActiveModel/Dirty.html), `from:` and `to:`, these aren't supported by the generator right now and we use them in our app a bit so I want to add support.

I didn't find an explicit compiler for `Dirty` but the `ActiveRecordColumns` compiler says it implements the `ActiveModel::Dirty` methods from various locations.

I checked and `ActiveRecord::TypedStore` doesn't [support parameters](https://github.com/byroot/activerecord-typedstore/blob/0dae8a971e4c347a90085cf3fac9730d37165e50/lib/active_record/typed_store/extension.rb#L39)

### Implementation

I went with adding an additional `kw_parameters` argument for the `create_method` call to try and ensure there's a pattern if needed in more places in the future.

### Tests

I updated existing tests that check this behaviour


DougEdey:
I have signed the CLA!

bitwise-aiden (approved):

lib/tapioca/dsl/compilers/active_record_columns.rb
```diff
@@ -153,12 +153,18 @@ def gather_constants
             methods_to_add: T.nilable(T::Array[String]),
             return_type: String,
             parameters: T::Array[[String, String]],
+            kw_parameters: T::Array[[String, String, String]],
           ).void
         end
-        def add_method(klass, name, methods_to_add, return_type: "void", parameters: [])
+        def add_method(klass, name, methods_to_add, return_type: "void", parameters: [], kw_parameters: [])
```
paracycle, line 159: Can we instead make the `parameters` parameter take a `T::Array[RBI::Param]` instead? I see just a single use of this method that supplies a parameter already, and it can easily supply `create_param(value, type: setter_type)` instead.

That way, there wouldn't be any split or extra processing for kw parameters.

lib/tapioca/dsl/compilers/active_record_columns.rb
```diff
@@ -254,6 +260,7 @@ def add_methods_for_attribute(klass, column_name, attribute_name = column_name,
             "#{attribute_name}_changed?",
             methods_to_add,
             return_type: "T::Boolean",
+            kw_parameters: [["from", setter_type, "T.unsafe(nil)"], ["to", setter_type, "T.unsafe(nil)"]],
```
paracycle, line 263: Based on my suggestion above, this would be:
```suggestion
            parameters: [
              create_kw_opt_param("from", type: setter_type, default: "T.unsafe(nil)"),
              create_kw_opt_param("to", type: setter_type, default: "T.unsafe(nil)")
            ],
```

lib/tapioca/dsl/compilers/active_record_columns.rb
```diff
@@ -277,6 +284,7 @@ def add_methods_for_attribute(klass, column_name, attribute_name = column_name,
             "#{attribute_name}_previously_changed?",
             methods_to_add,
             return_type: "T::Boolean",
+            kw_parameters: [["from", setter_type, "T.unsafe(nil)"], ["to", setter_type, "T.unsafe(nil)"]],
```
paracycle, line 287: Same as above.

lib/tapioca/dsl/compilers/active_record_columns.rb
DougEdey: Sure, only minor tweak is that it has to be a `T::Array[RBI::TypedParam]`, I'm not sure if that would be wildly/unexpectedly different from what you expected

paracycle (approved):

lib/tapioca/dsl/compilers/active_record_columns.rb
paracycle: Yeap, that's exactly what I meant to write actually. :) Thanks for doing the change.

PR #1571, opened 2023-07-19
Improve GraphQL Mutation compiler

paracycle:
### Motivation

We were generating `T.untyped` for some of the cases we could handle, which became apparent with the recent change to the mutation compiler to support argument with loads in #1556.

/cc @such

### Implementation

This PR does a few things:
1. It extends the `GraphqlTypeHelper#type_for` method to handle `NonNull` types and types that are just plain module references.
2. It simplifies the argument loads type mapping code.
3. It refactors, updates and slightly extends the existing tests.

### Tests

Updated tests


Tiedye:
This generates incorrect types for custom scalars

amomchilov:
See also: https://github.com/Shopify/tapioca/pull/1659

PR #1577, opened 2023-07-21
Refactor `Command` classes to make `run` method the public API

paracycle:
### Motivation

Finally getting around to https://github.com/Shopify/tapioca/pull/1218#pullrequestreview-1134759632 comment by @Morriar.

By making the abstract `execute` method be a private method, we can ensure that all commands are invoked via the `run` method, which allows us to wrap them in `Tapioca.silence_warnings` today and potentially do more in the future.

### Implementation

Add `Command#run` method that calls `Command#execute` and make `Command#execute` a private method. Encapsulate all `Tapioca.silence_warnings` calls.

### Tests

There should be no behaviour change.

### DIFF
--- a/lib/tapioca/commands/command.rb
+++ b/lib/tapioca/commands/command.rb
@@ -21,10 +21,17 @@ module Tapioca
         @file_writer = T.let(FileWriter.new, Thor::Actions)
       end
 
-      sig { abstract.void }
-      def execute; end
+      sig { void }
+      def run
+        Tapioca.silence_warnings do
+          execute
+        end
+      end
 
       private
+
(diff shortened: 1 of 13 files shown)


lib/tapioca/commands/command.rb
```diff
@@ -21,11 +21,18 @@ def initialize
         @file_writer = T.let(FileWriter.new, Thor::Actions)
       end
 
-      sig { abstract.void }
-      def execute; end
+      sig { void }
```
Morriar, line 24: Should we mark it as `final` so we're sure commands can't break the encapsulation?

PR #1585, opened 2023-07-27
Filter gem RBIs by version during `annotations` command

egiurleo:
### Motivation

This PR aims to solve the problem of versioning RBIs. Up until now, there is no default way to write RBI for different versions of the same gem. This can lead to confusing issues for Sorbet users where RBIs don't match the version of a gem they're using, and they have to spend time figuring out whether their gem is out of date, whether the RBI is wrong, or if it's even both! Having a way to specify RBI for multiple versions of the same gem would allow people to adopt and use Sorbet without having to go through this hassle.

### Implementation
The actual RBI filtering behavior is implemented in [the RBI repo](https://github.com/Shopify/rbi/pull/180). This uses the `FilterVersions` rewriter during the annotations command to remove parts of the annotation RBI that aren't relevant to a certain gem version.

### Tests
I have added a test for this functionality.

### DIFF
--- a/lib/tapioca/commands/annotations.rb
+++ b/lib/tapioca/commands/annotations.rb
@@ -46,22 +46,22 @@ module Tapioca
         GitAttributes.create_vendored_attribute_file(@outpath)
       end
 
-      sig { returns(T::Array[String]) }
+      sig { returns(T::Array[GemInfo]) }
       def list_gemfile_gems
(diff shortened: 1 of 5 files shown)


paracycle (commented):
I have a few inline comments, but additionally:

Is this still a "Proof of concept"? I think once we merge this, it will ship in the next release, right?

PR #1601, opened 2023-08-11
Deprecate the `todo` command

egiurleo:
### Motivation
The `todo` command had a valid use case when Tapioca was newer and couldn't generate RBI for constants in some cases. However, now that Tapioca is more mature, `todo` is more likely to cover up a type checking issue than fix it.

### Implementation
This commit adds a deprecation message to the `todo` command warning users that the command will be removed entirely in a future version.

### Tests
Changed existing tests.

### Next steps
- Release this deprecation in the next version of Tapioca and give users time to remove `tapioca todo` from their workflows
- Create a separate PR removing the command entirely
  - Remember to remove everything associated, such as the config options and todo file checks in `check_shims` command
- Release Tapioca again with the command removed


lib/tapioca/commands/todo.rb
```diff
@@ -6,6 +6,15 @@ module Commands
     class Todo < CommandWithoutTracker
       include SorbetHelper
 
+      DEPRECATION_MESSAGE = T.let(<<~DEPRECATION, String)
+        The `todo` command is deprecated and will be removed in a future release.
+
+        If your project is still missing type definitions for constants, try the following:
+        - Regenerate gem RBIs by running `bin/tapioca gem --all` and `bin/tapioca annotations`
```
st0012, line 13: From my understanding, these steps are best used together and in this order. If that's correct, should we make them an ordered list instead?

egiurleo: Yes this is an excellent point!

lib/tapioca/commands/todo.rb
```diff
@@ -6,6 +6,15 @@ module Commands
...
+        - Generate RBIs for DSLs by running `bin/tapioca dsl`
+        - Manually create an RBI shim defining the missing constants
```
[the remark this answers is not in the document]
egiurleo: Andy you're completely right. Thanks for this suggestion!

egiurleo:
I just pushed again with the following changes:
- Rebased on main
- Improved the deprecation message to talk explicitly about using `require.rb` and numbering the steps to make it clear that there is a preferred order to them
- Added a `run_with_deprecation` method to the `Todo` command that will display the deprecation warning
- In the `init` cli, instead of using Thor's `invoke` API, I create a new instance of the `Todo` command and just call the regular `run` method

st0012 (approved):

Morriar (approved):

PR #1604, opened 2023-08-14
Start generating `.gitattributes` file for generated and vendored RBI files

paracycle:
### Motivation

Fixes https://github.com/Shopify/ruby-dev-exp-issues/issues/651

The `linguist` gem added support for detecting generated and vendored RBI files in https://github.com/github-linguist/linguist/pull/6143, but it parses the RBI file header comments to understand that, which is quite brittle.

Tapioca can do a better job of this, since it already knows that `dsl` and `gem` commands will create "generated" RBI files in the output directory and the `annotations` command will create "vendored" RBI files. So, Tapioca should be generating the appropriate `.gitattributes` files in the relevant output directories to signal that the RBI files contained are generated/vendored.

### Implementation

1. Added a `GitAttributes` helper module that knows how to create `.gitattributes` files that state the `**/*.rbi` files in the folder are either generated or vendored RBI files.
2. Uses the helper module in `DslGenerate`, `GemGenerate`, `GemSync` and `Annotations` commands to create appropriate `.gitattributes` files in the output folder.

### Tests

Added extra test cases for checking generated `.gitattributes` files.

### DIFF
--- a/lib/tapioca/commands/dsl_generate.rb
+++ b/lib/tapioca/commands/dsl_generate.rb
@@ -32,6 +32,8 @@ module Tapioca
 
         say("All operations performed in working directory.", [:green, :bold])
         say("Please review changes and commit them.", [:green, :bold])
+      ensure
+        GitAttributes.create_generated_attribute_file(@outpath)
       end
     end
   end
--- a/lib/tapioca/helpers/git_attributes.rb
+++ b/lib/tapioca/helpers/git_attributes.rb
@@ -1,0 +1,34 @@
+# typed: strict
+# frozen_string_literal: true
+
+class GitAttributes
+  class << self
+    extend T::Sig
+
+    sig { params(path: Pathname).void }
+    def create_generated_attribute_file(path)
+      create_gitattributes_file(path, <<~CONTENT)
+        **/*.rbi linguist-generated=true
+      CONTENT
+    end
+
+    sig { params(path: Pathname).void }
+    def create_vendored_attribute_file(path)
+      create_gitattributes_file(path, <<~CONTENT)
+        **/*.rbi linguist-vendored=true
+      CONTENT
+    end
+
+    private
+
+    sig { params(path: Pathname, content: String).void }
+    def create_gitattributes_file(path, content)
+      # We don't want to start creating folders, just to write
+      # the `.gitattributes` file. So, if the folder doesn't
+      # exist, we just return.
+      return unless path.exist?
+
+      File.write(path.join(".gitattributes"), content)
+    end
+  end
+end
--- a/lib/tapioca/commands/annotations.rb
+++ b/lib/tapioca/commands/annotations.rb
@@ -23,6 +23,7 @@ module Tapioca
         typed_overrides: {}
       )
         super()
+        @outpath = T.let(Pathname.new(DEFAULT_ANNOTATIONS_DIR), Pathname)
         @central_repo_root_uris = central_repo_root_uris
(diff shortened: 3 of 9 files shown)


lib/tapioca/commands/dsl_generate.rb
```diff
@@ -32,6 +32,8 @@ def execute
 
         say("All operations performed in working directory.", [:green, :bold])
         say("Please review changes and commit them.", [:green, :bold])
+      ensure
```
Morriar, line 35: It's surprising that we're trying to create the file even if something went wrong earlier.

paracycle: Since `gitattribute` file creation makes sure that the output folder exists before creating a file, I figured it would be good to unconditionally generate a gitattribute file in case any RBI files have already been generated by the time we hit an error.

Morriar: If something went wrong earlier, we do not have any guaranty that `@outpath` even exist.

Also I'm wondering if the `.gitattribute` should be moved to `tapioca init` rather than rewriting it every time we generate the RBIs?

paracycle: > If something went wrong earlier, we do not have any guaranty that @outpath even exist.

That was exactly my point above. The way [`GitAttributes.create_gitattributes_file` method](https://github.com/Shopify/tapioca/pull/1604/files#diff-d9d26e65534c05831a2c22654a3e66df0b3cf6df8650068833598b08eaa8cae8R26-R29) works is to first check if the folder exists or not. If the folder does not exist, it does not create any file. So, this is the safest way to generate `.gitattributes` file for any RBI that may be generated in the output directory.

> Also I'm wondering if the .gitattribute should be moved to tapioca init rather than rewriting it every time we generate the RBIs?

I don't think that would be a great thing to add. People could delete their `rbi` folder, or do a bunch of other crazy things without needing to run `init` command again. A `gem` or `dsl` command should make sure a valid `.gitattributes` file is present in the output path if any RBI files are generated.

This would also make sure that people who upgrade their Tapioca version and generate new gem/dsl RBIs will get the `.gitattributes` file as well.

lib/tapioca/helpers/git_attributes.rb
```diff
@@ -0,0 +1,34 @@
...
+    private
+
+    sig { params(path: Pathname, content: String).void }
+    def create_gitattributes_file(path, content)
+      # We don't want to start creating folders, just to write
+      # the `.gitattributes` file. So, if the folder doesn't
+      # exist, we just return.
+      return unless path.exist?
+
+      File.write(path.join(".gitattributes"), content)
```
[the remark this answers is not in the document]
paracycle: Good question. I don't really want to go into the business of parsing gitattribute files.

I see two options:
1. We claim that the gem/dsl RBI folders belong completely to Tapioca and we can generate/overwrite whatever we want inside those folders.
2. We skip writing to the file if it exists.

So, far we've been operating based on the first option and not caring about existing files too much. It also allows us to update the contents of the file in the future without worrying about applying a changeset to it or whatever.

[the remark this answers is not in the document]
Morriar: > We claim that the gem/dsl RBI folders belong completely to Tapioca and we can generate/overwrite whatever we want inside those folders.

Ok, let's just overwrite then 👍

Morriar (approved):

PR #1610, opened 2023-08-16
Docs: Tapioca should also be in `test` group in Gemfile

andyw8:
It's common to use `check-shims` in CI, so I believe Tapioca should also be in the `test` group.


andyw8:
(description was missing, added)

Issue #1516, opened 2023-05-29
Merge gem annotations with gem RBIs to simplify user experience


KaanOzkan:
We can merge the annotation RBIs with the gem RBI during `tapioca gem` command instead of relying on `tapioca annotations`.

I think not worrying about the `annotations` command can simplify the user experience. Also this will result in a single gem related RBI which conceptually is more consistent especially with the future plans of [inline RBI versioning](https://github.com/Shopify/rbi/pull/180) for the annotations.

PR #1613, opened 2023-08-17
Combine `gem` and `annotations` commands

bitwise-aiden:
### Motivation

Closes: https://github.com/Shopify/tapioca/issues/1516

In an effort to simplify the UX of Tapioca, we want to combine the `annotation` and `gem` commands so that RBIs for gems are generated using source annotations as the base.

### Implementation

This implementation deprecates the old `annotation` command, raising an error if run. The logic for that command has been altered and moved inside the `AbstractGem` command to be used by the sync and generate codepaths. 

A few issues have spawned off of this one around how to cache annotations between requests, but for the time being the core logic from `annotation` for fetching files locally/http has been kept intact. This extra work gates our ability to get the `sha` for the annotation file, so for now we won't be updating with a version of annotations, just the location they were sourced from.

Along with the command options from `annotation`, `--exclude-annotation [gem, ...]` and `--annotations` (and subsequently `--no-annotations`) have been added to control when annotations are used as a base.

### Tests
Moved `annotation` tests into `gem`, updating for file layout, etc. Added some new tests for extra command args like `--exclude-annotation`


bitwise-aiden:
The `AbstractGem` code relating to files and net code will be abstracted out of the class once we have decided how to appropriately cache annotation sources.

README.md
```diff
@@ -72,7 +71,7 @@ $ tapioca help
 
 Commands:
   tapioca --version, -v      # Show version
-  tapioca annotations        # Pull gem RBI annotations from remote sources
+  tapioca annotations        # Deprecated, this command has been merged into `tapioca gem`. Please use that instead.
```
Morriar, line 74: I think we should keep the original description but add `deprecated`.

The error message when we run the command already explains what to run.

README.md
```diff
@@ -918,11 +867,17 @@ gem:
   rbi_max_line_length: 120
   environment: development
   halt_upon_load_error: true
+  annotations: true
+  annotation_sources:
+  - https://raw.githubusercontent.com/Shopify/rbi-central/main
+  annotation_netrc: true
+  annotation_netrc_file: ''
+  annotation_auth: ''
```
Morriar, line 875: Let's pluralize all of them, we pull more than one annotation. This will allow be more cohesive with `annotations: true`

lib/tapioca/cli.rb
```diff
@@ -250,6 +249,28 @@ def dsl(*constant_or_paths)
...
+      desc: "Use .netrc to authenticate to private annotation sources"
+    option :annotation_netrc_file, type: :string, desc: "Path to .netrc file"
+    option :annotation_auth,
+      type: :string,
+      default: nil,
+      desc: "HTTP authorization header for private annotation sources"
+    option :exclude_annotation,
+      type: :array,
+      banner: "gem [gem ...]",
+      desc: "Excludes annotation for gem while generating RBI",
```
Morriar, line 272: Here and all the previous options descriptions, let's pluralize:

```suggestion
      desc: "Excludes annotations for gem while generating RBI",
```

lib/tapioca/cli.rb
```diff
@@ -267,6 +288,14 @@ def gem(*gems)
         raise MalformattedArgumentError, "Option '--verify' must be provided without any other arguments" if verify
       end
 
+      if !options[:annotation_netrc] && options[:annotation_netrc_file]
+        raise Thor::Error, set_color(
+          "Options `--no-annotation-netrc` and `--annotation-netrc-file` can't be used together",
```
Morriar, line 293: Let's not forget to pluralize here too

lib/tapioca/commands/abstract_gem.rb
```diff
@@ -67,12 +77,98 @@ def initialize(
...
+        @annotation_sources.each do |uri|
+          index = fetch_annotation_source_index(uri, repo_number: multiple_repos ? repo_number : nil)
+          next unless index
+
+          indexes[uri] = index
+          repo_number += 1
+        end
+
+        if indexes.empty?
+          raise Thor::Error, set_color(
```
Morriar, line 161: I think we have `say_error` for formatting?

lib/tapioca/commands/abstract_gem.rb
```diff
@@ -307,6 +433,74 @@ def merge_with_exported_rbi(gem, file)
         say_error("\n\n  RBIs exported by `#{gem.name}` contain errors and can't be used:", :yellow)
         say_error("Cause: #{e.message} (#{e.location})")
       end
+
+      sig { params(annotation: RBI::Tree, file: RBI::File).void }
+      def merge_with_annotation(annotation, file)
+        file.root = RBI::Rewriters::Merge.merge_trees(
+          file.root,
+          annotation,
+          keep: RBI::Rewriters::Merge::Keep::LEFT,
```
Morriar, line 442: I wonder if that should be a CLI option?

spec/tapioca/cli/gem_spec.rb
```diff
@@ -207,6 +207,56 @@ class << self
...
+              include ::PP::ObjectMixin
+
+              def foo; end
+            end
+          RBI
+
+          repo.destroy!
+        end
+
+        it "errors if passing both --no-annotation-netrc and --annotation-netrc-file" do
```
Morriar, line 253: Let's not forget to pluralize the test descriptions too.

README.md
```diff
@@ -40,10 +40,9 @@ Tapioca makes it easy to work with [Sorbet](https://sorbet.org) in your codebase
     * [Excluding a gem from RBI generation](#excluding-a-gem-from-rbi-generation)
     * [Changing the strictness level of the RBI for a gem](#changing-the-strictness-level-of-the-rbi-for-a-gem)
     * [Keeping RBI files for gems up-to-date](#keeping-rbi-files-for-gems-up-to-date)
-  * [Pulling RBI annotations from remote sources](#pulling-rbi-annotations-from-remote-sources)
+    * [Pulling RBI annotations from remote sources](#pulling-rbi-annotations-from-remote-sources)
     * [Basic authentication](#basic-authentication)
```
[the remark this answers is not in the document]
bitwise-aiden: This is auto formatted off of the heading used. Given that annotations are now being rolled into gem, I'd updated the heading to be a part of the previous section. Do you think that we should leave it as a top level concept?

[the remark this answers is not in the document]
Morriar: The nesting is based on the heading level in the Markdown. You can add a `#` to increase the `<hX>` level and thus the nesting.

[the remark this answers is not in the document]
bitwise-aiden: I've misspoke. I intentionally removed the `#` causing this change as it seems fitting for this to be a sub heading of the gem command. I was wanting to know your thoughts on why we should keep it as a top level concept instead.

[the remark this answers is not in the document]
Morriar: I don't want it to be a top level I want the following sub-sections to still be nested under `Pulling RBI annotations from remote source`:

* `Basic authentication`
* `Using a .netrc file`

lib/tapioca/commands/abstract_gem.rb
bitwise-aiden: I think this is more about control flow. There are two other cases of this pattern in the `abstract_gem` file.

lib/tapioca/commands/abstract_gem.rb
bitwise-aiden: I'm having difficulty coming up with how we communicate this option. Would the possibilities be annotation or rbi as the base?

lib/tapioca/cli.rb
```diff
@@ -335,18 +367,12 @@ def check_shims
...
-      if !options[:netrc] && options[:netrc_file]
-        raise Thor::Error, set_color("Options `--no-netrc` and `--netrc-file` can't be used together", :bold, :red)
-      end
-
-      command = Commands::Annotations.new(
-        central_repo_root_uris: options[:sources],
-        auth: options[:auth],
-        netrc_file: netrc_file(options),
-        typed_overrides: options[:typed_overrides],
+      raise Thor::Error, set_color(
```
andyw8, line 370: nit: Doesn't 'deprecated' generally mean that a behaviour is still available, just not approved of?

README.md
```diff
@@ -321,70 +329,31 @@ Nothing to do, all RBIs are up-to-date.
...
-Removing annotations for gems that have been removed...  Nothing to do
-Fetching gem annotations from central repository...
+Since Tapioca does not perform any type inference, the RBI files generated for the gems do not contain any type signatures. Instead, Tapioca relies on the community to provide high-quality, manually written RBI annotations for public gems. These annotations are sourced and combined with the generated RBIs as a step during `tapioca gem`.
 
-  Fetched activesupport
-   created  sorbet/rbi/annotations/activesupport.rbi
-
-Done
-```
+By default, Tapioca will pull the annotations stored in the central repository located at https://github.com/Shopify/rbi-central. It is possible to use a custom repository by changing the value of the `--annotations-sources` option. For example if your repository is stored on Github:
```
andyw8, line 336:
```suggestion
By default, Tapioca will pull the annotations stored in the central repository located at https://github.com/Shopify/rbi-central. It is possible to use a custom repository by changing the value of the `--annotations-sources` option. For example if your repository is stored on GitHub:
```

andyw8 (approved):

bitwise-aiden:
This is being put on pause while I'm away on pat leave. 

The current state is that this is nearing completion and had some cursory tests against a large codebase.
We'd like to try and figure out caching / validation of which annotations go into generating the RBIs, or at least have a plan before shipping this.

bitwise-aiden:
Closing as this is likely very stale.

PR #1622, opened 2023-08-24
Update GraphQL mutation compiler to consider arguments with `replace_null_with_default`

dom-binti:
### Motivation

With `graphql-ruby` arguments, you can set `required: false, default_value: 'foo', replace_null_with_default: true` to fill a `null` sent by a client with the default value ([docs](https://graphql-ruby.org/fields/arguments.html#default-values)). The current graphql mutation compiler doesn't take this into account, and assumes that all `required: false` arguments are nilable, when in reality a non-nil `default_value` plus `replace_null_with_default` means the resolver will always receive a non-nil value.

### Implementation

I updated `GraphqlTypeHelper.type_for` to take in a `GraphQL::Schema::Argument`, rather than just the argument type, and check for the other `Argument` params when deciding if the type is nilable.

### Tests

Added a test with both a nil and a non-nil `default_value` to the existing graphql mutation compiler tests.

### DIFF
--- a/lib/tapioca/dsl/compilers/graphql_input_object.rb
+++ b/lib/tapioca/dsl/compilers/graphql_input_object.rb
@@ -55,7 +55,7 @@ module Tapioca
           root.create_path(constant) do |input_object|
             arguments.each do |argument|
               name = argument.keyword.to_s
-              input_object.create_method(name, return_type: Helpers::GraphqlTypeHelper.type_for(argument.type))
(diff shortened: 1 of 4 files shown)


dom-binti:
@KaanOzkan sgtm. Looks like CI is passing now. Let me know if there's anything else to do on my end!

KaanOzkan:
Looks like our monolith doesn't utilize `replace_null_with_default`. Existing RBIs are unchanged. Thanks for the contribution.

PR #1623, opened 2023-08-25
Add compiler for ActiveRecord::Store

aramgre:
### Motivation

Adds a DSL compiler for methods generated by [ActiveRecord::Store](https://api.rubyonrails.org/classes/ActiveRecord/Store.html). 

### Implementation

`ActiveRecord:Store` does not save the prefix or suffix after defining accessors, so this PR uses an extension to capture them. The extension only overrides `store_accessor`, [since that is also called indirectly by `store`](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/store.rb#L109), but I guess there's some risk it could break if the implementation of `store` was changed. 

### Tests

Added


KaanOzkan (approved):
LGTM

aramgre:
I have signed the CLA!

aramgre:
Sorry about that rebase fail, thanks for fixing it @KaanOzkan 🙏

PR #1629, opened 2023-08-30
Suggest bin/tapioca dsl -e test for CI failures

pedropb:
### Motivation

Different Rails environments can yield different RBIs when running `bin/tapioca dsl`, due to custom routes loaded only in specific environments (e.g.: development). See this [common example](https://github.com/Shopify/tapioca/issues/657#issuecomment-1003532239) that happens in Rails due to routes injected by the framework only in the development environment.

Related to #657 and #979 

### Implementation

- Passed the environment arg down the chain of calls from `cli` to `abstract_dsl`
- With this additional context, `abstract_dsl#report_diff_and_exit_if_out_of_date` can print a suggestion to regenerate the RBIs with the same runtime context that failed the verification:
```bash
bin/tapioca dsl --verify         # fails => then `bin/tapioca dsl` is suggested to update the RBIs
bin/tapioca dsl --verify -e test # fails => then `bin/tapioca dsl -e test` is suggested instead
```

### Tests

- Adapted existing tests assertions to expect the new message.
- Added a new test case to assert the new environment logic

### DIFF
--- a/lib/tapioca/cli.rb
+++ b/lib/tapioca/cli.rb
@@ -155,6 +155,7 @@ module Tapioca
         rbi_formatter: rbi_formatter(options),
         app_root: options[:app_root],
         halt_upon_load_error: options[:halt_upon_load_error],
+        environment: options[:environment],
       }
 
       command = if options[:verify]
--- a/lib/tapioca/commands/abstract_dsl.rb
+++ b/lib/tapioca/commands/abstract_dsl.rb
@@ -26,6 +26,7 @@ module Tapioca
           rbi_formatter: RBIFormatter,
           app_root: String,
           halt_upon_load_error: T::Boolean,
+          environment: String,
         ).void
       end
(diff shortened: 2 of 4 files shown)


pedropb:
Thanks @vinistock ! Our app is definitely missing configurations in `sorbet/tapioca/config.yml` 👍 

I'll have a look at them and see if they would be sufficient to eliminate our problems without requiring manual intervention like the suggestion in this PR.

pedropb:
@vinistock as discussed over Slack:
- making changes to `sorbet/tapioca/config` were unfruitful: our app cannot be changed to run in CI with `development` configurations
- however, any dev can run `bin/tapioca dsl -e test` to regenerate the RBIs and commit them to version control

Therefore I changed the approach and updated the PR description.
If you feel like these changes do not improving the dev experience, please let me know and I can adapt or close the PR.

PR #1632, opened 2023-09-01
Add Action Text compiler

ghiculescu:
### Motivation

Adds a compiler for Action Text.

### Implementation

I based most of this off the Active Storage compiler. Action Text is not very complex internally so the compiler was not too hard to write.

### Tests

Added tests and confirmed they match our home-made RBIs.

### DIFF
--- a/lib/tapioca/dsl/compilers/action_text.rb
+++ b/lib/tapioca/dsl/compilers/action_text.rb
@@ -1,0 +1,100 @@
+# typed: strict
+# frozen_string_literal: true
+
+begin
+  require "action_text"
+rescue LoadError
+  return
+end
+
+module Tapioca
+  module Dsl
+    module Compilers
+      # `Tapioca::Dsl::Compilers::ActionText` decorates RBI files for subclasses of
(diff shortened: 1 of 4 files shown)


KaanOzkan:
I can't replicate the failure locally on 3.0.6 😕

ghiculescu:
Maybe it was just flakey @KaanOzkan

KaanOzkan (approved):
LGTM. Could you add a test for `gather_constants` also similar to https://github.com/Shopify/tapioca/blob/2f20cd36d646f562b0194b7f0f7cfa2395d78924/spec/tapioca/dsl/compilers/active_storage_spec.rb#L23

KaanOzkan (approved):
Thanks it looks good. Can you please clean up the git history to atomic commits? We don't squash and merge in tapioca and it should help with git blame later on. Sorry for the back and forth didn't notice it earlier.

ghiculescu:
Done @KaanOzkan thank you for all your help!

ghiculescu:
I think the CI errors came from https://github.com/Shopify/tapioca/pull/1604

ghiculescu:
@KaanOzkan is this ok to merge?

PR #1644, opened 2023-09-06
Implement ActiveModel::Validations::Confirm compiler

Tonkpils:
### Motivation

Given a class implementing a `confirmation` validation from `ActiveModel::Validations`

```
class User < ApplicationRecord
  validates :password, confirmation: true
end
```

will generate virtual attributes `password_confirmation` and `password_confirmation=`. Without this compiler, usage of those methods will cause an error with Sorbet unless they are shimmed out.

### Implementation

The compiler is pretty straightforward iterating over all `ConfirmationValidator` and generating the necessary methods for rbi. It also skips `ActiveRecord::Base` since that class includes `ActiveModel::Validations` but does not implement any of its validations. 

### Tests

Tests were added


lib/tapioca/dsl/compilers/active_model_validations_confirmation.rb
```diff
@@ -0,0 +1,97 @@
...
+          }
+        end
+
+        class << self
+          sig { override.returns(T::Enumerable[Module]) }
+          def gather_constants
+            # Collect all the classes that include ActiveModel::Validations
+            all_classes.select do |c|
+              c < ActiveModel::Validations &&
+                c.name.to_s != "ActiveRecord::Base"
```
vinistock, line 66: Why is it necessary to explicitly exclude `ActiveRecord::Base`?

lib/tapioca/dsl/compilers/active_model_validations_confirmation.rb
```diff
@@ -0,0 +1,97 @@
...
+            end
+          end
+        end
+
+        sig { override.void }
+        def decorate
+          confirmation_validators = constant.validators.select do |v|
+            v.is_a?(ActiveModel::Validations::ConfirmationValidator)
+          end
```
vinistock, line 75: I _think_ you can do this.
```suggestion
          confirmation_validators = constant.validators.grep(ActiveModel::Validations::ConfirmationValidator)
```

lib/tapioca/dsl/compilers/active_model_validations_confirmation.rb
```diff
@@ -0,0 +1,97 @@
...
+
+          return if confirmation_validators.empty?
+
+          # Create a RBI definition for each class that includes Active::Model::Validations
+          root.create_path(constant) do |klass|
+            # Create RBI definitions for all the attributes that use confirmation validation
+            confirmation_validators.each do |validator|
+              validator.attributes.each do |attr_name|
+                klass.create_method("#{attr_name}_confirmation", return_type: "T.untyped")
```
vinistock, line 84: Would it be possible to figure out the right types based on the columns?

lib/tapioca/dsl/compilers/active_model_validations_confirmation.rb
Tonkpils: Without this the test was returning `ActiveRecord::Base` as a gathered class. I'm assuming it's because ActiveRecord::Base includes ActiveModel::Validations so it gets included there?

lib/tapioca/dsl/compilers/active_model_validations_confirmation.rb
Tonkpils: If the it's a subclass of ActiveRecord::Base, we can probably use one of ActiveRecord's API to figure out the column type and translate it to a proper return type. The same could be said for classes that implement ActiveModel::Attributes. 

However, since this validations can be added to POROs, we'd have to add logic to differentiate the two. Are we ok with doing that here?

lib/tapioca/dsl/compilers/active_model_validations_confirmation.rb
vinistock: But does it matter if it is collected? If would just not have any confirmation validations and not do anything, right?

lib/tapioca/dsl/compilers/active_model_validations_confirmation.rb
vinistock: Good point. Do these validations require you to use Active Model attributes or something that would allow us to fetch the type?

If not, I think it's fine to use `T.untyped`. Just trying to think if there's a way to get better typing support.

lib/tapioca/dsl/compilers/active_model_validations_confirmation.rb
Tonkpils: That's right. I'll go ahead and remove this then 😄

Tonkpils: All done! [b78f5bf](https://github.com/Shopify/tapioca/pull/1644/commits/b78f5bf7051bd99c0e73c8e01ec0da58c8258ed4)

Tonkpils:
> It appears you need to run bin/docs to update the docs.

All fixed 🙇

vinistock:
I think you might need to rebase the branch because of #1643. There are some typechecking and test errors.

Tonkpils:
@vinistock looks like this will be fixed by https://github.com/Shopify/tapioca/pull/1648?

vinistock:
Okay, rebasing now should fix the CI.

PR #1646, opened 2023-09-07
Support Composite Primary keys in ActiveRecordColumns compiler 

gmcgibbon:
### Motivation
Support Active Record composite primary keys in `ActiveRecordColumns` compiler.

### Implementation
I imagine using Rails edge may not be ideal, and if it isn't, we can look at merging this closer to when CPK is released. Due to how the compiler works, I would also consider renaming it to `ActiveRecordAttributes` instead.

### Tests
The implementation inadvertently fixed a bug where the generated attribute methods don't include `id` when using a non-id primary key. Active Record always has an `id` method defined via https://github.com/rails/rails/blob/bc2892ab4557136a959e95c329c2dcc0133c7585/activerecord/lib/active_record/attribute_methods/primary_key.rb.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_columns.rb
+++ b/lib/tapioca/dsl/compilers/active_record_columns.rb
@@ -110,8 +110,8 @@ module Tapioca
 
           root.create_path(constant) do |model|
             model.create_module(AttributeMethodsModuleName) do |mod|
-              constant.attribute_names.each do |column_name|
-                add_methods_for_attribute(mod, column_name)
+              (constant.attribute_names + ["id"]).uniq.each do |attribute_name|
+                add_methods_for_attribute(mod, attribute_name)
               end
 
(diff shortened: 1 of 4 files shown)


KaanOzkan (commented):
I think it's okay to move to Rails main. We currently [test against it](https://github.com/Shopify/tapioca/actions/runs/6099477084/job/16551728751#step:6:1165) so that logic would have to to be replaced with Rails 7: https://github.com/Shopify/tapioca/blob/main/gemfiles/Gemfile-rails-main.

There are some unrelated failures this PR will have to fix due to Rails main. It could be worth doing that work in another PR.

gmcgibbon:
This build is green now, and the CPK test is conditionalized to only run on main for now.

KaanOzkan (approved):

gmcgibbon:
Done. The output looks correct to me, so I'm going to merge.

PR #1648, opened 2023-09-11
Fix CI

KaanOzkan:
### Motivation

Resolving CI failures https://github.com/Shopify/tapioca/actions/runs/6124704298/job/16625214068

### Implementation

Reflected API changes similar to https://github.com/Shopify/tapioca/pull/1643

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -295,9 +295,9 @@ GEM
       zeitwerk (~> 2.5)
     rainbow (3.1.1)
     rake (13.0.6)
-    rbi (0.1.0)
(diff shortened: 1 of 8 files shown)


Morriar (approved):
We may need to require a rebase before merge?

bitwise-aiden (approved):

KaanOzkan:
Tests are using the latest version of YARP which isn't compatible with `rbi` gem. Will bump `rbi` once a new version with https://github.com/Shopify/rbi/pull/232 is released.

PR #1657, opened 2023-09-12
Add reusable GitHub Actions workflow for Dependabot RBI updates

andyw8:
~In draft while we verify on https://github.com/Shopify/ruby-lsp.~ ✔️ 

### Motivation

We can have a shared config rather than having to define it for each repo.

### Implementation

Copied directly from https://github.com/Shopify/ruby-lsp/blob/main/.github/workflows/dependabot_update_rbis.yml

### Tests

n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,31 +321,6 @@ Nothing to do, all RBIs are up-to-date.
 
 This option can be used on CI to make sure the RBI files are always up-to-date and ensure accurate type checking. **Warning**: doing so will break your normal Dependabot workflow as every pull-request opened to bump a gem version will fail CI since the RBI will be out-of-date and will require you to manually run `bin/tapioca gems` to update them.
 
-To avoid this problem, you can use a reusable GitHub Actions workflow to automatically update the RBIs:
-
-```yaml
-# .github/workflows/dependabot_update_rbis.yml
-
-name: Update gem RBIs on Dependabot PRs
-
-on:
-  pull_request:
-    types: [opened, synchronize]
-    paths:
-      - 'Gemfile.lock'
-
-permissions:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -321,6 +321,30 @@ Nothing to do, all RBIs are up-to-date.
...
+    paths:
+      - 'Gemfile.lock'
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  dependabot_update_rbis:
+    uses: Shopify/tapioca/.github/workflows/dependabot_update_rbis.yml@main
```
andyw8, line 343: If we have this within the tapoica report then we can't easily version it (e.g. `v1`), so that may an argument for moving it to its own repo.

andyw8:
We discovered on the ruby-lsp repo that this does not play well with auto-merging.

From @KaanOzkan:

> I believe the issue is that commits made from the github actions bot doesn’t trigger CI checks (CI checks aren’t associated with that commit). But the PR is waiting on required status checks on that commit. I assume this is intentional to prevent an infinite workflow -> commit -> workflow loop

https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow

README.md
andyw8: Updated to remove.

KaanOzkan: If you want to teach folks it's possible we could mention it in the wiki or readme without having a workflow pointing to tapioca's. We can tell them they can use our workflow as a starting point and highlight the GitHub access token requirement and Ruby version detection that they'd have to change. I just don't want to recommend a workflow that uses tapiocas reusable one.

PR #1659, opened 2023-09-18
Add support for custom GraphQL scalar coercion

nicoco007:
### Motivation
When trying to update our project to Tapioca v0.11.9, I ran into an issue with our custom scalars: Tapioca currently uses the GraphQL scalar class as the return type even if it defines a `coerce_input` method with a different return type. For example, if I have a scalar `MyScalarType` that inherits from `GraphQL::Schema::Scalar` and has a `coerce_input` that returns `MyScalar`, Tapioca will wrongly use `MyScalarType` in generated signatures for mutations and input objects. This is an attempt at fixing that issue.

I also shoehorned `GraphQL::Types::BigInt` here since I noticed it wasn't working properly either. Happy to separate that into its own PR if needed.

### Implementation
The approach I took is to simply check if a type signature exists on `coerce_input` and use the return type if it does. If not, it falls back to the current behaviour.

### Tests
I've added a test here and also checked against our project to make sure I didn't miss anything.

### DIFF
--- a/lib/tapioca/dsl/helpers/graphql_type_helper.rb
+++ b/lib/tapioca/dsl/helpers/graphql_type_helper.rb
@@ -28,7 +28,7 @@ module Tapioca
             type_for_constant(Float)
           when GraphQL::Types::ID.singleton_class, GraphQL::Types::String.singleton_class
             type_for_constant(String)
-          when GraphQL::Types::Int.singleton_class
+          when GraphQL::Types::Int.singleton_class, GraphQL::Types::BigInt.singleton_class
             type_for_constant(Integer)
           when GraphQL::Types::ISO8601Date.singleton_class
             type_for_constant(Date)
--- a/lib/tapioca/dsl/helpers/graphql_type_helper.rb
+++ b/lib/tapioca/dsl/helpers/graphql_type_helper.rb
@@ -45,6 +45,10 @@ module Tapioca
             else
               "T.any(#{value_types.join(", ")})"
             end
+          when GraphQL::Schema::Scalar.singleton_class
+            method = Runtime::Reflection.method_of(unwrapped_type, :coerce_input)
+            signature = Runtime::Reflection.signature_of(method)
+            signature&.return_type&.to_s || "T.untyped"
           when GraphQL::Schema::InputObject.singleton_class
             type_for_constant(unwrapped_type)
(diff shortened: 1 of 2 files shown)


lib/tapioca/dsl/helpers/graphql_type_helper.rb
```diff
@@ -45,6 +45,10 @@ def type_for(argument)
             else
               "T.any(#{value_types.join(", ")})"
             end
+          when GraphQL::Schema::Scalar.singleton_class
+            method = Runtime::Reflection.method_of(unwrapped_type, :coerce_input)
+            signature = Runtime::Reflection.signature_of(method)
+            signature&.return_type&.to_s || "T.untyped"
```
nicoco007, line 51: [The default behaviour of `Scalar` is to just return the raw value](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/scalar.rb#L9) so I _think_ it makes more sense to return `T.untyped` here rather than the `unwrapped_type`. Happy to discuss!

Tiedye:
@paracycle are you able to look at this?

lib/tapioca/dsl/helpers/graphql_type_helper.rb
paracycle: Yes, this makes sense to me. However, this needs more finesse, since `return_type` can be `Void` or `NotTyped`, neither of which make sense as a parameter type.

We need to do something similar to https://github.com/Shopify/tapioca/blob/9c9b0e730cce4f35e4478bb8dd80de66238ce456/lib/tapioca/dsl/helpers/active_record_column_type_helper.rb#L124-L125

paracycle (commented):
Generally this looks good to me but there is one aspect that needs improvement.

lib/tapioca/dsl/helpers/graphql_type_helper.rb
paracycle: Ok, I added this check in a commit, and will merge when the CI is green.

paracycle (approved):

paracycle:
> Oh sorry I wasn't paying attention, thanks for the change! I hadn't thought of those cases.

All good, thanks for your contribution. 🙏

PR #1677, opened 2023-09-27
fix generating incorrect type for custom scalars

Tiedye:
### Motivation

https://github.com/Shopify/tapioca/pull/1571 caused the graphql dsl compiler to return the incorrect type for custom scalars.  Custom scalars define two methods `coerce_input` and `coerce_result`, in particular `coerce_result` generates the data that is returned to the mutation and this can be _any_ type.  `T.untyped` was correct

### Implementation

Revert the offending change in https://github.com/Shopify/tapioca/pull/1571

### Tests

Updated the existing tests

### DIFF
--- a/lib/tapioca/dsl/helpers/graphql_type_helper.rb
+++ b/lib/tapioca/dsl/helpers/graphql_type_helper.rb
@@ -49,6 +49,10 @@ module Tapioca
             type_for_constant(unwrapped_type)
           when GraphQL::Schema::NonNull.singleton_class
             type_for(unwrapped_type.of_type)
+          when GraphQL::Schema::Scalar.singleton_class
+            "T.untyped"
+          when Module
+            Runtime::Reflection.qualified_name_of(unwrapped_type) || "T.untyped"
           else
             "T.untyped"
           end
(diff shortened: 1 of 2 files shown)


paracycle:
I am not sure if just reverting this change will fix things. Did you mean to revert the PR that adds `loads` support https://github.com/Shopify/tapioca/pull/1556?

Tiedye:
> I think instead of removing the whole `Module` fallback, what we really need to do is handle `GraphQL::Schema::Scalar` types explicitly right before that to generate `T.untyped`. Would that not fix the `CustomScalar` problem?

I'll give it a go

Tiedye:
> I think instead of removing the whole `Module` fallback, what we really need to do is handle `GraphQL::Schema::Scalar` types explicitly right before that to generate `T.untyped`. Would that not fix the `CustomScalar` problem?

Yep that worked

Tiedye:
@paracycle How's that look?

Tiedye:
> > I think instead of removing the whole Module fallback, what we really need to do is handle GraphQL::Schema::Scalar types explicitly right before that to generate T.untyped. Would that not fix the CustomScalar problem?
> 
> Oh, I did something similar in #1659 that I think addresses this.

Nice!  Yours is better, I'll close this if that gets merged

PR #1679, opened 2023-09-28
Generate RBIs for default gem symbols that are missing from the payload 

KaanOzkan:
### Motivation
Resolves https://github.com/Shopify/ruby-dev-exp-issues/issues/1057


KaanOzkan:
I can't reproduce locally but this is blocked by sorbet executable we run to bootstrap symbols seg faulting on CI.

PR #1682, opened 2023-10-04
Update AASM compiler for instance aasm method

bdewater:
### Motivation
This is useful for e.g. `model.aasm.may_fire_event?(params[:event].to_sym)`.

### Implementation
See https://github.com/aasm/aasm/blob/2e952ff2088082c9f00c95e3c1261afc826f737c/lib/aasm/aasm.rb#L67-L74

### Tests
✅


vinistock (commented):
Looks good to me. Just a question

PR #1694, opened 2023-10-16
Add postgres Uuid to be translated as String by tapioca dsl

pedrol3001:
### Motivation
The Rails application I'm working on utilizes UUIDs as described in this tutorial: [link to the tutorial](https://pawelurbanek.com/uuid-order-rails). Unfortunately, Tapioca's DSL couldn't correctly identify the column types generated by PostgreSQL. This led to a situation where primary and foreign keys were classified as T.untyped, disabling these attributes type checking.

### Implementation
To resolve this issue, I've introduced a specific verification for the column type, "ActiveRecord::ConnectionAdapters::PostgreSQL::OID::Uuid," which ensures that the affected columns are correctly recognized as ::String by Tapioca, thereby enabling accurate type checking.

### Tests
As i am new to this repository, I couldn't find an established location for adding tests since there are no existing ones for the active_record_column_type_helper.rb file. I would appreciate guidance on the appropriate method for adding tests to this file or whether it's a common practice in this context.

### DIFF
--- a/sorbet/rbi/shims/activerecord_postgresql.rbi
+++ b/sorbet/rbi/shims/activerecord_postgresql.rbi
@@ -1,5 +1,6 @@
 # typed: strict
 
 # These constants are dynamically loaded only when connecting to postgresql
+module ActiveRecord::ConnectionAdapters::PostgreSQL::OID::Uuid; end
 module ActiveRecord::ConnectionAdapters::PostgreSQL::OID::Array; end
 module ActiveRecord::ConnectionAdapters::PostgreSQL::OID::Hstore; end


pedrol3001:
I have signed the CLA!

KaanOzkan (approved):
Looking at sorbet-rails `string` [looks correct](https://github.com/chanzuckerberg/sorbet-rails/blob/5bfd6da31f105e64e4dba37fd2c723ade13c27f5/lib/sorbet-rails/model_column_utils.rb#L111). We don't utilize postgres internally so if this branch works on your project we can merge it.

PR #1702, opened 2023-10-30
Fix engine loading for Rails 7.2

vinistock:
### Motivation

Commit https://github.com/rails/rails/commit/ebfca905db14020589c22e6937382e6f8f687664 renamed `eager_load_paths` to `all_eager_load_paths`.

For Rails applications pointing to `main`, Tapioca ends up not loading all engine paths

### Implementation

The idea of the implementation is to switch to using the Rails 7.2 name of the configuration and just ensure it always exists even for older versions.

### Tests

I'm not sure how we can write a test for this given that 7.2 has no releases yet. I did test this in Core though.

### DIFF
--- a/lib/tapioca/loaders/loader.rb
+++ b/lib/tapioca/loaders/loader.rb
@@ -80,6 +80,8 @@ module Tapioca
       def load_rails_engines
         return if engines.empty?
 
+        normalize_eager_load_paths_configuration!
+
         with_rails_application do
           run_initializers
 
@@ -110,7 +112,7 @@ module Tapioca
         autoloader = Zeitwerk::Loader.new
 
         engines.each do |engine|
-          engine.config.eager_load_paths.each do |path|
+          engine.config.all_eager_load_paths.each do |path|
             # Zeitwerk only accepts existing directories in `push_dir`.
             next unless File.directory?(path)
             # We should not add directories that are already managed by a Zeitwerk loader.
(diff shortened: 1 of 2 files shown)


lib/tapioca/loaders/loader.rb
```diff
@@ -110,7 +112,7 @@ def load_engines_in_zeitwerk_mode
         autoloader = Zeitwerk::Loader.new
 
         engines.each do |engine|
-          engine.config.eager_load_paths.each do |path|
+          engine.config.all_eager_load_paths.each do |path|
```
st0012, line 115: Isn't `all_eager_load_paths` a new method added in the referenced commit? Should we be doing the opposite instead?
- Still calls `eager_load_paths`
- For Rails 7.2, we assign `eager_load_paths` with `all_eager_load_paths`'s value

vinistock: I think doing the opposite has potential to get us into weird situations. The `eager_load_paths` method still exists in Rails 7.2, it just returns something different than before.

If we patch it and Rails itself uses it in a way we don't expect, it could lead to strange scenarios. By making sure the new method exists even for old versions, we don't have to worry about it.

st0012: Ah sorry I forgot that Rails config doesn't need a static method definition, so I thought this would cause NoMethodError.

lib/tapioca/loaders/loader.rb
```diff
@@ -179,6 +181,19 @@ def engines
           .reject { |engine| gem_in_app_dir?(project_path, engine.config.root.to_path) }
       end
 
+      # Rails 7.2 renamed `eager_load_paths` to `all_eager_load_paths`, which maintains the same original functionality.
+      # The `eager_load_paths` method still exists, but doens't return all paths anymore and causes Tapioca to miss some
```
andyw8, line 185:
```suggestion
      # The `eager_load_paths` method still exists, but doesn't return all paths anymore and causes Tapioca to miss some
```

andyw8 (approved):

paracycle:
> I'm not sure how we can write a test for this given that 7.2 has no releases yet. I did test this in Core though.

We do test against Rails main, so this should have been caught already. But, it seems our existing tests weren't able to catch this failure, since Rails main builds only have a single unrelated failure even before this PR was merged. 🤔

PR #1707, opened 2023-11-03
Fix engine loading for Rails < 7.2

### LINKED ISSUE #1706: Gem RBI generation is missing Rails engines files after upgrading to v0.11.10
Like it says on the tin... I bumped Tapioca to v0.11.10 and ran `tapioca gem` and all (or at least many) files loaded from Rails engines are now ignored and no longer part of the generated RBIs.

E.g. we're using the [maintenance_tasks](https://github.com/Shopify/maintenance_tasks) gem, and now `sorbet/rbi/gems/maintenance_tasks@2.3.3.rbi` is missing definitions for `MaintenanceTasks::ApplicationController`, `MaintenanceTasks::Task`, etc.

AFAICT, same thing happens for all of our gems that include Rails engines (blazer, console1984, audits1984, turbo-rails...).

I assume this is caused by #1702 somehow. We are using Rails 7.0.8.

cc @vinistock

vinistock:
### Motivation

Closes #1706

In #1702, I was trying to avoid having to spread a bunch of conditionals in our code to handle the Rails 7.2 configuration rename. For some reason, the approach doesn't work properly for Rails < 7.2.

### Implementation

Switched the configuration patching to just use conditionals where we look for eager load paths.

### Tests

We have a [specific test](https://github.com/Shopify/tapioca/blob/1d509df3c68077254110a622d78f8e6530ad3e52/spec/tapioca/cli/gem_spec.rb#L1502) for verifying that RBIs include classes defined under the `app` folder in engines. I'm not sure why this isn't failing in the Rails main build.

I verified that this works on Rails 7.2 and older manually.


lib/tapioca/loaders/loader.rb
```diff
@@ -181,19 +191,6 @@ def engines
           .reject { |engine| gem_in_app_dir?(project_path, engine.config.root.to_path) }
       end
 
-      # Rails 7.2 renamed `eager_load_paths` to `all_eager_load_paths`, which maintains the same original functionality.
-      # The `eager_load_paths` method still exists, but doesn't return all paths anymore and causes Tapioca to miss some
-      # engine paths. The following commit is the change:
-      # https://github.com/rails/rails/commit/ebfca905db14020589c22e6937382e6f8f687664
```
andyw8, line 187: I think it would be useful to keep this reference in a comment?

lib/tapioca/loaders/loader.rb
```diff
@@ -112,7 +110,14 @@ def load_engines_in_zeitwerk_mode
         autoloader = Zeitwerk::Loader.new
 
         engines.each do |engine|
-          engine.config.all_eager_load_paths.each do |path|
+          # https://github.com/rails/rails/commit/ebfca905db14020589c22e6937382e6f8f687664
+          eager_load_paths = if Rails::VERSION::MAJOR >= 7 && Rails::VERSION::MINOR >= 2
+            engine.config.all_eager_load_paths
+          else
+            engine.config.eager_load_paths
+          end
```
paracycle, line 118: I am not a huge fan of version checks if we don't have to, feature detection is always nicer. 

Can we instead check for the existence of the `all_eager_load_paths` method instead and encapsulate the lookup in a method?

```suggestion
          eager_load_paths = eager_load_paths(engine)
```
and

```ruby
def eager_load_paths(engine)
  config = engine.config

  (config.respond_to?(:all_eager_load_paths) && config.all_eager_load_paths) || config.eager_load_paths
end
```

vinistock: Agreed. I made the change.

However, we still need to duplicate the logic for the symbol loader, which I don't exactly love.

vinistock:
I also upgraded `rbi` and restricted the `prism` version temporarily. There's a bug in [how we were generating the Prism RBIs](https://github.com/ruby/prism/pull/1767) and until that is fixed we can't use Prism v0.16 or higher in Tapioca (because the `tapioca gem --verify` would fail).

lib/tapioca/loaders/loader.rb
paracycle: Yes, that's not great, but I think the real fix is to extract an engine handler class to unify the logic. Not really a concern that needs to be addressed by this PR, though.

vinistock:
Note: the latest RBI has the fixes we need and I opened a PR to upgrade our dependency #1708, which I can merge immediately after this one.

paracycle:
> because the `tapioca gem --verify` would fail

I don't quite get why this would be the case. Can you explain more?

paracycle (approved):

PR #1714, opened 2023-11-09
Restore `type_for(type)` function

antoine-malliarakis:
### Motivation

Following https://github.com/Shopify/tapioca/pull/1622 the function `Tapioca::Dsl::Helpers::GraphqlTypeHelper#type_for` required an `GraphQL::Schema::Argument` (instead of a `GraphQL::Schema::Wrapper`) to be passed.

Though this did make perfectly sense, we actually used that method to explicitly compute types in our DSL process without this being necessarily "for a field". For example it could be for union types.

### Implementation

What I did was:

1. Rework the `type_for` method to be backward compatible with the pre https://github.com/Shopify/tapioca/pull/1622 code (e.g. keep accepting a sole `GraphQL::Schema::Wrapper` argument as a parameter)
2. Introduce an explicit `type_for_argument` which would be redirecting appropriately to keep https://github.com/Shopify/tapioca/pull/1622 feature

### Tests

I added tests for some introduced adherence.

### DIFF
--- a/lib/tapioca/dsl/compilers/graphql_input_object.rb
+++ b/lib/tapioca/dsl/compilers/graphql_input_object.rb
@@ -55,7 +55,10 @@ module Tapioca
           root.create_path(constant) do |input_object|
             arguments.each do |argument|
               name = argument.keyword.to_s
-              input_object.create_method(name, return_type: Helpers::GraphqlTypeHelper.type_for(argument, constant))
+              input_object.create_method(
+                name,
+                return_type: Helpers::GraphqlTypeHelper.type_for_argument(argument, constant),
+              )
             end
           end
         end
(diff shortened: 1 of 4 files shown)


antoine-malliarakis:
Flagging as draft as there were other refactorings which made this PR actually outdated. Will rework.

antoine-malliarakis:
@KaanOzkan rebased at head to take into account latest modifications. Not too impacting it seems.

@vinistock do you think you could have a look at this PR eventually ? (Or is there someone else I would need to contact to that extent ?)

vinistock (approved):
Changes look good to me. There are some CI failures though

antoine-malliarakis:
> Changes look good to me. There are some CI failures though

I believe I fixed the linting (or more specifically, the type check error). I'm not too sure about how to reproduce the errors reported for the other ruby versions though. Still checking.

KaanOzkan:
Failures were unrelated to this PR

PR #1720, opened 2023-11-23
 Support for custom GraphQL input preparation (Phase one)

DougEdey:
### Motivation

@eapache-opslevel kinda nerd sniped me on this one, I thought "how hard could it be?"

This just tackles the situation where there's a `prepare` method with a signature, there is a default `prepare` method: https://github.com/rmosolgo/graphql-ruby/blob/565b62f13d3cdd7d2cd880e82d189ab64278b1e8/lib/graphql/schema/input_object.rb#L47-L56

I am unsure if there's a way to define a signature if you use the `prepare: -> (input, context) {}` lambda style `prepare` definition though.

### Implementation
By default `prepare` will return the current object, so I kept that behaviour in this MR, if you define a `prepare` method, without a signature, we assumed you're returning the current object.

If you add a signature to the overriden prepare method, we'll use the return type

I'm not a big fan of my weird nesting in `type_for_constant`, I'm happy for feedback to improve that code.

### Tests

_Yes! Yes I did!_

### DIFF
--- a/lib/tapioca/dsl/helpers/graphql_type_helper.rb
+++ b/lib/tapioca/dsl/helpers/graphql_type_helper.rb
@@ -19,6 +19,7 @@ module Tapioca
           else
             argument.type
           end
+
           unwrapped_type = type.unwrap
 
           parsed_type = case unwrapped_type
@@ -78,6 +79,13 @@ module Tapioca
 
         sig { params(constant: Module).returns(String) }
         def type_for_constant(constant)
+          if constant.instance_methods.include?(:prepare)
+            prepare_method = constant.instance_method(:prepare)
+            if (prepare_signature = Runtime::Reflection.signature_of(prepare_method))
+              return prepare_signature.return_type&.to_s if prepare_signature.return_type
+            end
+          end
+
           Runtime::Reflection.qualified_name_of(constant) || "T.untyped"
         end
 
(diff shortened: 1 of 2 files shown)


lib/tapioca/dsl/helpers/graphql_type_helper.rb
```diff
@@ -78,6 +79,13 @@ def type_for(argument)
 
         sig { params(constant: Module).returns(String) }
         def type_for_constant(constant)
+          if constant.instance_methods.include?(:prepare)
+            prepare_method = constant.instance_method(:prepare)
+            if (prepare_signature = Runtime::Reflection.signature_of(prepare_method))
+              return prepare_signature.return_type&.to_s if prepare_signature.return_type
```
KaanOzkan, line 85: I think this could fall into a similar situation as https://github.com/Shopify/tapioca/pull/1659#discussion_r1383597406 with void/NotTyped. Can we account for those and add tests for them by changing the signature of `prepare`?

DougEdey: I added tests (small refactor to the `GraphqlTypeHelper` to make it reusable), the thing to note is that if the `prepare` method has no signature, I'm assuming it's the `InputObject#prepare` method from the graphql-ruby gem which returns the Input type, and not `Untyped`

KaanOzkan (approved):

PR #1723, opened 2023-11-28
Update the workaround for `debug` gem

st0012:
### Motivation

The current approach of disabling `debug` loading makes it impossible to debug Tapioca itself with the gem.

### Implementation

But by lazily activate the debugger, we can avoid the side-effect as well as allowing breakpoints to lazily load the debugger.

It's the same as doing `require "debug/prelude"`, which only defines the breakpoint methods.


Morriar (approved):

shioyama:
Thank you ❤️

PR #1726, opened 2023-12-05
WIP: Use a synthetic include for the CurrentAttributes generator

andyw8:


andyw8:
I verified against Shopify core by running `bin/tapioca dsl`. 18 RBI files were updated, and there the typecheck still passes.

vinistock (approved):

KaanOzkan (approved):

andyw8:
@paracycle I've push some updates to address your feedback, and removed the mention of 'synthetic' from the title/description.

lib/tapioca/dsl/compilers/active_support_current_attributes.rb
```diff
@@ -40,23 +40,29 @@ module Compilers
       # # typed: true
       #
       # class Current
-      #   sig { returns(T.untyped) }
-      #   def self.account; end
+      #   include GeneratedAttributeMethods
       #
-      #   sig { returns(T.untyped) }
-      #   def account; end
+      #   class << self
```
andyw8, line 45: Previously the comment was incorrectly showing the class methods being defined with `def self.`, but they are actually defined within `class << self`.

paracycle (approved):
Thanks!

PR #1729, opened 2023-12-12
Fix deprecation warning (fixture_path -> fixture_paths)

alexevanczuk:
Resolves:

```
ActiveSupport::DeprecationException: DEPRECATION WARNING: TestFixtures.fixture_path= is deprecated and will be removed in Rails 7.2. Use .fixture_paths= instead. (called from use_ui at /Users/alexevanczuk/.rbenv/versions/3.2.0/lib/ruby/site_ruby/3.2.0/rubygems/user_interaction.rb:47) (Parallel::UndumpableException)

```

Still need to add some tests!

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_fixtures.rb
+++ b/lib/tapioca/dsl/compilers/active_record_fixtures.rb
@@ -76,7 +76,12 @@ module Tapioca
             Class.new do
               T.unsafe(self).include(ActiveRecord::TestFixtures)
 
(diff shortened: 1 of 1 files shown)


andyw8:
:wave: Similar to @paracycle's comment [here](https://github.com/Shopify/tapioca/pull/1707#discussion_r1381940997), can we use feature discovery rather than a version check?

KaanOzkan (approved):
Thanks. Existing tests should cover this compiler and we have CI for Rails 6.1 and main. Main isn't fully green but the relevant tests are passing.

PR #1732, opened 2023-12-15
Account for new identity comparison hashes in Sorbet internals

vinistock:
### Motivation

The build in main is still breaking because of the changes in https://github.com/sorbet/sorbet/pull/7531.

Now that the hashes are compared by identity, we can no longer use the `object_id` to figure out if a signature is final or not.

### Implementation

I added a feature check so that we decide what to do based on the Sorbet version.

We couldn't force the hash to be compared by identity without breaking Sorbet's own internals, so I don't think there are any other approaches we can take.

### Tests


dirceu (approved):

lib/tapioca/gem/listeners/sorbet_signatures.rb
```diff
@@ -67,7 +67,11 @@ def compile_signature(signature, parameters)
         sig { params(signature: T.untyped).returns(T::Boolean) }
         def signature_final?(signature)
           modules_with_final = T::Private::Methods.instance_variable_get(:@modules_with_final)
-          final_methods = modules_with_final[signature.owner.object_id]
+          final_methods = if sorbet_supports?(:identity_comparison)
+            modules_with_final[signature.owner]
+          else
+            modules_with_final[signature.owner.object_id]
+          end
```
paracycle, line 74: Could we not have done:
```ruby
final_methods = modules_with_final[signature.owner] || modules_with_final[signature.owner.object_id]
```

amomchilov: Good call!

(For the record, this was fixed in #1734)

PR #1733, opened 2023-12-15
Compiler for `GraphQL::Schema`

amomchilov:
### Motivation

Which creates a strongly-typed sig for `#context`, if the context class was customized via the `context_class`  DSL.

For Shopify employees, please see the timeline for a link to a PR with a sample application of this compiler.

### Implementation

I copied the `GraphQLMutation` compiler and edited it.

### Tests

I copied the tests for the `GraphQLMutation` compiler and edited those, too. :)


amomchilov:
You set the context class on the Schema, but you don't actually ever read it from the Schema 🤦 

I'll rework this to add sigs to the correct classes, like `GraphQL::Schema::Resolver`, `GraphQL::Schema::Mutation`, etc.

PR #1734, opened 2023-12-18
Remove identity comparison feature check in favour of a conditional

vinistock:
### Motivation

Context: https://github.com/Shopify/tapioca/pull/1732/files#r1428547394

Instead of maintaining a feature check, we can just use a conditional and reduce complexity.

### Implementation

Got rid of the feature check and switch to a conditional. Also, left a comment with the PR and version it gets introduced since we may want to remove part of the conditional in the future if we make our Sorbet requirement more strict.

### Tests


paracycle (approved):
Thank you for making the change! 🙏

PR #1739, opened 2023-12-20
Small ActiveRecord column related changes

paracycle:
### Motivation

The main motivation for this PR is to make sure the new Rails 7.1 behaviour of the `id_value` attribute alias being defined too late is accounted for.

After [this change](https://github.com/Shopify/shopify/pull/471447/files#diff-df235247748c2d0fa248f44258702609d6fdd1d5b3afecfb88e394f3623e1458L163-L167) was merged in Rails, the `id_value` attribute alias is only defined while methods are being defined on the model. However, that creates a problem since an eager load of the application causes a different behaviour (since eager load loads the schema and defined the methods on models) than a non-eager load of the application. Thus, depending on how a model is loaded, it could list `id_value` as an alias to `id` in `attribute_aliases`, or not.

The only workaround is to trigger method definitions on each model, which unfortunately costs us some performance, but makes the behaviour predictable.

### Implementation

Call `define_method_attributes` on each model before attempting to discover attribute names and aliases.

I've also moved the check for the existence of `attribute_method_patterns` method to a constant, so that we don't have to keep doing the same check over and over again for each model.

### Tests

No additional tests.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_model_attributes.rb
+++ b/lib/tapioca/dsl/compilers/active_model_attributes.rb
@@ -70,7 +70,7 @@ module Tapioca
 
         sig { returns(T::Array[[::String, ::String]]) }
         def attribute_methods_for_constant
-          patterns = if constant.respond_to?(:attribute_method_patterns)
+          patterns = if Helpers::ActiveRecordColumnTypeHelper::HasAttributeMethodPatternsMethod
             # https://github.com/rails/rails/pull/44367
             T.unsafe(constant).attribute_method_patterns
           else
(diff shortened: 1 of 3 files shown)


lib/tapioca/dsl/compilers/active_model_attributes.rb
```diff
@@ -70,7 +70,7 @@ def gather_constants
 
         sig { returns(T::Array[[::String, ::String]]) }
         def attribute_methods_for_constant
-          patterns = if constant.respond_to?(:attribute_method_patterns)
+          patterns = if Helpers::ActiveRecordColumnTypeHelper::HasAttributeMethodPatternsMethod
```
st0012, line 73: I think the helper needs to be required in this file too to access the constant?

paracycle:
This work was done as part of #1742 in commit https://github.com/Shopify/tapioca/pull/1742/commits/0526dddea91372a75ad28f1906f2c559a65e7945

PR #1740, opened 2023-12-20
Bump our Sorbet dependency to `0.5.10820` to remove lots of compatibility workarounds

paracycle:
### Motivation

Our Sorbet version dependency was dating back to [July 21, 2022](https://rubygems.org/gems/sorbet-runtime/versions/0.5.10187) which meant that we had to carry a lot of patches and workaround to ensure compatibility with all versions since then. Keeping this burden around was adding complexity and well as adding considerable time to our CI runs due to the need to test with older Sorbet versions. This was making development on the gem increasingly hard. We can solve this by enforcing a newer version (but not too new) of Sorbet as our dependency.

I've picked the last version that we have patches or a feature flag check for as the new baseline version, which ended up being 0.5.10820 from [May 12, 2023](https://rubygems.org/gems/sorbet-runtime/versions/0.5.10820) which is already more than 6 months old.

### Implementation

Bump Sorbet version dependency, and remove all workaround. Read commit by commit to see what was removed and why.

### Tests

No additional tests, removed/modified tests that were exercising old behaviour.


lib/tapioca/sorbet_ext/backcompat_patches.rb
```diff
@@ -1,24 +1,7 @@
...
-      class << self
-        def [](type)
-          T.untyped
-        end
-      end
-    end
-  end
-end
+# NOTE: There is currently no need for any patches, so this file is empty.
+# If you add a backcompat patch please remove this comment.
```
andyw8, line 7:
```suggestion
# NOTE: If there is currently no need for any patches, this file will be empty.
```
(then we don't need to re-add/remove the comment in future).

andyw8 (approved):

PR #1742, opened 2023-12-20
Start targeting Rails 7.1 by default

paracycle:
### Motivation

We need to track the latest released version of Rails.

### Implementation

Bump our `Gemfile` dependencies to 7.1 series. Since we only support the latest 2 Rails versions, this also means that we move Rails 6.1 tests to instead target Rails 7.0.

### Tests

No new tests.


lib/tapioca/dsl/compilers/active_record_columns.rb
```diff
@@ -108,6 +108,11 @@ class ActiveRecordColumns < Compiler
         def decorate
           return unless constant.table_exists?
 
+          # We need to call this to ensure that some attribute aliases are defined, e.g.
+          # `id_value` as an alias for `id`.
+          # I think this is a regression on Rails 7.1, but we are where we are.
```
rafaelfranca, line 113: What do you mean regression on Rails 7.1?

rafaelfranca: To future reference. `model.attribute_name`s no returning `id_value` in non-eager loading behavior is a bug in Rails, that is what this comment means

paracycle: Yes, apologies for being cryptic, but I think we should have to trigger method definitions to be able to read the full set of aliases on a model, but the way Rails 7.1 works, we need to do it. That's what I mean by "regression".

PR #1746, opened 2023-12-31
prevent infinite recursion if a type is seen again

tdyas:
### Motivation

The algorithm used to force load all constants is vulnerable to an infinite loop. I hit just such a case when trying to bootstrap sorbet with tapioca in an internal Rails repository at my employer. Running any Tapioca command using this eager load algorithm hung indefinitely after printing that it was loading the Rails application.

I tracked down the issue by vendoring tapioca and adding debug logging statements. The execution continued until `Tapioca::Runtime::Trackers::Autoload.eager_load_all!` at which point the loop continued indefinitely for the same few constant names.

### Implementation

The fix is relatively straightforward: Just maintain a cache of constant names which have already been seen, and skip the processing of a constant name if it was already processed.

### Tests

[WIP] I know which internal code is triggering the issue, but still need to generate a test case from it.

### DIFF
--- a/lib/tapioca/runtime/trackers/autoload.rb
+++ b/lib/tapioca/runtime/trackers/autoload.rb
@@ -18,10 +18,16 @@ module Tapioca
           sig { void }
           def eager_load_all!
             with_disabled_exits do
+              seen_constant_name = {}
               until @constant_names_registered_for_autoload.empty?
                 # Grab the next constant name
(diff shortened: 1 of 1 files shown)


tdyas:
Yeah, I have no idea why the loop triggered. The code in question is internal code in one of our support gems so I will need to do some work to disentangle the code into a standalone reproduction.

PR #1749, opened 2024-01-02
Exclude testing of Ruby 3.0 against Rails main

andyw8:
The minimum is now 3.1.0:
https://github.com/rails/rails/commit/6ba2fdb2fe85751b573aadd05608471daf1a44ff


paracycle:
This is already implemented in the Ruby 3.3 upgrade PR.

PR #1750, opened 2024-01-05
Use types method instead of instance variable in fixed hash patch

vinistock:
### Motivation

The PR https://github.com/sorbet/sorbet/pull/7534 changed the instance variable names to allow for lazy initialization of types. That means the `@types` variable is now always `nil` and our CI is failing with the latest `sorbet-runtime`.

### Implementation

The fix is to use the method `types`. On old versions, it's an attr_reader for `@types` and on new versions it's a method that triggers the lazy initialization of the types.

### Tests

The tests should now pass.


paracycle:
Oh, thanks for the PR, but I think we don't actually need this patch anymore. It was a temporary patch until my upstream PR was merged and widely distributed. I am pretty sure that our minimum version of Sorbet runtime already has this in place. I'll check to make sure, but I am pretty sure that we can just remove this code.

paracycle:
Yes, indeed, the implementation of the method is the same on the minimum version that we enforce: https://github.com/sorbet/sorbet/blob/0.5.10820.20230511164542-d94b3e0b2/gems/sorbet-runtime/lib/types/types/fixed_hash.rb#L84-L94

So, can we update this PR to instead get rid of the patch?

egiurleo (approved):
🥳 Thanks for fixing, just ran into this!

PR #1756, opened 2024-01-10
Skip compiling constants with alias from another gem in namespace

egiurleo:
### Motivation

When compiling a gem, it is possible for the gem to define an alias that includes an alias from another gem in its namespace.

For example, let's say we have gems `foo` and `bar`:

```ruby
# foo.rb
module Foo; end
F = Foo

# bar.rb
module Foo
  module Bar; end
end
F::B = Foo::Bar
```

Then the RBIs might look something like this:
```ruby
# foo.rbi
module Foo; end
F = Foo

# bar.rbi
module Foo::Bar; end
F::B = Foo::Bar
Foo::B = Foo::Bar
```
There are two issues with this RBI:
1. There will be a Sorbet error because the alias F is defined in the RBI for both gems (once in the line `F = Foo` and another time in the line `F::B = Foo::Bar`)
2. The alias for `Foo::Bar` is defined twice in `bar.rbi`

This commit will cause the RBI for gem `bar` to skip over the line `F::B = Foo::Bar`, which will prevent the alias `F` from being defined twice in RBI.

### More Context
I found this issue when trying to generate RBI for the `eventmachine` and `em-synchrony` gems. `eventmachine` defines an alias `EM = EventMachine`, and `em-synchrony` defines another alias `EM::S = EventMachine::Synchrony`. The resulting RBI files cause a Sorbet error because `EM` is defined twice.

### Implementation
I implemented this change by adding an additional check to the `skip_alias?` method that determines whether the namespace of a constant contains an alias and skips compiling that constant if so.

### Tests
I added one test that fails without this change and passes with it. All other tests continue to pass. I have tested this in Shopify core and verified no new Sorbet errors.


lib/tapioca/runtime/reflection.rb
```diff
@@ -255,6 +255,35 @@ def child_module_for_parent_with_name(parent, name)
...
+      sig { params(symbol: String).returns(String) }
+      def namespace_of(symbol)
+        symbol_arr = symbol.split("::")
+        return "" if symbol_arr.length == 1
+
+        T.must(symbol_arr[0...-1]).join("::")
+      end
+
+      sig { params(symbol: String).returns(T::Boolean) }
+      def contains_alias?(symbol)
```
egiurleo, line 268: @paracycle Pointed out that there's already a [method in tapioca that basically does this](https://github.com/Shopify/tapioca/blob/1e3c9ac5ddec6a49b10486095be0baf11e5832b3/lib/tapioca/gem/pipeline.rb#L438-L448) and I could just adapt that to fit this use case. I'll reopen the PR once that's done!

paracycle, line 268: Can we rename this to be:
```suggestion
      def has_aliased_namespace?(name)
```

lib/tapioca/runtime/reflection.rb
```diff
@@ -255,6 +255,35 @@ def child_module_for_parent_with_name(parent, name)
       def method_defined_by_forwardable_module?(method)
         method.source_location&.first == Object.const_source_location(:Forwardable)&.first
       end
+
+      sig { params(symbol: String).returns(String) }
+      def namespace_of(symbol)
```
paracycle, line 260: This doesn't need to be a separate method, we can fold it into how the other method operates.

lib/tapioca/runtime/reflection.rb
```diff
@@ -255,6 +255,35 @@ def child_module_for_parent_with_name(parent, name)
...
+          dealiased_name = name_of(constant)
+          is_alias = dealiased_name != full_constant_name
+
+          return true if is_alias
+
+          namespace = dealiased_name
+        end
+
+        false
+      end
```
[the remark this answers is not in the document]
egiurleo: Worked perfectly! And it's much more elegant.

lib/tapioca/runtime/reflection.rb
```diff
@@ -255,6 +255,21 @@ def child_module_for_parent_with_name(parent, name)
...
+        name_parts = name.split("::")
+        name_parts.pop # drop the constant name, leaving just the namespace
+
+        name_parts.each_with_object([]) do |name_part, namespaces|
+          namespaces << "#{namespaces.last}::#{name_part}".delete_prefix("::")
+        end.any? do |namespace|
+          constant = constantize(namespace)
+          next unless Module === constant
+
+          name_of(constant) != namespace
```
Morriar, line 270: Should we extract this as a private `constant_name_differs_from_actual_namespace` method to facilitate reading? Maybe just adding a comment in the block would help.

Morriar (approved):

PR #1757, opened 2024-01-11
Remove duplicate block parameter in ActiveJob#perform_later

jenshenny:
### Motivation

If an `ActiveJob#perform` contains a `&block` argument, tapioca will generate a duplicate block argument.

### Before
```
class TestJob
  class << self
    sig do
      params(
        args: T.untyped,
        _arg1: T.untyped,
        block: T.untyped,
        block: T.nilable(T.proc.params(job: TestJob).void)
      ).returns(T.any(TestJob, FalseClass))
    end
    def perform_later(*args, **_arg1, &block, &block); end

    sig { params(args: T.untyped, _arg1: T.untyped).returns(T.untyped) }
    def perform_now(*args, **_arg1); end
  end
end
```

### After
```
class TestJob
  class << self
    sig do
      params(
        args: T.untyped,
        _arg1: T.untyped,
        block: T.nilable(T.proc.params(job: TestJob).void)
      ).returns(T.any(TestJob, FalseClass))
    end
    def perform_later(*args, **_arg1, &block); end

    sig { params(args: T.untyped, _arg1: T.untyped).returns(T.untyped) }
    def perform_now(*args, **_arg1); end
  end
end
```

### Implementation

In `perform_later_parameters`, it'll remove any existing block argument before adding one in. Not sure if there's a better way to approach it.

### Tests

Added a test case where the Job parameter has a block parameter.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_job.rb
+++ b/lib/tapioca/dsl/compilers/active_job.rb
@@ -83,6 +83,7 @@ module Tapioca
         end
         def perform_later_parameters(parameters, constant_name)
           if ::Gem::Requirement.new(">= 7.0").satisfied_by?(::ActiveJob.gem_version)
+            parameters.reject! { |typed_param| RBI::BlockParam === typed_param.param }
             parameters + [create_block_param(
               "block",
               type: "T.nilable(T.proc.params(job: #{constant_name}).void)",
--- a/spec/tapioca/dsl/compilers/active_job_spec.rb
+++ b/spec/tapioca/dsl/compilers/active_job_spec.rb
@@ -115,6 +115,36 @@ module Tapioca
               RBI
               assert_equal(expected, rbi_for(:NotifyJob))
             end
+
+            it "generates correct RBI file for subclass with block argument" do
+              add_ruby_file("job.rb", <<~RUBY)
+                class NotifyJob < ActiveJob::Base
+                  def perform(user_id, &blk)
+                    # ...
+                  end
+                end
+              RUBY
+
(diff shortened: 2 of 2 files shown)


lib/tapioca/dsl/compilers/active_job.rb
```diff
@@ -83,6 +83,7 @@ def decorate
         end
         def perform_later_parameters(parameters, constant_name)
           if ::Gem::Requirement.new(">= 7.0").satisfied_by?(::ActiveJob.gem_version)
+            parameters.reject! { |typed_param| typed_param.param.name == "block" }
```
paracycle, line 86: There is no guarantee that the original block parameter will be called `"block"`, so this won't work in the general case.

spec/tapioca/dsl/compilers/active_job_spec.rb
```diff
@@ -115,6 +115,36 @@ def perform_now(user_id); end
               RBI
               assert_equal(expected, rbi_for(:NotifyJob))
             end
+
+            it "generates correct RBI file for subclass with block argument" do
+              add_ruby_file("job.rb", <<~RUBY)
+                class NotifyJob < ActiveJob::Base
+                  def perform(user_id, &block)
```
paracycle, line 122: Can we add a case where the block parameter is NOT named `block` as well?

jenshenny: I updated the current test with a block parameter not called `block`. I don't see why we need two tests for the same behaviour (checking if the param is a block param).

Issue #1754, opened 2024-01-09
IdentityCache compiler misses a fetch_multi_* 


bitwise-aiden:
### Description

When using `cache_attribute` with multiple `by` fields, the resulting `fetch_multi_field_a_by_field_b_and_field_c` method is not created. 

It appears as though there is a [length check](https://github.com/Shopify/tapioca/blob/main/lib/tapioca/dsl/compilers/identity_cache.rb#L241) that is preventing this from being generated. Looking back to [where it appears to be introduced](https://github.com/Shopify/tapioca/pull/69), there isn't clear indication of what this check was intended for. 

One approach to solving could be removing the check and seeing what breaks, or adding a `create_attribute_fetch_by_methods` method to create for this case.

PR #1758, opened 2024-01-11
Update IdentityCache compiler to generate correct multi methods

bitwise-aiden:
### Motivation
Closes: https://github.com/Shopify/tapioca/issues/1754

### Implementation

While researching this, I noticed that there were some missing methods for the default `cache_index` case as well. So as @KaanOzkan suggested in the issue, removing the `length` check fixed both issues. 

### Tests
I've added some new tests and fixed old ones to illustrate the cases we are now expecting.


lib/tapioca/dsl/compilers/identity_cache.rb
```diff
@@ -165,6 +165,10 @@ def create_fetch_by_methods(field, klass)
           create_index_fetch_by_methods(field, klass) if is_cache_index
         end
 
+        # cache_index 1
+        # cache_index 2+
+        # cache_attribute n
+
```
vinistock, line 171: Do we need these?

vinistock (approved):

paracycle (approved):

PR #1762, opened 2024-01-15
Check if prepare argument is a symbol or string

KaanOzkan:
### Motivation

Leftover from the refactor https://github.com/Shopify/tapioca/pull/1714/files#diff-8b787801eb4d5ed2047d8b5d37043650443fae0500ae270f8d7ededeb311f27cL68. It causes TypeError's when the argument isn't a symbol or a string.

### Implementation

Bring back the previous check, call `to_sym` to please Sorbet.

### Tests

Added

### DIFF
--- a/lib/tapioca/dsl/helpers/graphql_type_helper.rb
+++ b/lib/tapioca/dsl/helpers/graphql_type_helper.rb
@@ -26,9 +26,9 @@ module Tapioca
           end
 
           prepare = argument.prepare
-          prepare_method = if prepare
+          prepare_method = if prepare.is_a?(Symbol) || prepare.is_a?(String)
             if constant.respond_to?(prepare)
-              constant.method(prepare)
+              constant.method(prepare.to_sym)
             end
           end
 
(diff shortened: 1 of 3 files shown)


lib/tapioca/dsl/helpers/graphql_type_helper.rb
```diff
@@ -26,9 +26,9 @@ def type_for_argument(argument, constant)
           end
 
           prepare = argument.prepare
-          prepare_method = if prepare
+          prepare_method = if prepare.is_a?(Symbol) || prepare.is_a?(String)
             if constant.respond_to?(prepare)
-              constant.method(prepare)
+              constant.method(prepare.to_sym)
```
andyw8, line 31: I think `method` takes either?

```
irb(main):002:0> String.new.method(:length)
=> #<Method: String#length()>
irb(main):003:0> String.new.method("length")
=> #<Method: String#length()>
```

KaanOzkan: https://github.com/sorbet/sorbet/pull/7602

PR #1765, opened 2024-01-15
Stop requiring stuff from DSL compilers

### LINKED ISSUE #1759: Support for granular `rails` framework usage
Some Rails application may choose not to `require 'rails/all'`, and instead require the various Rails frameworks more granularly (e.g. maybe `activerecord` is not needed). Rails is different from most other gems/ecosystems in that the recommended way forward for applications is to continue to depend on the `rails` gem, but instead change the `require` statements in `config/application.rb`, rather than changing the application to actually include the individual gems in `Gemfile`.

Currently, Tapioca does not serve this use case well, requiring workaround for certain features.

For example,

- If using annotations, all the framework annotations will be downloaded. Tapioca does not support an exclusion mechanism for annotations, requiring the host app to either not use annotations, or manually delete the ones it does not want.
- If using `DSLs`, many of the frameworks will be `require`d by DSL compilers, which may lead to DSL RBIs being generated for some of their constants, even if the application does not use them. For example, the compiler which generates RBI for classes which use an `ActiveSupport::Concern` may generate RBI for various `ActiveRecord` modules, even if the host application does not use `ActiveRecord`. The host application may not want to disable that compiler, if it is using `ActiveSupport::Concern` in other places, leading to it needing to manually delete those RBI files (which then breaks `tapioca dsl --verify`).

## How will we know this is fixed?

This problem is addressed if Tapioca offers a way to support such application in which the host application does not need to alter the RBI generated by Tapioca in any way. It is acceptable for the host application to need to provide config in `sorbet/tapioca/config.yml` to configure Tapioca's behaviour (ideally this would be well documented).

## How is this currently addressed?

An application at Shopify has worked around these limitations in the following way:

- Adding a post RBI generation step in which unwanted annotations and DSL RBI files are removed.
- Wrapping `tapioca dsl --verify` in another script which scans the output, and if the only errors are the missing RBI which were delete, pretends `tapioca dsl --verify` succeeded and exits `0`.

paracycle:
### Motivation

Fix #1759

DSL compilers should not change the object space of the application they run in, and thus should not really be doing any requires. This commit removes all requires from DSL compilers, and instead relies on the application (or, in our case, the tests) to require the necessary files.

A DSL compiler should be responsible for checking if the constants it is expecting to have been loaded are actually loaded, and if not, it should not be defining the DSL compiler class at all.

### Implementation

- Remove all the `require` lines from DSL compilers, replace them with early returns guarding for `defined?` for relevant constants.
- Add necessary `require` lines inside `before_setup` methods in tests, so that they get required after we fork (for isolation), but before we load the actual DSL compiler class.

### Tests

Updates tests


tasks/generator_documentation.rake
```diff
@@ -107,7 +100,7 @@ task generate_dsl_documentation: :yard_for_generate_documentation do
   end
 
   def main
-    YARD::Registry.load!
+    Dir.glob("lib/tapioca/dsl/{compiler.rb,compilers/**/*.rb}").each { YARD.parse(_1) }
```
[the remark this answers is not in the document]
paracycle: Ah neat!

sambostock:
When do the compiler **extensions** get loaded? I see they still use the "attempt to require, or return" approach, but maybe that's fine?

lib/tapioca/dsl/compilers/aasm.rb
```diff
@@ -1,12 +1,7 @@
 # typed: strict
 # frozen_string_literal: true
 
-begin
-  require "active_record"
-  require "aasm"
-rescue LoadError
-  return
-end
+return unless defined?(AASM)
```
vinistock, line 4: This was requiring Active Record as well. Does the compiler need `ActiveRecord` to be defined too?

lib/tapioca/dsl/compilers/action_text.rb
```diff
@@ -1,11 +1,7 @@
 # typed: strict
 # frozen_string_literal: true
 
-begin
-  require "action_text"
-rescue LoadError
-  return
-end
+return unless defined?(ActiveRecord::Base)
```
vinistock, line 4: Should this check for action text as well?

lib/tapioca/dsl/compilers/url_helpers.rb
```diff
@@ -1,13 +1,7 @@
...
 # frozen_string_literal: true
 
-begin
-  require "rails"
-  require "action_controller"
-  require "action_view"
-rescue LoadError
-  return
-end
+return unless defined?(Rails) && defined?(ActionDispatch::IntegrationTest) && defined?(ActionView::Helpers)
```
vinistock, line 4: Why the integration test? 🤔

Morriar: We have a direct reference to it [lower](https://github.com/Shopify/tapioca/pull/1765/files#diff-180fc447bbffd314ea8493148a280ac33a13b94c4006aacb339daaa97ec45b1dR99) but I wonder if the check could be moved there 🤔

lib/tapioca/dsl/compilers/active_record_fixtures.rb
```diff
@@ -62,7 +55,7 @@ class << self
 
           sig { override.returns(T::Enumerable[Module]) }
           def gather_constants
-            return [] unless Rails.application
+            return [] unless defined?(Rails.application) && Rails.application
```
andyw8, line 58: Why is `return [] unless defined?(Rails.application)` not sufficient here?

andyw8:
Would it be worth adding a cop to prevent any accidental future use of `require` within DSL compilers?

paracycle:
> ~When do the compiler **extensions** get loaded? I see they still use the "attempt to require, or return" approach, but maybe that's fine?~
> 
> _Nevermind, I just missed them somehow..._ 🤦‍♂️

You actually make a great point. I need to change how extensions work, since they are loaded _before_ the application is loaded. As things stand, these extensions just won't work.

paracycle:
Ok, I've reworked the extensions to do lazy loading, and made the `UrlHelpers` compiler to work in a more fine-grained way. This PR should be ready for another round of reviews.

lib/tapioca/dsl/compilers/active_record_fixtures.rb
```diff
@@ -1,14 +1,7 @@
...
 
-begin
-  require "rails"
-  require "active_record"
-  require "active_record/fixtures"
-  require "active_support/test_case"
-rescue LoadError
-  return
-end
+return unless defined?(Rails) && defined?(ActiveSupport::TestCase) && defined?(ActiveRecord::TestFixtures)
```
Morriar, line 4: I wonder if we could define a `defined?(*args)` in self?

Morriar (approved):

PR #1769, opened 2024-01-17
[WIP] Proposal to switch to 'conventional' spec syntax

andyw8:
View with whitespace changes ignored: https://github.com/Shopify/tapioca/compare/andyw8/avoid-class-for-specs?w=1

There was some discussion in https://github.com/Shopify/ruby-lsp/pull/1265 on whether mixing `class` and `spec` styles in Minitest was valid. I've opened this to show what would be involved in updating Tapioca if we were to switch to a 'conventional `spec` approach.

Overall it seems fairly straightforward. One downside is that we need to fully qualify some constant references.

cc @st0012

### DIFF
--- a/spec/executor_spec.rb
+++ b/spec/executor_spec.rb
@@ -3,76 +3,72 @@
 
 require "spec_helper"
 
-module Tapioca
-  class ExecutorSpec < Minitest::Spec
-    describe "Tapioca::Executor" do
-      before do
-        @queue = T.let((0...8).to_a, T::Array[Integer])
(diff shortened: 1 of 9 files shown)


paracycle:
Does this not break the Sorbet rewriter that processes spec style tests?

paracycle:
Based on https://github.com/Shopify/ruby-lsp/pull/1265#issuecomment-1939207483 I am going to close this.

This is a legitimate usage of the library and without these classes, we can't use our shared superclasses and all type-checking will fail.

PR #1775, opened 2024-01-24
WIP: Fix current attributes compiler

andyw8:
Due to https://github.com/rails/rails/commit/2cd4abcc876bdbe3adc2cbbfeee9712fc2bc0601

Note: This needs to be conditional so it also supports Rails < 7.2.

### DIFF
--- a/lib/tapioca/rbi_ext/model.rb
+++ b/lib/tapioca/rbi_ext/model.rb
@@ -100,7 +100,7 @@ module RBI
       )
       parameters.each do |param|
         method << param.param
-        sig << RBI::SigParam.new(param.param.name.delete_prefix("&"), param.type)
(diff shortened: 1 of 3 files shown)


lib/tapioca/rbi_ext/model.rb
```diff
@@ -100,7 +100,7 @@ def create_method(name, parameters: [], return_type: "T.untyped", class_method:
       )
       parameters.each do |param|
         method << param.param
-        sig << RBI::SigParam.new(param.param.name, param.type)
+        sig << RBI::SigParam.new(param.param.name.delete_prefix("&"), param.type)
```
paracycle, line 103: You shouldn't need this at all.

paracycle:
Implemented by #1778

PR #1776, opened 2024-01-24
Update `ActiveSupportCurrentAttributes` compiler to include block param on getters

egiurleo:
### Motivation
I'm working on regenerating RBIs on core, and there is a change where classes that inherit from `CurrentAttributes` are having `block` params added to their attribute getter methods. See this example from the `socket-socker` gem RBI:

Screenshot 2024-01-24 at 3 24 55 PM

This is causing an issue because these same classes have the methods re-defined in DSL RBI files, generated by the `ActiveSupportCurrentAttributes` compiler, without these block parameters. This causes a type checking error, which is preventing me from fully updating RBIs.

### Implementation
I updated the `ActiveSupportCurrentAttributes` DSL compiler to add this block param to getter methods.

### Tests
I modified the existing compiler tests.

### Questions
I'm not sure I understand why the gem RBIs are changing to have a block param added. Is that expected behavior? There's nothing that jumps out to me in the [current attributes source code](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/current_attributes.rb), but I'll keep looking.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_support_current_attributes.rb
+++ b/lib/tapioca/dsl/compilers/active_support_current_attributes.rb
@@ -125,7 +125,12 @@ module Tapioca
               return_type: "T.untyped",
             )
           else
-            klass.create_method(method, class_method: class_method, return_type: "T.untyped")
+            klass.create_method(
+              method,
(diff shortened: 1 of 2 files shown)


andyw8:
https://github.com/Shopify/tapioca/pull/1775

PR #1791, opened 2024-02-07
Simplify gemfiles

paracycle:
### Motivation

The current way of keeping extra gemfiles in sync with the main gemfile is more messy and leads to problems if we forget to update gem dependencies in all of them.

### Implementation

Instead of keeping separate gemfiles around, we can pass the desired Rails version using an env variable. This allows us to use a single gemfile that is able to build against multiple Rails versions on CI.

The only caveat is that we need to remove the lockfile if we are running against a non-current Rails version, which at the moment is Rails 7.1. This is needed since Setup Ruby installs the bundle using `--deployment` which prevents the lockfile from changing, thus breaking the ability to change Rails version changes. Removing the lockfile before installing the bundle, solves that problem. Moreover, this ends up being the same behaviour as before, since with multiple gemfiles, we never kept their lockfiles in the repository either, so they would end up creating their lockfiles from scratch every time.

### Tests

No new tests.


KaanOzkan (approved):

paracycle:
@KaanOzkan I made a small extra change to centralize the string "7.1" meaning "current" Rails into the Gemfile. Would be great if you could review again, or review just the last commit.

Gemfile
```diff
@@ -19,11 +21,17 @@ group :deployment, :development do
 end
 
 group :development, :test do
+  if rails_version == "main"
+    gem "rails", github: "rails/rails", branch: "main"
+  else
+    gem "rails", "~> #{rails_version}.0"
```
Morriar, line 29: Why do we anchor it on `.0`? Can't we just pass the whole version through `RAILS_VERSION`?

.github/workflows/ci.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       fail-fast: false
       matrix:
         ruby: ["3.0", "3.1", "3.2", "3.3", "head"]
-        rails: ["7.0", "7.1", "main"]
+        rails: ["7.0", "current", "main"]
```
Morriar, line 40: It doesn't read that well with this `current` change compared to `7.1`. What's the goal of this change?

Morriar (approved):

PR #1792, opened 2024-02-08
Document exporting signatures using `rbi/` folder

KaanOzkan:
### Motivation

Looking at the [implementation](https://github.com/Shopify/tapioca/pull/453) I don't think this is documented anywhere.


vinistock (approved):

README.md
```diff
@@ -40,6 +40,7 @@ Tapioca makes it easy to work with [Sorbet](https://sorbet.org) in your codebase
     * [Excluding a gem from RBI generation](#excluding-a-gem-from-rbi-generation)
     * [Changing the strictness level of the RBI for a gem](#changing-the-strictness-level-of-the-rbi-for-a-gem)
     * [Keeping RBI files for gems up-to-date](#keeping-rbi-files-for-gems-up-to-date)
+    * [Exporting signatures using `rbi/` folder](#exporting-signatures-using-rbi-folder)
```
paracycle, line 43: All of our documentation is written from the point of view of the consumer of types via RBI files, but this topic is written from the gem author's point of view, which is confusing.

I think what we are really trying to say here, for Tapioca users, is that Tapioca can use the signatures exported from gems in `rbi/` folders. So, I think we should word this, and the paragraph below to read that way.

paracycle (changes requested):

README.md
paracycle: The case for gem authors, and the fact that they can export an `/rbi` folder is already documented by Sorbet. I think we should be documenting that Tapioca will automatically import and merge those signatures when the gem exports them. Also, that one can opt out of that behaviour using the `--no-exported-gem-rbis` flag.

paracycle (approved):
Thank you!

PR #1793, opened 2024-02-12
Remove `pry`/`pry-byebug` and start using `irb`/`debug` exclusively

paracycle:
### Motivation

Since `irb` and `debug` can easily replace `pry` and `pry-byebug`, there is no reason to keep onto the old tools.

### Implementation

Remove all usages of `pry` and `pry-byebug` from the Gemfile, and make `bin/console` work with `irb` using a special name.

### Tests

No need for test changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,18 +223,18 @@ For gems that have a normal default `require` and that load all of their constan
 For example, suppose you are using the class `BetterHtml::Parser` exported from the `better_html` gem. Just doing a `require "better_html"` (which is the default require) does not load that type:
 
 ```shell
-$ bundle exec pry
+$ bundle exec irb
 
-[1] pry(main)> require 'better_html'
+irb(main):001> require 'better_html'
 => true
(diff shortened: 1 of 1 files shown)


st0012 (approved):
🎉

Gemfile
```diff
@@ -10,12 +10,11 @@ rails_version = ENV.fetch("RAILS_VERSION", CURRENT_RAILS_VERSION)
...
-gem "debug", require: false
-gem "pry"
-gem "pry-byebug"
-gem "rubocop-shopify", require: false
+gem "debug"
+gem "irb"
+gem "rubocop-shopify"
 gem "rubocop-sorbet", ">= 0.4.1"
-gem "rubocop-rspec", require: false
+gem "rubocop-rspec"
```
andyw8, line 17: Doesn't removing the `require: false` cause `bundle exec` to be slower?

PR #1794, opened 2024-02-12
Make unprocessable compilers a warning instead of an error

### LINKED ISSUE #1786: `tapioca dsl` rejects valid values from `--exclude` if they are not loaded
I use some automation scripts to simplify managing Sorbet across modules in a monorepo. 
One of these scripts uses the command `tapioca dsl --exclude ActiveRecordFixtures UrlHelpers` 

As of 0.12.0, this command still works fine for the modules that load Rails in its entirety; but on a few simpler gems, the command now fails with:
```
Error: Cannot find compiler 'ActiveRecordFixtures'
Error: Cannot find compiler 'UrlHelpers'
```

Since `tapioca dsl --list-compilers` does not list these entries when run against the impacted gems, modifying my script to check for the available compilers was easy enough; but it would be nice if the `--exclude` flag would accept allowable compilers even when they are not loaded.

paracycle:
### Motivation

Fix #1786

### Implementation

Since we change if the compilers are loaded depending on if various constants are defined, running the same `tapioca dsl` command in different contexts could end up loading a different set of compilers. However, this creates a problem if users are passing in `--only` or `--exclude` for compilers with a predetermined set of compiler names, since the set of compiler names could change between runs.

It is better to show a warning that the requested compilers are not available, and continue with the rest of the DSL generation process, rather than erroring out.

### Tests

Updates existing tests

### DIFF
--- a/lib/tapioca/commands/abstract_dsl.rb
+++ b/lib/tapioca/commands/abstract_dsl.rb
@@ -175,10 +175,11 @@ module Tapioca
 
         unless unprocessable_compilers.empty?
           message = unprocessable_compilers.map do |name, _|
-            set_color("Error: Cannot find compiler '#{name}'", :red)
+            set_color("Warning: Cannot find compiler '#{name}'", :yellow)
           end.join("\n")
 
-          raise Thor::Error, message
+          say(message)
+          say("")
         end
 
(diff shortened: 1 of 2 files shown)


egiurleo (approved):
This is a good solution!

PR #1799, opened 2024-02-20
Add support for overloaded signatures and add overloaded ActiveRecordRelations sigs

### LINKED ISSUE #1789: Is there a way to write a compiler that defines multiple `sig` for a single method?
For example for `T::Array` there are multiple overloads/sigs for `first` and `first(limit)`. I'd like to be able to do the same with a compiler.

There's `create_method`. I tried calling it twice with the same method name but that ends up creating two methods with two sigs, and Sorbet takes the last definition. I couldn't find anything in the API to do this... but maybe it's possible?

bdewater:
### Motivation
Close #1789

### Implementation
Based on the patch in aforementioned issue.

### Tests
Updated tests.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_relations.rb
+++ b/lib/tapioca/dsl/compilers/active_record_relations.rb
@@ -709,51 +709,121 @@ module Tapioca
             case method_name
             when :find_each
               order = ActiveRecord::Batches.instance_method(:find_each).parameters.include?([:key, :order])
-              create_common_method(
+              sigs = [
+                common_relation_methods_module.create_sig(
+                  parameters: [
+                    create_kw_opt_param("start", type: "T.untyped", default: "nil"),
+                    create_kw_opt_param("finish", type: "T.untyped", default: "nil"),
+                    create_kw_opt_param("batch_size", type: "Integer", default: "1000"),
+                    create_kw_opt_param("error_on_ignore", type: "T.untyped", default: "nil"),
+                    *(create_kw_opt_param("order", type: "Symbol", default: ":asc") if order),
+                    create_block_param("block", type: "T.proc.params(object: #{constant_name}).void"),
+                  ],
+                  return_type: "void",
+                ),
+                common_relation_methods_module.create_sig(
+                  parameters: [
+                    create_kw_opt_param("start", type: "T.untyped", default: "nil"),
+                    create_kw_opt_param("finish", type: "T.untyped", default: "nil"),
+                    create_kw_opt_param("batch_size", type: "Integer", default: "1000"),
+                    create_kw_opt_param("error_on_ignore", type: "T.untyped", default: "nil"),
+                    *(create_kw_opt_param("order", type: "Symbol", default: ":asc") if order),
+                  ],
+                  return_type: "T::Enumerator[#{constant_name}]",
+                ),
+              ]
+              common_relation_methods_module.create_method_with_sigs(
                 "find_each",
+                sigs: sigs,
                 parameters: [
-                  create_kw_opt_param("start", type: "T.untyped", default: "nil"),
-                  create_kw_opt_param("finish", type: "T.untyped", default: "nil"),
-                  create_kw_opt_param("batch_size", type: "Integer", default: "1000"),
-                  create_kw_opt_param("error_on_ignore", type: "T.untyped", default: "nil"),
-                  *(create_kw_opt_param("order", type: "Symbol", default: ":asc") if order),
-                  create_block_param("block", type: "T.nilable(T.proc.params(object: #{constant_name}).void)"),
+                  RBI::KwOptParam.new("start", "nil"),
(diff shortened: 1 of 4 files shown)


KaanOzkan:
@bdewater it's resolved on main, you can rebase.

lib/tapioca/rbi_ext/model.rb
```diff
@@ -87,24 +87,59 @@ def create_type_variable(name, type:, variance: :invariant, fixed: nil, upper: n
...
+        return_type: String,
+      ).returns(RBI::Sig)
+    end
+    def create_sig(parameters: [], return_type: "T.untyped")
+      sig = RBI::Sig.new(return_type: return_type)
+      parameters.each do |param|
+        sig << RBI::SigParam.new(param.param.name, param.type)
+      end
+      sig
+    end
```
paracycle, line 141:
```suggestion
    def create_sig(parameters: [], return_type: "T.untyped")
      params = parameters.map do |param|
        RBI::SigParam.new(param.param.name, param.type)
      end
      
      RBI::Sig.new(params: params, return_type: return_type)
    end
```

bdewater:
Thanks for the feedback @paracycle! Rebased and incorporated your suggestions.

paracycle (approved):
Thank you!

bdewater:
I fixed the last commit ([here](https://github.com/bdewater/tapioca/commit/4457b30d82b5f3b51d5dbc8c0467202cb2c00811)) but it doesn't show up yet because of https://www.githubstatus.com/incidents/wcl1sw4mzg60

lib/tapioca/dsl/compilers/active_record_relations.rb
```diff
@@ -560,12 +550,25 @@ def create_common_methods
...
+              end
+              sigs = [
+                common_relation_methods_module.create_sig(
+                  parameters: [create_param("args", type: args_type)],
+                  return_type: "T::Enumerable[#{constant_name}]",
+                ),
+                common_relation_methods_module.create_sig(
+                  parameters: [create_param("args", type: "T.untyped")],
+                  return_type: constant_name,
+                ),
```
KaanOzkan, line 566: This pattern doesn't seem to work
```ruby 
# test.rbi
class Foo
  sig { returns(T.untyped) }
  def foo; end

  sig { returns(T::Array[T.untyped]) }
  def bar; end
end

sig { params(args: T::Array[T.untyped]).returns(T::Enumerable[Foo]) }
sig { params(args: T.untyped).returns(Foo) }
def find(args); end
```

```ruby
# test.rb
T.reveal_type(find(Foo.new.foo)) # T::Enumerable[Foo]
T.reveal_type(find(Foo.new.bar)) # T::Enumerable[Foo]
```

First sig always takes precedence. I'm not sure why, maybe it's a sorbet bug. We could open a ticket and revert this change until its fixed.

bdewater: T.untyped argument would be matched against the first signature since it's a subtype of everything, as Jez pointed out in the linked Sorbet issue. 

Also noticed this particular issue was documented on https://sorbet.org/docs/overloads 😅 
> In the presence of untyped arguments, chances are high that the first overload is selected, which might not be desired.

bdewater: Some experimenting with the Rails console - `find(nil)` errors [here](https://github.com/rails/rails/blob/v7.1.3.2/activerecord/lib/active_record/relation/finder_methods.rb#L496-L497), `find({})` errors [here](https://github.com/rails/rails/blob/v7.1.3.2/activerecord/lib/active_record/connection_adapters/abstract/quoting.rb#L14-L28) which gives a pretty good list of types that can be passed.

lib/tapioca/dsl/compilers/active_record_relations.rb
```diff
@@ -560,12 +550,29 @@ def create_common_methods
...
+              sigs = [
+                common_relation_methods_module.create_sig(
+                  parameters: [create_param("args", type: id_types)],
+                  return_type: constant_name,
+                ),
+                common_relation_methods_module.create_sig(
+                  parameters: [create_param("args", type: array_type)],
+                  return_type: "T::Enumerable[#{constant_name}]",
+                ),
+              ]
```
amomchilov, line 571: This is missing one overload:

`find` can be passed a block, in which case it delegates to `Enumerable#find`, which returns a nilable result.

alexevanczuk:
@KaanOzkan Would you be able to cut a new release that includes this? Would love to add this to add overloads for some RBI generation we have today.

KaanOzkan:
@alexevanczuk I'm not sure what you mean, this has been released since [v0.13.0](https://github.com/Shopify/tapioca/releases/tag/v0.13.0). You should be seeing  overloaded sigs under `CommonRelationMethods` in a model RBI, for example the `find` method.

PR #1804, opened 2024-02-27
Stop ignoring `irb` in RBI generation

KaanOzkan:
### Motivation

Since https://github.com/sorbet/sorbet/commit/2f9fbc2236df51f21342231bd225f31bef25c221 superclass mismatch between IRB's tapioca generated RBIs and the sorbet payload can be resolved by using the new flag with the appropriate class name,
`--suppress-payload-superclass-redefinition-for=Foo`.

This flag can be added to sorbet/config.yml for convenience.


KaanOzkan:
@st0012 Most renamings should be fine but if there are plans to change superclasses then yes I'm happy to keep ignoring it until it's done. `irb` is not commonly referenced so dealing with superclass mismatches are probably not worth having the RBIs. If there are constants that may be common we could put them as annotations in `rbi-central` so that they don't have to be manually shimmed. Maybe something similar to what we have in our monolith?:

```rb
module IRB
  VERSION = T.let(T.unsafe(nil), String)

  class Statement
    class Command < Statement
    end

    class Expression < Statement
    end
  end
end
```

PR #1807, opened 2024-02-28
Enable untyped code highlighting

KaanOzkan:
### Motivation
Part of https://github.com/Shopify/team-ruby-dx/issues/1125

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -12,5 +12,6 @@
   "[ruby]": {
     "editor.defaultFormatter": "Shopify.ruby-lsp",
     "editor.formatOnSave": true
-  }
+  },
+  "sorbet.highlightUntyped": "everywhere-but-tests"
 }


.vscode/settings.json
```diff
@@ -12,5 +12,6 @@
   "[ruby]": {
     "editor.defaultFormatter": "Shopify.ruby-lsp",
     "editor.formatOnSave": true
-  }
+  },
+  "sorbet.highlightUntyped": "everywhere-but-tests"
```
andyw8, line 16: I think this won't work correctly for this repo?

> Sorbet treats any file in a folder named test or ending with the .test.rb suffix as a test file. This is not yet configurable

KaanOzkan: Good point, I'll open a PR to include `spec/` and `_spec`. In the meantime I don't think this is a blocker for the PR.

andyw8 (approved):

.vscode/settings.json
KaanOzkan: https://github.com/sorbet/sorbet/pull/7730

PR #1809, opened 2024-02-28
Bump `sorbet-static-and-runtime` requirement to v0.5.11087

Morriar:
### Motivation

Sorbet support for multiple RBI signatures on a same method was added at this version. We need this feature for new work such as https://github.com/Shopify/tapioca/pull/1799 and https://github.com/Shopify/tapioca/tree/uk-rbs-converter.

The v0.5.11087 is ~5 months old, it should be okay.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -8,7 +8,7 @@ GIT
 PATH
   remote: .
   specs:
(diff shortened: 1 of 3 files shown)


paracycle (approved):
Shipit!

PR #1817, opened 2024-03-12
Update README to clarify what --verify does

simonlevasseur:
### Motivation
I recently removed the `turbo-rails` gem from a repo and ran `bin/tapioca gem`. It removed the RBI files for that gem and `bin/tapioca gems --verify` succeeds. However, `bin/typecheck` fails because RBIs for other gems have references to Turbo in them and it can't resolve the constant. The fix is to run `bin/tapioca gems --all` which rewrites all of the gem RBIs but in doing so, it also generates a bunch of misc. changes (mostly in comments but also in cases where RBIs might have been impacted by other gems).

### Implementation
Update the README to clarify what `bin/tapioca gems --verify` is actually doing and why it might yield lots of misc. changes.


andyw8 (approved):
Looks good to me but let’s have the experts @KaanOzkan or @Morriar confirm.

PR #1818, opened 2024-03-12
Simplify PR label checks

Morriar:
### Motivation

When a PR is missing a required label, we currently create a comment in the PR discussion saying so. This creates a lot of noise and alert fatigue when we could use Github action statuses for the same result.

### Implementation

I moved and simplified the logic of the action into a Ruby script: `bin/check_pr_labels`.


paracycle (approved):
Hmm, let's get this merged, but I had gone for a bash script with the long term view of extracting this out to a org-wide shared task in the future.

PR #1820, opened 2024-03-14
Add ripper DSL

kddnewton:
### Motivation

This doesn't quite fit in to the normal idea of a DSL, but I think it fits best here. Basically, because ripper isn't a normal gem and doesn't have a version, there's not a great way to generate static RBIs for it, because it depends on your current version of Ruby.

If this doesn't belong here and can go elsewhere, let me know.

### Implementation

Loop over the defined constants at runtime, generate the definitions.

### Tests

I included tests.


spec/tapioca/dsl/compilers/ripper_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+              assert_includes(rbi, <<~RUBY)
+                sig { params(param0: T.untyped, param1: T.untyped, param2: T.untyped).returns(T.untyped) }
+                def on_binary(param0, param1, param2); end
+              RUBY
+            end
+
+            it "generates RBI definitions for ripper state constants" do
+              rbi = rbi_for(:Ripper)
+
+              assert_match(/^Ripper::EXPR_BEG = \d+$/, rbi)
```
KaanOzkan, line 34: This actually generates `Ripper::EXPR_BEG = T.let(T.unsafe(nil), ::Integer)` so this match fails

KaanOzkan (commented):
I'm fine with having this here since it can't be in ruby. Can you commit the resulting RBI file as well. After running `bundle exec tapioca dsl`.

vinistock (approved):
We have some test failures, but the code looks good to me

PR #1823, opened 2024-03-15
Fix return type of ActiveRecord::QueryMethods#extract_associated

bdewater:
### Motivation
See https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-extract_associated - this returns an array and the return type unknown since it depends on the association passed in.

### Implementation
Override when iterating over QUERY_METHODS to return the right thing seemed like the right place to hook into since it collects methods from the aforementioned Active Record module.

### Tests
Updated tests.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_relations.rb
+++ b/lib/tapioca/dsl/compilers/active_record_relations.rb
@@ -559,13 +559,29 @@ module Tapioca
           )
 
           QUERY_METHODS.each do |method_name|
-            create_relation_method(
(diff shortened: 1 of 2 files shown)


bdewater:
Unrelated protobuf errors.

KaanOzkan (approved):

KaanOzkan:
@bdewater can you rebase?

bdewater:
@KaanOzkan CI is green now :)

PR #1824, opened 2024-03-15
Improve typing for SecurePassword authenticate* methods

bdewater:
### Motivation
See example on https://github.com/rails/rails/blob/6f0d1ad14b92b9f5906e44740fce8b4f1c7075dc/activemodel/lib/active_model/secure_password.rb#L165-L186

### Tests
Updated tests.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_model_secure_password.rb
+++ b/lib/tapioca/dsl/compilers/active_model_secure_password.rb
@@ -78,7 +78,15 @@ module Tapioca
 
           root.create_path(constant) do |klass|
             methods.each do |method|
-              create_method_from_def(klass, constant.instance_method(method))
+              if method == :authenticate || method.start_with?("authenticate_")
+                klass.create_method(
+                  method.to_s,
+                  parameters: [create_param("unencrypted_password", type: "T.untyped")],
+                  return_type: "T.any(#{constant}, FalseClass)",
+                )
+              else
+                create_method_from_def(klass, constant.instance_method(method))
+              end
             end
           end
         end
--- a/spec/tapioca/dsl/compilers/active_model_secure_password_spec.rb
+++ b/spec/tapioca/dsl/compilers/active_model_secure_password_spec.rb
@@ -66,10 +66,10 @@ module Tapioca
                 # typed: strong
 
                 class User
-                  sig { params(unencrypted_password: T.untyped).returns(T.untyped) }
+                  sig { params(unencrypted_password: T.untyped).returns(T.any(User, FalseClass)) }
                   def authenticate(unencrypted_password); end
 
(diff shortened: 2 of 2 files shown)


bdewater:
Getting a bunch of unrelated `NoMethodError: undefined method `build' for an instance of Google::Protobuf::DescriptorPool`

lib/tapioca/dsl/compilers/active_model_secure_password.rb
```diff
@@ -78,7 +78,15 @@ def decorate
 
           root.create_path(constant) do |klass|
             methods.each do |method|
-              create_method_from_def(klass, constant.instance_method(method))
+              if method.start_with?("authenticate")
+                klass.create_method(
+                  method.to_s,
+                  parameters: [create_param("unencrypted_password", type: "T.nilable(String)")],
```
paracycle, line 84: So, here is a fun story:
- The dynamically defined `authenticate_xxx` method calls `BCrypt::Password#is_password?` with this argument: https://github.com/rails/rails/blob/6f0d1ad14b92b9f5906e44740fce8b4f1c7075dc/activemodel/lib/active_model/secure_password.rb#L177
- That `is_password?` method is an alias to `==` on `BCrypt::Password`: https://github.com/bcrypt-ruby/bcrypt-ruby/blob/6ac44974ba3c304de39eb06384b10facca191458/lib/bcrypt/password.rb#L79
- The `==` method passes this argument to `BCrypt::Engine.hash_secret` method: https://github.com/bcrypt-ruby/bcrypt-ruby/blob/6ac44974ba3c304de39eb06384b10facca191458/lib/bcrypt/password.rb#L77
- Finally, `hash_secret` method calls `.to_s` on this argument to turn it into a `String`: https://github.com/bcrypt-ruby/bcrypt-ruby/blob/master/lib/bcrypt/engine.rb#L68

So, despite the general usage, it is theoretically possible to pass something that is `to_s`able but isn't a `String` into this method, so a `T.untyped` might be a better call here.

Do you need to check that what you are passing in is a `String`?

lib/tapioca/dsl/compilers/active_model_secure_password.rb
```diff
@@ -78,7 +78,15 @@ def decorate
 
           root.create_path(constant) do |klass|
             methods.each do |method|
-              create_method_from_def(klass, constant.instance_method(method))
+              if method.start_with?("authenticate")
```
paracycle, line 81: Do you mind changing this check to:
```suggestion
              if method.start_with?("authenticate_") || method.to_s == "authenticate"
```
which I feel is slightly safer since it will match a narrower set.

paracycle (approved):
Thanks!

Issue #1819, opened 2024-03-13
API change in google-protobuf 4.26.0 breaks CI


KaanOzkan:
DSL compiler appears to be broken due to the changes https://github.com/Shopify/tapioca/actions/runs/8265727256/job/22622289790#step:5:802

KaanOzkan:
I think we're dealing with https://protobuf.dev/news/2023-12-27/#ruby-breaking-changes. It looks like changes introduced in https://protobuf.dev/news/2023-04-20/ (3.23.0) are now the only way and `build` is deprecated.

paracycle:
Do mind that the `Google::Protobuf::DescriptorPool#build` method is only being called in tests, so this means that protobuf _tests_ are broken and likely not the code.

We probably just need to change the tests to use the new generated syntax.

PR #1827, opened 2024-03-18
Upgrade protobuf and update tests

KaanOzkan:
### Motivation

Resolves https://github.com/Shopify/tapioca/issues/1819

### DIFF
--- a/lib/tapioca/helpers/test/content.rb
+++ b/lib/tapioca/helpers/test/content.rb
@@ -44,23 +44,6 @@ module Tapioca
           File.write(file_name, content)
           file_name
         end
-
-        sig { params(name: String, content: String, require_file: T::Boolean).void }
-        def add_proto_file(name, content, require_file: true)
-          proto_path = tmp_path("proto/#{name}.proto")
-          raise ArgumentError, "a file named '#{name}' was already added; cannot overwrite." if File.exist?(proto_path)
-
-          proto_dir = File.dirname(proto_path)
-          FileUtils.mkdir_p(proto_dir)
-          File.write(proto_path, content)
-
-          lib_path = tmp_path("lib")
-          FileUtils.mkdir_p(lib_path)
-          _, stderr, status = Open3.capture3("protoc --proto_path=#{proto_dir} --ruby_out=#{lib_path} #{proto_path}")
-          raise "Error executing protoc: #{stderr}" unless status.success?
-
-          Tapioca.silence_warnings { require("#{lib_path}/#{name}_pb.rb") } if require_file
-        end
       end
     end
   end
--- a/spec/tapioca/dsl/compilers/protobuf_spec.rb
+++ b/spec/tapioca/dsl/compilers/protobuf_spec.rb
@@ -536,6 +536,20 @@ module Tapioca
             end
           end
         end
+
+        private
+
+        sig { params(name: String, content: String, require_file: T::Boolean).void }
+        def add_proto_file(name, content, require_file: true)
+          add_content_file("proto/#{name}.proto", content).tap do |proto_path|
+            lib_path = tmp_path("lib")
+            proto_dir = File.dirname(proto_path)
+            _, stderr, status = Open3.capture3("protoc --proto_path=#{proto_dir} --ruby_out=#{lib_path} #{proto_path}")
+            raise "Error executing protoc: #{stderr}" unless status.success?
+
+            Tapioca.silence_warnings { require("#{lib_path}/#{name}_pb.rb") } if require_file
+          end
+        end
       end
     end
   end


lib/tapioca/helpers/test/content.rb
```diff
@@ -44,6 +44,23 @@ def add_content_file(name, content)
...
+          proto_dir = File.dirname(proto_path)
+          FileUtils.mkdir_p(proto_dir)
+          File.write(proto_path, content)
+
+          lib_path = tmp_path("lib")
+          FileUtils.mkdir_p(lib_path)
+          _, stderr, status = Open3.capture3("protoc --proto_path=#{proto_dir} --ruby_out=#{lib_path} #{proto_path}")
+          raise "Error executing protoc: #{stderr}" unless status.success?
+
+          Tapioca.silence_warnings { require("#{lib_path}/#{name}_pb.rb") } if require_file
```
paracycle, line 62: It would be great to reuse the `add_content` method here:
```suggestion
          add_content_file(tmp_path("proto/#{name}.proto"), content).tap do |proto_path|
            lib_path = tmp_path("lib")
            FileUtils.mkdir_p(lib_path)
            _, stderr, status = Open3.capture3("protoc --proto_path=#{proto_dir} --ruby_out=#{lib_path} #{proto_path}")
            raise "Error executing protoc: #{stderr}" unless status.success?

            Tapioca.silence_warnings { require("#{lib_path}/#{name}_pb.rb") } if require_file
          end
```

paracycle: I also wonder if this is not specific enough that it should be a private method in the protobuf spec file.

spec/tapioca/dsl/compilers/protobuf_spec.rb
```diff
@@ -161,26 +152,19 @@ def clear_cart_item_index; end
...
-              RUBY
+                message Cart {
+                  optional VALUE_TYPE value_type = 1;
+                }
+
+                enum VALUE_TYPE {
+                  NULL = 0;
+                  FIXED_AMOUNT = 1;
+                  PERCENTAGE = 2;
+                }
```
paracycle, line 166: What happens if we move this definition inside the `message Cart` definition? Can we replicate the `Cart` namespacing for the enum that way?

spec/tapioca/dsl/compilers/protobuf_spec.rb
```diff
@@ -549,60 +514,6 @@ def fields=(value); end
...
-
-                  sig { params(value: Google::Protobuf::Map[String, T.untyped]).void }
-                  def tables=(value); end
-                end
-              RBI
-
-              assert_equal(expected, rbi_for(:Cart))
-            end
-
             it "handles map types regardless of their name" do
```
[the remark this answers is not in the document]
paracycle: This was to test against a regression, so it is fine to keep it as is, IMO.

paracycle (commented):
Thank you for doing this work! 🙏

spec/tapioca/dsl/compilers/protobuf_spec.rb
```diff
@@ -225,64 +212,16 @@ def resolve(symbol); end
...
-              add_ruby_file("protobuf.rb", <<~RUBY)
-                require 'google/protobuf/timestamp_pb'
-                require 'google/protobuf/wrappers_pb'
-
-                Google::Protobuf::DescriptorPool.generated_pool.build do
-                  add_file("cart.proto", :syntax => :proto3) do
-                    add_message "MyCart" do
-                      optional :value_type, :enum, 1, "MyCart.MYVALUETYPE"
-                    end
-                    add_enum "MyCart.MYVALUETYPE" do
```
KaanOzkan, line 238: Similar to `T.untyped` test I don't think we can replicate this syntax, so I removed this test in favour of the previous one.

lib/tapioca/helpers/test/content.rb
KaanOzkan: Yeah I agree. I refactored like you suggested and added it as a private method instead.

spec/tapioca/dsl/compilers/protobuf_spec.rb
KaanOzkan: Good catch, moving it inside does keep the namespacing.

spec/tapioca/dsl/compilers/protobuf_spec.rb
```diff
@@ -664,6 +536,20 @@ def progress=(value); end
             end
           end
         end
+
+        private
+
+        sig { params(name: String, content: String, require_file: T::Boolean).void }
+        def add_proto_file(name, content, require_file: true)
+          add_content_file(tmp_path("proto/#{name}.proto"), content).tap do |proto_path|
```
paracycle, line 544: Ah, I just noticed something that might break using `add_content_file` like this. The `name` argument to `add_content_file` is used to build a tmp path under the `lib` folder, so when we supply a full temp path argument to the method now, it will add it under `lib` somewhere.

So, I think we might have either:
1. extract a `add_tmp_file` method that accepts a full path (not just a `name`) and use it here, or
2. check if the passed argument is already a temp path inside `add_content_file` and change behaviour accordingly.

I think 1 should be cleaner.

paracycle (approved):
The diff looks great. I have one small concern, but it doesn't affect the functionality it seems.

spec/tapioca/dsl/compilers/protobuf_spec.rb
KaanOzkan: Oh yeah it was weird. I ended up removing the `tmp_path` argument to `add_content_file` since it's already creating a tmp directory for us and `tmp_path` method returns the first created tmp directory. So now we have this starting with a temporary path.

```
λ tree
.
└── lib
    ├── content_pb.rb
    └── proto
        └── content.proto

3 directories, 2 files
```

Do you prefer `proto` being outside of `lib`?

paracycle: No, actually that looks good to me, thank you!

Issue #600, opened 2021-11-06
ActiveRecord Ids should not be nilable


elfassy:
According to the tests tapioca generates id columns as nilable:
https://github.com/Shopify/tapioca/blob/3ea298ea1cfc4d0ffc439a5f1ac46dc033aa2d22/spec/tapioca/compilers/dsl/active_record_columns_spec.rb#L80-L81

in our DB we have ids as NOT NULL
```
`id` bigint(20) NOT NULL AUTO_INCREMENT,
```

Whereas other columns respect the null status set in the DB
https://github.com/Shopify/tapioca/blob/3ea298ea1cfc4d0ffc439a5f1ac46dc033aa2d22/spec/tapioca/compilers/dsl/active_record_columns_spec.rb#L174

Logic is:

https://github.com/Shopify/tapioca/blob/main/lib/tapioca/helpers/active_record_column_type_helper.rb#L49
```ruby
    if column_name == @constant.primary_key ||
        column_name == "created_at" ||
        column_name == "updated_at"
      getter_type = "T.nilable(#{getter_type})"
    end
```

What makes primary_key columns special? Curious why they are not treated as other not null columns.

CraigChilds94:
@elfassy 

I believe this is to allow the case where you call `Model.new`

`id, created_at & updated_at` would be `nil`? 

Example:

```ruby
instance = Model.new

# this would be nil and so should be `T.nilable` ?
instance.id
```

elfassy:
@CraigChilds94 that's what I thought at first, but that would mean every column would be nilable, which is not the case, see https://github.com/Shopify/tapioca/blob/3ea298ea1cfc4d0ffc439a5f1ac46dc033aa2d22/spec/tapioca/compilers/dsl/active_record_columns_spec.rb#L174

CraigChilds94:
@elfassy

In this case **every** column should have `T.nilable` however if we were to do this it wouldn't reflect the application behaviour we might want from the type safety on our models.

I guess there's an assumption here that if you were to have a persisted model like you said, that it would not be "nilable" but in reality even if you `new` the Model without a value or `nil` it isn't covered by this DSL.

I'm not quite sure if theres a way we could actually cater for this

CraigChilds94:
@paracycle I think we can close this issue right? It appears as though this behaviour is expected?

connorshea:
[...]

PR #1831, opened 2024-03-20
load default compilers before user defined compilers

stathis-alexander:
### Motivation
Currently, the tapioca default compilers load _after_ the user defined compilers. This makes it impossible to extend or monkey patch a default compiler without monkey patching the load order in an initializer. See [here](https://github.com/Shopify/tapioca/issues/600#issuecomment-1973235728) for a comment concerning this behavior.

This PR switches the load order of default compilers to go _before_ the user defined compilers. This allows users to extend the default compilers or monkey patch custom behavior with compilers of their own.


KaanOzkan:
@stathis-alexander There is a CI failure

Issue #1845, opened 2024-03-26
v0.13.0 - Errors with custom ID types in ActiveRecord


leifg:
We are using [custom ID types](https://api.rubyonrails.org/classes/ActiveRecord/Attributes/ClassMethods.html#method-i-attribute) for our primary keys:

```ruby
class MyModel < ApplicationRecord
  attribute :id, MyModelId.new
end

MyModel.find(MyModelId.new("123"))
```

With 0.13.0 this results in type errors when regenerating DSL RBI files:

```ruby
Expected T.any(TrueClass, FalseClass, String, Symbol, ActiveSupport::Multibyte::Chars, Numeric, ActiveModel::Type::Binary::Data, ActiveRecord::Type::Time::Value, Date, Time, ActiveSupport::Duration, T::Class[T.anything]) but found MyModelId for argument args https://srb.help/7002                                                                                                                                                                                                                                                                                 
    26 |    MyModel.find(MyModelId.new("123"))
```

Is there anyway to support id types defined by `attribute :id`?

PR #1851, opened 2024-03-28
Use `type_for_attribute` to determine primary_key for `find`

alex-tan:
### Motivation

Issue #1845
We use custom ID types for our models. The current sigs for `find` don't reflect the use of `ActiveRecord::Attributes::ClassMethods.attribute`

### Implementation

I utilized `Tapioca::Dsl::Helpers::ActiveModelTypeHelper.type_for` which reflects on the signatures of the `ActiveModel::Type::Value` subclasses to determine what other potential types should be in the signatures.

### Tests

I have added tests.

### DIFF
--- a/spec/tapioca/dsl/compilers/active_record_relations_spec.rb
+++ b/spec/tapioca/dsl/compilers/active_record_relations_spec.rb
@@ -48,8 +48,33 @@ module Tapioca
             end
 
             it "generates proper relation classes and modules" do
+              add_ruby_file("schema.rb", <<~RUBY)
+                ActiveRecord::Migration.suppress_messages do
+                  ActiveRecord::Schema.define do
+                    create_table :posts do |t|
+                    end
+                  end
+                end
+              RUBY
+
+              add_ruby_file("custom_id.rb", <<~RUBY)
+                class CustomId < ActiveRecord::Type::Value
+                  extend T::Sig
+
+                  sig { params(value: T.untyped).returns(T.nilable(CustomId)) }
+                  def deserialize(value)
+                    CustomId.new(value) unless value.nil?
+                  end
+
+                  def serialize(value)
+                    value
+                  end
+                end
+              RUBY
+
               add_ruby_file("post.rb", <<~RUBY)
                 class Post < ActiveRecord::Base
+                  attribute :id, CustomId.new
                 end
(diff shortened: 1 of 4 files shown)


spec/tapioca/dsl/compilers/active_record_relations_spec.rb
```diff
@@ -105,8 +130,8 @@ def fifth; end
...
-                    sig { params(args: T::Array[T.any(String, Symbol, ::ActiveSupport::Multibyte::Chars, T::Boolean, BigDecimal, Numeric, ::ActiveRecord::Type::Binary::Data, ::ActiveRecord::Type::Time::Value, Date, Time, ::ActiveSupport::Duration, T::Class[T.anything])]).returns(T::Enumerable[::Post]) }
+                    sig { params(args: T.any(String, Symbol, ::ActiveSupport::Multibyte::Chars, T::Boolean, BigDecimal, Numeric, ::ActiveRecord::Type::Binary::Data, ::ActiveRecord::Type::Time::Value, Date, Time, ::ActiveSupport::Duration, T::Class[T.anything], T.nilable(::CustomId))).returns(::Post) }
+                    sig { params(args: T::Array[T.any(String, Symbol, ::ActiveSupport::Multibyte::Chars, T::Boolean, BigDecimal, Numeric, ::ActiveRecord::Type::Binary::Data, ::ActiveRecord::Type::Time::Value, Date, Time, ::ActiveSupport::Duration, T::Class[T.anything], T.nilable(::CustomId))]).returns(T::Enumerable[::Post]) }
```
bdewater-thatch, line 134: The code should not generate a nilable type, since `Model.find(nil)` will raise:
`[...]/activerecord-7.1.3.2/lib/active_record/relation/finder_methods.rb:497:in 'find_with_ids': Couldn't find Model without an ID (ActiveRecord::RecordNotFound)`

alex-tan: True, this is because the `deserialize` [sig](https://github.com/Shopify/tapioca/pull/1851/files) says that's the return value so it's extracting from that. I could do a Regex replace of `T.nilable(` and `)`?

alex-tan:
I have signed the CLA!

PR #1852, opened 2024-03-28
Start downloading the latest protoc binary

paracycle:
### Motivation

Instead of hard-coding a certain protoc binary URL into our CI code, we can use the `gh` tool to download the latest Linux x64 binary of `protoc` and unzip it to the correct location. This should make this step less brittle.

### Implementation

Start using the `gh release download` command to download the latest archive and only extract the `protoc` binary from it.

### Tests

No tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,6 +51,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Install protobuf
+        env:
+          GH_TOKEN: ${{ github.token }}
         run: |
           gh release download --repo google/protobuf --pattern '*-linux-x86_64*' --output protoc.zip
           unzip protoc.zip bin/protoc -d $HOME/.local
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -51,9 +51,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Install protobuf
+        env:
+          GH_TOKEN: ${{ github.token }}
         run: |
-          curl -OL https://github.com/google/protobuf/releases/download/v25.3/protoc-25.3-linux-x86_64.zip
-          unzip protoc-25.3-linux-x86_64.zip -d $HOME/.local
+          gh release download --repo google/protobuf --pattern '*-linux-x86_64*' --output protoc.zip
```
KaanOzkan, line 57: Looks like this version isn't compatible with the "import wrapper" call https://github.com/Shopify/tapioca/actions/runs/8474508328/job/23221014800?pr=1852#step:6:722

PR #1854, opened 2024-03-29
Add `reset_association` method signature for singular AR associations

cbothner:
### Motivation
This generated method is missing from the rbis. It's added with metaprogramming here:
https://github.com/rails/rails/blob/a2a870a7360e9590e14e1be989792cd5b607f944/activerecord/lib/active_record/associations/builder/singular_association.rb#L23-L25

And documented here:
https://github.com/rails/rails/blob/a2a870a7360e9590e14e1be989792cd5b607f944/activerecord/lib/active_record/associations.rb#L1556-L1557

### Implementation
I put it right next to `reload_association`

### Tests
Updated 👍🏻

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_associations.rb
+++ b/lib/tapioca/dsl/compilers/active_record_associations.rb
@@ -196,6 +196,10 @@ module Tapioca
             "reload_#{association_name}",
             return_type: association_type,
           )
+          klass.create_method(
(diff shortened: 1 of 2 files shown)


KaanOzkan (approved):
Thanks. Could be good to add it to the class documentation in the beginning of the file. You probably have to run `bin/docs` after.

PR #1864, opened 2024-04-10
Compiler for ViewComponent::Slotable

franklinhu:
### Motivation

Rails projects that use Tapioca often also use [ViewComponents](https://viewcomponent.org/) to encapsulate view behavior. ViewComponents supports composition through "slots" to render other ViewComponent classes.

### Implementation

- Adds a Tapioca compiler that looks at `ViewComponent::Slotable`'s registered slots that are defined via DSL

### Tests

- Adds some tests for basic usages of `ViewComponent` with `renders_one` and `renders_many`


franklinhu:
I have signed the CLA!

KaanOzkan:
Hey @franklinhu, thank you for this PR but unfortunately every new DSL compiler inside tapioca is another feature that we have to maintain moving forward. So we try to keep the compilers to DSL's that are commonly used within Shopify. As a result, I will be declining this PR. I see the value here and I suggest moving it to a separate gem or ideally within `view_components` repo for visibility for other tapioca users.

PR #1871, opened 2024-04-15
Infer types for active record fixtures

TobiasBales:
### Motivation
It has been bugging me some time that fixtures are T.untyped.
After a conversation today with @vinistock we realized that method overloading might work - and lo and behold - it does work.

The generated signatures return `Model` for `models(:one)` and `T::Array[Model]` for `models(:one, :two)` by overloading it with nil as the rest arg type, hat tip to @vinistock for the idea.

### Implementation
The signature of the fixture methods changed from `users(*fixture_names)` to `users(fixture_name, *other_fixutres)` in order to allow overloading based on the type of the rest args (e.g. `nil` or `T.any(String, Symbol`).

The mapping for fixture method (e.g. `users`) to model name (e.g. `User`) happens by creating a lookup table (once) for a mapping from underscore name to name. All descendants of `ActiveRecord::Base` go into the lookup table.
This is required since there are potential conflicts with the fixture method name between `FooBar` and `Foo::Bar` which both map to `foo_bars`.
By building this lookup table once we can map this reasonably efficiently and correctly.

Should no matching model be found it falls back to T.untyped.
Do we have some way of collecting metrics if/how often this happens?

I added an ad-hoc implementation for turning the `Blog::Post` into `blog_posts` - I would assume there is something like that already but I did not find it - any pointers would be more than welcome!

### Tests
I modified the class documentation and tests to account for the changed signature and added the correct return types.
I also modified one test case to have a namespaces model `Post` -> `Blog::Post` to ensure that namespaced models are handled correctly and I also added a test that ensures the fallback to `T.untyped` happens as expected.

I also ran this on a personal project of mine (rails app with ~10k lines of ruby) and it worked as expected and `srb tc` still passed (meaning I used the fixtures as expected but also, at least for my case, it generated reasonable rbis)

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_fixtures.rb
+++ b/lib/tapioca/dsl/compilers/active_record_fixtures.rb
@@ -133,13 +133,37 @@ module Tapioca
           model_name_from_fixture_files = fixture_file_class_mapping[fixture_name]
           return model_name_from_fixture_files if model_name_from_fixture_files
 
-          model_name_from_fixture_sets = T.unsafe(fixture_loader).fixture_sets[fixture_name]
-          if model_name_from_fixture_sets
-            model_name = ActiveRecord::FixtureSet.default_fixture_model_name(model_name_from_fixture_sets)
-            return model_name if Object.const_defined?(model_name)
+          if fixture_loader.respond_to?(:fixture_sets)
+            model_name_from_fixture_sets = T.unsafe(fixture_loader).fixture_sets[fixture_name]
+            if model_name_from_fixture_sets
+              model_name = ActiveRecord::FixtureSet.default_fixture_model_name(model_name_from_fixture_sets)
+              return model_name if Object.const_defined?(model_name)
+            end
           end
 
+          active_record_base_class = fixture_active_record_base_class_mapping[fixture_name]
+          return active_record_base_class if active_record_base_class
+
           "T.untyped"
+        end
+
+        sig { returns(T::Hash[String, String]) }
+        def fixture_active_record_base_class_mapping
+          @fixture_class_mapping ||= T.let(
+            begin
+              ActiveRecord::Base.descendants.each_with_object({}) do |model_class, mapping|
+                class_name = model_class.name
+
+                fixture_name = class_name.underscore.gsub("/", "_")
+                fixture_name = fixture_name.pluralize if ActiveRecord::Base.pluralize_table_names
+
+                mapping[fixture_name] = class_name
+
+                mapping
+              end
+            end,
+            T.nilable(T::Hash[String, String]),
+          )
         end
 
         sig { returns(T::Hash[String, String]) }
--- a/lib/tapioca/dsl/compilers/active_record_fixtures.rb
+++ b/lib/tapioca/dsl/compilers/active_record_fixtures.rb
@@ -1,7 +1,10 @@
 # typed: strict
 # frozen_string_literal: true
 
-return unless defined?(Rails) && defined?(ActiveSupport::TestCase) && defined?(ActiveRecord::TestFixtures)
+return unless defined?(Rails) &&
+  defined?(ActiveSupport::TestCase) &&
+  defined?(ActiveRecord::TestFixtures) &&
+  defined?(ActiveRecord::FixtureSet)
 
 module Tapioca
   module Dsl
@@ -127,15 +130,43 @@ module Tapioca
 
         sig { params(fixture_name: String).returns(String) }
         def return_type_for_fixture(fixture_name)
+          model_name_from_fixture_files = fixture_file_class_mapping[fixture_name]
+          return model_name_from_fixture_files if model_name_from_fixture_files
+
           model_name_from_fixture_sets = T.unsafe(fixture_loader).fixture_sets[fixture_name]
-
           if model_name_from_fixture_sets
-            model_name = T.unsafe(ActiveRecord::FixtureSet).default_fixture_model_name(model_name_from_fixture_sets)
-
+            model_name = ActiveRecord::FixtureSet.default_fixture_model_name(model_name_from_fixture_sets)
             return model_name if Object.const_defined?(model_name)
           end
 
           "T.untyped"
+        end
+
+        sig { returns(T::Hash[String, String]) }
+        def fixture_file_class_mapping
+          @fixture_file_class_mapping ||= T.let(
+            begin
+              fixture_paths = if T.unsafe(fixture_loader).respond_to?(:fixture_paths)
+                T.unsafe(fixture_loader).fixture_paths
+              else
+                T.unsafe(fixture_loader).fixture_path
+              end
+
+              Array(fixture_paths).each_with_object({}) do |path, mapping|
+                Dir["#{path}{.yml,/{**,*}/*.yml}"].select do |file|
+                  next unless ::File.file?(file)
+
+                  ActiveRecord::FixtureSet::File.open(file) do |fh|
+                    next unless fh.model_class
+
+                    fixuture_name = file.delete_prefix(path.to_s).delete_prefix("/").delete_suffix(".yml")
+                    mapping[fixuture_name] = fh.model_class
+                  end
+                end
+              end
+            end,
+            T.nilable(T::Hash[String, String]),
+          )
         end
       end
     end
--- a/lib/tapioca/dsl/compilers/active_record_fixtures.rb
+++ b/lib/tapioca/dsl/compilers/active_record_fixtures.rb
@@ -3,8 +3,7 @@
 
 return unless defined?(Rails) &&
   defined?(ActiveSupport::TestCase) &&
-  defined?(ActiveRecord::TestFixtures) &&
-  defined?(ActiveRecord::FixtureSet)
+  defined?(ActiveRecord::TestFixtures)
 
 module Tapioca
   module Dsl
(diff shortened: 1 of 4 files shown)


bdewater:
> Should no matching model be found it falls back to T.untyped.

I was poking at this one as well and noticed it is possible to specify a [model class](https://api.rubyonrails.org/classes/ActiveRecord/FixtureSet.html#class-ActiveRecord::FixtureSet-label-Configure+the+fixture+model+class) in the fixture file itself. Could try `ActiveRecord::FixtureSet::File.open` to read the YAML file and call `#model_class` on it before falling back to `T.untyped`.

bdewater:
Yep it does and matches the logic here: https://github.com/rails/rails/blob/6d3fd5b98cc1c700fe9733521118abc7540546e7/activerecord/lib/active_record/fixtures.rb#L791-L793

lib/tapioca/dsl/compilers/active_record_fixtures.rb
```diff
@@ -24,8 +24,10 @@ module Compilers
       # # test_case.rbi
       # # typed: true
       # class ActiveSupport::TestCase
-      #   sig { params(fixture_names: T.any(String, Symbol)).returns(T.untyped) }
-      #   def posts(*fixture_names); end
+      #   sig { params(fixture_name: T.any(String, Symbol), other_fixtures: NilClass)).returns(Post) }
```
Morriar, line 27:
```suggestion
      #   sig { params(fixture_name: T.any(String, Symbol), other_fixtures: NilClass).returns(Post) }
```

lib/tapioca/dsl/compilers/active_record_fixtures.rb
```diff
@@ -24,8 +24,10 @@ module Compilers
...
+      #   sig { params(fixture_name: T.any(String, Symbol), other_fixtures: T.any(String, Symbol)))
```
Morriar, line 28:
```suggestion
      #   sig { params(fixture_name: T.any(String, Symbol), other_fixtures: T.any(String, Symbol))
```

lib/tapioca/dsl/compilers/active_record_fixtures.rb
```diff
@@ -104,12 +106,53 @@ def method_names_from_eager_fixture_loader
 
         sig { params(mod: RBI::Scope, name: String).void }
         def create_fixture_method(mod, name)
-          mod.create_method(
+          model_class = active_record_models[name] || "T.untyped"
+          mod.create_method_with_sigs(
             name,
-            parameters: [create_rest_param("fixture_names", type: "T.any(String, Symbol)")],
-            return_type: "T.untyped",
+            sigs: [
```
Morriar, line 112: fwiw we added a more elegant way to do this in RBI: https://github.com/Shopify/rbi/pull/284.

lib/tapioca/dsl/compilers/active_record_fixtures.rb
```diff
@@ -104,12 +106,53 @@ def method_names_from_eager_fixture_loader
...
           )
         end
+
+        sig { returns(T::Hash[String, String]) }
+        def active_record_models
+          @active_record_models ||= T.let(
+            ActiveRecord::Base.descendants.map do |model|
+              [fixture_name(model.name), model.name]
+            end.to_h,
+            T.nilable(T::Hash[String, T::Class[ActiveRecord::Base]]),
```
Morriar, line 141: Is this type correct? It looks like we store the `model.name` right?

```suggestion
            T.nilable(T::Hash[String, String]),
```

lib/tapioca/dsl/compilers/active_record_fixtures.rb
```diff
@@ -104,12 +106,53 @@ def method_names_from_eager_fixture_loader
...
+          )
+        end
+
+        sig { params(class_name: String).returns(String) }
+        def fixture_name(class_name)
```
Morriar, line 146: Since this compiler is for `ActiveRecord`,  is it safe to assume `ActiveSupport` is also loaded? It looks like we have a runtime dependency: https://rubygems.org/gems/activerecord/versions/7.1.3.2/dependencies.

Could we use https://api.rubyonrails.org/classes/String.html#method-i-singularize instead?

andyw8: Do we need to take consider custom inflections into account? If so then the `gsub` approach could fail.

lib/tapioca/dsl/compilers/active_record_fixtures.rb
TobiasBales: awesome

lib/tapioca/dsl/compilers/active_record_fixtures.rb
TobiasBales: yeah, at first it was a mapping to the classes, I am also going to check why it did not complain about the types because it really should have

lib/tapioca/dsl/compilers/active_record_fixtures.rb
TobiasBales: I'll look at how the rails code generates it and see if it uses the inflections.
And good call out on the ActiveSupport usage

TobiasBales: @andyw8 for the inflection, as far as I can tell, fixtures names are generated [here](https://github.com/rails/rails/blob/6d3fd5b98cc1c700fe9733521118abc7540546e7/activerecord/lib/active_record/fixtures.rb#L546) which seems to only depend on the configs (e.g. `ActiveRecord::Base`'s `pluralize_table_names` value.
Does that seem correct to you?

TobiasBales: @Morriar after some digging  I think we can use `ActiveRecord::FixtureSet.default_fixture_model_name` directly (that is what is used by the fixtures themselves).
Somehow sorbet complains about FixtureSet not being defined, it executes just fine though - any ideas on why that might be?

TobiasBales:
The tests are now failing on rails 7.0 due to missing `#fixture_sets` methods.
I will see if I can find something based on what we do for rails 7.0 to get the same data.

Should that turn out to not be possible - thoughts on just supporting this for rails 7.1+ and going with `T.untyped` for 7.0?
It might be worth the reduced complexity if it is an acceptable trade

lib/tapioca/dsl/compilers/active_record_fixtures.rb
```diff
@@ -104,10 +109,96 @@ def method_names_from_eager_fixture_loader
...
+          model_name_from_fixture_files = fixture_file_class_mapping[fixture_name]
+          return model_name_from_fixture_files if model_name_from_fixture_files
+
+          model_name_from_fixture_set = fixture_class_name_from_fixture_set(fixture_name)
+          return model_name_from_fixture_set if model_name_from_fixture_set
+
+          model_name_from_active_record_base_class = fixture_active_record_base_class_mapping[fixture_name]
+          return model_name_from_active_record_base_class if model_name_from_active_record_base_class
+
+          "T.untyped"
```
Morriar, line 142: I wonder if this would make it easier to read:

```suggestion
          fixture_file_class_mapping[fixture_name] ||
            fixture_class_name_from_fixture_set(fixture_name) ||
            fixture_active_record_base_class_mapping[fixture_name] ||
            "T.untyped"
```

Morriar (approved):
One minor suggestion but it looks good to me 👍

lib/tapioca/dsl/compilers/active_record_fixtures.rb
```diff
@@ -104,10 +109,96 @@ def method_names_from_eager_fixture_loader
...
+
+              Array(fixture_paths).each_with_object({}) do |path, mapping|
+                Dir["#{path}{.yml,/{**,*}/*.yml}"].select do |file|
+                  next unless ::File.file?(file)
+
+                  ActiveRecord::FixtureSet::File.open(file) do |fh|
+                    next unless fh.model_class
+
+                    fixuture_name = file.delete_prefix(path.to_s).delete_prefix("/").delete_suffix(".yml")
+                    mapping[fixuture_name] = fh.model_class
```
bdewater, line 196: Nit:
```suggestion
                    fixture_name = file.delete_prefix(path.to_s).delete_prefix("/").delete_suffix(".yml")
                    mapping[fixture_name] = fh.model_class
```

TobiasBales: lol thank you, I completely missed that

lib/tapioca/dsl/compilers/active_record_fixtures.rb
TobiasBales: I am very much torn on if it is better or worse to read - but I did think about it.

I'll just adopt it and rename the methods slightly so that that invocation makes more sense.

Whats the etiquette on merging here?
Squash / rebase / merge?

Morriar: No strong opinion, you can keep it as is

TobiasBales: after looking at it again I do prefer it, pushed

now the build failed with something something sqlite3 gem, going to rebuild

lib/tapioca/dsl/compilers/active_record_fixtures.rb
```diff
@@ -104,10 +109,90 @@ def method_names_from_eager_fixture_loader
...
+
+              Array(fixture_paths).each_with_object({}) do |path, mapping|
+                Dir["#{path}{.yml,/{**,*}/*.yml}"].select do |file|
+                  next unless ::File.file?(file)
+
+                  ActiveRecord::FixtureSet::File.open(file) do |fh|
+                    next unless fh.model_class
+
+                    fixutre_name = file.delete_prefix(path.to_s).delete_prefix("/").delete_suffix(".yml")
+                    mapping[fixutre_name] = fh.model_class
```
Morriar, line 190:
```suggestion
                    mapping[fixture_name] = fh.model_class
```

PR #1875, opened 2024-04-18
Constrain sqlite3 gem dependency

andyw8:
@TobiasBales was seeing [failures](https://github.com/Shopify/tapioca/actions/runs/8735153192/job/23967349760) relating to the sqlite3 v2.0.0 release. I think we need to constrain it, [as we did for ruby-lsp-rails](https://github.com/Shopify/ruby-lsp-rails/pull/344).


TobiasBales (approved):
👍 thank you

PR #1876, opened 2024-04-18
Trigger matrix status check for push and PR events

KaanOzkan:
### Motivation
Github uses commit status to set the status for the branch so we need this to run on pushes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,7 +71,7 @@ jobs:
         continue-on-error: ${{ !!matrix.experimental }}
 
   buildall:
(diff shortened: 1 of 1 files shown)


Morriar (approved):

paracycle:
I am sorry, I don't understand what this PR is trying to fix. We do have proper CI status for commits on `main`, for example. What's broken?

Morriar:
> I am sorry, I don't understand what this PR is trying to fix. We do have proper CI status for commits on `main`, for example. What's broken?

Isn't it to avoid skipping the final step that displays the aggregated build status:
image

PR #1877, opened 2024-04-19
Introduce `skip_constants` flag to `dsl`

KaanOzkan:
### Motivation

There might be need for an escape hatch that will prevent tapioca from generation DSL RBIs for specific constants.

### Implementation

Remove specified constants from the list of processable constants coming from DSL compilers

### Tests

Included

### DIFF
--- a/README.md
+++ b/README.md
@@ -500,6 +500,7 @@ Options:
                                                                                                    # Default: .
              [--halt-upon-load-error], [--no-halt-upon-load-error], [--skip-halt-upon-load-error]  # Halt upon a load error while loading the Rails application
                                                                                                    # Default: true
+             [--skip-constant=constant [constant ...]]                                             # Do not generate RBI definitions for the given application constant(s)
   -c,        [--config=<config file path>]                                                         # Path to the Tapioca configuration file
                                                                                                    # Default: sorbet/tapioca/config.yml
   -V,        [--verbose], [--no-verbose], [--skip-verbose]                                         # Verbose output for debugging purposes
@@ -933,6 +934,7 @@ dsl:
   list_compilers: false
   app_root: "."
   halt_upon_load_error: true
+  skip_constant: []
 gem:
   outdir: sorbet/rbi/gems
   file_header: true
--- a/lib/tapioca/cli.rb
+++ b/lib/tapioca/cli.rb
@@ -135,6 +135,11 @@ module Tapioca
       type: :boolean,
       desc: "Halt upon a load error while loading the Rails application",
       default: true
+    option :skip_constant,
+      type: :array,
+      banner: "constant [constant ...]",
+      desc: "Do not generate RBI definitions for the given application constant(s)",
+      default: []
     def dsl(*constant_or_paths)
       set_environment(options)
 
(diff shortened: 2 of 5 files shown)


sambostock:
With regards to naming, I would go so far as to say that we should _deprecate_ `exclude` (and similar options) in both `dsl` and `gem` (in a separate PR) in favour of more explicit `exclude_compilers` and `exclude_gems` options, respectively. IMO it's non-obvious that `exclude` in `dsl` means to exclude _compilers_ rather than files or constants.

egiurleo:
I like the idea of eventually deprecating `exclude` and only using `exclude_xyz` instead. Either way we'll have to do a release a deprecation and then a breaking change to there's time to fix stuff later.

README.md
```diff
@@ -500,6 +500,7 @@ Options:
                                                                                                    # Default: .
              [--halt-upon-load-error], [--no-halt-upon-load-error], [--skip-halt-upon-load-error]  # Halt upon a load error while loading the Rails application
                                                                                                    # Default: true
+             [--skip-constant=constant [constant ...]]                                             # Skip the given application constant(s) when generating RBIs
```
Morriar, line 503: While the name of the option is okay, I think we should use something else than `skip` in the description to better explain what it is doing:

```suggestion
             [--skip-constant=constant [constant ...]]                                             # Do not generate RBI definition for the given application constant(s)
```

lib/tapioca/commands/abstract_dsl.rb
```diff
@@ -36,6 +37,7 @@ def initialize(
         exclude:,
         file_header:,
         tapioca_path:,
+        skip_constant:,
```
Morriar, line 40: I think this could be optional with a default value?

```suggestion
        skip_constant: [],
```

lib/tapioca/cli.rb
```diff
@@ -149,6 +154,7 @@ def dsl(*constant_or_paths)
         exclude: options[:exclude],
         file_header: options[:file_header],
         tapioca_path: TAPIOCA_DIR,
+        skip_constant: options[:skip_constant],
```
Morriar, line 157: Should we check that `requested_constants` and `skip_constant` do not conflict? Or maybe add a note about which one takes precedence?

lib/tapioca/commands/abstract_dsl.rb
```diff
@@ -124,6 +127,7 @@ def create_pipeline
           error_handler: ->(error) {
             say_error(error, :bold, :red)
           },
+          skipped_constants: constantize(@skip_constant),
```
Morriar, line 130: Would it make sense to also path `ignore_missing` to constantize?

spec/tapioca/cli/dsl_spec.rb
```diff
@@ -1371,6 +1371,67 @@ def body=(body); end
...
+
+                sig { override.returns(T::Enumerable[Module]) }
+                def self.gather_constants
+                  [Post, Job]
+                end
+              end
+            end
+          RB
+
+          result = @project.tapioca("dsl --skip-constant Job")
```
Morriar, line 1408: Let's add a test with more than one value for `--skip-constant` so we do not accidentally break this in the future.

Morriar (approved):

PR #1885, opened 2024-04-27
Detach RBI header comment from top-level node (when generating YARD docs)

dduugg:
### Motivation

See https://github.com/Homebrew/brew/issues/17167
When generating YARD docs that include both source and rbi files, the tapioca header overwrites the source comments for the file's top-level node. 

### Implementation

YARD detaches comments that have more than two newlines of separation from the source code that follows. Thus, I've added a blank line following the file header comment, and verified that the generated yard docs do not include the header.

### Tests

Tests have been updated.


Morriar (approved):
I guess you're right, those files are generated, we shouldn't care too much about Rubocop.

Issue #1890, opened 2024-05-06
`bin/tapioca dsl` failing on undefined method `silenced=` for class `ActiveSupport::Deprecation`


bravehager:
It seems like Tapioca relies on instance delegation on `ActiveSupport::Deprecation.silenced=` that was removed in https://github.com/rails/rails/pull/49624 to [silence deprecations](https://github.com/Shopify/tapioca/blob/923a1cedb073189694028f47971d439ab215d905/lib/tapioca/loaders/loader.rb#L193) during code loading.

## Steps to reproduce

From the root directory of a Rails application:

```ruby
# typed: true
# frozen_string_literal: true

require "bundler/inline"

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails", github: "rails/rails", branch: "main"
  gem "tapioca", github: "Shopify/tapioca", branch: "main"
end

require "active_support/deprecation"
require "tapioca/internal"

Tapioca::Loaders::Dsl.load_application(tapioca_path: "")
```

Which results in:

```
undefined method `silenced=' for class ActiveSupport::Deprecation (NoMethodError)

          ActiveSupport::Deprecation.silenced = true
```

PR #1891, opened 2024-05-06
Remove `ActiveSupport::Deprecation` silencing from loader

bravehager:
### Motivation

See #1890

### Implementation

~~Just check if the deprecated method is actually defined or not.~~ Remove deprecation silencing.

There's probably a longer term solution using `Rails.application.deprecators.silence { ... }` but it seems like the current solution assumes we can globally silence all `ActiveSupport::Deprecation` warnings _before_ the Rails application is loaded.

### Tests

N/A (I could be missing something, but I don't see any existing tests for this behavior)


KaanOzkan:
Thanks

Issue #1893, opened 2024-05-07
Breaking changes for Rails fixtures


ltribolet:
sorbet version: 0.5.11369
tapioca version: 0.14.0
ruby version: 3.3.1
rails: 7.1.3.2

Since version 0.14.0 fixtures signatures changed from:

```ruby
def my_fixture(*fixture_names); end
```

to

```ruby
def my_fixture(fixture_name, *other_fixtures); end
```

Which cause the following error:

```
./my/test/file.rb:1: Not enough arguments provided for method ActiveSupport::TestCase#my_fixture. Expected: 1+, got: 0 https://srb.help/7004
    1 |        assert_equal 1, my_fixure.length
                                                                ^
    ./sorbet/rbi/dsl/active_support/test_case.rbi:150: ActiveSupport::TestCase#my_fixture defined here
     150 |  sig { params(fixture_name: T.any(String, Symbol), other_fixtures: NilClass).returns(MyRecord) }
```

[Documentation](https://api.rubyonrails.org/classes/ActiveRecord/FixtureSet.html) says you can omit the parameter to fetch all fixtures within the fixture file.

PR #1895, opened 2024-05-08
ActiveRecordFixtures compiler should support passing no args

KaanOzkan:
### Motivation

Resolves https://github.com/Shopify/tapioca/issues/1893


TobiasBales (approved):
❤️ thank you
I somehow didn't realize you can call it without any arguments

Issue #1837, opened 2024-03-22
Support postgres `cidr` attribute type


alex-tan:
When running `tapioca dsl` on attributes with Postgres' `cidr` [type](https://www.postgresql.org/docs/current/datatype-net-types.html#DATATYPE-CIDR), it treats it as `T.untyped`. It looks like the underlying type in Rails is Ruby's `IPAddr`.

PR #1907, opened 2024-05-21
Support Postgres `cidr` column type

alex-tan:
### Motivation

#1837 This adds types for `cidr` columns

### Implementation

I added a `when` to `type_for_activerecord_value`, following the pattern used for `Uuid`.

### Tests

I didn't because I don't believe there's an established way in the test suite to test against Postgres directly. I verified this worked in our app that has a `Cidr` column.


andyw8:
For anyone else wondering what CIDR means:

https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing

PR #1911, opened 2024-05-28
Protobuf include Google::Protobuf::MessageExts

andyw8:


lib/tapioca/dsl/compilers/protobuf.rb
```diff
@@ -86,6 +88,7 @@ def decorate
             elsif constant == Google::Protobuf::Map
               create_type_members(klass, "Key", "Value")
             else
+              klass.create_include("Google::Protobuf::MessageExts")
```
paracycle, line 91: Hmm, I think we should instead make these classes subclass from `Google::Protobuf::AbstractMessage` instead, which pulls in this include as well: https://github.com/protocolbuffers/protobuf/blob/5d40c4f43409bbcdf7ab2d24cbb4b1c363cfa7a0/ruby/lib/google/protobuf/message_exts.rb#L29

It is a private constant, but since it will only be inside our generated RBI files, that shouldn't be a problem.

paracycle: Also, at runtime, that is the correct description of these message classes: https://github.com/protocolbuffers/protobuf/blob/5d40c4f43409bbcdf7ab2d24cbb4b1c363cfa7a0/ruby/lib/google/protobuf/ffi/message.rb#L35

andyw8:
Updated with @Morriar's suggestion, but looking into the 1 failing test.

andyw8:
Passing on Core.

andyw8:
(merging this commit history was unintentional, I was used to the auto-squash in other repos).

PR #1912, opened 2024-05-30
Do not report duplicate for shim or annotations classes redefining the superclass

Morriar:
### Motivation

Sometimes we want to use shims to define the parent class that Tapioca may have missed. (Example [1](https://github.com/Shopify/shop-server/pull/77554/files#r1588551662), [2](https://github.com/Shopify/shopify/pull/509499/files#diff-e252f62eeb853d302e699ae346524b6bc2f94ed73d1e9677d639e9f014328265R81-R84))

Consider this file generated by Tapioca for a gem:

```rb
# sorbet/rbi/gems/foo@1.0.0.rbi
class Foo
end
```

The user knows that `Foo` actually inherits from `Baz` and wants to fix this in a shim:

```rb
# sorbet/rbi/shims/foo.rbi
class Foo < Bar
end
```

Tapioca `check-shims` command will report a useless duplicated shims, yet this redefinition should be allowed:

```
          Duplicated RBI for ::Foo:
           * sorbet/rbi/shims/foo.rbi:1:0-2:3
           * sorbet/rbi/gems/foo@1.0.0.rbi:1:0-2:3
```

### Implementation

Before reporting a shim class as a duplicate, we should check if it redefines the super class compared to the definitions we can find in the generated files.

### Tests

See automated tests.

### DIFF
--- a/lib/tapioca/helpers/rbi_files_helper.rb
+++ b/lib/tapioca/helpers/rbi_files_helper.rb
@@ -174,7 +174,21 @@ module Tapioca
       shims_or_todos = extract_shims_and_todos(nodes, shim_rbi_dir: shim_rbi_dir, todo_rbi_file: todo_rbi_file)
       return false if shims_or_todos.empty?
 
+      not_shims_or_todos = nodes - shims_or_todos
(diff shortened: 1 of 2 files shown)


amomchilov (approved):
Love it

PR #1921, opened 2024-06-13
Handle fixtures with no associated model

vinistock:
### Motivation

Some projects use YAML files inside the `test/fixtures` directory to represent arbitrary data that is not associated to any specific model.

This is a valid use case that Rails allows, but our DSL compiler fails when any fixtures like that are present because it assumes there will be a model.

### Implementation

There are essentially two parts to the fix:

1. Started rescuing the fixture formatting error that gets raised when no model is associated with it and we invoke `fh.model_class`. We don't want generation to stop because we found one of those fixtures
2. In addition to not crashing, we also don't want to generate RBI methods for fixtures that don't have an associated model or that contain invalid formatting. So we need to remove them from the list of methods to be created. I started memoizing `method_names` so that we can remove invalid fixtures

### Tests

Added a test that reproduces the crash.


spec/tapioca/dsl/compilers/active_record_fixtures_spec.rb
```diff
@@ -55,6 +55,40 @@ class User
...
+              add_content_file("test/fixtures/serialized_data.yml", <<~YAML)
+                ---
+                field1: 123
+                name: Hello
+              YAML
+
+              add_content_file("test/fixtures/posts.yml", <<~YAML)
+                super_post:
+                  title: An incredible Ruby post
+                  author: Johnny Developer
```
KaanOzkan, line 68: 👏

KaanOzkan (approved):

lib/tapioca/dsl/compilers/active_record_fixtures.rb
```diff
@@ -190,10 +196,14 @@ def fixture_class_mapping_from_fixture_files
...
                   ActiveRecord::FixtureSet::File.open(file) do |fh|
+                    fixture_name = file.delete_prefix(path.to_s).delete_prefix("/").delete_suffix(".yml")
                     next unless fh.model_class
 
-                    fixture_name = file.delete_prefix(path.to_s).delete_prefix("/").delete_suffix(".yml")
                     mapping[fixture_name] = fh.model_class
+                  rescue ActiveRecord::Fixture::FormatError
+                    # For fixtures that are not associated to any models and just contain raw data or fixtures that
+                    # contain invalid formatting, we want to skip them and avoid crashing
+                    method_names.delete(T.must(fixture_name))
```
paracycle, line 206: Hmm, I am not sure I really like the fact that we are modifying the array we are iterating on when we've called this method here.

I'd rather have us filter invalid fixtures way earlier in the process, even if it means iterating on them twice.

Alternatively, we could turn this method into something that builds the methods we should be iterating on or something.

vinistock: I switched to filtering ahead of time. We already had a map of `fixture_name => Model`, so we can use that to filter.

PR #1922, opened 2024-06-13
Use latest rbtrace to avoid build failures

vinistock:
### Motivation

The build in main is failing because building `rbtrace` v0.4.14 native extensions is crashing. Updating to the latest v0.5.1 fixes the issue.

Also, I added an assertion so that the error is less obscure if this fails again in the future.


KaanOzkan (approved):

vinistock:
Closing this as it turns out these issues are related to Bundler somehow #1923

PR #1924, opened 2024-06-14
Support array types in `typed_store` compiler

### LINKED ISSUE #1910: `ActiveRecord::TypedStore` DSL compiler does not respect `array: true` for columns
If you're using the `ActiveRecord::TypedStore` gem ([reference](https://github.com/byroot/activerecord-typedstore)) and set a column as being `array: true`, the generated rbi will not respect the array indication. Example:

```ruby

class Shop < ActiveRecord::Base
  typed_store :settings do |s|
    s.string(:tags, array: true, default: [], null: false)
  end
end

### Generated RBI:

class Shop
  module StoreAccessors
    sig { returns(String) }
    def tags; end

    # ...
  end
end

```

The expected would be to have the columns that are defined as `array` to have their type wrapped in a `T::Array` type. So for the example above, I would expect the output to be:

```ruby
class Shop
  module StoreAccessors
    sig { returns(T::Array[String]) }
    def tags; end

    # ...
  end
end
```

amomchilov:
### Motivation

Closes #1910

### Implementation

Decide to refactor the `#type_for` method to take the whole `ActiveRecord::TypedStore::Field` object, instead of just its `type_sym`, because I would have needed to also pass in `field.array`. This way it can just access what it needs.

This has the nice benefit of moving the `if field.null` responsibility out of the `decorate` method.

### Tests

I added tests for both `array: true, null: true` and `array: true, null: false`. I didn't assert the type of _all_ the methods (since they're mostly the same), just the getter/setter (just like how other tests check them).

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_typed_store.rb
+++ b/lib/tapioca/dsl/compilers/active_record_typed_store.rb
@@ -98,8 +98,7 @@ module Tapioca
             stores.values.each do |store_data|
               store_data.accessors.each do |accessor, name|
                 field = store_data.fields.fetch(accessor)
-                type = type_for(field.type_sym)
-                type = as_nilable_type(type) if field.null
+                type = type_for(field)
                 name ||= field.name # support < 1.5.0
 
                 generate_methods(store_accessors_module, name.to_s, type)
@@ -136,9 +135,23 @@ module Tapioca
           T::Hash[Symbol, String],
         )
 
-        sig { params(attr_type: Symbol).returns(String) }
-        def type_for(attr_type)
-          TYPES.fetch(attr_type, "T.untyped")
+        sig { params(field: ActiveRecord::TypedStore::Field).returns(String) }
+        def type_for(field)
+          type = TYPES.fetch(field.type_sym, "T.untyped")
+
+          type = if field.array
+            # `null: false` applies to the array itself, not the elements, which are always nilable.
+            # https://github.com/byroot/activerecord-typedstore/blob/2f3fb98/spec/support/models.rb#L46C34-L46C45
+            # https://github.com/byroot/activerecord-typedstore/blob/2f3fb98/spec/active_record/typed_store_spec.rb#L854-L857
+            nilable_element_type = as_nilable_type(type)
+            "T::Array[#{nilable_element_type}]"
+          else
+            type
+          end
+
+          type = as_nilable_type(type) if field.null
+
+          type
         end
 
         sig do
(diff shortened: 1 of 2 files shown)


lib/tapioca/dsl/compilers/active_record_typed_store.rb
```diff
@@ -136,9 +135,21 @@ def gather_constants
...
+            # `null: false` applies to the array itself, not the elements, which always nilable.
+            # https://github.com/byroot/activerecord-typedstore/blob/2f3fb98/spec/support/models.rb#L46C34-L46C45
+            # https://github.com/byroot/activerecord-typedstore/blob/2f3fb98/spec/active_record/typed_store_spec.rb#L854-L857
+            nilable_element_type = as_nilable_type(type)
+            "T::Array[#{nilable_element_type}]"
+          else
+            type
+          end
+
+          field.null ? as_nilable_type(type) : type
```
paracycle, line 152: Small nitpick, do you mind changing this to:
```suggestion
          type = as_nilable_type(type) if field.null
          
          type
```
which IMO is more readable and more idiomatic Ruby than the ternary.

lib/tapioca/dsl/compilers/active_record_typed_store.rb
```diff
@@ -136,9 +135,21 @@ def gather_constants
...
 
-        sig { params(attr_type: Symbol).returns(String) }
-        def type_for(attr_type)
-          TYPES.fetch(attr_type, "T.untyped")
+        sig { params(field: ActiveRecord::TypedStore::Field).returns(String) }
+        def type_for(field)
+          type = TYPES.fetch(field.type_sym, "T.untyped")
+
+          type = if field.array
+            # `null: false` applies to the array itself, not the elements, which always nilable.
```
paracycle, line 143:
```suggestion
            # `null: false` applies to the array itself, not the elements, which are always nilable.
```

paracycle (approved):
Thanks this looks like a good change. I have a few inline comments.

amomchilov:
@ipvalverde Nice catch. I lost it somehow, but was able to revive it back from my editor's local history.

Can you have a look and review? Cheers

amomchilov:
I ran these changes on core ([cherry-picked in isolation](https://github.com/Shopify/tapioca/commit/b01409f6d1cd1828ab295d5f199eb03a4148907f), on top of the `v0.14.3` version that core)

https://github.com/Shopify/shopify/commit/36d05ee84d0084e452f644d2b07e01eb54795a8a

It type-checks without any changes needed.

PR #1927, opened 2024-06-17
Ignore RubyMine files, and skip `Minitest::Reporters`

amomchilov:
Add RubyMime-related project files to `.gitignore`, and don't use `Minitest::Reporters` for RubyMine ([issue](https://youtrack.jetbrains.com/issue/RUBY-33007/Using-Minitest-should-not-require-changes-to-testhelper.rb)).


andyw8:
I think this is better suited for people's personal `~/.gitignore`.

.gitignore
```diff
@@ -8,3 +8,81 @@
 /spec/reports/
 /tmp/
 /vendor/
+
+# Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio, WebStorm and Rider
+# Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
+
+# User-specific stuff
+.idea/**/workspace.xml
```
paracycle, line 16: I think we'd rather just ignore the whole `.idea` folder

spec/spec_helper.rb
```diff
@@ -21,7 +21,9 @@
 backtrace_filter.add_filter(%r{gems/railties})
 backtrace_filter.add_filter(%r{tapioca/helpers/test/})
 
-Minitest::Reporters.use!(SpecReporter.new(color: true), ENV, backtrace_filter)
+unless ENV["RM_INFO"]
```
paracycle, line 24: I'd rather not special case RubyMine here. I'd be open to changing the default reporter, though, since our custom reporter doesn't work that well.

.gitignore
amomchilov: Do you think I should do that here at all, or leave it to each dev to configure as @andyw8 suggested?

andyw8: (this is also what's recommended on https://git-scm.com/docs/gitignore#:~:text=Patterns%20which%20a,is%20used%20instead. )

spec/spec_helper.rb
paracycle: Let's try not using it at all, and see what happens?

spec/spec_helper.rb
```diff
@@ -21,7 +21,9 @@
...
+  Minitest::Reporters.use!(SpecReporter.new(color: true), ENV, backtrace_filter)
+end
```
paracycle, line 26: Can we make this to be just:
```suggestion
Minitest.backtrace_filter = backtrace_filter
```

amomchilov: Can you explain the reasoning? Do we not want to use `SpecReporter`?

spec/spec_helper.rb
amomchilov: Left is without, right is with.

I see 2 slight benefits to using this SpecReporter:

1. The passing tests are summarized more tersely with simple green dots (though with less detail, like the name and runtime)
2. There's a backtrace filter applied

The stock one is more detailed, and perhaps better overall, but that's subjective

image

PR #1931, opened 2024-06-17
Refactor DSL compiler options to be a flat namespaced hash and document them

paracycle:
I've been thinking about this for a while, and I think namespacing compiler options was a little heavy handed. I've been thinking about cases where a single option might need to be consumed by multiple Active Record compilers, so I think it makes more sense to have a flat hash of options that can be consumed by any compiler.

This makes the code simpler and makes it easier to pass these options on the command line interface, as well.

I've also used this opportunity to improve the documentation, as promised

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ Tapioca makes it easy to work with [Sorbet](https://sorbet.org) in your codebase
     * [Changing the typed strictness of annotations files](#changing-the-typed-strictness-of-annotations-files)
   * [Generating RBI files for Rails and other DSLs](#generating-rbi-files-for-rails-and-other-dsls)
     * [Keeping RBI files for DSLs up-to-date](#keeping-rbi-files-for-dsls-up-to-date)
+    * [Using DSL compiler options](#using-dsl-compiler-options)
     * [Writing custom DSL compilers](#writing-custom-dsl-compilers)
     * [Writing custom DSL extensions](#writing-custom-dsl-extensions)
   * [RBI files for missing constants and methods](#rbi-files-for-missing-constants-and-methods)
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -545,6 +547,16 @@ if Rails.env.development?
...
 
+#### Using DSL compiler options
+
+Some DSL compilers are able to change their behaviour based on the options passed to them. For example, the
+`ActiveRecordColumns` compiler can be configured to change how it generates types for method related to Active Record
+column attributes. To pass options during DSL RBI generation, use the `--compiler-options` flag:
+```shell
+$ bin/tapioca dsl --compiler-options=ActiveRecordColumnTypes:untyped
+```
+which will make the `ActiveRecordColumns` compiler generate untyped signatures for column attribute methods.
```
Morriar, line 558: Can you add an example on how to pass the same option using the `tapioca/config.yml` file? 🙏

lib/tapioca/dsl/compilers/active_record_columns.rb
```diff
@@ -15,6 +15,21 @@ module Compilers
       # created for columns and virtual attributes that are defined in the Active Record
       # model.
       #
+      # This compiler accepts a `ActiveRecordColumnTypes` option that can be used to specify
+      # how the types of the column related methods should be generated. The option can be one of
```
Morriar, line 19:
```suggestion
      # how the types of the column related methods should be generated. The option can be one of the following:
```

lib/tapioca/dsl/compilers/active_record_columns.rb
```diff
@@ -15,6 +15,21 @@ module Compilers
...
+      #  column type as the return type. This means that if the column is a string, the method return type
+      #  will be `String`, but if the column is also nullable, then the return type will be `T.nilable(String)`. This
+      #  mode basically treats each model as if it was a valid and persisted model. Note that this makes typing
+      #  Active Record models easier, but does not match the behaviour of non-persisted or invalid models, which can
+      #  have all kinds of non-sensical values in their column attributes.
+      #  - `nilable`: All column methods will be generated with `T.nilable` return types. This is strictly the most
+      #  correct way to type the methods, but it can make working with the models more cumbersome, as you will have to
+      #  handle the `nil` cases explicitly using `T.must` or the safe navigation operator `&.`, even for valid
+      #  persisted models.
+      #  - `untyped`: The methods will be generated with `T.untyped` return types. This mode is useful if you are not
```
Morriar, line 30:
```suggestion
      #  - `untyped`: The methods will be generated with `T.untyped` return types. This mode is practical if you are not
```

lib/tapioca/dsl/compilers/active_record_columns.rb
```diff
@@ -150,13 +176,25 @@ def gather_constants
...
-                "types",
-                Helpers::ActiveRecordColumnTypeHelper::ColumnTypeOption::Persisted.serialize,
-              ),
-            ),
-            T.nilable(Helpers::ActiveRecordColumnTypeHelper::ColumnTypeOption),
+            begin
+              column_type_option = Helpers::ActiveRecordColumnTypeHelper::ColumnTypeOption::Persisted
+
+              if options.key?("ActiveRecordColumnTypes")
+                value = options["ActiveRecordColumnTypes"]
```
Morriar, line 183:
```suggestion
              value = options["ActiveRecordColumnTypes"]
              if value
```

Morriar: We could even do something like `options["ActiveRecordColumnTypes"] || "persisted"` but this forces using the string rather than the constant for `"persisted"`.

Morriar: I wonder if we could move this logic in the enum itself: `ColumnTypeOption.from_option`?

PR #1933, opened 2024-06-18
fix: `ActiveSupportConcern` recursively checks for `ClassMethod`

spencewenski:
### Motivation

`ActiveSupportConcern` uses [`const_defined?`](https://apidock.com/ruby/Module/const_defined%3f) to check if `ClassMethods` exists on a module. By default, `const_defined?` checks the module and its ancestors for the existence of the constant. This means that `const_defined?` will return `true` when a module does not have `ClassMethods` defined, but one of its ancestors does. This can result in `ActiveSupportConcern` generating `mixes_in_class_methods.*ClassMethods` lines for invalid `.*ClassMethods` modules.

### Implementation

`const_defined?` takes a second parameter that controls whether to include ancestors in the check. Setting this parameter to `false` to exclude ancestors.

### Tests

The first commit in this PR adds a negative test case that should fail with the existing code and pass with the fix from the second commit.


amomchilov (approved):

spencewenski:
> Thanks. I'll merge after the linting fix.

This should be fixed now. Thanks!

PR #1937, opened 2024-06-20
In ActiveRecordColumns persisted mode, remove T.nilable from reflected sigs

alex-tan:
### Motivation

We use types inheriting from `ActiveModel::Type::Value` where the `deserialize` sig says it returns `T.nilable([custom type])`. However when using the new dsl `persisted` mode, we can assume the type won't be nilable.

### Implementation

Apply `as_non_nilable_type` to `getter_type` and `setter_type` only in persisted mode and when the column is not nullable.

### Tests

I've added tests.

### DIFF
--- a/lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
+++ b/lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
@@ -114,7 +114,10 @@ module Tapioca
             end
 
           if @column_type_option.persisted? && !column&.null
-            [getter_type, setter_type]
+            # It's possible that when ActiveModel::Type::Value is used, the signature being reflected on in
+            # ActiveModelTypeHelper.type_for(type_value) may say the type can be nilable. However, if the type is
+            # persisted and the column is not nullable, we can assume it's not nilable.
+            [as_non_nilable_type(getter_type), as_non_nilable_type(setter_type)]
           else
             getter_type = as_nilable_type(getter_type) unless not_nilable_serialized_column?(column_type)
             [getter_type, as_nilable_type(setter_type)]
--- a/spec/tapioca/dsl/compilers/active_record_columns_spec.rb
+++ b/spec/tapioca/dsl/compilers/active_record_columns_spec.rb
@@ -984,6 +984,53 @@ module Tapioca
                 assert_includes(rbi_for(:Post), expected)
               end
 
+              it "strips T.nilable from reflected signatures method for non-nilable columns in persisted mode" do
+                add_ruby_file("schema.rb", <<~RUBY)
+                  ActiveRecord::Migration.suppress_messages do
+                    ActiveRecord::Schema.define do
+                      create_table :posts do |t|
+                        t.decimal :cost, null: false
+                      end
+                    end
+                  end
+                RUBY
+
+                add_ruby_file("custom_type.rb", <<~RUBY)
+                  class CustomType
+                    attr_accessor :value
+
+                    def initialize(number = 0.0)
+                      @value = number
+                    end
+
+                    class Type < ActiveRecord::Type::Value
+                      extend(T::Sig)
+
+                      sig { params(value: T.nilable(Numeric)).returns(T.nilable(::CustomType))}
+                      def deserialize(value)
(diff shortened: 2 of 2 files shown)


lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
```diff
@@ -114,7 +114,10 @@ def column_type_for(column_name)
             end
 
           if @column_type_option.persisted? && !column&.null
-            [getter_type, setter_type]
+            # It's possible that when ActiveModel::Type::Value is used, the signature being reflected on in
+            # ActiveModelTypeHelper.type_for(type_value) may say the type can be nilable. However, if the type is
+            # persisted and the column is not nullable, we can assume it's not nilable.
+            [as_non_nilable_type(getter_type), as_non_nilable_type(setter_type)]
```
KaanOzkan, line 120: I'm wary of adding this overhead to every type by default to cover for an edge case. I'd like to understand why changing the signature of `deserialize` and [other methods](https://github.com/Shopify/tapioca/blob/a4ec8e1260c7c841ca2422f5736cfa710d9f78b0/lib/tapioca/dsl/helpers/active_model_type_helper.rb#L17-L20) is not an acceptable solution.

alex-tan: Sure thing. So we use custom types pretty widely both for `id` columns and as nilable foreign keys. I'm not sure if there's a way around it for the use case where the type can be used both for nullable and non-nullable columns:

```ruby
sig { params(value: T.nilable(Numeric)).returns(T.nilable(::CustomType))}
def deserialize(value)
  CustomType.new(value) if value
end
```

I could look at moving the call into just the branch of `type_for_activerecord_value` that covers this case. I think it would just mean needing to pass in `@column_type_option` and `column` into that method.

alex-tan: (I guess just column since the other is an instance variable)

lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
```diff
@@ -121,8 +121,8 @@ def column_type_for(column_name)
           end
         end
 
-        sig { params(column_type: T.untyped).returns(String) }
-        def type_for_activerecord_value(column_type)
+        sig { params(column_type: T.untyped, null: T::Boolean).returns(String) }
+        def type_for_activerecord_value(column_type, null:)
```
KaanOzkan, line 125:
```suggestion
        def type_for_activerecord_value(column_type, column_nullability:)
```

Something like this will be better.

lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
```diff
@@ -180,9 +180,16 @@ def type_for_activerecord_value(column_type)
...
-            "T::Array[#{type_for_activerecord_value(column_type.subtype)}]"
+            "T::Array[#{type_for_activerecord_value(column_type.subtype, null:)}]"
           else
-            ActiveModelTypeHelper.type_for(column_type)
+            base_type = ActiveModelTypeHelper.type_for(column_type)
+
+            # It's possible that when ActiveModel::Type::Value is used, the signature being reflected on in
+            # ActiveModelTypeHelper.type_for(type_value) may say the type can be nilable. However, if the type is
+            # persisted and the column is not nullable, we can assume it's not nilable.
+            return as_non_nilable_type(base_type) if @column_type_option.persisted? && !null
```
KaanOzkan, line 190: Since you've also contributed the `EncryptedAttributeType` case in line 136, this could affect that too right? So we should return non nilable types if possible in that case too.

KaanOzkan (commented):
Could you also amend the first commit so that it doesn't include the now outdated logic.

lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
alex-tan: Ah yes, thanks for pointing that out 👍

KaanOzkan (approved):

PR #1939, opened 2024-06-21
Generate correct return types in IdentityCache compiler

egiurleo:
There is a difference in behavior between the return types of the existing IdentityCache compiler and how IdentityCache compiler actually works.

In the aliased fetch methods,[ IdentityCache will return an array unless the field has the `unique` property](https://github.com/Shopify/identity_cache/blob/main/lib/identity_cache/cached/attribute.rb#L63), even if it is not a multi-type. Previously, our IdentityCache compiler was always setting the return type as a single object for those methods.


bitwise-aiden (approved):
🙏🏻

KaanOzkan (approved):

PR #1944, opened 2024-07-04
Fix AASM compiler's global callback method signatures

tubaxenor:
### Motivation

AASM global callbacks takes splat argument as first argument instead of a nilable symbol: https://github.com/aasm/aasm/blob/0e03746a2b86558ee1bf7bd7db873938cbb3b29b/lib/aasm/base.rb#L145-L171

### Implementation

Change AASM compiler to use correct params definition.

### Tests

Added accordingly.


tubaxenor:
I have signed the CLA!

KaanOzkan (approved):

PR #1946, opened 2024-07-08
Fix missing array signature for ActiveRecordRelation #create

bitwise-aiden:
### Motivation

If passed an array, ActiveRecord's `#create` method will return an array of that models type. This is missing in the current implementation.

### Implementation

I've split the create methods so that the ones that are able to take arrays can now do so and return the type too. 

### Question for reviewers

Is there a preference on how to define constants for the single vs multi?


lib/tapioca/dsl/compilers/active_record_relations.rb
```diff
@@ -991,14 +992,26 @@ def create_common_methods
...
+                },
+                return_type: constant_name,
+              ),
+              common_relation_methods_module.create_sig(
+                parameters: {
+                  attributes: "T::Array[T.untyped]",
+                  block: "T.nilable(T.proc.params(objects: #{constant_name}).void)",
+                },
+                return_type: "T::Array[#{constant_name}]",
+              ),
```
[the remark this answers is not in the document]
bitwise-aiden: I did wonder if that would be the case; thanks for confirming, will swap them over :D

lib/tapioca/dsl/compilers/active_record_relations.rb
```diff
@@ -991,25 +991,50 @@ def create_common_methods
...
-            block_type = "T.nilable(T.proc.params(object: #{constant_name}).void)"
-            create_common_method(
-              method_name,
-              parameters: [
-                create_param("attributes", type: "T.untyped"),
-                create_block_param("block", type: block_type),
-              ],
-              return_type: constant_name,
+            sigs = [
+              common_relation_methods_module.create_sig(
```
paracycle, line 995: Do you mind leaving a comment on top of this sig (or the `T.untyped` sig) explaining why it needs to be the first entry (or the last entry, if you want to do it for the other one)?

It will be helpful for future onlookers to know that there is intent behind the order of the sigs.

lib/tapioca/dsl/compilers/active_record_relations.rb
```diff
@@ -991,25 +991,50 @@ def create_common_methods
...
           BUILDER_METHODS.each do |method_name|
-            create_common_method(
-              method_name,
-              parameters: [
-                create_opt_param("attributes", type: "T.untyped", default: "nil"),
-                create_block_param("block", type: "T.nilable(T.proc.params(object: #{constant_name}).void)"),
-              ],
-              return_type: constant_name,
+            sigs = [
+              common_relation_methods_module.create_sig(
```
paracycle, line 1019: Same comment for this part.

paracycle (approved):
Changes look good! Can we just have comments to convey the intent behind the sig ordering?

PR #1949, opened 2024-07-09
Fix gem RBI generation for Spoom's generic

Morriar:
Before this change, we were enabling the generic type registry _after_ loading Spoom, meaning that any generic type found inside Spoom's code was effectively ignored.

This change reorders the requires so the generic type registry is loaded before Spoom and stores the `E = type_member` so they can be pushed to the generated RBI.


lib/tapioca/internal.rb
```diff
@@ -29,14 +28,16 @@
 require "tapioca/helpers/gem_helper"
 
 require "tapioca/helpers/git_attributes"
```
paracycle, line 30: I guess we can also move these helpers below the `runtime` and `sorbet_ext` requires, right?

paracycle (approved):

PR #1955, opened 2024-07-11
Generate *_changed? and *_previously_changed? sigs for belongs_to relations

cheshire137:
### Motivation

Models with a `belongs_to` relation now have a `_changed?` and `_previously_changed?` method since https://github.com/rails/rails/pull/42751. This shipped in Rails 7, see https://github.com/rails/rails/blob/b84fa1083ca7e6422356fdf80cb33751d0a23eff/activerecord/CHANGELOG.md#rails-700alpha1-september-15-2021.

I noticed in my project Sorbet was giving me an error about my model not having a `foo_changed?` method when I had a `belongs_to :foo`, even though a `respond_to?(:foo_changed?)` returned true.

### Implementation

I referenced existing `_changed?` method generation for regular columns on a model:

https://github.com/Shopify/tapioca/blob/df5079fd5063b61a3f110e330329c2eacd039ecb/lib/tapioca/dsl/compilers/active_record_columns.rb#L296-L305

While @erinnachen pointed out https://github.com/Shopify/tapioca/blob/main/lib/tapioca/dsl/compilers/active_record_associations.rb looked to be the compiler for methods coming from `belongs_to` relations, so I added the new method generation there.

I verified in the Rails implementation in https://github.com/rails/rails/pull/42751/files#diff-084ec489ba3b9f2e388734a99876f0d94776111278acfe4901235dca570040c5 that the new methods take no parameters.

### Tests

I updated existing tests to expect the new method signatures to be present.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_associations.rb
+++ b/lib/tapioca/dsl/compilers/active_record_associations.rb
@@ -206,6 +206,16 @@ module Tapioca
             "reset_#{association_name}",
             return_type: "void",
           )
+          if reflection.is_a?(ActiveRecord::Reflection::BelongsToReflection)
+            klass.create_method(
+              "#{association_name}_changed?",
+              return_type: "T::Boolean",
+            )
+            klass.create_method(
+              "#{association_name}_previously_changed?",
+              return_type: "T::Boolean",
+            )
+          end
           unless reflection.polymorphic?
             klass.create_method(
               "build_#{association_name}",
--- a/spec/tapioca/dsl/compilers/active_record_associations_spec.rb
+++ b/spec/tapioca/dsl/compilers/active_record_associations_spec.rb
@@ -124,6 +124,12 @@ module Tapioca
                         sig { params(attributes: T.untyped).returns(T.untyped) }
                         def author_attributes=(attributes); end
 
+                        sig { returns(T::Boolean) }
+                        def author_changed?; end
+
+                        sig { returns(T::Boolean) }
+                        def author_previously_changed?; end
+
                         sig { params(args: T.untyped, blk: T.untyped).returns(::User) }
                         def build_author(*args, &blk); end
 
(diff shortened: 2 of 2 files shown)


lib/tapioca/dsl/compilers/active_record_associations.rb
```diff
@@ -206,6 +206,16 @@ def populate_single_assoc_getter_setter(klass, association_name, reflection)
             "reset_#{association_name}",
             return_type: "void",
           )
+          if reflection.is_a?(ActiveRecord::Reflection::BelongsToReflection)
```
cheshire137, line 209: Is there a good way of checking the version of Rails being targeted? We could avoid generating these signatures for pre-Rails 7 if so.

cheshire137:
I have signed the CLA!

lib/tapioca/dsl/compilers/active_record_associations.rb
abraham: I'm not sure if it's recommended but it looks like you could do [this](https://stackoverflow.com/a/19012037/26406)

```suggestion
          if reflection.is_a?(ActiveRecord::Reflection::BelongsToReflection) && Rails::VERSION::MAJOR >= 7
```

cheshire137: Ooh, thanks! I'll look around the repo, see if there are other such checks like this and, if so, I'll add this one.

cheshire137: Hmm, no such uses:

screenshot of VS Code search for Rails::VERSION

I do see a #rails_version used in tests, so I can investigate that.

cheshire137: That worked great, thank you! I'll keep this in draft till I see whether tests will pass on CI (they passed locally).

PR #1957, opened 2024-07-12
Allow Tapioca commands to accept environment variables

paracycle:
This change allows Tapioca commands to accept environment variables via the command line or the configuration file. This is useful for setting up the environment for the Tapioca commands, such as setting the custom environment variables that affect how an application is loaded.

### Motivation

https://github.com/Shopify/tapioca/pull/1952#issuecomment-2220850850

### Implementation

Add a `class_option :env` that allows all commands to accept environment variables.

### Tests

Added a DSL CLI test

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,7 @@ Commands:
 Options:
   -c, [--config=<config file path>]                  # Path to the Tapioca configuration file
                                                      # Default: sorbet/tapioca/config.yml
+      [--env=key:value]                              # Environment variables to set before running Tapioca
   -V, [--verbose], [--no-verbose], [--skip-verbose]  # Verbose output for debugging purposes
                                                      # Default: false
 
@@ -122,6 +123,7 @@ Usage:
 Options:
   -c, [--config=<config file path>]                  # Path to the Tapioca configuration file
                                                      # Default: sorbet/tapioca/config.yml
+      [--env=key:value]                              # Environment variables to set before running Tapioca
   -V, [--verbose], [--no-verbose], [--skip-verbose]  # Verbose output for debugging purposes
                                                      # Default: false
 
@@ -202,6 +204,7 @@ Options:
                                                                                                      # Default: true
   -c,          [--config=<config file path>]                                                         # Path to the Tapioca configuration file
                                                                                                      # Default: sorbet/tapioca/config.yml
+               [--env=key:value]                                                                     # Environment variables to set before running Tapioca
   -V,          [--verbose], [--no-verbose], [--skip-verbose]                                         # Verbose output for debugging purposes
                                                                                                      # Default: false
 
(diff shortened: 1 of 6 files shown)


lib/tapioca/helpers/env_helper.rb
```diff
@@ -8,10 +8,20 @@ module EnvHelper
...
-      ENV["RAILS_ENV"] = ENV["RACK_ENV"] = options[:environment]
+      set_all_environment_variables(options[:env])
+      ENV["RAILS_ENV"] = ENV["RACK_ENV"] = options[:environment] if options[:environment]
       ENV["RUBY_DEBUG_LAZY"] = "1"
     end
+
+    sig { params(options: T.nilable(T::Hash[String, String])).void }
+    def set_all_environment_variables(options) # rubocop:disable Naming/AccessorMethodName
+      (options || {}).each do |key, value|
+        ENV[key] = value
```
Morriar, line 23: So the `--env` values are overriding the `ENV`, is this really what we want?

If so, I think we should document this in the option description or the README.

spec/tapioca/cli/dsl_spec.rb
```diff
@@ -2677,6 +2677,17 @@ def title=(title); end
             RACK ENVIRONMENT: staging
           OUT
         end
+
+        it "must accept and set custom environment variables" do
+          @project.write!("lib/post.rb", <<~RB)
+            $stderr.puts "ENVIRONMENT: \#{ENV.to_h.inspect}"
+          RB
+
+          result = @project.tapioca("dsl --env Foo:Bar --env Baz:1")
```
Morriar, line 2686: Can we add a test showing the behaviour of overriding system env?

README.md
```diff
@@ -88,6 +88,7 @@ Commands:
 Options:
   -c, [--config=<config file path>]                  # Path to the Tapioca configuration file
                                                      # Default: sorbet/tapioca/config.yml
+      [--env=key:value]                              # Environment variables to set before running Tapioca
```
vinistock, line 91: Nitpick: it might be my lack of familiarity with Thor, but it wasn't immediately obvious to me how we'd be passing multiple environment variables to the same command (I understood after reading the tests).

Could we add an example somewhere in the README of how to pass these?

lib/tapioca/helpers/env_helper.rb
vinistock: Also, if this is indeed what we want, is there a reason not to use `ENV.merge!(options)`?

README.md
paracycle: So, we already have a few `hash` type options and how they work is documented here: https://github.com/rails/thor/wiki/Method-Options#types-for-method_options

The main way is to use the `--opt=key:value key2:value2` way, but if you mark the option as `repeatable: true`, then you can also do `--opt=key:value --opt=key2:value2` too.

I guess I can add a section for how to pass env variables via command line or config file to make that more clear and link to the Thor docs for details.

lib/tapioca/helpers/env_helper.rb
paracycle: Yeah, I didn't think of `ENV.merge!`, that sounds better!

paracycle: > So the `--env` values are overriding the `ENV`, is this really what we want?

What else could they be doing? Tapioca sets those env variables as the user has passed them and we don't care if they override some values or not. `bin/tapioca dsl --env FOO:42` is the same thing that happens when someone runs `FOO=42 bin/tapioca dsl`, in either case we override whatever the value of the `FOO` env variable might have been to make it `42` for the duration of that run.

PR #1962, opened 2024-07-18
Add `ActiveSupportTimeExt` compiler

paracycle:
This compiler generates an RBI file for Active Support's Time extensions. Currently this is limited to the return type of the `Time.current` method which changes its return type depending on if `Time.zone` is set or not.

### Motivation

https://github.com/Shopify/rbi-central/pull/262#issuecomment-2230885870

### Implementation

Add a new compiler that only handles `Time` constant and is only loaded if `Time.current` method is defined.

### Tests

Added tests for the new compiler.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_support_time_ext.rb
+++ b/lib/tapioca/dsl/compilers/active_support_time_ext.rb
@@ -1,0 +1,68 @@
+# typed: strict
+# frozen_string_literal: true
+
+return unless defined?(Time.current) && defined?(ActiveSupport::TimeWithZone)
+
+module Tapioca
+  module Dsl
+    module Compilers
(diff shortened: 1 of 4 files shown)


lib/tapioca/dsl/compilers/active_support_time_ext.rb
```diff
@@ -0,0 +1,68 @@
...
+
+        ConstantType = type_member { { fixed: T.class_of(::Time) } }
+
+        sig { override.void }
+        def decorate
+          return unless constant.respond_to?(:zone)
+
+          root.create_path(constant) do |mod|
+            return_type = if ::Time.zone
+              "::ActiveSupport::TimeWithZone"
```
KaanOzkan, line 48: Should we check for existence of `ActiveSupport::TimeWithZone` in case someone defines their own `Time.zone` without ActiveSupport?

egiurleo (approved):
Nice!

PR #1966, opened 2024-07-21
Fix ActiveRecordRelation compiler's handling of composite primary keys

bitwise-aiden:
### Motivation

During `ActiveRecordRelation` compiler's decorate it generates the wrong method signature of the `#find` method for records with composite keys. The signature for the method appears to have been updated when multiple values are passed (i.e. array of arrays), but the singular value still shows as `T.any(....)` when it should be wrapped in an array.

### Implementation

I've updated the handling so if `#composite_primary_key?` is present and returns a truthy value then the value of ids will be updated to be an array of instead. This then works forward so that the value for an array of composite keys is correct too.

### Tests

Updated tests to reflect changes.


paracycle (approved):
Thanks, good catch!

PR #1935, opened 2024-06-19
Optimize `URLHelpers.gather_constants`

amomchilov:
### Motivation

Profiling `tapioca DSL SomeConst` identified this method as a hot spot, taking roughly 1 out of the 6 total second spent by Tapioca (not including application load time).

The changes in this PR speed up running `tapioca dsl` against Shopify core by **1 entire second!**

## Benchmark results (writeup WIP)

| run   | gather_constants | User time | Syscall time | Total |
|-------|------------------|-----------|--------------|-------|
| Master          | 2s 350ms | 37s 383ms | 6s 760ms | 33s 10ms  |
| [Original improvements](https://github.com/Shopify/tapioca/commit/2344fdf731b73ad01fd6be71317e57e46518790e)        | 1s 333ms | 36s 383ms | 6s 970ms | 32s 216ms |
| [Ufuk suggestion](https://github.com/Shopify/tapioca/pull/1935/commits/70d7fb5491fe655715da1e7f52dbd8ac1085418e) | 1s 473ms | 36s 593ms | 6s 750ms | 32s 313ms |
| [Final improvement](https://github.com/Shopify/tapioca/pull/1935/commits/ef5e6c9674fb77f59cd2f8a4f1003f83a442d73e) | **1s 330ms** | 35s 533ms | 6s 420ms | **31s 27ms** |
| [Set (dumb)](https://github.com/Shopify/tapioca/compare/70d7fb5491fe655715da1e7f52dbd8ac1085418e..684c6d7a98a612f268f814d8f079f0c6756a216b)      | 1s 530ms | 36s 710ms | 6s 657ms | 32s 308ms |
| [Set (smart)](https://github.com/Shopify/tapioca/compare/70d7fb5491fe655715da1e7f52dbd8ac1085418e..c31a9ef73c59a8965ec0679440b10ee7c3e90379)     | 1s 523ms | 37s 260ms | 6s 770ms | 32s 901ms |

<details><summary>Benchmark run details</summary>

### Master

| run   | gather_constants | User time | Syscall time | Total |
|-------|------------------|-----------|--------------|-------|
| run 1 | 2s 350ms | 37s 420ms | 6s 710ms | 32s 889ms |
| run 2 | 2s 410ms | 37s 210ms | 6s 830ms | 32s 992ms |
| run 3 | 2s 290ms | 37s 520ms | 6s 740ms | 33s 149ms |
|  avg  | 2s 350ms | 37s 383ms | 6s 760ms | 33s 10ms  |

[...]

PR #1969, opened 2024-07-23
Flip subtype checks (`<` → `>`)

amomchilov:
### Motivation

In my work on #1935, I discovered that `<` wasn't behaving as expected for some modules, which needs special handling like in #677

e.g.

```ruby
require "xpath"

class DefinitelyNotXPath; end

if XPath < DefinitelyNotXPath
  puts "This shouldn't be called, but it is"
end

# Happens because `XPath#<` is overridden, and always returns a truthy value:
XPath < DefinitelyNotXPath
# => #<XPath::Expression:0x0000000128f39c78
#  @arguments=[:<, #<XPath::Expression:0x0000000128f39e58 @arguments=[], @expression=:this_node>, String],
#  @expression=:binary_operator
# >
```

### Implementation

Just flip the operators around, so the receiver is always the known-good classes which don't override the `Module#>`.

### Tests

I haven't detected any concrete problems caused by this (other than the one already solved in #1935, so I haven't added any test cases.

## Other considerations

It would really be nice to have an automated way of detecting and preventing such cases, but. I can't think of a way to do so without too many false positives. E.g. blanket banning `<` with a RuboCop rule would be too aggressive.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_support_concern.rb
+++ b/lib/tapioca/dsl/compilers/active_support_concern.rb
@@ -73,7 +73,7 @@ module Tapioca
                 # not singleton classes
                 !mod.singleton_class? &&
                 # extend ActiveSupport::Concern, and
-                mod.singleton_class < ActiveSupport::Concern &&
+                ActiveSupport::Concern > mod.singleton_class &&
                 # have dependencies (i.e. include another concern)
                 !dependencies_of(mod).empty?
             end
--- a/lib/tapioca/dsl/compilers/active_support_concern.rb
+++ b/lib/tapioca/dsl/compilers/active_support_concern.rb
@@ -66,18 +66,17 @@ module Tapioca
 
           sig { override.returns(T::Enumerable[Module]) }
           def gather_constants
-            # Find all Modules that are:
             all_modules.select do |mod|
-              # named (i.e. not anonymous)
-              name_of(mod) &&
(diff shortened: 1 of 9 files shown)


lib/tapioca/dsl/compilers/active_support_concern.rb
```diff
@@ -66,19 +66,18 @@ class << self
 
           sig { override.returns(T::Enumerable[Module]) }
           def gather_constants
-            # Find all Modules that are:
```
amomchilov, line 69: The comments here are superfluous IMO, better to clarify the code itself (e.g. by adding a `has_dependencies?` helper)

lib/tapioca/dsl/compilers/smart_properties.rb
```diff
@@ -86,8 +86,9 @@ class << self
           def gather_constants
             all_modules.select do |c|
               name_of(c) &&
-                c != ::SmartProperties::Validations::Ancestor &&
-                c < ::SmartProperties && ::SmartProperties::ClassMethods === c
+                ::SmartProperties > c &&
```
amomchilov, line 89: Checking this first will disqualifier more constants earlier, evaluating fewer operands in this `&&` chain.

It also lines things up nicely :)

KaanOzkan (approved):

Issue #1972, opened 2024-07-29
Suppress "WARNING: Multiple constants with the same name" in LSP mode


amomchilov:
When running the Tapioca `DSLGenerate` from the Ruby LSP, we're seeing these warnings:

> WARNING: Multiple constants with the same name: ...

https://github.com/Shopify/tapioca/blob/21a9a6dcfa9a58ccf794f09ef8a2604859dc8506/lib/tapioca/dsl/pipeline.rb#L182-L183

This didn't used to happen before, because each run was invoked by a separate `bin/tapioca dsl ...` command, in a fresh process, with an independent copy of the Rails server. When running in the LSP, and the Rails server is reused, with multiple `DSLGenerate` calls happening within the lifetime of that single process.

This interacts poorly with Zeitwerk reloading. Each reload of `example.rb` would create a new instance of `Class` called `Example`. Surprisingly, the old copy of `Class` does _not_ go away. When Tapicoa uses `ObjectSpace.each_object(Module)`, it'll find both the old `Example` class, and the new one.

There would have been a risk that Tapioca would non-deterministically render RBIs for a stale version of the constant, but luckily that doesn't happen because it can find the newest one with this `constantize`:

https://github.com/Shopify/tapioca/blob/21a9a6dcfa9a58ccf794f09ef8a2604859dc8506/lib/tapioca/dsl/pipeline.rb#L190

I.e. whichever `Module`/`Class` is currently bound to that constant name, is the newer one.

So because this isn't an issue, perhaps this warning should be suppressed when it happens on the second run of a pipeline? I.e. it'll trigger up to once in the LSP mode, but then never again. For CLI runs, it'll continue to trigger on every run.

PR #1975, opened 2024-07-30
Track and filter out previously seen constants

amomchilov:
### Motivation

Exploring one potential solution for #1972

### Implementation

Every time we run `DSLGenerate`, we'll take note of all the constants we've seen. On future runs, anytime there are duplicate constants, we can filter the already-seen ones, to know which ones were newly introduced.

#### Pitfall:

If there are 2 reloads between Tapioca runs, then there 3 copies of the `Module`/`Class`. We'll remove the 1 we saw last run, but we have no way to decide which is the newer of the remaining 2.

#### `WeakMap`

We could have used a `Set`, but that would keep stale classes alive unnecessarily. There's no such thing as a `WeakSet`, so we just use a `WeakMap` that always has `true` values. ([this is what Set does, anyway](https://github.com/ruby/set/blob/4e95d55294f0e577529ca8702c4d2c3ae90247e7/lib/set.rb#L511-L512)).

#### Memory usage

In core, we have about 3623 named constants that end up in this list.

`WeakMap` uses 20% more memory than a `WeakKeyMap`:

```
Constant  count: 3623
WeakMap    size: 172752 bytes
WeakKeyMap size: 143768 bytes
Hash       size: 114848 bytes
```

<details><summary>memory benchmark</summary>

```ruby
#!/usr/bin/ruby

require "objspace"

module FakeModules; end

CONSTANTS = 3623.times.map do
  FakeModules.const_set("Module#{_1}", Module.new)
end

HASH = {}
WEAK_MAP = ObjectSpace::WeakMap.new
WEAK_KEY_MAP = ObjectSpace::WeakKeyMap.new

CONSTANTS.each do |c|
  HASH[c] = true
  WEAK_MAP[c] = true
  WEAK_KEY_MAP[c] = true
end

memory_stats = [
  ["Hash      ", ObjectSpace.memsize_of(HASH)],
  ["WeakMap   ", ObjectSpace.memsize_of(WEAK_MAP)],
  ["WeakKeyMap", ObjectSpace.memsize_of(WEAK_KEY_MAP)],
].sort_by { |stats| -stats[1] }

lines = memory_stats.map { |stats| "#{stats[0]} size: #{stats[1]} bytes" }

puts <<~MSG
  ===== Memory stats

  Constant  count: #{CONSTANTS.count}
  #{lines.join("\n")}
MSG
```

</details>

### Lookup time

`WeakMap` look-ups were faster than `WeakKeyMap` and `Hash`:

```
===== CPU benchmark
ruby 3.3.0 (2023-12-25 revision 5124f9ac75) [arm64-darwin23]
Comparison:
     WeakMap#member?:   472578.9 i/s
        WeakMap#key?:   467961.4 i/s - same-ish: difference falls within error
    WeakMap#include?:   467399.7 i/s - same-ish: difference falls within error
          WeakMap#[]:   459978.7 i/s - same-ish: difference falls within error
             Hash#[]:   325692.1 i/s - 1.45x  slower
       Hash#include?:   297177.0 i/s - 1.59x  slower
        Hash#member?:   296108.1 i/s - 1.60x  slower
           Hash#key?:   293849.8 i/s - 1.61x  slower
     WeakKeyMap#key?:   281934.9 i/s - 1.68x  slower
       WeakKeyMap#[]:   277954.9 i/s - 1.70x  slower
```

The `1.7x` speedup on lookup is probably worth the `1.2x` memory use.

<details><summary>cpu time benchmark</summary>

```ruby
#!/usr/bin/ruby

require "objspace"
require "benchmark/ips"

module FakeModules; end

CONSTANTS = 3623.times.map do
  FakeModules.const_set("Module#{_1}", Module.new)
end

HASH = {}
WEAK_MAP = ObjectSpace::WeakMap.new
WEAK_KEY_MAP = ObjectSpace::WeakKeyMap.new

CONSTANTS.each do |c|
  HASH[c] = true
  WEAK_MAP[c] = true
  WEAK_KEY_MAP[c] = true
end

puts "===== CPU benchmark\n"

Benchmark.ips do |x|
  x.config(warm: 0.1, time: 10)
  
  x.report("Hash#include?") do |times|
    i = 0
    constants = CONSTANTS.sample(62)
    
    while (i += 1) <= times
      constants.each { HASH.include?(_1) }
    end
  end
  
  x.report("Hash#member?") do |times|
    i = 0
    constants = CONSTANTS.sample(62)
    
    while (i += 1) <= times
      constants.each { HASH.member?(_1) }
    end
  end
  
  x.report("Hash#key?") do |times|
    i = 0
    constants = CONSTANTS.sample(62)
    
    while (i += 1) <= times
      constants.each { HASH.key?(_1) }
    end
  end
  
  x.report("Hash#[]") do |times|
    i = 0
    constants = CONSTANTS.sample(62)
    
    while (i += 1) <= times
      constants.each { HASH[_1] }
    end
  end
  
  x.report("WeakMap#include?") do |times|
    i = 0
    constants = CONSTANTS.sample(62)
    
    while (i += 1) <= times
      constants.each { WEAK_MAP.include?(_1) }
    end
  end
  
 x.report("WeakMap#member?") do |times|
    i = 0
    constants = CONSTANTS.sample(62)
    
    while (i += 1) <= times
      constants.each { WEAK_MAP.member?(_1) }
    end
  end
  
 x.report("WeakMap#key?") do |times|
    i = 0
    constants = CONSTANTS.sample(62)
    
    while (i += 1) <= times
      constants.each { WEAK_MAP.key?(_1) }
    end
  end
  
 x.report("WeakMap#[]") do |times|
    i = 0
    constants = CONSTANTS.sample(62)
    
    while (i += 1) <= times
      constants.each { WEAK_MAP[_1] }
    end
  end
  
   x.report("WeakKeyMap#key?") do |times|
    i = 0
    constants = CONSTANTS.sample(62)
    
    while (i += 1) <= times
      constants.each { WEAK_KEY_MAP.key?(_1) }
    end
  end
  
   x.report("WeakKeyMap#[]") do |times|
    i = 0
    constants = CONSTANTS.sample(62)
    
    while (i += 1) <= times
      constants.each { WEAK_KEY_MAP[_1] }
    end
  end
  
  x.compare!
end
```

<details>

### DIFF
--- a/lib/tapioca/dsl/pipeline.rb
+++ b/lib/tapioca/dsl/pipeline.rb
@@ -1,6 +1,8 @@
 # typed: strict
 # frozen_string_literal: true
 
+require "objspace"
+
 module Tapioca
   module Dsl
     class Pipeline
@@ -166,15 +168,34 @@ module Tapioca
 
       sig { params(constants: T::Set[Module]).returns(T::Set[Module]) }
       def filter_anonymous_and_reloaded_constants(constants)
+        # There's no such thing as  "WeakSet", so we just use a `WeakMa[Module, true]` instead.
+        $previously_seen_constants ||= ObjectSpace::WeakMap.new
+
         # Group constants by their names
         constants_by_name = constants
           .group_by { |c| Runtime::Reflection.name_of(c) }
           .select { |name, _| !name.nil? }
 
+        # TODO: optimize
+        constants_by_name.each do |name, constant_versions|
+          constant_versions.each do
+            $previously_seen_constants[_1] = true
+          end
+        end
+
+        puts "Weak map count: #{$previously_seen_constants.size}, size: #{ObjectSpace.memsize_of($previously_seen_constants)} bytes"
+
         constants_by_name = T.cast(constants_by_name, T::Hash[String, T::Array[Module]])
 
         # Find the constants that have been reloaded
-        reloaded_constants = constants_by_name.select { |_, constants| constants.size > 1 }.keys
+        reloaded_constants = constants_by_name
+          .select do |name, constant_versions|
+            # If there are 2 versions of the same constant, discard all the ones we've seen before.
+            # FIXME: Doesn't work if there are 3 or more version of the same constant,
+            # e.g. if 2 reloads happen between Tapioca runs.
+            constant_versions.count { !$previously_seen_constants.key?(_1) } > 1
+          end
+          .keys
 
         unless reloaded_constants.empty?
           reloaded_constant_names = reloaded_constants.map { |name| "`#{name}`" }.join(", ")


amomchilov:
Wait a minute, there's already a solution for this, several lines down 😆 

https://github.com/Shopify/tapioca/blob/2bdef14fc422b56ea3b217a497c4f21d64358031/lib/tapioca/dsl/pipeline.rb#L211-L212

paracycle:
Yes, see https://github.com/Shopify/tapioca/pull/869 and https://github.com/Shopify/tapioca/issues/867

PR #1978, opened 2024-08-02
Add a sig with no args for ActiveRecordRelations builder methods

bitwise-aiden:
### Motivation

When a model's builder methods are called with no arguments it will default to the first signature that is now `T::Array` param and return. This is obviously incorrect as it will only be instantiating one object. 

### Implementation

I've added a new sig for these methods that omits the `attributes` param, handling this first.

As `#create!` is a part of the builders this will also create the sig for that method. I'm _sure_ there is a way to have a valid call to `#create!` with no params, but even if not I think that it's fine to keep generating this sig for the method.

### Tests

Updated existing tests to reflect changes.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_relations.rb
+++ b/lib/tapioca/dsl/compilers/active_record_relations.rb
@@ -1004,6 +1004,11 @@ module Tapioca
               method.add_opt_param("attributes", "nil")
               method.add_block_param("block")
 
+              method.add_sig do |sig|
+                sig.add_param("block", "T.nilable(T.proc.params(object: #{constant_name}).void)")
+                sig.return_type = constant_name
+              end
+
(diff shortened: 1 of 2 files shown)


bitwise-aiden:
Tangential to this PR, when calling the builder methods with a value that resolves to `T.untyped` the `T::Array` return will be the signature that is picked by default. I know the order we have the sigs at the moment is essential for `T::Array` to work, but is there a sig combo that doesn't result in having to type a bunch of inputs?

paracycle (approved):

KaanOzkan:
> Tangential to this PR, when calling the builder methods with a value that resolves to T.untyped the T::Array return will be the signature that is picked by default. I know the order we have the sigs at the moment is essential for T::Array to work, but is there a sig combo that doesn't result in having to type a bunch of inputs?

Brought this up again in https://github.com/sorbet/sorbet/issues/7744

PR #1985, opened 2024-08-09
Fix ActiveRecordRelation new defined out of class

bitwise-aiden:
### Motivation

Following this PR: https://github.com/Shopify/tapioca/pull/1983

### Implementation

This fixes the issue with the previous where the method of `new` was defined as an instance method instead of a class method.

### Tests

Updated tests to reflect.


paracycle (approved):
Lol, good catch! I can't believe I missed this the first time around. 🤦

PR #1989, opened 2024-08-14
Iterate over requested_constants in DSL helpers

KaanOzkan:
### Motivation

Speedup gather_constants performance when a constant is supplied by not iterating over the ObjectSpace. This PR shouldn't change the functionality because we intersect the constants received from the ObjectSpace iteration with the requested ones https://github.com/Shopify/tapioca/blob/df6e272f4b4dcab5d1a21b2f9aabab1741743603/lib/tapioca/dsl/pipeline.rb#L156

Running `tapioca dsl Constant` before and after in the monolith:
hyperfine tapioca dsl Shop (main)

hyperfine tapioca dsl Shop (perf branch)

### Tests

Most of the tests in the dsl_spec supply arguments to `tapioca dsl` command.

### DIFF
--- a/lib/tapioca/dsl/compiler.rb
+++ b/lib/tapioca/dsl/compiler.rb
@@ -25,6 +25,8 @@ module Tapioca
       sig { returns(T::Hash[String, T.untyped]) }
       attr_reader :options
 
+      @@requested_constants = T.let([], T::Array[Module]) # rubocop:disable Style/ClassVars
+
       class << self
         extend T::Sig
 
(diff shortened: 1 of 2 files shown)


lib/tapioca/dsl/compiler.rb
```diff
@@ -44,20 +46,51 @@ def processable_constants
...
+          type_parameters(:U)
+            .params(klass: T.all(T::Class[T.anything], T.type_parameter(:U)))
+            .returns(T::Array[T.type_parameter(:U)])
+        end
+        def descendants_of(klass)
+          if @@requested_constants.any?
+            T.cast(
+              @@requested_constants.select do |k|
+                k < klass && !k.singleton_class?
+              end,
```
amomchilov, line 66: I'm not sure this will be sufficient.

Imagine we're editing a superclass, and we make changes that'll effect the RBIs for the subclasses. Those subclasses will need to change, but they might not be in our requested constants (only the superclass would be)

lib/tapioca/dsl/compiler.rb
```diff
@@ -25,6 +25,8 @@ class Compiler
       sig { returns(T::Hash[String, T.untyped]) }
       attr_reader :options
 
+      @@requested_constants = T.let([], T::Array[Module])
```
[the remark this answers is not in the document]
amomchilov: This sketches me out a little, but there's a really really nice benefit to this approach: all the existing DSL Compilers will mostly just work, without needing to be updated.

What if we loosened that constraint: could we make a better API if we were willing to change the compilers?

e.g. what if the main module selection was done by something like:

```ruby
sig { abstract.params(candidates: T::Enumerable[Module]).returns(T::Enumerable[Module]) }
def select_constants_to_decorate(candidates)
```

This would replace `gather_constants`. By default, we would call it with `select_constants_to_decorate(all_modules)` (which would work similar to how gather_constants works in practice today), otherwise we could do `select_constants_to_decorate(requested_constants)`.

amomchilov (commented):
Here's my $0.02!

lib/tapioca/dsl/compiler.rb
KaanOzkan: What you're describing is a possibility with requested constants as whole. This PR is just making existing functionality faster. We filter the constants coming from `gather_constants` to match `requested_constants` [here](https://github.com/Shopify/tapioca/blob/main/lib/tapioca/dsl/pipeline.rb#L156).

I think you're coming from the LSP Addon perspective and yes if we want it to cover all cases it'd have to be smarter at identifying which constants to generate. But that's not the scope of this PR.

amomchilov (approved):

PR #1990, opened 2024-08-14
Remove polymorphic check when generating associating methods

matthewarkin:
### Motivation
We noticed when adding some typing to our code that association methods for polymorphic associations were returning a generic `ActiveRecord::Associations::CollectionProxy` instead of the `Model::PrivateCollectionProxy` that is used elsewhere in tapioca.

### Implementation
I believe checking for the polymorphicness of the association is actually unnecessary since it shouldn't ever be the case that the association for a has_many would return a collection of different types.

### Tests
I updated the existing tests to reflect the updated behavior. I don't think I'm missing a test case but thats mostly because I can't think of a case where you woudn't be able to pin the method to a specific model/type.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_associations.rb
+++ b/lib/tapioca/dsl/compilers/active_record_associations.rb
@@ -287,7 +287,7 @@ module Tapioca
         def type_for(reflection)
           validate_reflection!(reflection)
 
-          return "T.untyped" if !constant.table_exists? || polymorphic_association?(reflection)
+          return "T.untyped" if !constant.table_exists?
 
           T.must(qualified_name_of(reflection.klass))
         end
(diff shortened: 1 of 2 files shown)


matthewarkin:
I have signed the CLA!

lib/tapioca/dsl/compilers/active_record_associations.rb
```diff
@@ -287,7 +287,7 @@ def populate_collection_assoc_getter_setter(klass, association_name, reflection)
         def type_for(reflection)
           validate_reflection!(reflection)
 
-          return "T.untyped" if !constant.table_exists? || polymorphic_association?(reflection)
+          return "T.untyped" if !constant.table_exists?
 
           T.must(qualified_name_of(reflection.klass))
```
amomchilov, line 292: Hmmm, I'm getting a failure when running against our core monolith:

```
Error: `Tapioca::Dsl::Compilers::ActiveRecordAssociations` failed to generate RBI for `OurModel`
--
  | /tmp/bundle/ruby/3.4.0+0/bundler/gems/rails-f0fd6ab1259b/activerecord/lib/active_record/reflection.rb:500:in 'ActiveRecord::Reflection::AssociationReflection#compute_class': Polymorphic associations do not support computing the class. (ArgumentError)
  |  
  | raise ArgumentError, "Polymorphic associations do not support computing the class."
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | from /tmp/bundle/ruby/3.4.0+0/bundler/gems/rails-f0fd6ab1259b/activerecord/lib/active_record/reflection.rb:439:in 'ActiveRecord::Reflection::MacroReflection#_klass'
  | from /tmp/bundle/ruby/3.4.0+0/bundler/gems/rails-f0fd6ab1259b/activerecord/lib/active_record/reflection.rb:431:in 'ActiveRecord::Reflection::MacroReflection#klass'
  | from /tmp/bundle/ruby/3.4.0+0/bundler/gems/tapioca-b0eab561b24b/lib/tapioca/dsl/compilers/active_record_associations.rb:292:in 'Tapioca::Dsl::Compilers::ActiveRecordAssociations#type_for'

```

Where `OurModel` looks something like this, with a polymorphic `belongs_to`:

```ruby
class OurModel < ApplicationRecord
   belongs_to :reference, polymorphic: true
end
```

amomchilov (commented):
Hey @matthewarkin, thanks for your first contribution!

We'd be happy to merge this in, but there's an error in our CI that I don't have time to troubleshoot and fix myself. Could you take a look?

lib/tapioca/dsl/compilers/active_record_associations.rb
```diff
@@ -287,7 +287,7 @@ def populate_collection_assoc_getter_setter(klass, association_name, reflection)
         def type_for(reflection)
           validate_reflection!(reflection)
 
-          return "T.untyped" if !constant.table_exists? || polymorphic_association?(reflection)
+          return "T.untyped" if !constant.table_exists?
```
amomchilov, line 290: Lint fix:

```suggestion
          return "T.untyped" unless constant.table_exists?
```

PR #1993, opened 2024-08-18
switch association nil behavior on new column type config option

stathis-alexander:
### Motivation

With [the introduction of the column type option for the ActiveRecordColumn](https://github.com/Shopify/tapioca/pull/1888), we are finally _almost_ able to migrate from the now deprecated Sorbet Rails only Tapioca for our AR model class RBIs. However, the option only affects active record columns that have either a non-null DB index or a `presence: true` validation.

Another common, related pattern that makes Tapioca difficult to adopt is that `has_one` or `belongs_to` associations can also be declared as non-`nil` through the `required` and `optional` kwargs, respectively. Presumably the justification for this behavior is identical to that of the active record columns; i.e., newly instantiated models may not have these relationships setup yes, despite the standard case being that the model is actually persisted and fetched from the DB. 

This PR proposes re-using the introduced column type option to switch the behavior of associations RBIs such that they generated as non-nilable when the association is marked as required or non-optional.

### Implementation

This is not intended to be the final solution - ideally we would DRY up the column type option code here. We should likely rename the column type option here as well - assuming adoption has not been too broad and too swift - to better indicate the behavior.  Instead, I wanted to demonstrate the idea, and I am committed to making whatever adjustments are required to get this merged into the codebase. 

### Tests

I would be glad to go through the effort of adding a robust suite of tests once there is sufficient indication that this change is desirable.


lib/tapioca/dsl/compilers/active_record_associations.rb
```diff
@@ -148,6 +148,20 @@ def gather_constants
 
         private
 
+        ColumnTypeOption = Helpers::ActiveRecordColumnTypeHelper::ColumnTypeOption
+
+        def column_type_option
+          @column_type_option ||= T.let(
+            ColumnTypeOption.from_options(options) do |value, default_column_type_option|
```
KaanOzkan, line 155: Yes this would have to be its own option similar to the previous implementation.

lib/tapioca/dsl/compilers/active_record_associations.rb
```diff
@@ -292,6 +306,29 @@ def type_for(reflection)
...
+        end
+        def single_association_type_for(reflection)
+          return "T.untyped" if column_type_option.untyped?
+
+          association_class = type_for(reflection)
+          return as_nilable_type(association_class) unless column_type_option.persisted?
+
+          case reflection
+          when ActiveRecord::Reflection::BelongsToReflection
+            return association_class if reflection.options[:optional].nil? || !reflection.options[:optional]
```
KaanOzkan, line 322: I was confusing so noting that `nil?` looks at more than expected https://github.com/rails/rails/blob/ac104520ee8518441bb4f8c2c7835d0bb8fd6800/activerecord/lib/active_record/relation/query_attribute.rb#L35-L40

KaanOzkan (commented):
I think this is fine to go ahead with the implementation. I don't have context on the reflection options but it looks normal.

lib/tapioca/dsl/compilers/active_record_associations.rb
stathis-alexander: Hm, apologies but I don't follow your comment. I added the `nil?` check because often times these options are not passed to the association. In that case, we should fall back to default behavior. Let me know if I'm missing something.

stathis-alexander:
@KaanOzkan Thank you for the comments! I have rewritten this for clarity, splitting the type option out into its own proper option. I have chosen a default of "nilable", since this retains the existing default behavior.

spec/tapioca/dsl/compilers/active_record_associations_spec.rb
```diff
@@ -1603,6 +1603,249 @@ def shop_previously_changed?; end
...
+                  RUBY
+
+                  add_ruby_file("user.rb", <<~RUBY)
+                    class User < ActiveRecord::Base
+                    end
+                  RUBY
+
+                  add_ruby_file("default_required_post.rb", <<~RUBY)
+                    class DefaultRequiredPost < ActiveRecord::Base
+                      def self.belongs_to_required_by_default
```
stathis-alexander, line 1735: It seems important to me to test that we respect the default configuration, although this feels like a weak way to test this behavior. I could not figure out how to set the Rails configuration option directly.

KaanOzkan (approved):

stathis-alexander:
@KaanOzkan , apologies for the delay here. I was traveling last week. I have updated the documentation. I believe this is ready for release.

lib/tapioca/dsl/compilers/active_record_associations.rb
```diff
@@ -121,6 +146,50 @@ def initialize(class_name)
           end
         end
 
+        class AssociationTypeOption < T::Enum
+          extend T::Sig
+
+          enums do
+            Nilable = new("nilable")
+            Persisted = new("persisted")
```
KaanOzkan, line 154: Docs talk about an `untyped` option as well and I think it's good to add and test.

stathis-alexander: Apologies - that was a bad copy / paste and an unintended inclusion. I've updated the docs. The two options - nilable and persisted - have tests.

lib/tapioca/dsl/compilers/active_record_associations.rb
```diff
@@ -121,6 +144,50 @@ def initialize(class_name)
...
+            extend T::Sig
+
+            sig do
+              params(
+                options: T::Hash[String, T.untyped],
+                block: T.proc.params(value: String, default_association_type_option: AssociationTypeOption).void,
+              ).returns(AssociationTypeOption)
+            end
+            def from_options(options, &block)
+              column_type_option = Nilable
```
paracycle, line 165:
```suggestion
              association_type_option = Nilable
```

lib/tapioca/dsl/compilers/active_record_associations.rb
```diff
@@ -121,6 +144,50 @@ def initialize(class_name)
...
+              value = options["ActiveRecordAssociationTypes"]
+
+              if value
+                if has_serialized?(value)
+                  column_type_option = from_serialized(value)
```
paracycle, line 170:
```suggestion
                  association_type_option = from_serialized(value)
```

lib/tapioca/dsl/compilers/active_record_associations.rb
```diff
@@ -121,6 +144,50 @@ def initialize(class_name)
...
+                else
+                  block.call(value, column_type_option)
+                end
+              end
+
+              column_type_option
```
paracycle, line 176:
```suggestion
              association_type_option
```

spec/tapioca/dsl/compilers/active_record_associations_spec.rb
paracycle: I think you should be able to set `ActiveRecord::Base.belongs_to_required_by_default = true` in the `schema.rb` file to set this this as default.

This is what Active Record tests seems to be doing: https://github.com/rails/rails/blob/8b0c8079c6ee8460354f8b3cbe2296d3f3f3b21e/activerecord/test/cases/associations/required_test.rb#L27-L40

lib/tapioca/dsl/compilers/active_record_associations.rb
```diff
@@ -292,6 +372,45 @@ def type_for(reflection)
...
+        # if the association's attribute has an unconditional presence validation, it should also be considered
+        # non-optional.
+        sig { params(reflection: ReflectionType).returns(T::Boolean) }
+        def belongs_to_and_non_optional_reflection?(reflection)
+          return false unless reflection.belongs_to?
+
+          required_by_default = !!reflection.active_record.belongs_to_required_by_default
+          return required_by_default if reflection.options[:optional].nil?
+
+          !reflection.options[:optional]
```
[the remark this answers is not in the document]
stathis-alexander: Ah yes, I was operating more from my intiution and not looking directly at the rails source. 

I've used slightly different logic from the. above that makes the hierarchy of the `required` then `optional` keys more obvious.

egiurleo:
Hi, @stathis-alexander, we just fixed CI on Ruby 3.3. Would you mind rebasing on main and force pushing your branch?

stathis-alexander:
Hi @egiurleo , I was under the impression based on the conversation [here](https://github.com/Shopify/tapioca/pull/1996#issuecomment-2347013001) that this pull request would not be accepted.

PR #1996, opened 2024-08-19
generate non-nilable types for columns with unconditional presence validations

stathis-alexander:
### Motivation
Tapioca recently added support for generating active record attribute types in `:persisted` mode, thus removing one of the more annoying aspects of Tapioca's AR compilers. See [here](https://github.com/Shopify/tapioca/pull/1888). However, this does not go far enough, since it is common in the Rails community to prefer model level validations over database level validations. See [here](https://guides.rubyonrails.org/active_record_validations.html#why-use-validations-questionmark). 

This extends that to also generate non-nilable types when there is an unconditional presence validation on the column, thus respecting the application level constraints as well as the database level ones. 

### Implementation
We check the validators exposed by `ActiveRecord::Base` integration directly. 

### Tests
Tests will be added if the change is willing to be accepted.


lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
```diff
@@ -235,6 +235,25 @@ def not_nilable_serialized_column?(column_type)
...
+          column = @constant.columns_hash[column_name]
+          return false if column && !column.null
+          return true unless @constant.respond_to?(:validators_on)
+
+          has_unconditional_presence_validator = @constant.validators_on(column_name).any? do |validator|
+            case validator
+            when ActiveRecord::Validations::PresenceValidator
+              !validator.options.key?(:if) && !validator.options.key?(:unless) && !validator.options.key?(:on)
+            else
+              true
```
KaanOzkan, line 251: Shouldn't this be `false`?

KaanOzkan (commented):
Thanks, I think it makes sense to not mark it as nilable if there are presence validators.

lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
stathis-alexander: Ah yes, thank you. I changed how this was written and the conditions flipped.

stathis-alexander:
@KaanOzkan Thank you for your earlier comments. I have updated the code here to improve readability and added specs. Please let me know if there is anything additional required. 

I believe this change may be breaking - the default behavior prior has been to generate nilable types for these cases, and an unexpected change here may be surprising to others. I will defer to your judgement on versioning.

lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
```diff
@@ -113,7 +115,7 @@ def column_type_for(column_name)
               getter_type
             end
 
-          if @column_type_option.persisted? && !column&.null
+          if @column_type_option.persisted? && (virtual_attribute?(column_name) || !nilable_column)
```
stathis-alexander, line 118: The `virtual_attribute?` check here was surprising to me (I do not have much experience with these), but there was a spec enforcing it, so I made it more explicit and retained the behavior.

KaanOzkan: Oh the default is persisted https://github.com/Shopify/tapioca/blob/f2010b52a1980b51f649ac4277f37a98f0accce5/lib/tapioca/dsl/helpers/active_record_column_type_helper.rb#L32

stathis-alexander: Yes, sorry for the confusion. I just meant that my intuition would be that virtual attributes would be nil-able by default, but the  spec specifically enforced that they were non-nil.

KaanOzkan: So before `!column&.null` would return `false` and we didn't enter this branch. But using `!nilable_column` we are entering this branch and you need the extra conditional correct? I'd like to understand why this difference is happening and make sure it's not a bug in the new implementation. Is there an implicit presence validator in the test model?

stathis-alexander: Hi @KaanOzkan . Yes, this is super subtle and confused me as well. 

Put succinctly, the confusing behavior is that `!column&.null` is `true` if `column` is `nil`, which is the case if the attribute is virtual. This means that we were always falling into the branch for virtual attributes. I don't know whether this is desirable, except that there was a spec relying on this behavior. 

In my code, `has_non_null_database_constraint?` is only `true` if a column exists because we return `false` early if the `column` is `nil`, thus separating out the virtual and non-virtual cases (specifically whether a `column` is found or not), so I need to add an additional explicit check (`virtual_attribute?`) that checks for `nil`-ness of the `column` to retain the existing behavior. I believe this is ultimately clearer and easier to follow.

I hope that helps clarify things.

KaanOzkan: Thank you for the explanation. I'm not sure if the test is enforcing this or it just happened to follow this code path. cc @bdewater

I'm okay with keeping the current behaviour but I'd prefer if it was cleaner. Could we remove `return false if column.nil?` instead in `has_non_null_database_constraint`?

stathis-alexander: > I'm okay with keeping the current behaviour but I'd prefer if it was cleaner. Could we remove return false if column.nil? instead in has_non_null_database_constraint?

I'm happy to make this change if you so prefer, but to me, it's cleaner to make the two conditions explicit and separate. I believe the reason it was confusing in the current code is that they were subtly intertwined. Additionally, it seems to me that a method named `has_non_null_database_constraint?` should be `false` if there is no database column to be constrained. 

Again, happy to make the change to align with the rest with the project and its maintainers standards and preferences, so let me know if my logic falls short for you.

stathis-alexander: > It felt too specific of a concept for type helper to know but I think you're right. I'll wait on @egiurleo's review before shipping. 

Sounds good, I appreciate the time and thoughts here!

> Also there's second commit doesn't have a message, I think it can be squashed in to the first.

Ah yes, we use squash and commit here at AL, and I forget that not everyone does that. Let me update the branch.

stathis-alexander:
@KaanOzkan / @egiurleo  Anything I can do to help proceed this along?

stathis-alexander:
cc: @paracycle - Here I extend the `:persisted` `ColumnTypeOption` that you introduced to respect Active Record validations as well as database level constraints (similar to the functionality introduced [here](https://github.com/Shopify/tapioca/pull/1993)). Any comments or feedback would be much appreciated!

paracycle (changes requested):
To be honest, I am not sure I agree that this compiler should have any opinions on the validators that exist on the column, so I don't think this is something that I would like to accept into Tapioca.

The whole point in the `persisted` option was to only respect what the database schema said about what the field could be, nothing more. As a matter of fact, if we change the nilability of the attribute based on the validations that exist for the attribute, there is a chance that you could break custom validators that might need to guard against the attribute value being `nil`. In that case, Sorbet will be complaining about an unreachable branch, when we all know that the validator could be called with a `nil` attribute. That is already a small risk with `persisted` option, and I don't think we should increase that risk even more.

Also, the code that is reaching into the validators and checking for a validator on this field with no conditions is reaching too much into other parts of the Active Record code base, that, IMO, don't have anything to do with columns/attributes. 

There is also the case where I could have a custom validator that basically does the same presence check, but that would not be considered for nilability, which makes this an inconsistent check, in my opinion.

rafaelfranca:
> It seems to me that allowing one to configure the behavior of the compilers is not too unreasonable and allows users to adopt Tapioca more easily. Ultimately, it seems that would be good for Tapioca?

Not if users are adopting the wrong behavior. With type systems, either you are correct, or you are incorrect. Tapioca should not ever generate incorrect type information. There should not be an option to do so. The option to allow users to adopt Tapioca more easily is `:untyped`.



> Hm. That is too bad. Is there somewhere I can view this discussion? Maybe slack or GitHub issue? I looked but could not find it.

There isn't any discussion. But the reasoning is the one I gave above, but I can expand a little bit more. We can't generate incorrect type information because we are just allowing users to incorrectly type their system. Differently from an "gradual approach", the wrong type information does incur costs to the database, either by making people fell safe when they aren't, or by making harder to use the correct information. The "gradual approach" is `:untyped`.

Just to make it clear, `:persisted` was added as an option for backward compatibility because we by mistake made it the default behavior before we added the options.

stathis-alexander:
@rafaelfranca Thank you for the additional context. I appreciate the time. 

It seems the only option here then is to write our own compilers that produce the types we'd prefer (accurate or inaccurate). It seems there is some desire from others, so that it likely makes sense to create a separate repository in which to host and share these.

marknuzz:
@stathis-alexander I'm a little bummed by the decision but I think it was the right one, and the reasoning makes sense to me. That being said, I think there is a huge need for more compilers and that it would be great for people in the community (i.e. those like yourself and I) to work on them. This goes not just for compilers but rbi definitions as well. 

If you'd be interested I'd love to collaborate on such a project. I think it could help with a lot of the rough areas of adopting and using Sorbet.

stathis-alexander:
Thanks @iMacTia. Definitely a nice approach and looking forward to seeing how it develops. 

For others looking to migrate from `sorbet-rails` to `Tapioca` without having to wait for a proper solution, I've started a gem called [Boba](https://github.com/angellist/boba). 

I've published the custom and extended default compilers we're using to do it there. I've implemented the changes in this PR as well as some others there that allow for generating rbis more consistent with the "persisted" behavior of the `sorbet-rails` ones. It's in early form, but may be enough to get your project off of `sorbet-rails`, which is rapidly becoming out of date (we have / had several monkey patches to support behavior inconsistent with latest rails & ruby).

PR #2010, opened 2024-09-03
Support Postgres Interval, Range, Bit, and BitVarying

alex-tan:
### Motivation

Adds support for this column type.

### Implementation

I followed the other implementations of the postgres column types.

### Tests

No tests since there's not an established way of doing that in the repo for this type of feature.

### DIFF
--- a/lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
+++ b/lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
@@ -196,6 +196,11 @@ module Tapioca
                    ActiveRecord::ConnectionAdapters::PostgreSQL::OID::Array === type
                }
             "T::Array[#{type_for_activerecord_value(column_type.subtype, column_nullability:)}]"
+          when ->(type) {
+                 defined?(ActiveRecord::ConnectionAdapters::PostgreSQL::OID::Range) &&
+                   ActiveRecord::ConnectionAdapters::PostgreSQL::OID::Range === type
+               }
+            "T::Range[#{type_for_activerecord_value(column_type.subtype, column_nullability:)}]"
           else
             as_non_nilable_if_persisted_and_not_nullable(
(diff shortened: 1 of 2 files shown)


amomchilov (approved):
Thanks!

Are there any other PG column types that we should add here?

alex-tan:
I tested these three new columns however [`ActiveRecord::Point`](https://github.com/rails/rails/blob/6b3c4781b31953884371a3bbd7a4aa06457762f7/activerecord/lib/active_record/connection_adapters/postgresql/oid/point.rb#L4) doesn't resolve after the rbis are generated. Is there a place I should add a shim for that?

amomchilov:
@alex-tan It seems to fall under this category, so perhaps you can just declare it in `activerecord_postgresql.rbi` as well?

https://github.com/Shopify/tapioca/blob/28b06aac4f9c97783bb5c23512ae09d50cec45b9/sorbet/rbi/shims/activerecord_postgresql.rbi#L3

```ruby
class Point < Struct.new(:x, :y)
   sig { returns(FloatButImNotSure) } # TODO: Fill in correct types
   def x; end
   
   sig { params(x:FloatButImNotSure).returns(FloatButImNotSure) }
   def x=(x); end
   
   sig { returns(FloatButImNotSure) }
   def y; end

   sig { params(x:FloatButImNotSure).returns(FloatButImNotSure) }
   def y=(x); end
end
```

amomchilov:
Ahh, you're right, these are just shims for Tapioca's own internal use.

Hmmm... I'll ask for guidance internally. To get you unblocked quicker, would you like to roll this PR back to just the original support for `Interval`? I'll approve/merge that right away, then you can split the rest in their own new PR

alex-tan:
Sounds good, thanks 👍

amomchilov (approved):

alex-tan:
@egiurleo sure thing, done 👍

PR #2029, opened 2024-10-02
Fix broken CI on Ruby 3.1

dirceu:
We have a test that was pointing to rails#main; however, main / Rails 8 has dropped support for Ruby 3.1 so the test started failing. The test does seem valuable so we're pointing to rails#7-2-stable instead.

~This won't solve _everything_ though: the CI matrix combination of ruby 3.1 + rails main will be broken.~

~Another option here could be to remove 3.1 from CI; this might be OK given we no longer support that version at Shopify.~

https://github.com/Shopify/tapioca/pull/2029#issuecomment-2389487729

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,6 +37,9 @@ jobs:
       matrix:
         ruby: ["3.1", "3.2", "3.3", "head"]
         rails: ["7.0", "current", "main"]
+        exclude:
+          - ruby: "3.1"
+            rails: "main"
         include:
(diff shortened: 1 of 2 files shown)


KaanOzkan (approved):
>This won't solve everything though: the CI matrix combination of ruby 3.1 + rails main will be broken.

`rails main` should be experimental and always passing. But if not we can skip this test in that particular setup.

EDIT: Okay the issue is experimental only plays a part in running tests and not during `setup-ruby`.

PR #2030, opened 2024-10-03
[Tapioca add-on] Add add-on boiler plate

vinistock:
### Motivation

Create the add-on boiler plate with basic activation and getting a handle to the Rails add-on's runtime client.

### Implementation

This is just boiler plate code for the basic add-on.

### Tests

We are discussing how to test the add-on. Due to the dependency on the Rails add-on, it fails to activate without a Rails app, so we may need to add a dummy one. We can also explore stubbing to see if we can avoid that.


paracycle (approved):
We should be able to add a single-file Rails application into the repo somewhere, right? It doesn't have to be an app with the normal Rails structure and stuff.

andyw8 (approved):

PR #2031, opened 2024-10-03
[Tapioca Add-on] Trigger DSL generation upon file changes

KaanOzkan:
### Motivation

Merge part of tapioca add-on logic
### Implementation

- Bump ruby-lsp and ruby-lsp-rails
- Use new API in ruby-lsp
- Implement `workspace_did_change_watched_files` method that uses the Ruby LSP index to find potentially modified constants in a file and sends them to the Rails app
### Tests

We'll add tests when we can test rails client

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,8 +15,8 @@ gem "irb"
 gem "rubocop-shopify"
 gem "rubocop-sorbet", ">= 0.4.1"
 gem "rubocop-rspec"
-gem "ruby-lsp", ">= 0.19.0"
-gem "ruby-lsp-rails", ">= 0.3.17"
+gem "ruby-lsp", ">= 0.19.1"
+gem "ruby-lsp-rails", ">= 0.3.18"
 
 group :deployment, :development do
   gem "rake"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -303,12 +303,12 @@ GEM
       rubocop (~> 1.51)
     rubocop-sorbet (0.8.5)
       rubocop (>= 1)
-    ruby-lsp (0.19.0)
+    ruby-lsp (0.19.1)
       language_server-protocol (~> 3.17.0)
       prism (>= 1.1, < 2.0)
       rbs (>= 3, < 4)
       sorbet-runtime (>= 0.5.10782)
-    ruby-lsp-rails (0.3.17)
+    ruby-lsp-rails (0.3.18)
       ruby-lsp (>= 0.19.0, < 0.20.0)
     ruby-progressbar (1.13.0)
     shopify-money (3.0.0)
@@ -394,8 +394,8 @@ DEPENDENCIES
   rubocop-rspec
   rubocop-shopify
   rubocop-sorbet (>= 0.4.1)
-  ruby-lsp (>= 0.19.0)
-  ruby-lsp-rails (>= 0.3.17)
+  ruby-lsp (>= 0.19.1)
+  ruby-lsp-rails (>= 0.3.18)
   shopify-money
   sidekiq
   smart_properties
--- a/sorbet/rbi/gems/ruby-lsp-rails@0.3.18.rbi
+++ b/sorbet/rbi/gems/ruby-lsp-rails@0.3.18.rbi
@@ -31,13 +31,13 @@ class RubyLsp::Rails::Addon < ::RubyLsp::Addon
   sig { void }
   def initialize; end
 
-  # source://ruby-lsp-rails/lib/ruby_lsp/ruby_lsp_rails/addon.rb#49
-  sig { override.params(global_state: ::RubyLsp::GlobalState, message_queue: ::Thread::Queue).void }
-  def activate(global_state, message_queue); end
+  # source://ruby-lsp-rails/lib/ruby_lsp/ruby_lsp_rails/addon.rb#50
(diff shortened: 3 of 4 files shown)


lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -60,6 +60,22 @@ def version
 
       sig { params(changes: T::Array[{ uri: String, type: Integer }]).void }
       def workspace_did_change_watched_files(changes)
+        constants = changes.filter_map do |change|
+          path = change[:uri].gsub("file://", "")
+
+          entries = T.must(@index).entries_for(path, RubyIndexer::Entry::Class, RubyIndexer::Entry::Module)
```
KaanOzkan, line 66: Using `grep`, getting subclasses and then filtering them out feels weird to me. Wdyt of an API like this where you have a rest arg and use `==` instead? Or making the second argument optional and returning all entries by default?

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -23,15 +23,15 @@ class Addon < ::RubyLsp::Addon
       def initialize
         super
 
-        @global_state = T.let(nil, T.nilable(RubyLsp::GlobalState))
         @rails_runner_client = T.let(nil, T.nilable(RubyLsp::Rails::RunnerClient))
+        @index = T.let(nil, T.nilable(RubyIndexer::Index))
```
vinistock, line 27: I suggest we continue storing the global state. It contains every important information from the LSP, like configurations, encoding and so on. While we currently only use the index, we may need to access other aspects of it.

lib/ruby_lsp/tapioca/addon.rb
vinistock: You can just use the parent class to filter both at the same time.

```suggestion
          entries = T.must(@index).entries_for(path, RubyIndexer::Entry::Namespace)
```

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -60,6 +60,22 @@ def version
 
       sig { params(changes: T::Array[{ uri: String, type: Integer }]).void }
       def workspace_did_change_watched_files(changes)
+        constants = changes.filter_map do |change|
+          path = change[:uri].gsub("file://", "")
```
vinistock, line 64: For URIs, never manipulate the strings directly as it may produce incorrect results in certain operating systems.


```suggestion
          path = URI(change[:uri]).to_standardized_path
          next unless path
```

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -60,6 +60,22 @@ def version
...
+
+          entries = T.must(@index).entries_for(path, RubyIndexer::Entry::Class, RubyIndexer::Entry::Module)
+          next unless entries
```
vinistock, line 67: This shouldn't be needed. If `entries` is `nil`, then it will already not add anything to the result.

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -60,6 +60,22 @@ def version
...
+
+        return if constants.empty?
+
+        T.must(@rails_runner_client).send_notification("reload", {})
+        T.must(@rails_runner_client).send_notification(
+          "server_addon/delegate",
+          request_name: "dsl",
+          server_addon_name: "Tapioca",
+          constants: constants,
+        )
```
vinistock, line 78: We should add a proper API for this in the Rails add-on. Something like `@rails_runner_client.delegate()`.

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -60,6 +60,22 @@ def version
...
+        constants = changes.filter_map do |change|
+          path = change[:uri].gsub("file://", "")
+
+          entries = T.must(@index).entries_for(path, RubyIndexer::Entry::Class, RubyIndexer::Entry::Module)
+          next unless entries
+        end.flatten
+
+        return if constants.empty?
+
+        T.must(@rails_runner_client).send_notification("reload", {})
```
vinistock, line 72:
```suggestion
        T.must(@rails_runner_client).trigger_reload
```

lib/ruby_lsp/tapioca/addon.rb
KaanOzkan: I can't only use the parent class because it includes singleton classes. So I'm trying to avoid this filtering with a different API in Ruby LSP https://github.com/Shopify/tapioca/compare/main...andyw8/tapioca-lsp#diff-8243d0258619c7955b4277251f121d4b6788da80bcae39fe5130f27019d143bbR64

KaanOzkan: I think for our use case being able to omit the types and do our own filtering in tapioca makes sense. Will you accept that change in Ruby LSP?

vinistock: I mean, we created that API for the Tapioca add-on, so we can change it to whatever makes sense.

andyw8 (approved):

vinistock:
We need to also update the `Addon.get` call to constraint the version to one that defines the new APIs:

```ruby
# We need to change this to >= 0.3.18
addon = T.cast(::RubyLsp::Addon.get("Ruby LSP Rails", ">= 0.3.17", "< 0.4"), ::RubyLsp::Rails::Addon)
```

PR #2032, opened 2024-10-04
Allow control over Thor's output

andyw8:
Addresses https://github.com/Shopify/team-ruby-dx/issues/1297

See `lib/tapioca/logging.rb` for the important parts.

Notes:

* We might want to introduce a RuboCop cop to prevent accidental use of `puts`, `say`, `say_error` etc.
* There are still some places where the output is generated from within Thor, e.g. `create_file` and `remove_file`. I'm setting `verbose: false` if `TAPIOCA_DEVELOPMENT` is set, but this isn't ideal.
* There were a couple of places in the tests where the whitespace had to change, need to look into why.
* There is a type failure for argument forwarding (`...`) that I need still address.
* It would be nice if the tests could help us validate that we are always outputting to the logger and never directly to stdout, but that could be tricky.

Here is an example of the output written to the logs:

```
# Logfile created on 2024-10-08 12:54:33 -0400 by logger.rb/v1.6.1
I, [2024-10-08T12:54:33.421152 #10336]  INFO -- : Removing RBI files of gems that have been removed:
I, [2024-10-08T12:54:33.422190 #10336]  INFO -- : Nothing to do.
I, [2024-10-08T12:54:33.422234 #10336]  INFO -- : Generating RBI files of gems that are added or updated:
I, [2024-10-08T12:54:33.422327 #10336]  INFO -- : Nothing to do.
I, [2024-10-08T12:54:33.422427 #10336]  INFO -- : Checking generated RBI files... 
I, [2024-10-08T12:54:33.805008 #10336]  INFO -- :  Done
I, [2024-10-08T12:54:33.805374 #10336]  INFO -- :   No errors found


I, [2024-10-08T12:54:33.805389 #10336]  INFO -- : All operations performed in working directory.
I, [2024-10-08T12:54:33.805397 #10336]  INFO -- : Please review changes and commit them.
```

### DIFF
--- a/lib/tapioca/loaders/gem.rb
+++ b/lib/tapioca/loaders/gem.rb
@@ -58,7 +58,7 @@ module Tapioca
 
       sig { void }
       def require_gem_file
-        say("Requiring all gems to prepare for compiling... ")
+        logger.info("Requiring all gems to prepare for compiling... ")
         begin
           load_bundle(@bundle, @prerequire, @postrequire, @halt_upon_load_error)
         rescue LoadError => e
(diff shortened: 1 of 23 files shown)


lib/tapioca/gemfile.rb
```diff
@@ -179,7 +179,7 @@ def contains_path?(path)
       sig { void }
       def parse_yard_docs
         files.each do |path|
-          YARD.parse(path.to_s, [], Logger::Severity::FATAL)
+          YARD.parse(path.to_s, [], ::Logger::Severity::FATAL)
```
andyw8, line 182: (to avoid clashing with the `Logger` gem)

lib/tapioca/loaders/gem.rb
```diff
@@ -58,7 +58,7 @@ def initialize(bundle:, prerequire:, postrequire:, default_command:, halt_upon_l
 
       sig { void }
       def require_gem_file
-        say("Requiring all gems to prepare for compiling... ")
+        Tapioca.configuration.logger.info("Requiring all gems to prepare for compiling... ")
```
andyw8, line 61: (obviously we will make this less verbose)

PR #2035, opened 2024-10-07
[Tapioca add-on] Add `ServerAddon` boilerplate

KaanOzkan:
### Motivation

Add ServerAddon usage to be called by Rails app based on https://github.com/Shopify/tapioca/compare/main...andyw8/tapioca-lsp


lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -62,6 +63,9 @@ def version
 
       sig { params(changes: T::Array[{ uri: String, type: Integer }]).void }
       def workspace_did_change_watched_files(changes)
+        File.write("output.txt", "\nFiles changed", mode: "a")
```
vinistock, line 66: ✂️

lib/ruby_lsp/tapioca/server_addon.rb
```diff
@@ -0,0 +1,36 @@
...
+# frozen_string_literal: true
+
+begin
+  # The Tapioca add-on depends on the Rails add-on to add a runtime component to the runtime server. We can allow the
+  # add-on to work outside of a Rails context in the future, but that may require Tapioca spawning its own runtime
+  # server
+  require "ruby-lsp-rails"
+rescue LoadError
+  return
+end
```
vinistock, line 11: Is this necessary given that we don't activate the add-on at all if `ruby-lsp-rails` is not present?

lib/ruby_lsp/tapioca/server_addon.rb
```diff
@@ -0,0 +1,36 @@
...
+
+require "tapioca/internal"
```
vinistock, line 13: Now that the runtime functionality is all here, I think we can remove the internals require from the `addon.rb` file. We're not interacting with Tapioca there, only here.

vinistock (approved):

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -62,6 +61,9 @@ def version
 
       sig { params(changes: T::Array[{ uri: String, type: Integer }]).void }
       def workspace_did_change_watched_files(changes)
+        return unless T.must(@global_state).experimental_features
+        return unless @rails_runner_client # Client is not ready
```
KaanOzkan, line 65: I personally prefer multiple `unless` calls over combining them and thinking of De Morgan's Laws 😬

PR #2037, opened 2024-10-08
[Tapioca add-on] Generate DSL RBIs in add-on mode

KaanOzkan:
### Motivation

Cleaned up and brought changes from https://github.com/Shopify/tapioca/compare/main...ko/tapioca-lsp
### Tests

Tested briefly manually, not extensive.

### DIFF
--- a/lib/ruby_lsp/tapioca/server_addon.rb
+++ b/lib/ruby_lsp/tapioca/server_addon.rb
@@ -20,7 +20,8 @@ module RubyLsp
       private
 
       def dsl(params)
-        # TODO
+        load("tapioca/cli.rb") # Reload the CLI to reset thor defaults between requests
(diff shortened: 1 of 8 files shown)


lib/ruby_lsp/tapioca/server_addon.rb
```diff
@@ -20,7 +20,8 @@ def execute(request, params)
       private
 
       def dsl(params)
-        # TODO
+        load("tapioca/cli.rb") # Reload the CLI to reset thor defaults between requests
```
vinistock, line 23: Ugh, this is a bit odd. I think this is fine for the first iteration, but it would be nice to invest in separating the CLI from the add-on interface.

vinistock (approved):

PR #2038, opened 2024-10-09
WIP: handles_path

andyw8:


lib/tapioca/dsl/compilers/protobuf.rb
```diff
@@ -174,6 +174,11 @@ def gather_constants
             results = results.concat(enum_modules)
             results.any? ? results + [Google::Protobuf::RepeatedField, Google::Protobuf::Map] : []
           end
+
+          sig { params(path: String).returns(T::Boolean) }
+          def handles_path?(path)
+            File.fnmatch("**/*.proto", path, File::FNM_PATHNAME | File::FNM_EXTGLOB)
+          end
```
KaanOzkan, line 181: I'm not sure this is needed since I think after the `.proto` change user would regenerate the relevant ruby file and trigger the file watching with appropriate constants. Mentioned verifying this as part of the issue.

andyw8: Yeah that would sense. I've never worked with `.proto` files. (cc @vinistock ).

andyw8:
Will close since we're not actively working. Still useful for future reference.

PR #2041, opened 2024-10-11
[tapioca add-on] Bump depend_on_ruby_lsp! version

andyw8:
So that I can point to Ruby LSP's `main` during development.

### DIFF
--- a/lib/ruby_lsp/tapioca/addon.rb
+++ b/lib/ruby_lsp/tapioca/addon.rb
@@ -1,7 +1,7 @@
 # typed: strict
 # frozen_string_literal: true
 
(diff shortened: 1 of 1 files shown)


lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -1,7 +1,7 @@
 # typed: strict
 # frozen_string_literal: true
 
-RubyLsp::Addon.depend_on_ruby_lsp!(">= 0.19.1", "< 0.20")
+RubyLsp::Addon.depend_on_ruby_lsp!(">= 0.20", "< 0.21")
```
KaanOzkan, line 4: Is it possible to change it locally instead during development? I'm afraid of it degrading the internal testing since Ruby LSP isn't automatically bumped right away in my experience.

PR #2044, opened 2024-10-15
Add hash return type for relation groupchain size

marknuzz:
### Motivation
`Model.where(...).group(...).size` returns a hash, not a scalar number. 

### Implementation
This is a special case where `ActiveRecord::Calculations` doesn't handle the size method. 

https://github.com/rails/rails/blob/6f57590388ca38ed2b83bc1207a8be13a9ba2aef/activerecord/lib/active_record/relation.rb#L273-L280

```ruby
# activerecord-7.1.4/lib/active_record/relation.rb
def size
  if loaded?
    records.length
  else
    count(:all)
  end
end
```

```ruby
Model.group('created_at').count(:all) == Model.group('created_at').size
=> true
```

We can handle this by adding a special case for the size method as we can't rely on the Calculations class to provide the name of the method.

### Tests
Yes

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_relations.rb
+++ b/lib/tapioca/dsl/compilers/active_record_relations.rb
@@ -372,7 +372,7 @@ module Tapioca
             return_type: "T.self_type",
           )
 
-          CALCULATION_METHODS.each do |method_name|
+          (CALCULATION_METHODS + [:size]).each do |method_name|
             case method_name
             when :average, :maximum, :minimum
               klass.create_method(
@@ -400,9 +400,9 @@ module Tapioca
                 ],
                 return_type: "T::Hash[T.untyped, Integer]",
               )
-            when :sum
+            when :sum, :size
               klass.create_method(
-                "sum",
+                method_name.to_s,
                 parameters: [
                   create_opt_param("column_name", type: "T.nilable(T.any(String, Symbol))", default: "nil"),
                   create_block_param("block", type: "T.nilable(T.proc.params(record: T.untyped).returns(T.untyped))"),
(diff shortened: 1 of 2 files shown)


lib/tapioca/dsl/compilers/active_record_relations.rb
```diff
@@ -376,7 +376,7 @@ def create_group_chain_methods(klass)
             return_type: "T.self_type",
           )
 
-          CALCULATION_METHODS.each do |method_name|
+          (CALCULATION_METHODS + [:size]).each do |method_name|
```
amomchilov, line 379: Is there any reason to not add `:size` into the `CALCULATION_METHODS` array itself?

amomchilov: Hey Mark!

> where the method name doesn't fit nicely into into an array and the method should be defined outside of the scope of the case statement looped over an array.

That seems reasonable to me.

> If you give an argument like User.where(foo: 1) you do not in fact get to use the methods on the PrivateAssociationRelationWhereChain returned.

I can't test this right now, could you explain the issue?

> And now I can fully appreciate how difficult the job is of the Tapioca team!

Haha thanks! Trying to tame Ruby meta-programming with types is a formidable challenge :)

marknuzz:
I've been extremely busy lately so I may not have time to get to it this week but will do what I can

lib/tapioca/dsl/compilers/active_record_relations.rb
amomchilov: Regarding this:

> And when looking into how the .to_a worked, I'm now finding that the GeneratedRelationMethods#where chain typing is actually incorrect. If you give an argument like User.where(foo: 1) you do not in fact get to use the methods on the PrivateAssociationRelationWhereChain returned. Additionally, you can't call :to_a on an object that is a where chain (User.where.not.to_a), unlike what the generated rbi suggests.

Ufuk has opened a PR to make "where chains" no longer be relations: #553048

egiurleo:
@marknuzz When you're ready, it's worth rebasing on main to get the latest CI fixes and see if that fixes your PR too.

PR #2046, opened 2024-10-15
Bump expected rack version in tests

KaanOzkan:
### Motivation

Fixes a CI failure https://github.com/Shopify/tapioca/actions/runs/11338670378/job/31571952145?pr=2043#step:6:71


amomchilov (approved):
Would it be preferable to make these expectations "looser", so this doesn't need to be maintained?

KaanOzkan:
Yeah I think it's fine to add support for a `assert_rbi_exists` in spoom with a wildcard. I'll take a look.

PR #2049, opened 2024-10-16
Assert project file existence using patterns

KaanOzkan:
### Motivation

Came up in https://github.com/Shopify/tapioca/pull/2046

### Implementation

Use the built in pattern support for `Dir[]`

### DIFF
--- a/spec/spec_with_project.rb
+++ b/spec/spec_with_project.rb
@@ -104,6 +104,13 @@ module Tapioca
       assert(@project.file?(path))
     end
 
+    # Assert that `path` exists inside project.
+    # `Path` can include patterns such as `*`, useful for testing RBIs for real gems
+    sig { params(path: String).void }
+    def assert_project_file_match(path)
+      assert(Dir[path].any?)
+    end
+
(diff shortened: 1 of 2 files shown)


spec/spec_with_project.rb
```diff
@@ -104,6 +104,13 @@ def assert_project_file_exist(path)
       assert(@project.file?(path))
     end
 
+    # Assert that `path` exists inside project.
+    # `Path` can include patterns such as `*`, useful for testing RBIs for real gems
+    sig { params(path: String).void }
+    def assert_project_file_exist_with_pattern(path)
```
Morriar, line 110:
```suggestion
    def assert_project_file_match_pattern(path)
```

Morriar: Or even shorter:

```suggestion
    def assert_project_file_match(path)
```

Morriar (approved):

spec/spec_with_project.rb
KaanOzkan: Thanks. Changed the name.

PR #2051, opened 2024-10-18
Gems: Fix gem RBI compilation for aliased methods where the original method has a sig

marknuzz:
### Motivation
I was working on porting a small gem to use Sorbet, with sigs on all of the methods defined within the gem. I noticed that the generated RBI from tapioca gem did not produce the correct output for the methods which were aliased. The gem contained a couple of aliased methods. It seemed worth looking into so I gave it a shot, and was able to fix it. 

### Implementation
I made sure to play it safe, by confirming that it is an alias method, and that the original method is a valid method that can be referenced. We use the signature for the original method instead of the alias method, and only if the alias method has no signature directly available (which may be true for 100% of alias methods, but I didn't want to assume). 

sorbet-runtime does have a way to internally build the signature of an aliased method, but it proved to be too deep in the private namespace and I abandoned that idea. If you actually call the aliased method, not once, but twice, the method will be unwrapped and you can get the signature of the aliased method through sorbet's API. But I couldn't identify any way to trigger that without calling the method, so we just use the signature of the original method if it is available. 

If it is an abstract method, we set the aliased signature to nil (as it already would have been before this change), as unintended effects may occur from an aliasing an abstract method. I'm not sure if anyone does it, but in case they do, we don't want to cause a breaking change for them.

### Tests
Added unit tests, and tested the gems command (with this fix) on all of the gems I had installed locally. Gems typically don't use signatures in methods, and of those that do, fewer will alias them. So I could only trigger this for my own gem I was working on locally. However, this may have more of a noticeable effect for companies that are using private gems internally.

### DIFF
--- a/lib/tapioca/gem/listeners/methods.rb
+++ b/lib/tapioca/gem/listeners/methods.rb
@@ -73,15 +73,29 @@ module Tapioca
           return unless method_owned_by_constant?(method, constant)
           return if @pipeline.symbol_in_payload?(symbol_name) && !@pipeline.method_in_gem?(method)
 
+          is_alias = method.name != method.original_name
           signature = lookup_signature_of(method)
-          method = T.let(signature.method, UnboundMethod) if signature
+          parameters = T.let(nil, T.nilable(T::Array[[Symbol, T.nilable(Symbol)]]))
+
+          if signature
+            method = T.let(signature.method, UnboundMethod)
+          elsif is_alias && signature.nil? && constant.method_defined?(method.original_name)
+            alias_source_method = constant.instance_method(method.original_name)
+            signature = lookup_signature_of(alias_source_method)
+
+            # Skip abstract methods if they are defined this way.
+            # Aliasing abstract methods is likely to yield unwanted results
+            signature = nil if signature&.mode == "abstract"
+            parameters = T.let(signature&.method&.parameters, T.nilable(T::Array[[Symbol, T.nilable(Symbol)]]))
+          end
 
           method_name = method.name.to_s
           return unless valid_method_name?(method_name)
           return if struct_method?(constant, method_name)
           return if method_name.start_with?("__t_props_generated_")
 
-          parameters = T.let(method.parameters, T::Array[[Symbol, T.nilable(Symbol)]])
+          method = T.let(signature.method, UnboundMethod) if signature
+          parameters ||= T.let(method.parameters, T::Array[[Symbol, T.nilable(Symbol)]])
 
           sanitized_parameters = parameters.each_with_index.map do |(type, name), index|
             fallback_arg_name = "_arg#{index}"


lib/tapioca/gem/listeners/methods.rb
```diff
@@ -73,15 +73,29 @@ def compile_method(tree, symbol_name, constant, method, visibility = RBI::Public
...
+          if signature
+            method = T.let(signature.method, UnboundMethod)
+          elsif is_alias && signature.nil? && constant.method_defined?(method.original_name)
+            alias_source_method = constant.instance_method(method.original_name)
+            signature = lookup_signature_of(alias_source_method)
+
+            # Skip abstract methods if they are defined this way.
+            # Aliasing abstract methods is likely to yield unwanted results
+            signature = nil if signature&.mode == "abstract"
+            parameters = T.let(signature&.method&.parameters, T.nilable(T::Array[[Symbol, T.nilable(Symbol)]]))
```
KaanOzkan, line 89:
```suggestion
            parameters = signature&.method&.parameters
```
I don't think we need this

KaanOzkan: I see. We don't enforce `strong` so it's fine if these are `T.untyped` but given we iterate over it types are useful.

KaanOzkan (approved):
Thanks for the contribution.

lib/tapioca/gem/listeners/methods.rb
```diff
@@ -73,15 +73,29 @@ def compile_method(tree, symbol_name, constant, method, visibility = RBI::Public
...
+          end
 
           method_name = method.name.to_s
           return unless valid_method_name?(method_name)
           return if struct_method?(constant, method_name)
           return if method_name.start_with?("__t_props_generated_")
 
-          parameters = T.let(method.parameters, T::Array[[Symbol, T.nilable(Symbol)]])
+          method = T.let(signature.method, UnboundMethod) if signature
```
KaanOzkan, line 97: Why was this moved after the checks on line 92?

lib/tapioca/gem/listeners/methods.rb
```diff
@@ -73,15 +73,29 @@ def compile_method(tree, symbol_name, constant, method, visibility = RBI::Public
...
 
+          is_alias = method.name != method.original_name
           signature = lookup_signature_of(method)
-          method = T.let(signature.method, UnboundMethod) if signature
+          parameters = T.let(nil, T.nilable(T::Array[[Symbol, T.nilable(Symbol)]]))
+
+          if signature
+            method = T.let(signature.method, UnboundMethod)
+          elsif is_alias && signature.nil? && constant.method_defined?(method.original_name)
+            alias_source_method = constant.instance_method(method.original_name)
```
paracycle, line 83: In general, this is incorrect since alias binding happens at call site and doesn't get mutated. See: https://bsky.app/profile/fxn.bsky.social/post/3lbehp5rtes2j

```ruby
class A
  def foo = 42

  alias :bar :foo
end

class B < A
  def foo = 1
end

B.new.foo # => 1
B.new.bar # => 42
```

paracycle (changes requested):

PR #2052, opened 2024-10-18
[Tapioca Add-on] Optimize File Save Handling

alexcrocha:
### Motivation

Currently, each file save triggers DSL generation, even when there are no code changes. This leads to unnecessary processing. This PR addresses the issue by optimizing the save handling process.
 
### Implementation

Implemented a checksum-based approach using `Zlib.crc32`. By calculating and storing file checksums, we can skip DSL generation for unchanged files, reducing redundant processing.

### DIFF
--- a/lib/ruby_lsp/tapioca/addon.rb
+++ b/lib/ruby_lsp/tapioca/addon.rb
@@ -12,6 +12,8 @@ rescue LoadError
   return
 end
 
+require "zlib"
+
 module RubyLsp
   module Tapioca
     class Addon < ::RubyLsp::Addon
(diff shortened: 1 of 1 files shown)


lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -71,6 +74,21 @@ def workspace_did_change_watched_files(changes)
           path = URI(change[:uri]).to_standardized_path
           next if path.end_with?("_test.rb", "_spec.rb")
 
+          case change[:type]
+          when Constant::FileChangeType::CREATED, Constant::FileChangeType::CHANGED
+            content = File.read(path)
+            current_checksum = Zlib.crc32(content).to_s
+
+            if change[:type] == Constant::FileChangeType::CHANGED && @file_checksums[path] == current_checksum
+              $stderr.puts "File has not changed. Skipping #{path}"
```
andyw8, line 83: This _might_ be a little noisy to show every time, but we can keep it for now.

alexcrocha: Good point. Let's monitor it and I'll remove if it's too much.

PR #2057, opened 2024-10-29
Remove rubocop-rspec

andyw8:
I noticed in one of the Dependabot updates that we have this unnecessary dependency.

(the project [initially used RSpec](https://github.com/Shopify/tapioca/pull/66))


andyw8:
Ah, I see.

PR #2060, opened 2024-10-30
Use Prism to determine valid names

vinistock:
### Motivation

While hunting for memory optimizations, I noticed that a lot of memory was being retained by `RubyVM::InstructionSequence`. The only place in Tapioca where we use this is to determine if method or parameter names are valid to avoid generating incorrect RBIs.

We can ditch this internal API and rely on Prism to achieve the same, which won't retain memory and is actually faster.

### Implementation

Switched to using Prism for determining if a method/parameter name is valid. I benchmarked the current approach vs the one on this branch and these were the results:

```
Tapioca
  RBIHelper
ruby 3.3.4 (2024-07-09 revision be1089c8ec) [arm64-darwin23]
Warming up --------------------------------------
                 new   474.000 i/100ms
Calculating -------------------------------------
                 new      4.767k (± 3.8%) i/s  (209.79 μs/i) -     24.174k in   5.079176s

Comparison:
                 new:     4766.7 i/s
                 old:     3113.2 i/s - 1.53x  slower
```

### Tests

Current tests should cover it.


lib/tapioca/internal.rb
```diff
@@ -23,6 +23,7 @@
 require "thor"
 require "yaml"
 require "yard-sorbet"
+require "prism"
```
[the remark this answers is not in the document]
vinistock: I think it should be fine since we will always depend on `rbi` and in turn it will always depend on `prism`.

However, I would like to start a discussion about folding the `rbi` gem into Tapioca, since it's the only consumer of the gem. It would make maintenance a bit easier, fewer dependencies to manage, not need for RBI extensions in Tapioca and so on.

lib/tapioca/helpers/rbi_helper.rb
```diff
@@ -107,36 +107,22 @@ def as_non_nilable_type(type)
...
-      iseq = RubyVM::InstructionSequence.compile("def #{name}; end", nil, nil, 0, false)
-      # pull out the first operation in the instruction sequence and its first argument
-      op, arg, _data = iseq.to_a.dig(-1, 0)
-      # make sure that the operation is a method definition and the method that was
-      # defined has the expected name, for example, for `def !foo; end` we don't get
-      # a syntax error but instead get a method defined as `"foo"`
-      op == :definemethod && arg == name.to_sym
-    rescue SyntaxError
-      false
+      # Special case: Prism supports parsing `def @foo; end`, but the Sorbet parser doesn't. This condition can go away
```
amomchilov, line 110: > Prism supports parsing `def @foo; end`

Is that a bug?

lib/tapioca/helpers/rbi_helper.rb
```diff
@@ -107,36 +107,22 @@ def as_non_nilable_type(type)
...
+      # once Sorbet is using Prism under the hood as it will no longer result in an RBI that Sorbet can't parse
+      return false if name.start_with?("@")
+
+      result = Prism.parse("def #{name}(a); end")
```
amomchilov, line 114: Heh, clever.

lib/tapioca/helpers/rbi_helper.rb
```diff
@@ -107,36 +107,22 @@ def as_non_nilable_type(type)
...
-      # pull out the first operation in the instruction sequence and its first argument and data
-      op, arg, data = iseq.to_a.dig(-1, 0)
-      # make sure that:
-      # 1. a method was defined, and
-      # 2. the method has the expected method name, and
-      # 3. the method has a keyword parameter with the expected name
-      op == :definemethod && arg == sentinel_method_name && data.dig(11, :keyword, 0) == name.to_sym
-    rescue SyntaxError
-      false
+      Prism.parse("def sentinel_method_name(#{name}:); end").success?
```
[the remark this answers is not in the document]
vinistock: That feature is not released yet.

lib/tapioca/helpers/rbi_helper.rb
paracycle, line 114: Why not test `Prism.parse("def self.#{name}(a); end")` instead? I think it would address both the `@foo` concern and the `self.foo` one with just the parse.

vinistock: Oh, great idea https://github.com/Shopify/tapioca/pull/2062.

lib/tapioca/helpers/rbi_helper.rb
kddnewton: https://github.com/ruby/prism/pull/3197

PR #2063, opened 2024-10-31
WIP: [Tapioca Addon] Support gem RBI generation

alexcrocha:
# WIP

This Draft PR is still a work in progress.

### Motivation

To support gem RBI generation, we needed a way to detect changes in `Gemfile.lock`. Currently, changes to this file cause the Ruby LSP to restart, resulting in loss of access to any previous state information. This limitation prevents monitoring and processing of gem changes.

### Implementation

To address this, a snapshot file of `Gemfile.lock` is created to persist data across server restarts. Upon restart, both the snapshot and the current `Gemfile.lock` are parsed using `Bundler::LockfileParser`. If differences are detected, the relevant gem names and specifications are extracted, allowing us to then trigger the gem RBI generation.

This implementation handles gem additions, spec changes, and gem removals.


vinistock:
Independent of the code changes themselves, I still have doubts about the UX here. Take this scenario for example:

1. You are working on a branch developing some feature
2. Now, you need to rebase your branch on the latest main
3. The latest main has lockfile changes, but those weren't accompanied by the necessary RBI updates
4. After the rebase, the LSP will immediately generate gem RBIs that are completely unrelated to your work, resulting in untracked changes

I believe we need to try to come up with a strategy to be able to tell if the current developer made the update to the lockfile vs it being updated by something else (like git pull/switching branches/rebasing). Otherwise, it will result in a pretty annoying experience to constantly get unrelated changes.

Maybe, we could remember the launch timestamp of the add-on. When the lockfile is updated, it will trigger a restart and then the add-on can check if the update to the lockfile happened after the previous launch (which should indicate that the current developer made the update). I don't know, there might be better options on how to achieve this.

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -106,6 +111,60 @@ def workspace_did_change_watched_files(changes)
...
+        # If the Ruby LSP was stopped shortly after the last git checkout/pull operation, we don't need to regenerate
+        # RBIs since any Gemfile.lock changes were likely from version control, not from running bundle install
+        return if (ruby_lsp_stop_time - git_timestamp) <= GIT_OPERATION_THRESHOLD
+
+        process_gemfile_changes
+      end
+
+      sig { returns(T.nilable(Time)) }
+      def fetch_last_git_operation_time
+        git_reflog_output = %x(git reflog --date=iso | grep -E "checkout|pull" | head -n 1).strip
```
andyw8, line 138:
```suggestion
        git_reflog_output = %x(git reflog --date=iso --grep-reflog='^\(checkout\|pull\)' -1).strip
```
This approach has a couple of advantages:

* It helps prevent mismatches due to a commit messages co-incidentally containing the word 'checkout' or 'pull'. (The `^` means match only at the start of the line).
* It avoids the need for external programs which may not be available on Windows.

alexcrocha:
Closed in favour of https://github.com/Shopify/tapioca/pull/2081

PR #2070, opened 2024-11-08
Use method overloading to improve WhereChain definition

paracycle:
### Motivation

`ActiveRecord::QueryMethods#where` returns a `ActiveRecord::QueryMethods::WhereChain` instance only when it is called with no arguments. When called with arguments, it returns the same `ActiveRecord::Relation` instance.

Previously, we didn't have proper overloading support in RBI files, so the best thing we could do was act like `ActiveRecord::QueryMethods::WhereChain` was a subclass of the related relation class. This was not ideal because it would allow users to call methods on `WhereChain` that were not actually defined on it. For example, one could do `User.all.where(name: "John").not` which is not valid.

### Implementation

Now that we have support for overloading, we can define `where` to return a `WhereChain` instance only when called with no arguments, and in all other cases, we can say that it returns the same relation instance.

### Tests

Updated existing tests.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_relations.rb
+++ b/lib/tapioca/dsl/compilers/active_record_relations.rb
@@ -210,10 +210,6 @@ module Tapioca
             query_methods |= ActiveRecord::SpawnMethods.instance_methods(false)
             # Remove the ones we know are private API
             query_methods -= [:all, :arel, :build_subquery, :construct_join_dependency, :extensions, :spawn]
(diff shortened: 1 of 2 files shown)


egiurleo (approved):
Nice! Thanks for fixing this. Not sure what's up with CI, but I can investigate if you need any support.

amomchilov (approved):
Nice!

egiurleo:
Force pushed to get CI fixes :)

PR #2073, opened 2024-11-14
Call `to_a` on cached `ObjectSpace` enumerator

### LINKED ISSUE #1974: `all_modules`/`all_classes` scan the object space every time
These memoizations don't work like they were likely intended to:

https://github.com/Shopify/tapioca/blob/21a9a6dcfa9a58ccf794f09ef8a2604859dc8506/lib/tapioca/dsl/compiler.rb#L49-L64

The Enumerator is cached, but it doesn't internally cache the objects it sees when enumerated. Each call will walk the ObjectSpace over again. Minimal demo:

```ruby
e = ObjectSpace.each_object(Object)
irb(main):026> e.count
38593
irb(main):027> e.count
42984 # Different count
irb(main):028> e.count
47371 # Different count
```

If the values were cached as intended, the counts should've been the same each time.

# Proposed solution

We should probably `to_a` these, and cache that instead.

```diff
-   sig { returns(T::Enumerable[T::Class[T.anything]]) } 
+   sig { returns(T::Array[T::Class[T.anything]]) } 
    def all_modules
      @all_classes ||= T.let( 
-      ObjectSpace.each_object(Class), 
+      ObjectSpace.each_object(Class).to_a, 
-      T.nilable(T::Enumerable[T::Class[T.anything]]),
+      T.nilable(T::Array[T::Class[T.anything]]),
      ) 
    end 
```

# Further ideas

We should probably also derive `all_classes` from `all_modules`. Filtering that smaller Array is faster than a second walk through the ObjectSpace.

```ruby
    sig { returns(T::Array[T::Class[T.anything]]) } 
    def all_modules
      @all_classes ||= T.let( 
        all_modules.filter { Class > _1 }, 
        T.nilable(T::Array[T::Class[T.anything]]),
      ) 
    end 
```

warhammerkid:
Based on the existence of the `@@requested_contants` code path, it seems safe for this method to return an `Array` in some cases, so simply caching the array of all `Modules` rather than the enumerator should be safe. Also, I have included the recommended optimization to derive `all_classes` from `all_modules`, which both simplifies things (no need for the `@@requested_constants` code path in both methods), as well as improves performance by skipping a full ObjectSpace traversal again.

### Motivation
Closes #1974

### Tests
This is simply an optimization to an existing tested behavior.

### DIFF
--- a/lib/tapioca/dsl/compiler.rb
+++ b/lib/tapioca/dsl/compiler.rb
@@ -74,11 +74,7 @@ module Tapioca
         sig { returns(T::Enumerable[T::Class[T.anything]]) }
         def all_classes
           @all_classes ||= T.let(
-            if @@requested_constants.any?
-              @@requested_constants.grep(Class)
-            else
-              ObjectSpace.each_object(Class)
-            end,
+            all_modules.grep(Class).freeze,
             T.nilable(T::Enumerable[T::Class[T.anything]]),
           )
         end
@@ -87,10 +83,10 @@ module Tapioca
         def all_modules
           @all_modules ||= T.let(
             if @@requested_constants.any?
-              @@requested_constants.select { |k| k.is_a?(Module) }
+              @@requested_constants.grep(Module)
             else
-              ObjectSpace.each_object(Module)
-            end,
+              ObjectSpace.each_object(Module).to_a
+            end.freeze,
             T.nilable(T::Enumerable[Module]),
           )
         end


warhammerkid:
I have signed the CLA!

lib/tapioca/dsl/compiler.rb
```diff
@@ -74,11 +74,10 @@ def descendants_of(klass)
         sig { returns(T::Enumerable[T::Class[T.anything]]) }
         def all_classes
           @all_classes ||= T.let(
-            if @@requested_constants.any?
-              @@requested_constants.grep(Class)
-            else
-              ObjectSpace.each_object(Class)
-            end,
+            T.cast(
+              all_modules.select { |k| k.is_a?(Class) },
```
amomchilov, line 78: I know it's kind of niche, but `.grep(Class)` is noticeably faster (last I benchmarked it), since it calls `Class#===` directly, without needing the overhead of a block call like `select` needs

```suggestion
              all_modules.grep(Class),
```

lib/tapioca/dsl/compiler.rb
```diff
@@ -89,7 +88,7 @@ def all_modules
             if @@requested_constants.any?
               @@requested_constants.select { |k| k.is_a?(Module) }
```
amomchilov, line 89: May as well change this too, to be consistent.

```suggestion
              @@requested_constants.grep(Module)
```

amomchilov (approved):
Thanks for your contribution! I have 2 small suggestions, but otherwise LGTM!

Once you ship this, I'll do a quick benchmark against our monolith. I'm expecting there should be pretty substantial gains!

lib/tapioca/dsl/compiler.rb
egiurleo: I actually didn't know this! Good to know.

amomchilov: 2 other nice benefits:

1. It removes the need for [this cast](https://github.com/Shopify/tapioca/compare/85751a539fc2df50baaa258fe32a289f7d5213f2..2dfbfa43f5c0b984cb7bc7fdf229608f78868310) because `Enumerable#grep` has [a special overload for metatypes](https://github.com/sorbet/sorbet/blob/c1678be7c00cd2beb40adc153ccce5c424fd57a2/rbi/core/enumerable.rbi#L751-L757).
2. The `k.is_a?(Class)` was calling `is_a?` on `k`, which could be arbitrary user code that breaks things. `Class === k` would always be on a receiver we control, `Class`.

egiurleo:
Rebased to get CI fixes!

lib/tapioca/dsl/compiler.rb
paracycle: Yes, @amomchilov is spot on, especially around the use of `===`. In this project, we prefer to use `===` over `is_a?` whenever it is possible, since the receiver of `===` is usually under our control, so the result can be more dependable.

paracycle (approved):

PR #2074, opened 2024-11-14
Various CI fixes

KaanOzkan:
### Motivation

Fix CI

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -27,6 +27,11 @@ group :development, :test do
     rails_version = CURRENT_RAILS_VERSION if rails_version == "current"
     gem "rails", "~> #{rails_version}.0"
   end
+  if rails_version == "7.0"
+    gem "sqlite3", "~> 1.4"
+  else
+    gem "sqlite3"
+  end
 
   gem "mutex_m"
   gem "smart_properties"
@@ -35,7 +40,6 @@ group :development, :test do
   gem "sprockets"
   gem "state_machines"
   gem "activerecord-typedstore"
-  gem "sqlite3", "~>1.4"
   gem "identity_cache"
   gem "cityhash",
     git: "https://github.com/csfrancis/cityhash.git",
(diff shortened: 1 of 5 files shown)


Gemfile
```diff
@@ -35,7 +35,7 @@ group :development, :test do
   gem "sprockets"
   gem "state_machines"
   gem "activerecord-typedstore"
-  gem "sqlite3", "~>1.4"
+  gem "sqlite3"
```
KaanOzkan, line 38: I'm hoping the issue in https://github.com/Shopify/tapioca/pull/1875/ (which I don't know) is resolved by now

spec/tapioca/cli/gem_spec.rb
```diff
@@ -1287,6 +1287,7 @@ module Foo; end
             class Foo::Engine < ::Rails::Engine
               class << self
                 def __callbacks; end
+                def __callbacks=(new_value); end
```
KaanOzkan, line 1290: I couldn't find where this is defined in Rails but I think it's a safe change

Gemfile
```diff
@@ -27,6 +27,11 @@ group :development, :test do
     rails_version = CURRENT_RAILS_VERSION if rails_version == "current"
     gem "rails", "~> #{rails_version}.0"
   end
+  if rails_version == "7.0"
+    gem "sqlite3", "~> 1.4"
+  else
+    gem "sqlite3"
+  end
```
KaanOzkan, line 34: This is needed because sqlite3 version needs to be consistent with the Rails version https://github.com/rails/rails/blob/dd8f7185faeca6ee968a6e9367f6d8601a83b8db/activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb#L14.

Gemfile
andyw8: Yes, this has been resolved in Rails.

PR #2079, opened 2024-11-20
Git hooks exploration

andyw8:
(using `lefthook` just for prototyping, this would be in `dev`)


lefthook.yml
```diff
@@ -0,0 +1,7 @@
+pre-commit:
+  commands:
+    tapioca-gem:
+      glob: "Gemfile.lock"
+      run: |
+        echo "Changes were detected in Gemfile.lock. Running `tapioca gem`"
+        bin/tapioca gem --verify && bin/tapioca gem
```
andyw8, line 7: (this doesn't actually work on `tapioca` itself since it needs to be `bundle exec tapioca`).

PR #2081, opened 2024-11-20
WIP: [Tapioca Addon] Support gem RBI generation on dirty lockfile

alexcrocha:
# WIP

This Draft PR is still a work in progress and an alternative implementation to #2063 

### Motivation

To support gem RBI generation, we needed a way to detect changes in `Gemfile.lock`. Currently, changes to this file cause the Ruby LSP to restart, resulting in loss of access to any previous state information. This limitation prevents monitoring and processing of gem changes.

### Implementation

To address this, we check git status to determine if `Gemfile.lock` has uncommitted changes. When changes are detected, we:

1. Compare the current `Gemfile.lock` with a snapshot stored in `tmp/tapioca/.gemfile_lock_snapshot`
2. Use `Bundler::LockfileParser` to parse both files and detect:
   - Removed gems (present in snapshot but not in current)
   - Changed gems (version differences between snapshot and current)
3. Process the changes in the server addon:
   - Remove RBI files for deleted gems
   - Generate new RBI files for changed gems using Tapioca
4. Update the snapshot with the current lockfile content

This approach specifically targets manual bundle updates while avoiding unnecessary RBI regeneration during normal git operations like branch switches.


lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -106,6 +110,42 @@ def workspace_did_change_watched_files(changes)
...
+          File.write(GEMFILE_LOCK_SNAPSHOT, current_lockfile)
+          return
+        end
+
+        return if current_lockfile == snapshot_lockfile
+
+        T.must(@rails_runner_client).delegate_notification(
+          server_addon_name: "Tapioca",
+          request_name: "gem",
+          snapshot_lockfile: snapshot_lockfile,
```
KaanOzkan, line 143: Will sending large lockfiles be an issue? Maybe getting timings on core is enough to answer this.

lib/ruby_lsp/tapioca/server_addon.rb
```diff
@@ -23,6 +25,34 @@ def dsl(params)
...
+        if changed_gems.any?
+          $stdout.puts("Generating RBIs for changed gems: #{changed_gems.join(", ")}")
+
+          load("tapioca/cli.rb") # Reload the CLI to reset thor defaults between requests
+          ::Tapioca::Cli.start(["gem"] + changed_gems)
+        end
+      end
+
+      def parse_lockfile(content)
+        return {} if content.to_s.empty?
```
KaanOzkan, line 52: I'm surprised by this code, isn't `content` string already? We are the ones sending the initial arguments to rails add-on so maybe we can just do `return unless content` here?

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -44,6 +46,8 @@ def activate(global_state, outgoing_queue)
           @rails_runner_client = addon.rails_runner_client
           outgoing_queue << Notification.window_log_message("Activating Tapioca add-on v#{version}")
           @rails_runner_client.register_server_addon(File.expand_path("server_addon.rb", __dir__))
+
+          handle_gemfile_changes
```
KaanOzkan, line 50: I think something more descriptive is better here. Example:
```suggestion
        generate_gem_rbis if gemfile_changed
```

In this example `gemfile_changed` will contain `handle_gemfile_changes and some of `process_gemfile_changes`. I think something like this reads better because if I'm skimming through I know its intent.

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -106,6 +110,42 @@ def workspace_did_change_watched_files(changes)
...
+
+        process_gemfile_changes
+      end
+
+      sig { returns(T.nilable(Integer)) }
+      def process_gemfile_changes
+        current_lockfile = File.read("Gemfile.lock")
+        snapshot_lockfile = File.read(GEMFILE_LOCK_SNAPSHOT) if File.exist?(GEMFILE_LOCK_SNAPSHOT)
+
+        unless snapshot_lockfile
```
KaanOzkan, line 131: Have we compared the snapshot approach to `git diff`? I can think of one scenario where you update the lockfile before snapshot is created so upon the consequent restart there are changes in `git status` however, snapshot doesn't exist yet so we don't generate. It's not a common scenario so not important. But feels like running `git diff` instead is a viable solution. Curious about the reasoning here. Does it have downsides for changing branches?

lib/ruby_lsp/tapioca/addon.rb
KaanOzkan: On that note, do we have to send both lockfiles? What if we do the processing done in `server_addon.rb` here and send removed and changed gem names to the server instead? It's a possibility but I don't see a strong reason to do it unless there are memory/speed issues. Maybe only worth looking after we ship the feature.

lib/ruby_lsp/tapioca/server_addon.rb
```diff
@@ -23,6 +25,34 @@ def dsl(params)
...
       end
+
+      def gem(params)
+        snapshot_specs = parse_lockfile(params[:snapshot_lockfile])
+        current_specs = parse_lockfile(params[:current_lockfile])
+
+        removed_gems = snapshot_specs.keys - current_specs.keys
+        changed_gems = current_specs.select { |name, version| snapshot_specs[name] != version }.keys
+
+        return $stdout.puts("No gem changes detected") if removed_gems.empty? && changed_gems.empty?
```
andyw8, line 36: Let's put the `return` on its own line, since returning the result of `puts` isn't meaningful.

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -106,6 +110,42 @@ def workspace_did_change_watched_files(changes)
...
       end
+
+      private
+
+      sig { void }
+      def handle_gemfile_changes
+        return unless File.exist?(".git")
+
+        gemfile_status = %x(git status --porcelain Gemfile.lock).strip
+        return if gemfile_status.empty?
```
andyw8, line 121: (🤔 should be continue only if the output begins with `M`, indicating a modification?)

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -106,6 +110,42 @@ def workspace_did_change_watched_files(changes)
...
         )
       end
+
+      private
+
+      sig { void }
+      def handle_gemfile_changes
+        return unless File.exist?(".git")
+
+        gemfile_status = %x(git status --porcelain Gemfile.lock).strip
```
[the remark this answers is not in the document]
alexcrocha: This is now an older implementation, but I do agree some timings would be nice to have. I'll take a look on Monday.

lib/ruby_lsp/tapioca/lockfile_diff_parser.rb
```diff
@@ -0,0 +1,44 @@
...
+    class LockfileDiffParser
+      GEM_PATTERN = /[+-](.*[\w\-]+)\s*\(/
+      ADDED_LINE_PATTERN = /^\+.*[\w\-]+ \(.*\)/
+      REMOVED_LINE_PATTERN = /^-.*[\w\-]+ \(.*\)/
+
+      attr_reader :added_or_modified_gems
+      attr_reader :removed_gems
+
+      def initialize(diff_content)
+        @diff_content = diff_content
```
KaanOzkan, line 15:
```suggestion
        @diff_content = diff_content.lines
```

If the only thing we do is call `.lines` on it we could do it here and do it only once instead of 2 calls below.

lib/ruby_lsp/tapioca/server_addon.rb
```diff
@@ -23,6 +26,25 @@ def dsl(params)
...
+
+      def gem(params)
+        gem_changes = LockfileDiffParser.new(params[:diff])
+
+        removed_gems = gem_changes.removed_gems
+        added_or_modified_gems = gem_changes.added_or_modified_gems
+
+        if removed_gems.any?
+          $stdout.puts("Removing RBIs for deleted gems: #{removed_gems.join(", ")}")
+          FileUtils.rm_f(Dir.glob("sorbet/rbi/gems/{#{removed_gems.join(",")}}@*.rbi"))
```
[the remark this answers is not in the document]
alexcrocha: I went on a little adventure to investigate what `Dir.glob` is doing. It took me to C land, which made it trickier for me to read, but I was able to find [this code](https://github.com/ruby/ruby/blob/master/dir.c#L3097) that handles our pattern matching (and I learned it's called brace expansion).

This makes me confident that the string we're passing to `Dir.glob` is being handled by Ruby itself and not passed to a shell 👍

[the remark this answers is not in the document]
andyw8: Ah, I didn't notice the `glob` call previously.

lib/ruby_lsp/tapioca/lockfile_diff_parser.rb
```diff
@@ -0,0 +1,43 @@
+# typed: true
```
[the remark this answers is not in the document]
KaanOzkan: You don't think the current `RubyLsp::Tapioca` is enough? What did you have in mind?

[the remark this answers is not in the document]
andyw8: Ah, nevermind, I was looking at a partial diff which didn't show the top-level.

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -45,6 +46,10 @@ def activate(global_state, outgoing_queue)
           @rails_runner_client = addon.rails_runner_client
           @outgoing_queue << Notification.window_log_message("Activating Tapioca add-on v#{version}")
           @rails_runner_client.register_server_addon(File.expand_path("server_addon.rb", __dir__))
+
+          if git_repo?
+            lockfile_changed? ? generate_gem_rbis : cleanup_orphaned_rbis
```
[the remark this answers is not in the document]
andyw8: Let's say you add a gem to try out. The LSP automatically generates its RBI. You don't commit anything. Then you decide you don't want the gem, so you reset the changes in `Gemfile.lock`. The LSP then restart, but `git diff` is blank, so nothing gets cleaned up.

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -127,6 +135,82 @@ def file_updated?(change, path)
 
         false
       end
+
+      sig { returns(T.nilable(T::Boolean)) }
```
andyw8, line 139: Can this be non-nilable?

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -127,6 +135,82 @@ def file_updated?(change, path)
...
+      def git_repo?
+        _, status = Open3.capture2e("git rev-parse --is-inside-work-tree")
+
+        status.success?
+      end
+
+      sig { returns(T::Boolean) }
+      def lockfile_changed?
+        !fetch_lockfile_diff.empty?
```
andyw8, line 148:
```suggestion
        fetch_lockfile_diff
        !@lockfile_diff.empty?
```
(to avoid violating https://en.wikipedia.org/wiki/Command%E2%80%93query_separation )

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -127,6 +135,82 @@ def file_updated?(change, path)
...
+      end
+
+      sig { returns(String) }
+      def fetch_lockfile_diff
+        @lockfile_diff = %x(git diff HEAD Gemfile.lock).strip
+      end
+
+      sig { void }
+      def generate_gem_rbis
+        gem_changes = LockfileDiffParser.new(@lockfile_diff)
```
andyw8, line 158:
```suggestion
        parser = LockfileDiffParser.new(@lockfile_diff)
```

andyw8 (changes requested):
Looks good but I think we need a few more tests for the main flows in `addon.rb` .

andyw8 (approved):
@KaanOzkan I think it's in a pretty good shape and I've marked as approved for if you want to release. It _does_ need some more tests, but it would valuable to ship it so that we can start getting feedback and iterating. We can expand on the tests while that is happening.

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -13,6 +13,9 @@
 end
 
 require "zlib"
+require "open3"
+
+require_relative "lockfile_diff_parser"
```
andyw8, line 18: This _should_ be the same as:
```suggestion
require "tapioca/lockfile_diff_parser"
```
(for consistency with how we use `require` elsewhere in the codebase)

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -13,6 +13,9 @@
 end
 
 require "zlib"
+require "open3"
```
andyw8, line 16: (nit) we could move this to just before the first use of `Open3` to reduce boot time by a tiny bit.

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -13,6 +13,7 @@
 end
```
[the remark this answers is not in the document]
KaanOzkan: That sounds good to me. I mentioned this in the issue as a future improvement.

spec/tapioca/ruby_lsp/tapioca/lockfile_diff_parser_spec.rb
```diff
@@ -0,0 +1,62 @@
+# typed: strict
+# frozen_string_literal: true
+
+require "spec_helper"
+require "ruby_lsp/tapioca/lockfile_diff_parser"
+
+module RubyLsp
+  module Tapioca
+    class LockFileDiffParserSpec < Minitest::Spec
```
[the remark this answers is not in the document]
alexcrocha: This case was indeed not being covered. I pushed an update to prevent this scenario from occurring. Thanks for catching that.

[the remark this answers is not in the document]
vinistock: Will the solution also work if the direct dependency doesn't appear in the diff?

We may need a more robust check, like using Bundler's lockfile parser to determine what dependencies exist.

```ruby
parser = Bundler::LockfileParser.new(Bundler.default_lockfile.read)
puts parser.dependencies
```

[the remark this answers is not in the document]
alexcrocha: You're right, the solution didn't cover that scenario.

Pushed an new update using your suggestion. Hopefully the third time is the charm 🤞

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -50,6 +52,15 @@ def activate(global_state, outgoing_queue)
...
           )
+
+          gem_rbi_check_result = RunGemRbiCheck.new.run
+          @outgoing_queue << Notification.window_log_message(
+            gem_rbi_check_result.stdout,
+          ) unless gem_rbi_check_result.stdout.empty?
+          @outgoing_queue << Notification.window_log_message(
+            gem_rbi_check_result.stderr,
+            type: Constant::MessageType::WARNING,
+          ) unless gem_rbi_check_result.stderr.empty?
```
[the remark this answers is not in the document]
alexcrocha: I like that idea

PR #2083, opened 2024-11-21
[Tapioca Add-on] Use gradual rollout API

KaanOzkan:


lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -32,8 +32,7 @@ def initialize
       sig { override.params(global_state: RubyLsp::GlobalState, outgoing_queue: Thread::Queue).void }
       def activate(global_state, outgoing_queue)
         @global_state = global_state
-        # TODO: Uncomment
-        # return unless @global_state.experimental_features
+        return unless @global_state.enabled_feature?(:tapioca_addon)
```
vinistock, line 35: This is camel case because it comes from the JSON settings and we don't convert it to snake case.

```suggestion
        return unless @global_state.enabled_feature?(:tapiocaAddon)
```

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -67,7 +66,7 @@ def version
       sig { params(changes: T::Array[{ uri: String, type: Integer }]).void }
       def workspace_did_change_watched_files(changes)
         # TODO: Uncomment
-        # return unless T.must(@global_state).experimental_features
+        return unless T.must(@global_state).enabled_feature?(:tapioca_addon)
```
vinistock, line 69: Need to also remove the `TODO` comment above.
```suggestion
        return unless T.must(@global_state).enabled_feature?(:tapiocaAddon)
```

PR #2085, opened 2024-11-22
[Tapioca Add-on] Limit what we require from the Rails add-on

andyw8:
The Tapioca add-on was requiring the entire `ruby-lsp-rails` gem, which in turns requires `rails`. As @st0012 identified, this caused a change to the behaviour of `.to_json` which resulted in the Dependencies view not being displayed correctly.

Closes https://github.com/Shopify/team-ruby-dx/issues/1317

cc @KaanOzkan


andyw8:
I've pushed an update, and have verified gem DSL generation on code-db.

st0012 (approved):

andyw8:
> We should remove https://github.com/Shopify/ruby-lsp-rails/blob/main/lib/ruby_lsp_rails/railtie.rb. We've had the deprecation warning for long enough

https://github.com/Shopify/ruby-lsp-rails/pull/535

PR #2089, opened 2024-11-25
Infer the worker count for `tapioca dsl`

rzane:
### Motivation

I want `tapioca dsl` to infer the appropriate number of workers automatically.

### Implementation

I noticed the CLI is setting `default: 2` for the DSL command, but the other commands don't provide a default. The inference defined [here](https://github.com/Shopify/tapioca/blob/5434fa6a65d5111b2a7e0adf581a9a43ab6e1da9/lib/tapioca/executor.rb#L18) never runs.

I'm assuming this is a mistake, but there could be a good reason for this hardcoded default that I'm not aware of.

### Tests

It would be difficult to write a test for this functionality because we'd need to assert on the assumed number of workers in a test for the CLI. The worker count determination happens deep in the stack.

### DIFF
--- a/README.md
+++ b/README.md
@@ -489,8 +489,7 @@ Options:
                                                                                                    # Default: false
   -q,        [--quiet], [--no-quiet], [--skip-quiet]                                               # Suppresses file creation output
                                                                                                    # Default: false
-  -w,        [--workers=N]                                                                         # Number of parallel workers to use when generating RBIs (default: 2)
-                                                                                                   # Default: 2
+  -w,        [--workers=N]                                                                         # Number of parallel workers to use when generating RBIs (default: auto)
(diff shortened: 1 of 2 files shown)


KaanOzkan:
This was manually set to 2 after being automatic in https://github.com/Shopify/tapioca/commit/2a5db224906007778f036080124497351bbdeb5e. So we'll have to figure out why.

In the meantime it'd be good to have a way to set `auto` in the `tapioca/config.yml` so you can still achieve what you want but specific to your project. Feel free to open a PR that does that. Although right now I'm not sure how it'll work because Thor requires a specific type for the option: https://github.com/rails/thor/wiki/method-options. Maybe we could use a value of `-1` to indicate auto 😬

KaanOzkan (approved):

PR #2099, opened 2024-11-27
Add dummy Rails app for testing Tapioca LSP

andyw8:
Part of https://github.com/Shopify/team-ruby-dx/issues/1327

The command I used. I inclued `--api` to skip the views, layouts, JS, etc.

```
rails new dummy --minimal --skip-ci --skip-rubocop --skip-git --skip-bundle --api --skip-tests --skip-keeps --no-solid --skip-kamal --skip-thruster
```

Then I removed some files that aren't relevant for our needs (see seperate commit).

We may need to refine it further, but let's gets this initial version merged first, then future chnages will be easier to review.


spec/dummy/Gemfile
```diff
@@ -0,0 +1,24 @@
...
+gem "kamal", require: false
+
+# Use Rack CORS for handling Cross-Origin Resource Sharing (CORS), making cross-origin Ajax possible
+# gem "rack-cors"
+
+group :development, :test do
+  # See https://guides.rubyonrails.org/debugging_rails_applications.html#debugging-with-the-debug-gem
+  gem "debug", platforms: %i[ mri windows ], require: "debug/prelude"
+end
+
```
andyw8, line 23: This whitespace is coming from the generator.

PR #2101, opened 2024-11-28
Add test for `lsp_add_on` flag in CLI

andyw8:
### Motivation

Part of https://github.com/Shopify/team-ruby-dx/issues/1327

### Implementation

Until now, most of Tapioca tests have been integration tests. Trying to verify the new LSP behaviour using this approach will be difficult, so I suggest we should consider a more mockist-style approach.

I'm choosing to use Mocha as it's the standard for Shopify, and it allows a lot flexibility.

I intend to add some integration tests later to verify the LSP behaviour as a whole

TODO:
- [x] I need to fix the typecheck, but wanted to open this up early for discussion.

### Tests

Indeed.


spec/tapioca/cli_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      before do
+        @command_stub = mock(run: nil)
+      end
+
+      it "passes through the `lsp_addon` flag to the DslGenerate command" do
+        Commands::DslGenerate.expects(:new).with do |options|
+          options[:lsp_addon] == true
+        end.returns(@command_stub)
+
+        capture_io { Cli.start(["dsl", "--lsp_addon"]) }
```
vinistock, line 18: Do we still print things even with the stub? I would've expected nothing to be printed given that we're stubbing `run` to do nothing.

spec/tapioca/cli_spec.rb
```diff
@@ -0,0 +1,30 @@
+# typed: true
+# frozen_string_literal: true
+
+require "spec_helper"
+
+module Tapioca
+  class CliSpec < SpecWithProject
+    describe "Tapioca::Cli" do
+      before do
+        @command_stub = mock(run: nil)
```
vinistock, line 10: This is just my preference, but I would avoid the extra dependency and just use `Minitest::Mock` here. I don't know what's the opinion for the rest of the team

spec/tapioca/cli_spec.rb
andyw8: In the second test is suppresses `Unknown switches "--another-flag"`. I can remove it from the first.

PR #2108, opened 2024-12-03
Add integration tests for add-on

andyw8:
Closes https://github.com/Shopify/team-ruby-dx/issues/1327

This PR adds an integration test for the flow of generating DSL RBIs for a particular compiler.

I chose to use the ActiveJob compiler to avoid any database complexity.

I've tried to reduce the Rails files to the minimum possible.

### DIFF
--- a/spec/tapioca/addon_spec.rb
+++ b/spec/tapioca/addon_spec.rb
@@ -1,0 +1,53 @@
+# typed: true
+# frozen_string_literal: true
+
+require "spec_helper"
+require "language_server-protocol"
+require "ruby_lsp/utils"
+require "ruby_lsp/ruby_lsp_rails/runner_client"
+require "minitest/hooks"
+
+module RubyLsp
+  module Tapioca
+    class AddonSpec < Minitest::HooksSpec
+      # The approach here is based on tests within the Ruby LSP Rails gem
+
+      # TODO: Replace with `before(:all)` once Sorbet understands it
+      def initialize(*args)
+        super(*T.unsafe(args))
+        @outgoing_queue = Thread::Queue.new
+        @client = T.let(nil, T.nilable(RubyLsp::Rails::RunnerClient))
+        FileUtils.chdir("spec/dummy") do
+          @client = RubyLsp::Rails::RunnerClient.new(@outgoing_queue)
+        end
+      end
+
+      after(:all) do
+        T.must(@client).shutdown
+
+        assert_predicate(@client, :stopped?)
+        @outgoing_queue.close
+      end
+
+      it "generates DSL RBIs for a given constant" do
+        addon_path = File.expand_path("lib/ruby_lsp/tapioca/server_addon.rb")
+        T.must(@client).register_server_addon(File.expand_path(addon_path))
+        T.must(@client).delegate_notification(
+          server_addon_name: "Tapioca",
+          request_name: "dsl",
+          constants: ["NotifyUserJob"],
+        )
+
+        begin
+          Timeout.timeout(10) do
+            sleep(1) until File.exist?("spec/dummy/sorbet/rbi/dsl/notify_user_job.rbi")
+          end
+        rescue Timeout::Error
+          flunk("RBI file was not generated")
+        end
+      ensure
+        FileUtils.rm_rf("spec/dummy/sorbet/rbi")
+      end
+    end
+  end
+end
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@
 /coverage/
 /doc/
 /pkg/
+/spec/dummy/log/
+/spec/dummy/tmp/
 /spec/examples.txt
 /spec/reports/
 /tmp/
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -47,6 +47,7 @@ Sorbet/TrueSigil:
     - "**/*.rake"
   Exclude:
     - "lib/ruby_lsp/tapioca/server_addon.rb"
+    - "spec/dummy/**/*.rb"
 
 Style/CaseEquality:
   Enabled: false
--- a/sorbet/rbi/shims/minitest.rbi
+++ b/sorbet/rbi/shims/minitest.rbi
@@ -1,0 +1,8 @@
+# typed: true
+
+# Can be removed once https://github.com/Shopify/rbi-central/pull/300 ships
+
+class Minitest::HooksSpec
+  sig { params(arg: Symbol, block: T.proc.bind(T.attached_class).void).void }
+  def self.after(arg, &block); end
+end
--- a/spec/dummy/app/jobs/notify_user_job.rb
+++ b/spec/dummy/app/jobs/notify_user_job.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+
+class NotifyUserJob < ActiveJob::Base
+  def perform(user)
+  end
+end
--- a/spec/dummy/bin/rails
+++ b/spec/dummy/bin/rails
@@ -1,0 +1,14 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+# This command will automatically be run when you run "rails" with Rails gems
+# installed from the root of your application.
+
+ENGINE_ROOT = File.expand_path("..", __dir__)
+APP_PATH = File.expand_path("../config/application", __dir__)
+
+# Set up gems listed in the Gemfile.
+ENV["BUNDLE_GEMFILE"] ||= File.expand_path("../Gemfile", __dir__)
+require "bundler/setup" if File.exist?(ENV["BUNDLE_GEMFILE"])
+
+require "rails/engine/commands"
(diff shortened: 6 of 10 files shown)


spec/tapioca/addon_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+        @client.shutdown
+
+        # On Windows, the server process sometimes takes a lot longer to shutdown and may end up getting force killed,
+        # which makes this assertion flaky
+        assert_predicate(@client, :stopped?) unless Gem.win_platform?
+        @outgoing_queue.close
+      end
+
+      it "generates DSL RBIs for a gem" do
+        raise "RBI already exists" if File.exist?("sorbet/rbi/dsl/create_comment_input.rbi")
```
andyw8, line 37: (stan suggests switch something built-in such as ActiveJob)

andyw8:
Looking into the test failures, seems unrelated to this PR though:

```
    it compile RBIs with location from gem source                   FAIL (1.20s)
        --- expected
        +++ actual
        @@ -1,66 +1,66 @@
        -"# source://the-default-gem//lib/bar.rb#1
        +"# source://the-default-gem/lib/bar.rb#1
```

spec/tapioca/addon_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+        @outgoing_queue = Thread::Queue.new
+        @client = T.let(RubyLsp::Rails::RunnerClient.new(@outgoing_queue), RubyLsp::Rails::RunnerClient)
+      end
+
+      after(:all) do
+        @client.shutdown
+
+        # On Windows, the server process sometimes takes a lot longer to shutdown and may end up getting force killed,
+        # which makes this assertion flaky
+        assert_predicate(@client, :stopped?) unless Gem.win_platform?
```
KaanOzkan, line 24: We don't develop or run CI on windows so I think we can remove this. Unless you think it's worth it for external contributors.

spec/tapioca/addon_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+        @client.register_server_addon(File.expand_path(addon_path))
+        @client.delegate_notification(
+          server_addon_name: "Tapioca",
+          request_name: "dsl",
+          constants: ["NotifyUserJob"],
+        )
+
+        begin
+          Timeout.timeout(10) do
+            until File.exist?(EXPECTED_RBI_PATH)
```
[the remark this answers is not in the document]
KaanOzkan: I was curious how to do this in Ruby with as little work as possible and ran into these gems that allow you to ask the OS to notify upon file changes. Pretty cool but don't think we need it.

There's [rb-inotify](https://github.com/guard/rb-inotify/tree/main) and [rb-fsevent](https://github.com/guard/rb-fsevent) which work for linux & macos respectively. There's also [this gem](https://github.com/ttilley/fssm/tree/master) that tries to install the relevant one depending on platform

spec/tapioca/addon_spec.rb
andyw8: I'll remove. I think it's unlikely that someone using Windows would want to contribute to developer tooling that doesn't run on Windows.

spec/tapioca/addon_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+        @client.shutdown
+
+        assert_predicate(@client, :stopped?)
+        @outgoing_queue.close
+        FileUtils.rm("bin/rails")
+      end
+
+      EXPECTED_RBI_PATH = "spec/dummy/sorbet/rbi/dsl/notify_user_job.rbi"
+      it "generates DSL RBIs for a gem" do
```
KaanOzkan, line 35:
```suggestion
      it "generates DSL RBIs for a given constant" do
```

andyw8:
(I will squash before merging).

spec/addon_spec_helper.rb
```diff
@@ -0,0 +1,43 @@
...
+
+require "minitest/spec"
+require "minitest/autorun"
+require "sorbet-runtime"
+require "language_server-protocol"
+require "ruby_lsp/utils"
+
+module ActiveSupport
+  class TestCase
+    def pop_result(server)
```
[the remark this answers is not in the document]
KaanOzkan: I think we can to this now?

spec/tapioca/addon_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+require "addon_spec_helper"
+require "ruby_lsp/ruby_lsp_rails/runner_client"
+require "minitest/hooks"
+
+module RubyLsp
+  module Tapioca
+    class AddonSpec < Minitest::HooksSpec
+      # The approach here is based on tests within the Ruby LSP Rails gem
+
+      # TODO: Replace by `before(:all)` once Sorbet understands it
```
KaanOzkan, line 13: I don't think this TODO is very helpful but

```suggestion
      # TODO: Replace with `before(:all)` once Sorbet understands it
```

spec/dummy/app/models/notify_user_job.rb
vinistock, line 1: This is a job, but it's under `app/models`. Is that intentional?

spec/tapioca/addon_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      def initialize(*args)
+        super(*T.unsafe(args))
+        FileUtils.cp("spec/dummy/bin/rails", "bin/rails")
+        @outgoing_queue = Thread::Queue.new
+        @client = T.let(nil, T.nilable(RubyLsp::Rails::RunnerClient))
+        FileUtils.chdir("spec/dummy") do
```
vinistock, line 19: If we change the directory to boot the client here, why do we need to copy the `bin/rails` file above? Wouldn't the chdir already deal with that?

spec/tapioca/addon_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+require "ruby_lsp/ruby_lsp_rails/runner_client"
+require "minitest/hooks"
+
+module RubyLsp
+  module Tapioca
+    class AddonSpec < Minitest::HooksSpec
+      # The approach here is based on tests within the Ruby LSP Rails gem
+
+      # TODO: Replace by `before(:all)` once Sorbet understands it
+      def initialize(*args)
```
vinistock, line 14: Could we use class instance variables for the client to avoid overriding `initialize`?

spec/tapioca/addon_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+        FileUtils.cp("spec/dummy/bin/rails", "bin/rails")
+        @outgoing_queue = Thread::Queue.new
+        @client = T.let(nil, T.nilable(RubyLsp::Rails::RunnerClient))
+        FileUtils.chdir("spec/dummy") do
+          @client = RubyLsp::Rails::RunnerClient.new(@outgoing_queue)
+        end
+      end
+
+      after(:all) do
+        # TODO: Remove `bind` once Sorbet understands `after(:all)`
```
vinistock, line 25: Sorbet can already understand it. We just need a signature for it (which is a nice contribution to rbi-central).

```ruby
class Minitest::HooksSpec
  sig { params(arg: Symbol, block: T.proc.bind(T.attached_class).void).void }
  def self.after(arg, &block); end
end
```

spec/tapioca/addon_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+        T.bind(self, AddonSpec)
+
+        T.must(@client).shutdown
+
+        assert_predicate(@client, :stopped?)
+        @outgoing_queue.close
+        FileUtils.rm("bin/rails")
+      end
+
+      EXPECTED_RBI_PATH = "spec/dummy/sorbet/rbi/dsl/notify_user_job.rbi"
```
vinistock, line 35: Does this need to be a constant?

spec/tapioca/addon_spec.rb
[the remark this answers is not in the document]
vinistock: Unless I'm misunderstanding what Kaan said, I don't think the code matches the suggestion.

```ruby
# The code is doing this, which sleeps and then enters a loop that will max
# CPU to 100%

sleep 1
until File.exist?(EXPECTED_RBI_PATH)
end

# Kaan's suggestion was to sleep on each iteration to avoid maximizing CPU usage
sleep 1 until File.exist?(EXPECTED_RBI_PATH)
```

spec/tapioca/addon_spec.rb
andyw8: https://github.com/Shopify/rbi-central/pull/300

spec/tapioca/addon_spec.rb
andyw8: Leftover from a previous approach, now removed.

spec/tapioca/addon_spec.rb
andyw8: Possibly, but I'm using the same approach as the exising `spec_with_project`.

andyw8:
@vinistock everything should have been addressed now.

andyw8:
(squashed)

Issue #1965, opened 2024-07-20
Add explicit ability to choose order of Tapioca compiler execution


bitwise-aiden:
### Description 

At present when compilers for DSL generation are loaded they are alphabetized and run in that order:

https://github.com/Shopify/tapioca/blob/8678e2ab7b853ccbf3f092ccf553e58773eaa38b/lib/tapioca/dsl/pipeline.rb#L115-L123

What this effectively means is that if two compilers create the same method with different signatures, whichever compiler was run last will be the one to define the signature used. If someone were wanting to write a different signature than a Tapioca defined one, their only recourse would be to name their compiler something that would come after `Tapioca::Dsl::Compilers::*` alphabetically.

Having the ability to push the compiler forward or back in the run order would make it possible to have the explicit control over the order rather than relying on discovering the above code to then know to name it something that puts it in the right place. Naming is already hard enough, but name golf to get your compiler to run in order although fun seems a little counter intuitive.

#### Order value

One solution would be to have a `order` value which is an integer representing where it is expected to run. This could then update the code above to order based on this value first, then alphabetically next to ensure a stable order run. The order value could be an integer between 0-100 and by default if not set, the could be 50 to allow for room either side. 

This would then allow projects to set the value of their compilers based on what they are needed and even override ones from their gems/Tapioca to ensure the order needed for them.

#### Method override flags

Given that the heart of the problem is that a method needs to be defined after an existing to override it's signature, maybe when generating it, optional flags can be passed to determine how it interacts with existing / future methods of the same name. It would have to be able to handle future cases too because of the aforementioned ordering alphabetically.

bitwise-aiden:
> For the different signature use case is it enough to run custom compilers after default ones?

Yes, I think providing the way to run them after would likely hit 99+% of use cases. Separating them out would hopefully fix that.

paracycle:
I'd like to question the premise of the issue, though. What is a real life example where 2 compilers want to type the same method in different ways?

I'd be wary of solutionizing before we understand if the problem is really a problem.

paracycle:
This should allow you to solve that: https://github.com/Shopify/tapioca/pull/1967

paracycle:
I also added a DSL CLI test that shows how to use the extensibility point inside a custom "compiler".

bitwise-aiden:
[...]

PR #2112, opened 2024-12-05
Provide a hook for custom `ActiveModel` types

rzane:
### Motivation

We use `ActiveModel::Attributes` extensively. In many cases, we're passing complex values to these models, such as a `User`. It wouldn't be practical in our situation to define an `ActiveModel::Type` class for each and every possible value that gets passed into a model.

Any of the following solutions would solve my issue:
1. Explicitly tell Tapioca the type of a given attribute.
2. Explicitly tell Tapioca to skip type generation for a given attribute, which would allow me to write my own compiler for these specific attributes.
3. Allow compilers to override the types defined by other compilers (similar to #1965).

### Implementation

This PR allows instances of `ActiveModel::Type::Value` to define an optional method `#__tapioca_type`. It returns a type.

For example, you could use this hook to implement an interface like this:

```ruby
class GenericType < ActiveModel::Type::Value
  def initialize(type)
    @type = type
  end

  def __tapioca_type = @type
end

class User
end

class Foo
  include ActiveModel::Attributes

  attribute :user, GenericType.new(User)
end
```

### Tests

Tests are included.

### DIFF
--- a/lib/tapioca/dsl/helpers/active_model_type_helper.rb
+++ b/lib/tapioca/dsl/helpers/active_model_type_helper.rb
@@ -14,7 +14,8 @@ module Tapioca
           def type_for(type_value)
             return "T.untyped" if Runtime::GenericTypeRegistry.generic_type_instance?(type_value)
 
-            type = lookup_return_type_of_method(type_value, :deserialize) ||
+            type = lookup_sorbet_type(type_value) ||
+              lookup_return_type_of_method(type_value, :deserialize) ||
               lookup_return_type_of_method(type_value, :cast) ||
               lookup_return_type_of_method(type_value, :cast_value) ||
               lookup_arg_type_of_method(type_value, :serialize) ||
@@ -37,6 +38,11 @@ module Tapioca
           sig { params(type: T.untyped).returns(T::Boolean) }
           def meaningful_type?(type)
             !MEANINGLESS_TYPES.include?(type)
+          end
+
+          sig { params(obj: T.untyped).returns(T.nilable(T::Types::Base)) }
+          def lookup_sorbet_type(obj)
+            T::Utils.coerce(obj.__sorbet_type) if obj.respond_to?(:__sorbet_type)
           end
 
           sig { params(obj: T.untyped, method: Symbol).returns(T.nilable(T::Types::Base)) }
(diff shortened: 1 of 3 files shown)


lib/tapioca/dsl/compilers/active_model_attributes.rb
```diff
@@ -99,8 +99,17 @@ def handle_method_pattern?(pattern)
           HANDLED_METHOD_TARGETS.include?(target.to_s)
         end
 
+        sig { params(attribute_name: ::String, attribute_type_value: T.untyped).returns(::String) }
+        def type_for(attribute_name, attribute_type_value)
+          type = if constant.respond_to?(:__tapioca_attribute_types)
```
KaanOzkan, line 104: This change is first of its kind so we'd like to exhaust other options first.

rzane: Totally understandable. Which other options are worth considering?

rzane: I think eventually, no matter what, we'll need to be able to override the types that Tapioca is generating via compilers.

I'm dealing with another situation right now where the column type is a string, but a gem overrides the reader to return an object.

rzane: Like I said in the description, this is meant to serve as an integration point for libraries. I don't think we'll have much luck guessing the type.

So for example, this interface would allow me to implement something like this in user land:
```ruby
attribute :user, sig: -> { User }
```

I don't think Tapioca should get into the runtime business.

paracycle:
> It wouldn't be practical in our situation to define an `ActiveModel::Type` class for each and every possible value that gets passed into a model.

You can define a "generic" `ActiveModel::Type` subclass and use it to supply such a hook, provided we add support for it in Tapioca.

For example:
```ruby
class GenericType < ActiveModel::Type::Value
  def initialize(type)
    @type = type
  end

  def __sorbet_type = @type
end

class User
end

class Foo
  include ActiveModel::Attributes

  attribute :foo, GenericType.new(User)
end

p Foo.attribute_types.first.last.__sorbet_type # => User
```

rzane:
Thanks @amomchilov, updated 👍

rzane:
@amomchilov are there any remaining blockers here?

amomchilov:
Nope, looks good! Thanks for the changes.

I'll deprecate `#sorbet_type` and rename it to match

PR #2114, opened 2024-12-06
Compilers can define their execution order

rzane:
### Motivation

Tapioca generates types for ActiveModel and ActiveRecord attributes, and usually does the right thing. However, there are some cases where more flexibility is necessary. It would be useful to be able to override the definitions provided by Tapioca's default compilers in very special cases.

#### Example 1: ActiveModel attributes without a type

Not all ActiveModel attributes can be represented with ActiveModel's built-in types. It also isn't practical to define an `ActiveModel::Type` for every type of value that might be passed to an ActiveModel.

For example, many ActiveModel instances take a `User` as a parameter (e.g. `attribute :user`). I'd like to be able to tell Sorbet that this is a `User`. Unfortunately, the Tapioca compiler defines this field as `T.untyped`. Without forking the ActiveModel compiler, I have no way of overriding the type defined by Tapioca.

#### Example 2: Column types don't necessarily represent method return types

We use a gem called [`steady_state`](https://github.com/Betterment/steady_state). It defines a state machine and stores the state in a string column on the record.

This gem overrides the attribute reader to return an instance of [`SteadyState::Attribute::State`](https://github.com/Betterment/steady_state/blob/main/lib/steady_state/attribute/state.rb), which is basically an [`ActiveSupport::StringInquirer`](https://api.rubyonrails.org/classes/ActiveSupport/StringInquirer.html).

Without forking the ActiveRecordColumns compiler, I have no way of overriding the type defined by Tapioca.

#### Step 1: Allow DSL compilers to control execution order

This PR implements an optional compiler method `.run_after_compilers`. Compilers can tell Tapioca which other compilers they expect to precede them.

So, for example, my SteadyState compiler could say:

```ruby
class Tapioca::Dsl::Compilers::SteadyState < Tapoica::Dsl::Compiler
  def self.run_after_compilers
    [Tapioca::Dsl::Compilers::ActiveRecordColumns]
  end

  # ... etc ...
end
```

Tapioca can use [`TSort`](https://docs.ruby-lang.org/en/3.3/TSort.html) to ensure that the steady state compiler runs after `ActiveRecordColumns`.

#### Step 2: Allow DSL compilers to redefine methods

This PR implements `RBI::Tree#remove_method`, which removes a method from the tree by name.

### Tests

This PR includes tests.

### DIFF
--- a/lib/tapioca/dsl.rb
+++ b/lib/tapioca/dsl.rb
@@ -3,6 +3,7 @@
 
 require "rbi"
 require "spoom"
+require "tsort"
 
 require "tapioca"
 require "tapioca/runtime/reflection"
--- a/lib/tapioca/dsl/compiler.rb
+++ b/lib/tapioca/dsl/compiler.rb
@@ -38,6 +38,11 @@ module Tapioca
         sig { abstract.returns(T::Enumerable[Module]) }
         def gather_constants; end
 
+        sig { overridable.returns(T::Array[T.class_of(Compiler)]) }
+        def run_after_compilers
+          []
+        end
+
         sig { returns(T::Set[Module]) }
         def processable_constants
           @processable_constants ||= T.let(
--- a/lib/tapioca/dsl/pipeline.rb
+++ b/lib/tapioca/dsl/pipeline.rb
@@ -137,7 +137,21 @@ module Tapioca
         active_compilers = compilers
         active_compilers -= excluded_compilers
         active_compilers &= requested_compilers unless requested_compilers.empty?
-        active_compilers
+        tsort_active_compilers(active_compilers)
+      end
+
+      sig do
+        params(compilers: T::Array[T.class_of(Compiler)]).returns(T::Array[T.class_of(Compiler)])
+      end
+      def tsort_active_compilers(compilers)
+        dependencies = compilers
+          .sort_by { |compiler| T.must(compiler.name) }
+          .to_h { |compiler| [compiler, compiler.run_after_compilers] }
+
+        TSort.tsort(
+          lambda { |&block| dependencies.each_key(&block) },
+          lambda { |compiler, &block| dependencies.fetch(compiler).each(&block) },
+        )
       end
 
       sig do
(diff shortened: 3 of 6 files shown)


rzane:
This PR is an alternative to https://github.com/Shopify/tapioca/pull/2112

paracycle:
> Without forking the ActiveModel compiler, I have no way of overriding the type defined by Tapioca.

I don't think this is true. You could prepend a module in the `ActiveModelTypeHelper` module and hook the `type_for` method to do any processing that you want. You could ship this as a "compiler" in your own application and thus modify the built-in behaviour of Tapioca.

Since the last few versions, Tapioca gives you the guarantee that it will load its internal compilers before loading custom compilers, so that this kind of modification is possible.

> Without forking the ActiveRecordColumns compiler, I have no way of overriding the type defined by Tapioca.

Same comment here. Injecting behaviour into built-in Tapioca compilers is possible.

See this test for more details on how that is possible: https://github.com/Shopify/tapioca/blob/cc54939d48adeb7dc9c68866a3fa1bedc5565e3a/spec/tapioca/cli/dsl_spec.rb#L2250-L2308

In general, I really don't want to add any kind of ordering guarantees to compilers, since that will be brittle and full of race conditions.

paracycle:
> Can you explain why sorting the compilers would be brittle and full of race conditions?

Because Tapioca does not control a central list of compilers, they could be coming from the app, any of the gems it includes, or any other source. If we were to add the ability to declare ordering, that implies:

1. That there is a way to globally reference another compiler by name. A constant reference like in your example isn't enough, since that assumes the declaring compiler will be loaded after the other one, which isn't always correct.
2. That no two compilers will need to depend on a particular compiler to load and, yet, still perform conflicting operations. That is, if compiler B and C both depend on A but have different ideas about what should happen with A's definitions, then they will still be conflicting. In order to break that conflict, you'd have to add ordering between B and C, but they could be coming from wildly different sources and the fact that they would have to know about each other is a code smell.

All of this is highly problematic, and means additional code and workflows to support. Instead, it is already possible to modify the behaviour of built-in compilers, and until we meet a case where that is not enough, we shouldn't have to take on the burden of adding a new feature.

Issue #2118, opened 2024-12-07
`PrivateRelation` is not aware of class methods


rzane:
ActiveRecord provides two ways to define scopes. The first is `scope`, which Tapioca generates RBI definitions for. ActiveRecord also allows scopes to be defined as class methods:

```ruby
class Post < ActiveRecord::Base
  def self.published
    where(published: true)
  end
end
```

Tapioca's `PrivateRelation` doesn't include these scopes. Referring to one of these scopes produces a type error:

```ruby
Post.all.published
         ^^^^^^^^^ Method published does not exist on Post::PrivateRelation
```

paracycle:
This is not how Active Record works, so this will not be represented by Tapioca by adding these methods to relation classes.

See https://github.com/Shopify/tapioca/pull/2119#issuecomment-2526341932 and https://github.com/Shopify/tapioca/pull/1994#issuecomment-2302624697 for more information.

PR #2119, opened 2024-12-07
`PrivateRelation` needs to be aware of class methods

rzane:
### Motivation

See https://github.com/Shopify/tapioca/issues/2118

### Implementation

This PR looks up the available class methods and copies them to the generated relation modules using `create_method_from_def`.

One possible issue with this PR is how it might interact with other compilers. For example, ActiveRecord's enum defines a class method `.statuses`. The enum compiler will define a class method for this enum with types, but this compiler will also define that class method without types.

A better solution would be to define a module `GeneratedClassMethods` and `include` it explicitly in `PrivateRelation`. I'm not totally sure how to achieve this.

### Tests

This PR includes tests.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_scope.rb
+++ b/lib/tapioca/dsl/compilers/active_record_scope.rb
@@ -46,9 +46,9 @@ module Tapioca
 
         sig { override.void }
         def decorate
-          method_names = scope_method_names
+          scope_method_names = gather_method_names
 
-          return if method_names.empty?
+          return if scope_method_names.empty?
 
           root.create_path(constant) do |model|
             relations_enabled = compiler_enabled?("ActiveRecordRelations")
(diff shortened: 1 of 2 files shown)


paracycle:
Thanks for your contribution, but this is not something Tapioca should be doing. 

Class methods work on relations because they are delegated to (using `method_missing`) by the relation classes. That does not mean that the class methods are a part of the relation class for the model. For that reason, that delegation needs a different treatment than adding the methods to the relation class. An ideal solution would be some kind of annotation in Sorbet like `delegates_missing_methods_to_type { Foo }` which Tapioca could use to describe that delegation.

In the absence of such annotations, Tapioca should not be lying about class methods on models being defined as methods on relation classes.

magni-:
@paracycle what's the suggested workaround for dynamic scopes? I took a look at the `aasm` compiler to see how its [automatic scopes](https://github.com/aasm/aasm?tab=readme-ov-file#automatic-scopes) were implemented, but they're absent from the compiler.

PR #2122, opened 2024-12-10
Revert "Don't attempt to load built-in compilers multiple times"

KaanOzkan:
Reverts Shopify/tapioca#2069, makes it explicit in the comment

### DIFF
--- a/lib/tapioca/loaders/dsl.rb
+++ b/lib/tapioca/loaders/dsl.rb
@@ -85,7 +85,7 @@ module Tapioca
       def load_dsl_compilers
         say("Loading DSL compiler classes... ")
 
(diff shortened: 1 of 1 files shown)


lib/tapioca/loaders/dsl.rb
```diff
@@ -85,7 +85,12 @@ def load_dsl_extensions
       def load_dsl_compilers
         say("Loading DSL compiler classes... ")
 
-        # Load built-in compilers and custom compilers exported from gems
+        # Load all built-in compilers
```
paracycle, line 88: If you want, we can be more explicit in this comment, so that future onlookers will not be confused.

paracycle (approved):

PR #2125, opened 2024-12-12
Rename `#sorbet_type` → `#__tapioca_type`

amomchilov:
### Motivation

Rename the `#sorbet_type` method called by the `JsonApiClient::Resource` compiler, to match our naming decision in #2112:

1. Prefixed with `__` to discourage the method from being called from regular user code (i.e. only Tapioca should be calling it).
2. Renamed to emphasize Tapioca over Sorbet, to clarify how this ends up being used. Should hopefully be an easier term to search for, with fewer false matches.

### Implementation

Print a warning when we find `#sorbet_type`, while still honouring it. If there's only a definition for `#__tapioca_type`, we'll use that instead.

After some migration time, we can drop support for `#sorbet_type`.

### DIFF
--- a/lib/tapioca/dsl/compilers/json_api_client_resource.rb
+++ b/lib/tapioca/dsl/compilers/json_api_client_resource.rb
@@ -147,7 +147,17 @@ module Tapioca
           return "T.untyped" if type.nil?
 
           sorbet_type = if type.respond_to?(:sorbet_type)
+            line, file = type.method(:sorbet_type).source_location
+
+            $stderr.puts <<~MESSAGE
+              WARNING: `#sorbet_type` is deprecated. Please rename your method to `#__tapioca_type`."
+
+              Defined on line #{line} of #{file}
+            MESSAGE
+
             type.sorbet_type
+          elsif type.respond_to?(:__tapioca_type)
+            type.__tapioca_type
           elsif type == ::JsonApiClient::Schema::Types::Integer
             "::Integer"
           elsif type == ::JsonApiClient::Schema::Types::String
--- a/spec/tapioca/dsl/compilers/json_api_client_resource_spec.rb
+++ b/spec/tapioca/dsl/compilers/json_api_client_resource_spec.rb
@@ -330,7 +330,46 @@ module Tapioca
                 assert_equal(expected, rbi_for(:Post))
               end
 
-              it "honours types that declare sorbet_type" do
+              it "honours types that respond to #__tapioca_type" do
+                add_ruby_file("post.rb", <<~RUBY)
+                  class CustomType
+                    def self.__tapioca_type
+                      "Integer"
+                    end
+                  end
+
+                  class Post < JsonApiClient::Resource
+                    property :comment_count, type: :custom_type
+                    property :tag_count, type: :custom_type, default: 0
+                  end
+                RUBY
+
+                expected = <<~RBI
+                  # typed: strong
+
+                  class Post
+                    include JsonApiClientResourceGeneratedMethods
+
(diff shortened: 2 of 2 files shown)


lib/tapioca/dsl/compilers/json_api_client_resource.rb
```diff
@@ -147,7 +147,17 @@ def type_for(property)
           return "T.untyped" if type.nil?
 
           sorbet_type = if type.respond_to?(:sorbet_type)
+            line, file = type.method(:sorbet_type).source_location
```
amomchilov, line 150: This assumes that the `source_location` is available (i.e. this was defined in Ruby, and not in a C extension). Is that safe to assume, or should I guard against it?

Doing so is a bit uglier:

```ruby
msg = <<~MESSAGE
  WARNING: `#sorbet_type` is deprecated. Please rename your method to `#__tapioca_type`."
MESSAGE

if (source_location = type.method(:sorbet_type).source_location)
  msg.concat("Defined on line #{line} of #{file}"
end

msg = $stderr.puts
```

spec/tapioca/dsl/compilers/json_api_client_resource_spec.rb
```diff
@@ -368,6 +368,47 @@ def tag_count=(tag_count); end
 
                 assert_equal(expected, rbi_for(:Post))
               end
+
+              it "#sorbet_type is deprecated and prints a warning" do
```
andyw8, line 372:
```suggestion
              it "prints a warning that #sorbet_type is deprecated" do
```

andyw8 (approved):

spec/tapioca/dsl/compilers/json_api_client_resource_spec.rb
```diff
@@ -368,6 +368,73 @@ def tag_count=(tag_count); end
...
+                    end
+                  end
+
+                  class Post < JsonApiClient::Resource
+                    property :comment_count, type: :custom_type
+                    property :tag_count, type: :custom_type, default: 0
+                  end
+                RUBY
+
+                rbi_for(:Post) # Should not raise
```
amomchilov, line 436: We don't have `assert_nothing_raised`, so a comment will suffice.

spec/tapioca/dsl/compilers/json_api_client_resource_spec.rb
```diff
@@ -368,6 +368,73 @@ def tag_count=(tag_count); end
...
+                add_ruby_file("post.rb", <<~RUBY)
+                  class CustomType
+                    def self.sorbet_type
+                      "Integer"
+                    end
+
+                    def self.method(name)
+                      m = super
+                      # Fake the source location, as if this was defined in a C extension.
+                      def m.source_location
```
amomchilov, line 423: It's surprisingly hard to make a method with no `source_location` ahha. Native methods from C extensions would have it, but other than that, every way to define a Ruby method ended up having it defined. I tried:

1. `define_singleton_method`
2. `define_method`
3. `eval`
4. `class_eval`
5. `Class.new { ... }`

and they all had it defined.

andyw8: Interesting.

andyw8 (approved):

spec/tapioca/dsl/compilers/json_api_client_resource_spec.rb
amomchilov: IKR? I could have sworn you needed to explicitly pass in the `___FILE__` and `__LINE__` else you got no info at all

PR #2129, opened 2024-12-18
Support Bundler 2.6

### LINKED ISSUE #2132: tapioca gem fails with bundler 2.6.0
Works fine with bundler 2.5.23, but when upgrading to bundler 2.6.0:

```
% be tapioca gem
bundler: failed to load command: tapioca (/Users/bmulholland/.local/share/mise/installs/ruby/3.3.6/bin/tapioca)
/Users/bmulholland/.local/share/mise/installs/ruby/3.3.6/lib/ruby/gems/3.3.0/gems/bundler-2.6.0/lib/bundler/spec_set.rb:143:in `missing_specs': undefined method `flat_map' for nil (NoMethodError)

      @materializations.flat_map(&:completely_missing_specs)
                       ^^^^^^^^^
	from /Users/bmulholland/.local/share/mise/installs/ruby/3.3.6/lib/ruby/gems/3.3.0/gems/tapioca-0.16.5/lib/tapioca/gemfile.rb:71:in `materialize_deps'
```

deivid-rodriguez:
### Motivation

Bundler 2.6 has just been released and unfortunately I noticed that it does not play well with tapioca.

### Implementation

The main incompatibility I noticed was introduced at https://github.com/rubygems/rubygems/pull/8251.

In that PR, missing specs were moved from the spec set of materialized specifications to the spec set holding the original resolution.

Unfortunately, that's necessary to fix the issue I wanted to solve, and I'm not sure how to fix the original issue while making the change backwards compatible with what tapioca was doing here. So I'm trying to change tapioca instead to play nice with Bundler 2.6.

I also found another test only incompatibility introduced by https://github.com/rubygems/rubygems/pull/8165 in Bundler 2.5.23. That's because `Gem::Specification.stubs` would hold a `Bundler::SpecSet` before that PR (similar to a hash) while after the PR it was change to hold an array of specs (the same thing it holds when `bundler/setup` has not been loaded). Since this only affects specs, I changed the specs to play nice with that change as well.

### Tests

My hope is to run CI with the latest Bundler and get tests green.

I'm also planning to introduce a basic tapioca CLI check to Bundler's CI so that I can detect this issue in advance in the future.

This PR closes #2132.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
         uses: ruby/setup-ruby@v1
         with:
           bundler-cache: true
-          rubygems: 3.5.23
+          rubygems: 3.6.2
       - name: Run type check
         run: bin/typecheck
       - name: Lint Ruby files
@@ -36,6 +36,7 @@ jobs:
       matrix:
         ruby: ["3.1", "3.2", "3.3", "head"]
         rails: ["7.0", "current", "main"]
+        rubygems: ["3.6.2"]
         exclude:
           - ruby: "3.1"
             rails: "main"
@@ -44,6 +45,9 @@ jobs:
             experimental: true
           - ruby: "head"
             experimental: true
+          - rails: "current"
+            ruby: "3.3"
+            rubygems: "3.5.23"
     name: Ruby ${{ matrix.ruby }} - Rails ${{ matrix.rails }}
     env:
       RAILS_VERSION: ${{ matrix.rails }}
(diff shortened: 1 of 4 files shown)


deivid-rodriguez:
I'm also adding a basic tapioca check to our CI: https://github.com/rubygems/rubygems/pull/8346.

KaanOzkan:
Had to use `--force` for rebasing instead of `force-with-lease` but I think the diff is as expected.

PR #2146, opened 2025-01-08
Allow Active Record `select` to return nilable

dirceu:
Follow-up to https://github.com/Shopify/tapioca/pull/2103; it turns out you can send a nilable to `select`.

`QueryMethods#select`:

```ruby
> model.features.class
=> Feature::ActiveRecord_Associations_CollectionProxy
> model.features.select { nil }
=> []
```

`Enumerable#select`:

```ruby
> a = T.let([1, nil, 3], T::Array[T.nilable(Integer)])
>
> T.reveal_type(a)
>
* a.select do |num|
*   num.to_c
> end
=> [1, nil, 3]
```

PR #2147, opened 2025-01-08
ActiveRecord `select` block returns `BasicObject`

jakebrady5:
### Motivation

This change matches the block return value to that of the [sig](https://github.com/sorbet/sorbet/blob/104d0cee5299e5cb19c731945f5a52634b78951b/rbi/core/array.rbi#L2304) on `Array#select`.

In practice, updating our large Rails app surfaced many violations of this sig that would cause (in my opinion) undesirable code changes in order to satisfy the `T::Boolean` return type. It does seem quite historically idiomatic for Ruby code to rely on the implicit truthiness/falsiness check, rather than coerce it explicitly.

### Implementation

Continuing the original [implementation](https://github.com/Shopify/tapioca/pull/2103), this is just a change to the return type of the block sig.

### Tests

Existing tests are updated to reflect this change and seem to have sufficient coverage.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_relations.rb
+++ b/lib/tapioca/dsl/compilers/active_record_relations.rb
@@ -597,7 +597,7 @@ module Tapioca
                     sig.return_type = mod == relation_methods_module ? RelationClassName : AssociationRelationClassName
                   end
                   method.add_sig do |sig|
-                    sig.add_param("blk", "T.proc.params(record: #{constant_name}).returns(T::Boolean)")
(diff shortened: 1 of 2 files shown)


paracycle (approved):
Thanks, this makes a lot of sense. 

I would have gone for returning `T.untyped` instead, but I agree that matching the `Array#select` sig is a better choice.

jakebrady5:
I was late on the search, but just commenting that another contributor had a similar though differing idea [here](https://github.com/Shopify/tapioca/pull/2146).

PR #2152, opened 2025-01-14
Load DSL compilers and extensions ahead of time in add-on mode

vinistock:


lib/tapioca/loaders/dsl.rb
```diff
@@ -25,17 +25,15 @@ def load_application(
           halt_upon_load_error: true,
           lsp_addon: false
         )
-          loader = new(
+          # Loaded ahead of time when using the add-on to avoid reloading multiple times
+          return if lsp_addon
```
KaanOzkan, line 29: What do you think of doing this in AbstractDsl before calling `load_application`? So then this method doesn't need `lsp_addon` argument. But maybe this describes the intent better since it's closer to the relevant logic.

PR #2154, opened 2025-01-16
Update active record relations DSL count to accept a block argument

dmarsh-gusto:
### Motivation

While working on a Rails application, I have code that is of the format `Model.count {|m| m.foo_relation.boolean?}`.

I don't want to write this as `.select{}.count` as that would violate the [Rubocop::Cop::Performance::Cop](https://github.com/rubocop/rubocop-performance/blob/master/lib/rubocop/cop/performance/count.rb#L6-L48)

The fix is to update the type signatures DSL for active record relations to allow `count` to accept an optional block argument.

see https://edgeapi.rubyonrails.org/classes/ActiveRecord/Calculations.html#method-i-count

### Implementation

Added to a `create_block_param` call for active record relations `count` method, `lib/tapioca/dsl/compilers/active_record_relations.rb`

### Tests

Updated existing tests for active record relations `count` method to accept an optional block parameter.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_relations.rb
+++ b/lib/tapioca/dsl/compilers/active_record_relations.rb
@@ -402,6 +402,7 @@ module Tapioca
                 "count",
                 parameters: [
                   create_opt_param("column_name", type: "T.untyped", default: "nil"),
+                  create_block_param("block", type: "T.nilable(T.proc.params(record: T.untyped).returns(T.untyped))"),
                 ],
                 return_type: "T::Hash[T.untyped, Integer]",
               )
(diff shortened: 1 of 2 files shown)


lib/tapioca/dsl/compilers/active_record_relations.rb
```diff
@@ -402,6 +402,7 @@ def create_group_chain_methods(klass)
                 "count",
                 parameters: [
                   create_opt_param("column_name", type: "T.untyped", default: "nil"),
+                  create_block_param("block", type: "T.nilable(T.proc.params(record: T.untyped).returns(T.untyped))"),
```
KaanOzkan, line 405: I don't know why groupchain `count` is different than regular `count` definition. I think it'll be good to take a closer look and be consistent if appropriate. Regular one uses overloaded sigs instead https://github.com/Shopify/tapioca/blob/91ad27251ad8c422f116cd0eafe9a5bd82ea7d70/lib/tapioca/dsl/compilers/active_record_relations.rb#L838.

Issue #2158, opened 2025-01-20
CI errors on Rails 7.0


KaanOzkan:
https://github.com/Shopify/tapioca/actions/runs/12872412643/job/35887831035?pr=2157

andyw8:
https://github.com/rails/rails/issues/54260

PR #2159, opened 2025-01-20
Update concurrent-ruby

andyw8:
Closes https://github.com/Shopify/tapioca/issues/2158 (see https://github.com/rails/rails/issues/54260)


andyw8:
See https://github.com/Shopify/tapioca/pull/2161 instead.

PR #2161, opened 2025-01-20
Add workaround for for Rails 7.0 logger issue

andyw8:
Closes https://github.com/Shopify/tapioca/issues/2158 (see https://github.com/rails/rails/issues/54260)

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -7,6 +7,7 @@ ENV["DEFAULT_TEST"] = "spec/**/*_spec.rb"
 require "bundler/setup"
 require "debug/prelude"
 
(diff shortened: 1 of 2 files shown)


vinistock:
Sure, but in that case let's change the comments to reflect what we need to do. They are saying "once it's fixed", but this is already fixed we just haven't upgraded.

paracycle:
Yeah, I think we should drop 7.0 support at this point.

According to https://rubyonrails.org/maintenance the bug-fix supported versions of Rails is 8.0 and 7.2. We should move to testing on:
- `7.2`
- `current` (meaning `8.0`)
- `main`

andyw8:
Comment updated.

PR #2163, opened 2025-01-22
[Tapioca Add-on] Trigger `ActiveSupport::TestCase` DSL compiler on fixture changes

andyw8:
Addresses https://github.com/Shopify/team-ruby-dx/issues/1305

TODO:

- [x] Point to ruby-lsp-rails once released (and update RBI)
- [ ] ~Verify performance in Core~

Note: This doesn't work on Core because the fixtures is compiler is disabled and Core uses a custom compiler.

### DIFF
--- a/lib/ruby_lsp/tapioca/addon.rb
+++ b/lib/ruby_lsp/tapioca/addon.rb
@@ -84,6 +84,7 @@ module RubyLsp
         return unless @rails_runner_client # Client is not ready
 
         has_route_change = T.let(false, T::Boolean)
+        has_fixtures_change = T.let(false, T::Boolean)
 
         constants = changes.flat_map do |change|
           path = URI(change[:uri]).to_standardized_path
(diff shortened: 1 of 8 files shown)


andyw8:
I still need to test on Core but feel free to review.

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -103,14 +110,18 @@ def workspace_did_change_watched_files(changes)
           end
         end.compact
 
-        return if constants.empty? && !has_route_change
+        return if constants.empty? && !has_route_change && !has_fixtures_change
```
andyw8, line 113: If we continue adding other things to watch then this will get unwieldy, but I think it's ok for now.

paracycle: Why can't we fold them both into a single boolean?

paracycle: Ah, I see that we run different DSLs based on the flags.

spec/dummy/app/models/user.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+class User < ActiveRecord::Base
+end
```
KaanOzkan, line 4: If this model needed for fixtures I think the name should be consistent, we can have a users fixture instead of `blah`

PR #2165, opened 2025-01-23
Run URL helpers compiler when routes are modified

vinistock:
### Motivation

Start generating URL helper DSLs upon route modifications. This is one of the compiler cases where there's not a constant, but rather a file pattern that triggers generation.

### Implementation

1. Started noticing modifications to any `routes.rb` file and then triggering a delegate notification
2. The server add-on side will collect relevant constants and then pass those to the DSL method. Note that the constants collected by the compiler are a mix of application constants and artificial ones created by Tapioca itself for the URL helper modules

### Tests

Added tests and made some adjustments to make them more deterministic and avoid sleeping.

EDIT: couldn't avoid sleeping for some reason. Tests could use some future improvement.

### DIFF
--- a/lib/ruby_lsp/tapioca/addon.rb
+++ b/lib/ruby_lsp/tapioca/addon.rb
@@ -79,10 +79,17 @@ module RubyLsp
         return unless T.must(@global_state).enabled_feature?(:tapiocaAddon)
         return unless @rails_runner_client # Client is not ready
 
+        has_route_change = T.let(false, T::Boolean)
+
         constants = changes.flat_map do |change|
           path = URI(change[:uri]).to_standardized_path
           next if path.end_with?("_test.rb", "_spec.rb")
           next unless file_updated?(change, path)
+
+          if File.basename(path) == "routes.rb" || File.fnmatch?("**/routes/**/*.rb", path, File::FNM_PATHNAME)
+            has_route_change = true
+            next
+          end
 
           entries = T.must(@index).entries_for(change[:uri])
           next unless entries
@@ -92,14 +99,21 @@ module RubyLsp
           end
         end.compact
 
-        return if constants.empty?
+        return if constants.empty? && !has_route_change
 
         @rails_runner_client.trigger_reload
-        @rails_runner_client.delegate_notification(
-          server_addon_name: "Tapioca",
-          request_name: "dsl",
-          constants: constants,
-        )
+
+        if has_route_change
+          @rails_runner_client.delegate_notification(server_addon_name: "Tapioca", request_name: "route_dsl")
+        end
+
+        if constants.any?
+          @rails_runner_client.delegate_notification(
+            server_addon_name: "Tapioca",
+            request_name: "dsl",
+            constants: constants,
+          )
+        end
       end
 
       private
--- a/lib/ruby_lsp/tapioca/server_addon.rb
+++ b/lib/ruby_lsp/tapioca/server_addon.rb
@@ -2,6 +2,7 @@
 # frozen_string_literal: true
 
 require "tapioca/internal"
+require "tapioca/dsl/compilers/url_helpers"
 
 module RubyLsp
   module Tapioca
@@ -22,17 +23,28 @@ module RubyLsp
             halt_upon_load_error: false,
           ).load_dsl_extensions_and_compilers
         when "dsl"
+          fork { dsl(params[:constants]) }
+        when "route_dsl"
           fork do
-            dsl(params)
+            constants = ::Tapioca::Dsl::Compilers::UrlHelpers.gather_constants
+            dsl(constants.map(&:name), "--only=Tapioca::Dsl::Compilers::UrlHelpers", "ActiveSupportConcern")
           end
         end
       end
 
       private
 
-      def dsl(params)
+      def dsl(constants, *args)
         load("tapioca/cli.rb") # Reload the CLI to reset thor defaults between requests
-        ::Tapioca::Cli.start(["dsl", "--lsp_addon", "--workers=1"] + params[:constants])
+
+        # Order here is important to avoid having Thor confuse arguments. Do not put an array argument at the end before
+        # the list of constants
+        arguments = ["dsl"]
+        arguments.concat(args)
+        arguments.push("--lsp_addon", "--workers=1")
+        arguments.concat(constants)
+
+        ::Tapioca::Cli.start(arguments)
       end
     end
   end
--- a/spec/tapioca/addon_spec.rb
+++ b/spec/tapioca/addon_spec.rb
@@ -3,8 +3,9 @@
 
 require "spec_helper"
 require "language_server-protocol"
-require "ruby_lsp/utils"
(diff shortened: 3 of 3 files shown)


lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -92,14 +99,21 @@ def workspace_did_change_watched_files(changes)
...
 
         @rails_runner_client.trigger_reload
-        @rails_runner_client.delegate_notification(
-          server_addon_name: "Tapioca",
-          request_name: "dsl",
-          constants: constants,
-        )
+
+        if has_route_change
+          @rails_runner_client.delegate_notification(server_addon_name: "Tapioca", request_name: "route_dsl")
```
KaanOzkan, line 107: How do the logs look with multiple forked processes running at the same time? Any issues of intertwined output?

lib/ruby_lsp/tapioca/server_addon.rb
```diff
@@ -22,17 +23,20 @@ def execute(request, params)
             halt_upon_load_error: false,
           ).load_dsl_extensions_and_compilers
         when "dsl"
+          fork { dsl(params[:constants]) }
+        when "route_dsl"
           fork do
-            dsl(params)
+            constants = ::Tapioca::Dsl::Compilers::UrlHelpers.gather_constants
+            dsl(constants.map(&:name))
```
KaanOzkan, line 30: If we have a different request we can take advantage of it and only run the specific compiler we are interested in. It should be faster as we'll skip `gather_constants` from other ones. It'll require more work but I suggest triggering `dsl` with a specific compiler and changing the existing logic to not delete all DSL RBIs when you specify a single compiler (This can be a different PR). Wdyt?

lib/ruby_lsp/tapioca/addon.rb
vinistock: I don't think they would run at the same time because if the file is `routes.rb`, we skip looking for declarations in it. It's not very common for people to define modules or classes that need DSLs as part of routes.

If they were to run at the same time, then indeed the outputs would be mixed.

lib/ruby_lsp/tapioca/server_addon.rb
vinistock: I can give that a shot. I'll need to take a look at how much code is necessary to trigger only one compiler manually.

I agree, it would be a considerable improvement.

andyw8 (approved):

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -79,11 +79,18 @@ def workspace_did_change_watched_files(changes)
...
         return unless @rails_runner_client # Client is not ready
 
+        has_route_change = T.let(false, T::Boolean)
+
         constants = changes.flat_map do |change|
           path = URI(change[:uri]).to_standardized_path
           next if path.end_with?("_test.rb", "_spec.rb")
           next unless file_updated?(change, path)
 
+          if File.basename(path) == "routes.rb"
```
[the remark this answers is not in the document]
andyw8: (also used for some Shopify apps, e.g. `billing`).

lib/ruby_lsp/tapioca/server_addon.rb
```diff
@@ -2,6 +2,7 @@
 # frozen_string_literal: true
 
 require "tapioca/internal"
+require "tapioca/dsl/compilers/url_helpers"
```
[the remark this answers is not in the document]
vinistock: No because we execute the generation in a forked process, which would mean we would load it every single time.

spec/tapioca/addon_spec.rb
```diff
@@ -3,25 +3,78 @@
...
+              tapiocaAddon: true,
+            },
+          },
+        })
+
+        addon = Addon.new
+        addon.instance_variable_set(:@rails_runner_client, @client)
+        addon.instance_variable_set(:@global_state, global_state)
+        addon.instance_variable_set(:@index, global_state.index)
+        addon.instance_variable_set(:@outgoing_queue, @outgoing_queue)
```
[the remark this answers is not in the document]
vinistock: I agree. This is a side effect of the test currently spawning the Rails add-on runner client manually, rather than creating real add-on instances.

It limits our testing, because then we can only test the logic in `server_addon` and not in the editor side `addon`.

I can follow up with some improvements to the test.

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -79,11 +79,18 @@ def workspace_did_change_watched_files(changes)
...
         return unless @rails_runner_client # Client is not ready
 
+        has_route_change = T.let(false, T::Boolean)
+
         constants = changes.flat_map do |change|
           path = URI(change[:uri]).to_standardized_path
           next if path.end_with?("_test.rb", "_spec.rb")
           next unless file_updated?(change, path)
 
+          if File.basename(path) == "routes.rb" || File.fnmatch?("**/routes/**/*.rb", path)
```
andyw8, line 89: Does this need `File::FNM_PATHNAME | File::FNM_EXTGLOB`?

lib/ruby_lsp/tapioca/server_addon.rb
vinistock: I took a stab at this. It requires a bit of a complex refactor in Tapioca because the responsibility of generating a DSL RBI is split between the `Pipeline` and `AbstractDsl`. One generates the RBI contents and the other one further decorates it and writes to disk. I think it's an improvement we can attempt later.

However, I was successful in passing the `--only` option to DSL limiting to only 2 compilers for routes (the concerns one is necessary because it triggers some action view helpers generation).

lib/ruby_lsp/tapioca/addon.rb
vinistock: I don't believe so. We're not using the `{}` syntax and we're declaring all path parts we're interested in.

What scenario is this going to fail without those flags?

andyw8: I think we need `File::FNM_PATHNAME` but not the other:

```
irb(main):005> File.fnmatch?("**/routes/**/*.rb", "routes/foo.rb")
=> false
irb(main):006> File.fnmatch?("**/routes/**/*.rb", "routes/foo.rb", File::FNM_PATHNAME)
=> true
```

vinistock: Good catch.

KaanOzkan (approved):

lib/ruby_lsp/tapioca/server_addon.rb
KaanOzkan: `--only` sounds good. Thanks.

lib/ruby_lsp/tapioca/addon.rb
vinistock: The only possibility is if you switch branches, git pull, run `rubocop -a` or anything that will trigger multiple files to be updated all at once.

During normal options, like saving files manually, it should never happen.

PR #2170, opened 2025-01-28
Prevent CI from running on both `push` and `pull_request` events

KaanOzkan:
### Motivation

Duplicate CI runs waste resources

### Implementation

 Push events will only be triggered on `main`. As a result pushes to your branches without a PR will not trigger a build.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,10 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+    - main
+  pull_request:
 
 env:
   SRB_SKIP_GEM_RBIS: true


paracycle:
I am not sure if the reasoning is correct. AFAIK, the `pull_request` build runs against a merge commit of the current branch onto the current state of `main` which gives you more confidence that your PR won't cause a conflict if merged to `main` at that point in time.

So, there is a separate value we get from `pull_request` builds.

Ref: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request

> Note that GITHUB_SHA for this event is the last merge commit of the pull request merge branch. If you want to get the commit ID for the last commit to the head branch of the pull request, use github.event.pull_request.head.sha instead.

paracycle (approved):
Thanks!

paracycle:
> Updated the configuration to favour `pull_request` builds by default instead. As a result pushed branches without a PR don't trigger CI.

Except pushes to `main`

PR #2176, opened 2025-01-29
Add progress indication for DSL operations

vinistock:
### Motivation

Sometimes RBI generation can take a while and the user has no hints that anything is happening. This PR adds progress notifications, so that the user has visual hints that Tapioca is running when a file is saved.

### Implementation

There's not much, I just wrapped our execution in the progress block.

### DIFF
--- a/lib/ruby_lsp/tapioca/server_addon.rb
+++ b/lib/ruby_lsp/tapioca/server_addon.rb
@@ -24,21 +24,35 @@ module RubyLsp
             halt_upon_load_error: false,
           ).load_dsl_extensions_and_compilers
         when "dsl"
-          fork { dsl(params[:constants]) }
+          fork do
+            with_notification_wrapper("dsl", "Generating DSL RBIs") do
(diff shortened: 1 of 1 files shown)


lib/ruby_lsp/tapioca/server_addon.rb
```diff
@@ -24,22 +24,38 @@ def execute(request, params)
             halt_upon_load_error: false,
           ).load_dsl_extensions_and_compilers
         when "dsl"
-          fork { dsl(params[:constants]) }
+          fork do
+            with_notification_wrapper("dsl", "Generating DSL RBIs") do
+              dsl(params[:constants])
+              end_progress("dsl")
```
KaanOzkan, line 30: Why doesn't the wrapper end it after yielding the block?

KaanOzkan: Not a big deal here but it would be cleaner in the future. Looks easy to forget to do.

PR #2178, opened 2025-01-31
Reload DSL compilers in the workspace automatically

vinistock:
### Motivation

If the user edits custom DSL compilers that exist in the codebase, we never reload them and thus may end up generating incorrect RBI files.

This is not a _super_ common scenario as most people aren't working on custom compilers on a regular basis, but we can still perform automatic reloading for them.

### Implementation

We do not need to care about compilers defined in gems or in Tapioca itself. If a gem is bumped, the entire LSP is restarted and so we will pick up the changes.

So I started paying attention to changes to compilers inside the workspace and then firing a reload notification.

The implementation simply lists all descendants of `Tapioca::Dsl::Compiler` that are defined inside of the codebase. Then we proceed to remove the constant, delete the file entry from `$LOADED_FEATURES` and re-require the file to load the current state.

**Note**

I did not figure out a way to reload DSL extensions. Those are usually used to monkeypatch classes of the application itself or even create constants that don't exist elsewhere, so I'm not sure what we could do that would be guaranteed to work on every case.

### Tests

Added a test.

### DIFF
--- a/lib/ruby_lsp/tapioca/server_addon.rb
+++ b/lib/ruby_lsp/tapioca/server_addon.rb
@@ -16,40 +16,18 @@ module RubyLsp
         case request
         when "reload_workspace_compilers"
           with_notification_wrapper("reload_workspace_compilers", "Reloading DSL compilers") do
-            compiler_paths = Dir.glob(
-              "#{params[:workspace_path]}/**/tapioca/**/compilers/**/*.rb",
-              File::FNM_PATHNAME | File::Constants::FNM_EXTGLOB,
-            )
-
-            # Remove all loaded compilers that are inside the workspace
-            ::Tapioca::Dsl::Compiler.descendants.each do |compiler|
-              name = compiler.name
-              next unless name && compiler_paths.include?(Module.const_source_location(name).first)
-
-              *parts, unqualified_name = name.split("::")
-
-              if parts.empty?
-                Object.send(:remove_const, unqualified_name)
-              else
-                parts.join("::").safe_constantize.send(:remove_const, unqualified_name)
-              end
-            end
-
-            # Remove from $LOADED_FEATURES each workspace compiler file and then re-required it to reload
-            compiler_paths.each do |path|
-              $LOADED_FEATURES.delete(path)
-              require File.expand_path(path)
-            end
+            @loader&.reload_custom_compilers
           end
         when "load_compilers_and_extensions"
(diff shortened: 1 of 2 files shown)


vinistock:
* **#2178** <a href="https://app.graphite.dev/github/pr/Shopify/tapioca/2178?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/tapioca/2178?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

KaanOzkan (approved):
I think it's fine to not reload extensions for now they aren't common. Can we also test it in core?

lib/ruby_lsp/tapioca/server_addon.rb
```diff
@@ -14,6 +14,33 @@ def name
...
+              else
+                parts.join("::").safe_constantize.send(:remove_const, unqualified_name)
+              end
+            end
+
+            # Remove from $LOADED_FEATURES each workspace compiler file and then re-required it to reload
+            compiler_paths.each do |path|
+              $LOADED_FEATURES.delete(path)
+              require File.expand_path(path)
+            end
```
paracycle, line 42: I would very much like to see this encapsulated in the DSL pipeline logic that actually does the compiler loading and here is why:

The DSL pipeline already knows the exact paths and names of the compilers that it has loaded, and it can keep track of those for this purpose. When we ask the pipeline to reload its compilers, we could give it an (optional) list of paths that have changed, and it can decide if any of the compilers will be loaded or not. And the LSP logic will be much clearer.

vinistock: I agree with you, but we once again hit the problem that we lose the whole pipeline state due to forking.

In the latest commit, I moved the logic to `Loaders::Dsl`, since we can easily hold a reference to a loader and use it repeatedly. Let me know what you think or you feel like I'm missing something.

paracycle: Yes, sorry, that was exactly what I had in mind, thank you. I misremembered it as the pipeline loading the compilers, but it wasn't, it was the loader, of course.

This looks great, thanks for the refactor. IMO, this looks much nicer and is more encapsulated.

PR #2187, opened 2025-02-10
Collect telemetry about add-on usage

vinistock:


Morriar (approved):
Would it be useful to also log the `activate` event?

vinistock:
I bumped the Ruby LSP requirement to v0.23.10, otherwise we cannot use the new machine ID. This will certainly cause some cases of the Tapioca add-on not being turned on until updates roll out to everybody, but it shouldn't be too long.

I also started producing a telemetry event for the incompatible API case, so that we can see how quickly the upgrades roll out.

st0012 (approved):

vinistock:
### Merge activity

* **Feb 11, 10:46 AM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Shopify/tapioca/2187).

PR #2190, opened 2025-02-10
Use more accurate heuristic to ignore test files in lsp addon

KaanOzkan:
### Motivation

Constants defined in test helpers will not be available to us in add-on mode and we don't have to attempt to generate them.
### Implementation

Widen the net by ignoring common test folders completely, `test` for minitest, `spec` for rspec and `features` for cucumber.
### Tests

Manual testing


lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -89,7 +89,7 @@ def workspace_did_change_watched_files(changes)
 
         constants = changes.flat_map do |change|
           path = URI(change[:uri]).to_standardized_path
-          next if path.end_with?("_test.rb", "_spec.rb")
+          next if File.fnmatch?("**/{test,spec,features}/**/*", path, File::FNM_PATHNAME | File::FNM_EXTGLOB)
```
andyw8, line 92: Won't this cause us to miss changes to `test/fixtures`?

vinistock: Actually, instead of trying to come up with a more complicated pattern, I think we can just move this down, no?

When we check for the custom DSL compilers, we skip to the next iteration if it matches. Same thing for routes and fixtures.

We can just move this check before trying to retrieve constants. WDYT?

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -89,7 +89,14 @@ def workspace_did_change_watched_files(changes)
 
         constants = changes.flat_map do |change|
           path = URI(change[:uri]).to_standardized_path
-          next if path.end_with?("_test.rb", "_spec.rb")
+
+          # NOTE: We only get notification for fixtures if ruby-lsp-rails is v0.3.31 or higher
```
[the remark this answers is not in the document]
andyw8: Feel free to remove. I wrote it as a reminder in case someone was trying to troubleshoot fixture handling notw orking.

vinistock (approved):

PR #2201, opened 2025-02-12
Allow ignoring gem constants

gmalette:
### Motivation

In one of our repos we're loading the same gem `foo` both in a dependency (gem `bar`), and directly in the host project. This `foo` gem is then generated with tapioca both as part of `sorbet/rbi/gem/foo@version.rbi` and `bar@version.rbi`.

The problem lies in that `foo` signatures that can't be loaded twice by Sorbet, such at `type_member` and `alias`.

I've been able to sidestep the problem so far by having `typed_overrides -> "ignore"` and manually creating annotations for the gem, but the list of constants required grows every day and this is untenable. What I really need is to be able to say "ignore `foo` as generated in `bar`". This is what this PR does.

I haven't exposed this to the `config.yml` file yet

### Implementation

This PR is the most bare-bones it can be; I just want to have a starting point for the discussion.

### Tests

Basic test to show what I expect it to do.

### DIFF
--- a/lib/tapioca/gem/pipeline.rb
+++ b/lib/tapioca/gem/pipeline.rb
@@ -22,19 +22,22 @@ module Tapioca
           error_handler: T.proc.params(error: String).void,
           include_doc: T::Boolean,
           include_loc: T::Boolean,
+          ignore: T::Array[String],
         ).void
       end
       def initialize(
         gem,
         error_handler:,
         include_doc: false,
-        include_loc: false
+        include_loc: false,
+        ignore: []
       )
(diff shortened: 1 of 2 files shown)


paracycle:
> This `foo` gem is then generated with tapioca both as part of `sorbet/rbi/gem/foo@version.rbi` and `bar@version.rbi`.

I think we should fix this root problem, instead of adding ignores for namespaces.

If the attribution of methods/constants to gems is wrong, we should fix that.

I actually think #293 might already fix this problem.

PR #2207, opened 2025-02-20
Raise `Thor::Error` instead of `exit`

KaanOzkan:
### Motivation

Exiting in add-on mode causes progress notifications to not be cleaned up

### Implementation

Raise a `Thor::Error` instead. It's rescued by thor and we control its behaviour using `exit_on_failure?` [here](https://github.com/Shopify/tapioca/blob/73381f35d990302e61822a32b709e598912d29db/lib/tapioca/cli.rb#L392)

### DIFF
--- a/lib/tapioca/commands/abstract_dsl.rb
+++ b/lib/tapioca/commands/abstract_dsl.rb
@@ -131,15 +131,22 @@ module Tapioca
 
       sig { returns(Tapioca::Dsl::Pipeline) }
       def create_pipeline
+        error_handler = if @lsp_addon
+          ->(error) {
+            say(error)
+          }
+        else
+          ->(error) {
+            say_error(error, :bold, :red)
+          }
+        end
         Tapioca::Dsl::Pipeline.new(
           requested_constants:
             constantize(@requested_constants) + constantize(constants_from_requested_paths, ignore_missing: true),
           requested_paths: @requested_paths,
           requested_compilers: constantize_compilers(@only),
           excluded_compilers: constantize_compilers(@exclude),
-          error_handler: ->(error) {
-            say_error(error, :bold, :red)
-          },
+          error_handler: error_handler,
           skipped_constants: constantize(@skip_constant, ignore_missing: true),
           number_of_workers: @number_of_workers,
           compiler_options: @compiler_options,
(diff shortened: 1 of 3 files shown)


lib/tapioca/commands/abstract_dsl.rb
```diff
@@ -131,15 +131,22 @@ def load_application
 
       sig { returns(Tapioca::Dsl::Pipeline) }
       def create_pipeline
+        error_handler = if @lsp_addon
+          ->(error) {
+            say(error, :bold, :red)
```
KaanOzkan, line 136: This is an adjacent change to show errors in lsp output. `say_error` messages aren't visible in the logs, I assume due to `stderr`, but we'd like to have them.

lib/tapioca/dsl/pipeline.rb
```diff
@@ -78,6 +78,7 @@ def run(&blk)
             No classes/modules can be matched for RBI generation.
             Please check that the requested classes/modules include processable DSL methods.
           ERROR
+          raise Thor::Error, ""
```
andyw8, line 81:
```suggestion
          raise Thor::Error, "Exiting"
```
(or something like that)

andyw8: Oh, is the argument displayed to the user?

(not a blocker, I was suggesting it just to ease possible future troubleshooting. If we saw that without any backtrace then it could be pain to track where it came from).

lib/tapioca/commands/abstract_dsl.rb
KaanOzkan: Good point.

paracycle (approved):
I don't love that Tapioca internals know about `Thor::Error`, but we can refactor that later. For now, this should be fine to merge.

KaanOzkan:
> I don't love that Tapioca internals know about Thor::Error, but we can refactor that later. For now, this should be fine to merge.

I opened https://github.com/Shopify/tapioca/issues/2208, feel free to make suggestions in the issue

KaanOzkan:
We have a CI failure due to Sorbet payload, it might be the github action cache or something is preventing bumping Sorbet in some workflows. I'm going to go ahead since this fix is time sensitive.

PR #2209, opened 2025-02-20
Migrate to RBS signatures supported by Sorbet

Morriar:
This PRs bump our Sorbet version to v0.5.11842 so we can start using RBS comment signatures instead of the `sig{}` syntax 🎉 

I migrated all the signatures using `spoom srb sigs translate`.

Note that for the moment we're not translating signatures that contain `abstract`.


will:
Just curious @Morriar and @paracycle, now that you've been using the rbs comments instead of the sorbet sig syntax for several months now, what is your verdict on them? I'm still just using sigs, but wouldn't be opposed to switching.

PR #2210, opened 2025-02-20
Always reload routes in UrlHelpers

cquinones100:
### Motivation
We are upgrading to Rails 8 and the generated RBIs for `GeneratedUrlHelpersModule` and `GeneratedPathHelpersModule` did not include any of our configured routes. The previous version of the `UrlHelpers` compiler links to https://github.com/rails/rails/pull/52012, so I think the change was made to anticipate lazy loaded routes, but it does not seem to work in practice.

### Implementation
`Rails.application.reload_routes!` produced the result I was looking for and I found that this also worked in Rails 7.

### Tests
The old code did not break tests when CURRENT_RAILS_VERSION was set to 7.1 or 8.0.0, so some guidance on how to write a test for this would be appreciated. I am surprised that [these tests](https://github.com/shopify/tapioca/blob/main/spec/tapioca/dsl/compilers/url_helpers_spec.rb#L215) weren't breaking.

### DIFF
--- a/lib/tapioca/dsl/compilers/url_helpers.rb
+++ b/lib/tapioca/dsl/compilers/url_helpers.rb
@@ -102,9 +102,11 @@ module Tapioca
           def gather_constants
             return [] unless defined?(Rails.application) && Rails.application
 
-            # Load routes if they haven't been loaded yet (see https://github.com/rails/rails/pull/51614).
-            routes_reloader = Rails.application.routes_reloader
-            routes_reloader.execute_unless_loaded if routes_reloader&.respond_to?(:execute_unless_loaded)
+            # In rails 8, `Rails.application.routes_reloader.execute_unless_loaded`
+            # is not behaving as it did in Rails 7.
+            # With this workaround we force the routes to be reloaded regardless
+            # of whether they have been loaded or not
+            Rails.application.reload_routes!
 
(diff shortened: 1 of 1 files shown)


paracycle:
I am not sure that I understand how this is a problem. Even [the follow up change in Rails 8](https://github.com/rails/rails/pull/52353) should still be loading all the routes when `execute_unless_loaded ` is called. This is [exactly the call that loads all the routes](https://github.com/rails/rails/commit/cdb283566c5d5251e2f7e663e655e57e8e10abbc#diff-6d1396aacca78513730c56d8c5ca4dae5697e59d7a5698762a3d2806a182ea4bR180) if the application is being loading in `eager_load` mode, e.g. in production, so that should work in Tapioca, as well.

Can you give a minimal example of a Rails 8 app that doesn't work properly?

paracycle:
Thanks for digging into this. The explanation makes sense but doesn't quite explain to me how your application is able to fully load its routes when eager loading, for example in production mode.

As I linked before, the call that gets made is again the `execute_unless_loaded` method, and the behaviour should be the same.

I suspect Tapioca's behaviour to be indicative of your app's production behaviour in this case. If my suspicion is correct, this might be preventing you from a production error.

Could you try to run your application either in production mode, or in eager load mode to see if your routes work fine?

Issue #2212, opened 2025-02-20
CI failure - sorbet payload


KaanOzkan:
https://github.com/Shopify/tapioca/actions/runs/13446065995/job/37571549416?pr=2211#step:6:765

PR #2213, opened 2025-02-20
Fix `fetch_annotation` signature

Morriar:
This code works by miracle.

We call `fetch_annotation` within a `select` but the method return `void`. It only works because the `T::Private::Types::Void::VOID` used by `sorbet-runtime` to replace the actual value is truthy.


paracycle (approved):

KaanOzkan (approved):

KaanOzkan:
CI issue is known https://github.com/Shopify/tapioca/issues/2212. Tried a quick filtering fix but the assertion requires a Spoom::ExecResult and not a String

PR #2221, opened 2025-02-27
Bump Ruby LSP upper bound to < 0.25

vinistock:
### Motivation

The changes in https://github.com/Shopify/ruby-lsp/pull/3252 will be breaking because the return of `document.parse_result` will change. However, that doesn't impact Tapioca and we will not include any breaking changes that do impact Tapioca in v0.24.

Let's bump our upper requirement ahead of time, so that people can already start upgrading and so we can eliminate any windows where the Tapioca add-on wouldn't work.

### Implementation

Just bumped the upper bound to < 0.25.

### DIFF
--- a/lib/ruby_lsp/tapioca/addon.rb
+++ b/lib/ruby_lsp/tapioca/addon.rb
@@ -1,7 +1,7 @@
 # typed: strict
 # frozen_string_literal: true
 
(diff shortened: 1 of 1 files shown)


andyw8 (approved):

vinistock:
### Merge activity

* **Feb 27, 9:17 AM EST**: [Graphite](https://app.graphite.dev/github/pr/Shopify/tapioca/2221) couldn't merge this PR because it failed for an unknown reason.

PR #2225, opened 2025-03-04
Ensure delegated type with single type generates safely

cam-stitt:
### Motivation
If a delegated type is defined with a single type class, the RBI that is generated fails because the `T.any` has only one argument.

### Implementation
A simple condition to either place `T.any` or the singular class name, depending on the number of types provided.

### Tests
I added a test case to confirm the expected outcome for a singular delegated type.


cam-stitt:
As much as I would like to agree to the CLA, I am not providing you my address.

paracycle:
Closing in favor of #2335

PR #2236, opened 2025-03-14
Add support for RBS signature comments

Morriar:
### Motivation

Before this PR, when generating RBI files for gems using RBS comments, Tapioca was unable to generate signatures.

This PR introduce a preprocessing of all loaded files that rewrites the RBS comments into actual Sorbet signatures blocks.

Thus a file looking like this:

```rb
class SomeGem
  #: -> void
  def foo; end
end
```

Will be translated into this at load time:

```rb
class SomeGem
  sig { void }
  def foo; end
end
```

This will enable `sorbet-runtime` to properly see the signature and wrap the method as it would do normally.

Then Tapioca will be able to pick up the signature and generate properly types RBIs for gems using RBS comment.

### Implementation

We rely on `ruby-next` for the rewriting. While I not convinced by the gem itself, it does work. We could potentially rewrite the mechanism with an ad-hoc implementation later.

### Tests

See automated tests.


lib/tapioca/internal.rb
```diff
@@ -1,6 +1,11 @@
 # typed: strict
 # frozen_string_literal: true
 
+# The rewriter needs to be loaded very early so RBS comments within Tapioca itself are rewritten
+require "spoom"
+require "tapioca/rbs/rewriter"
```
Morriar, line 6: We need this to happen super early so we also rewrite the RBS comments we find inside Tapioca

lib/tapioca/loaders/loader.rb
```diff
@@ -160,7 +160,7 @@ def with_rails_application(&blk)
         Rails.app_class = Rails.application = rails_application
       end
 
-      #: -> Array[singleton(Rails::Engine)]
+      T::Sig::WithoutRuntime.sig { returns(T::Array[T.class_of(Rails::Engine)]) }
```
Morriar, line 163: Now that we re-execute the signatures, having a reference to `Rails` is a problem. For now I changed them back to a `T::Sig::WithoutRuntime`. We can add a `@without-runtime` annotation later.

lib/tapioca/rbs/rewriter.rb
```diff
@@ -0,0 +1,34 @@
...
+        context.track!(self)
+        Spoom::Sorbet::Sigs.rbs_to_rbi(source)
+      end
+    end
+  end
+end
+
+RubyNext::Language.include_patterns.clear
+RubyNext::Language.include_patterns << "**/*.rb"
+RubyNext::Language.rewriters = [Tapioca::RBS::Rewriter]
```
Morriar, line 34: Note that we're replacing any list of rewriters that may have been set

spec/tapioca/gem/pipeline_spec.rb
```diff
@@ -4556,5 +4556,134 @@ class Foo; end
...
+
+      output = template(<<~RBI)
+        class Foo
+          sig { params(a: ::Integer, b: ::String).void }
+          def bar(a, b:); end
+
+          sig { returns(::String) }
+          def foo; end
+
+          def foo=(_arg0); end
```
Morriar, line 4591: Note that writer were already untyped before this PR.

lib/tapioca/rbs/rewriter.rb
```diff
@@ -0,0 +1,34 @@
...
+module Tapioca
+  module RBS
+    # Transpiles RBS comments back into Sorbet's `sig{}` blocks
+    class Rewriter < RubyNext::Language::Rewriters::Text
+      NAME = "rbs_rewriter"
+
+      #: (String source) -> String
+      def rewrite(source)
+        # Do not try to parse files that don't have RBS comments
+        return source unless source =~ /^\s*#\s*typed: (ignore|false|true|strict|strong|__STDLIB_INTERNAL)/
```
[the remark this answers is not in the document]
Morriar: The more I'm thinking about it and the more I think it's a mistake. I already got confused while writing tests because I forgot the sigil. Matching on `#:` seems safer and future proof 🤔

Morriar:
> We would gain significant performance improvement for tapioca gem with arguments if we could rewrite only the gems we are generating RBIs for. It would help with the performance loss of the early return when we match #: instead.

Yeah it's a good idea, though we would need to register the rewriter much later so we can have access to the list of gems which means no signatures for Tapioca itself. 

If we want to go this route I think it would mean exporting the RBI for Tapioca itself with `spoom srb sigs export` like we do for `rbi` and `spoom`. 

This would also solve the issue of the signatures I had to revert to `sig { ...checked(:never) }`.

lib/tapioca/bundler_ext/auto_require_hook.rb
```diff
@@ -37,7 +37,7 @@ def override_require_false(exclude:, &blk)
         end
       end
 
-      #: -> untyped
+      sig { returns(T.untyped).checked(:never) }
```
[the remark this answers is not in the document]
KaanOzkan: Thanks, I missed the `checked(:never)` calls. I think this is fine for now, we can look at speeding it up later. 

Locally I reverted some of these sigs to comments but I don't get any test failures. I'm wary that we'll forget to use sigs in this case and only encounter the issue after we release/merge. Is there a way we can run into this failure in CI and create a check?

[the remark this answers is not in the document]
KaanOzkan: Oh yeah. I tested it with RBI generation which strangely succeeds but CI fails. Thanks.

vinistock (commented):
The approach we're taking here is to do a full rewriter pass to translate RBS -> RBI and then letting Tapioca do its thing with no functional changes.

Would it be possible to change Tapioca to handle the RBS comments directly instead? For example, when we find a method, we could search its comments, find an RBS signature and then build a signature entry for it.

If this suggestion is feasible, I _think_ (without actually benchmarking anything) we may be able to achieve better performance on top of avoiding the `ruby-next` dependency.

What do you think? Am I missing context?

lib/tapioca/rbs/rewriter.rb
```diff
@@ -0,0 +1,34 @@
...
+
+        context.track!(self)
+        Spoom::Sorbet::Sigs.rbs_to_rbi(source)
+      end
+    end
+  end
+end
+
+RubyNext::Language.include_patterns.clear
+RubyNext::Language.include_patterns << "**/*.rb"
```
vinistock, line 33: Why do these need to be cleared first? Does it include something other than Ruby files by default?

We could do

```suggestion
RubyNext::Language.include_patterns = ["**/*.rb"]
```

or just append if it's already empty.

lib/tapioca/rbs/rewriter.rb
```diff
@@ -0,0 +1,34 @@
...
+
+# We need to include `T::Sig` very early to make sure that the `sig` method is available since gems using RBS comments
+# are unlikely to include `T::Sig` in their own classes.
+Module.include(T::Sig)
+
+module Tapioca
+  module RBS
+    # Transpiles RBS comments back into Sorbet's `sig{}` blocks
+    class Rewriter < RubyNext::Language::Rewriters::Text
+      NAME = "rbs_rewriter"
```
vinistock, line 18: Out of curiosity, is this used by `ruby-next`?

vinistock (approved):
Thanks for explaining the other considerations. I think using the indexer to detect the RBS signatures would be nice, but it seems like it would be a much more involved change that expands simply adopting it.

I would rather us move forward with solving the problem and then iterate on improving the performance / designing a longer term solution

KaanOzkan (approved):

lib/tapioca/rbs/rewriter.rb
paracycle: Yes, the patterns list does include other patterns by default, but yeah, agreed, we could just assign the array instead of clearing and appending.

Morriar: The writer is [not exposed](https://github.com/ruby-next/ruby-next/blob/da255b8896d3378aee844508d0e8d37c3c9980d3/lib/ruby-next/language.rb#L69).

About the clear, we _could_ keep [the default](https://github.com/ruby-next/ruby-next/blob/da255b8896d3378aee844508d0e8d37c3c9980d3/lib/uby-next.rb#L12), I'm just not taking any change in case it changes.

vinistock (approved):
Nice. I like the simplification of just using require-hooks

Morriar:
I had to add support for Bootsnap or files loaded from cache won't trigger the source transformation.

PR #2240, opened 2025-03-19
Translate all `T.let` to RBS inline comments supported by Sorbet

Morriar:
Using `spoom srb assertions translate` all assertions like:

```rb
x = T.let(foo, String)
```

can be translated to:

```rb
x = foo #: String
```

and understood by Sorbet with the option `--enable-experimental-rbs-assertions`.


paracycle (approved):
👍

PR #2243, opened 2025-03-27
Add `from` and `to` kwargs types to generated `saved_change_to_attribute?` and `will_save_change_to_attribute?` methods

dewski:
### Motivation

This PR updates the built-in `Tapioca::Dsl::Compilers::ActiveRecordColumns` compiler to add types for the `from` and `to` optional kwargs in the generated `saved_change_to_attribute?` and `will_save_change_to_attribute?` methods for each attribute in an ActiveRecord model.

### Implementation

There was precedence for supporting the `from` and `to` kwargs in `ActiveModel::Dirty` attributes in the generated `attribute_changed?` method:

https://github.com/Shopify/tapioca/blob/2144c9c669d344489b2ebac2420951709e49023b/lib/tapioca/dsl/compilers/active_record_columns.rb#L283-L292

`ActiveRecord::AttributeMethods::Dirty` generates the [`saved_change_to_attribute?` and `will_save_change_to_attribute?`](https://github.com/rails/rails/blob/621aa49cefce3313fbc672e2f3681103da686a0e/activerecord/lib/active_record/attribute_methods/dirty.rb#L52-L59) methods the same way `ActiveModel::Dirty` generates [`attribute_changed?`](https://github.com/rails/rails/blob/621aa49cefce3313fbc672e2f3681103da686a0e/activemodel/lib/active_model/dirty.rb#L241) methods.

Support for the `from` and `to` kwargs was added in [Rails v7.0](https://github.com/rails/rails/blob/v7.0.0/activerecord/lib/active_record/attribute_methods/dirty.rb#L20-L27).

### Tests

Tests were added showing the `from` and `to` types respect the nullability of the column from the database.

### DIFF
--- a/spec/tapioca/dsl/compilers/active_record_columns_spec.rb
+++ b/spec/tapioca/dsl/compilers/active_record_columns_spec.rb
@@ -103,7 +103,7 @@ module Tapioca
                       sig { returns(T.nilable([::Integer, ::Integer])) }
                       def id_change_to_be_saved; end
 
-                      sig { params(from: ::Integer, to: ::Integer).returns(T::Boolean) }
+                      sig { params(from: T.untyped, to: T.untyped).returns(T::Boolean) }
                       def id_changed?(from: T.unsafe(nil), to: T.unsafe(nil)); end
 
                       sig { returns(T.nilable(::Integer)) }
@@ -112,7 +112,7 @@ module Tapioca
                       sig { returns(T.nilable([::Integer, ::Integer])) }
                       def id_previous_change; end
 
-                      sig { params(from: ::Integer, to: ::Integer).returns(T::Boolean) }
+                      sig { params(from: T.untyped, to: T.untyped).returns(T::Boolean) }
                       def id_previously_changed?(from: T.unsafe(nil), to: T.unsafe(nil)); end
 
                       sig { returns(T.nilable(::Integer)) }
@@ -143,7 +143,7 @@ module Tapioca
                       sig { returns(T.nilable([::Integer, ::Integer])) }
                       def id_value_change_to_be_saved; end
 
-                      sig { params(from: ::Integer, to: ::Integer).returns(T::Boolean) }
+                      sig { params(from: T.untyped, to: T.untyped).returns(T::Boolean) }
                       def id_value_changed?(from: T.unsafe(nil), to: T.unsafe(nil)); end
 
                       sig { returns(T.nilable(::Integer)) }
@@ -152,7 +152,7 @@ module Tapioca
                       sig { returns(T.nilable([::Integer, ::Integer])) }
                       def id_value_previous_change; end
 
-                      sig { params(from: ::Integer, to: ::Integer).returns(T::Boolean) }
+                      sig { params(from: T.untyped, to: T.untyped).returns(T::Boolean) }
                       def id_value_previously_changed?(from: T.unsafe(nil), to: T.unsafe(nil)); end
 
                       sig { returns(T.nilable(::Integer)) }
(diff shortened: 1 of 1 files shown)


dewski:
I have signed the CLA!

KaanOzkan (commented):
Thanks. There is a missing expectation update caught in CI. I'm not sure why "Rails current" CI is passing given it should be running on Rails 7.1

lib/tapioca/dsl/compilers/active_record_columns.rb
```diff
@@ -264,12 +264,20 @@ def add_methods_for_attribute(klass, attribute_name, column_name = attribute_nam
             "saved_change_to_#{attribute_name}?",
             methods_to_add,
             return_type: "T::Boolean",
+            parameters: [
+              create_kw_opt_param("from", type: setter_type, default: "T.unsafe(nil)"),
+              create_kw_opt_param("to", type: setter_type, default: "T.unsafe(nil)"),
```
paracycle, line 269: I don't think there is any guarantee that `from` or `to` will have the `setter_type`. As a matter of fact, even [their default value](https://github.com/rails/rails/blob/main/activemodel/lib/active_model/attribute_mutation_tracker.rb#L44-L48) is a [sentinel object](https://github.com/rails/rails/blob/main/activemodel/lib/active_model/attribute_mutation_tracker.rb#L8), and they are also [processed through `type_cast`](https://github.com/rails/rails/blob/main/activemodel/lib/active_model/attribute_mutation_tracker.rb#L46), so I think they can be passed as any arbitrary type.

```suggestion
              create_kw_opt_param("from", type: "T.untyped", default: "T.unsafe(nil)"),
              create_kw_opt_param("to", type: "T.untyped", default: "T.unsafe(nil)"),
```

lib/tapioca/dsl/compilers/active_record_columns.rb
```diff
@@ -264,12 +264,20 @@ def add_methods_for_attribute(klass, attribute_name, column_name = attribute_nam
...
+            ],
           )
           add_method(
             klass,
             "will_save_change_to_#{attribute_name}?",
             methods_to_add,
             return_type: "T::Boolean",
+            parameters: [
+              create_kw_opt_param("from", type: setter_type, default: "T.unsafe(nil)"),
+              create_kw_opt_param("to", type: setter_type, default: "T.unsafe(nil)"),
```
paracycle, line 279: Same as above:
```suggestion
              create_kw_opt_param("from", type: "T.untyped", default: "T.unsafe(nil)"),
              create_kw_opt_param("to", type: "T.untyped", default: "T.unsafe(nil)"),
```

paracycle (changes requested):

spec/tapioca/dsl/compilers/active_record_columns_spec.rb
```diff
@@ -182,24 +182,24 @@ def restore_id_value!; end
                       sig { returns(T.nilable([::Integer, ::Integer])) }
                       def saved_change_to_id; end
 
-                      sig { returns(T::Boolean) }
-                      def saved_change_to_id?; end
+                      sig { params(from: ::Integer, to: ::Integer).returns(T::Boolean) }
```
paracycle, line 185: Shouldn't we be changing these expectations to use `T.untyped` as well?

dewski: Yep! Was in the process of pushing those changes up in e1197875112762b4100db126182d8978d2d1f72d and c29aa9ba028db33b54fd4f1bc95ec2ff945422b8.

paracycle (approved):
Thanks!

PR #2245, opened 2025-03-28
Fix check shims to allow methods with different parameters

Morriar:
### Motivation

@vinistock got stuck the other day trying to override an incorrect payload definition:

```rb
# in payload

class Foo
  def foo(a, b); end
end
```

```rb
# in shim.rbi

class Foo
  def foo(*args); end
end
```

this shouldn't raise a duplicate error.


lib/tapioca/helpers/rbi_files_helper.rb
```diff
@@ -152,49 +152,78 @@ def parse_and_index_files(index, files, number_of_workers:)
...
-        next true unless parent_name
+      # First let's look into scopes (classes, modules, sclass) for duplicates
+      shims_or_todos_scopes = shims_or_todos.grep(RBI::Scope)
+      if shims_or_todos_scopes.any?
+        # Extract the empty scopes from the shims or todos
+        # We do not care about non-empty scopes because they hold definitions that we will check against Tapioca's
+        # generated RBI files in another iteration.
+        shims_or_todos_empty_scopes = shims_or_todos_scopes.select(&:empty?)
+
+        shims_or_todos_empty_scopes.each do |scope|
```
amomchilov, line 176: Can this also use an `any?`?

lib/tapioca/helpers/rbi_files_helper.rb
```diff
@@ -152,49 +152,78 @@ def parse_and_index_files(index, files, number_of_workers:)
...
-      shims_or_todos_empty_scopes.select! do |scope|
-        # Empty modules are always duplicates
-        next true unless scope.is_a?(RBI::Class)
-
-        # Empty classes without parents are also duplicates
-        parent_name = scope.superclass_name
-        next true unless parent_name
+      # First let's look into scopes (classes, modules, sclass) for duplicates
+      shims_or_todos_scopes = shims_or_todos.grep(RBI::Scope)
+      if shims_or_todos_scopes.any?
```
amomchilov, line 170: Could we extract helper functions for these, so that this methods can return something like:

```
    has_duplicate_scopes(shims_or_todos)
      || has_duplicate_mixins(shims_or_todos)
      || has_duplicate_methods(shims_or_todos)
```

lib/tapioca/helpers/rbi_files_helper.rb
```diff
@@ -152,49 +152,78 @@ def parse_and_index_files(index, files, number_of_workers:)
...
+            # If the node has parameters, we compare them against all the other methods
+            # If at least one of them has the same parameters, it's a duplicate
+            return true if other_nodes.grep(RBI::Method).any? { |other| shim_or_todo_prop.params == other.params }
+          end
+
+          # We compare the shim or todo prop with all the other props of the same name
+          other_nodes.each do |node|
+            # Another prop has the same sig, we have a duplicate
+            return true if shim_or_todo_prop.sigs.any? { |sig| node.sigs.include?(sig) }
+          end
```
amomchilov, line 222: This looks `O(n^2)`, could that be a risk for long files?

lib/tapioca/helpers/rbi_files_helper.rb
Morriar: Good idea, done 👍

vinistock (approved):

PR #2271, opened 2025-05-12
use primary key type for ActiveRecord::Calculations#ids return type

ebarajas:
### Motivation
This PR updates the built-in Tapioca::Dsl::Compilers::ActiveRecordRelations compiler to add a narrowed type for the return value of `#ids`. Before this was simply set to `Array`, so each item in the array was resolving to `T.untyped`. 

### Implementation
I duplicated some existing work done for the `#find` method to determine the type of the primary key column. 

### Tests
I've updated the existing test to assert the RBI is generated as expected.


ebarajas:
I have signed the CLA!

KaanOzkan (approved):
I can't trigger CI from the GitHub UI, hooks might have ran into an issue. Can you rebase and push?

PR #2283, opened 2025-05-21
Use Zeitwerk::Loader.all_dirs to load all directories

### LINKED ISSUE #2281: Latest Tapioca (0.16.11) is incompatible with latest Zeitwerk (2.7.3)
As of Zeitwerk 2.7.3 is it no longer possible to call #flat_map on Zeitwerk::Registry.loaders as is done [here](https://github.com/Shopify/tapioca/blob/3008dfee2bf44483ff4fecfcd2410cf3a4062bde/lib/tapioca/loaders/loader.rb#L116). 

This is the 2.7.3 Zeitwerk code for Registry loaders: https://github.com/fxn/zeitwerk/blob/v2.7.3/lib/zeitwerk/registry/loaders.rb

st0012:
### Motivation

`Zeitwerk::Registry.loaders` is marked as private and is more unstable and just [changed the return type in 2.7.3](https://github.com/fxn/zeitwerk/commit/bd846963bbe2d81852f6f54febc01d38fde890aa#diff-ef2e28f094ca3837114a7116ca0e5a1e9a3591694d2069678d6af2b21894ec17R371).
We should use the public API `Zeitwerk::Loader.all_dirs` instead.

This is required for Tapioca to work with projects using Zeitwerk 2.7.3+

Closes #2281


st0012:
@fxn Thanks for the suggestion! I've opened https://github.com/Shopify/tapioca/pull/2285 for it 😄

PR #2287, opened 2025-05-23
Load the runtime patches before Spoom

Morriar:
### Motivation

We need these to be loaded before, if not generating the RBI for `Spoom` won't contain generics.


paracycle (approved):

KaanOzkan (approved):

st0012:
The build failure is unrelated to change (caused by rubygem's rdoc plugin) so I'm merging this and cut a release to unblock upgrade.

PR #2304, opened 2025-06-02
Migrate Ruby LSP add-on to RBS comments

vinistock:
### Motivation

Migrate all of the Tapioca add-on to RBS comments since we're going to remove the Ruby LSP's dependency on the `sorbet-runtime` gem. This will make sure that Tapioca is ready for the breaking change ahead of time.

### Implementation

Just migrated everything to the RBS comment syntax. I also made some minor style modifications.

### Tests

Current tests should cover it.

### DIFF
--- a/lib/ruby_lsp/tapioca/addon.rb
+++ b/lib/ruby_lsp/tapioca/addon.rb
@@ -18,8 +18,6 @@ require "ruby_lsp/tapioca/run_gem_rbi_check"
 module RubyLsp
   module Tapioca
     class Addon < ::RubyLsp::Addon
-      extend T::Sig
-
       #: -> void
       def initialize
         super
@@ -44,7 +42,7 @@ module RubyLsp
           # Get a handle to the Rails add-on's runtime client. The call to `rails_runner_client` will block this thread
           # until the server has finished booting, but it will not block the main LSP. This has to happen inside of a
           # thread
-          addon = T.cast(::RubyLsp::Addon.get("Ruby LSP Rails", ">= 0.4.0", "< 0.5"), ::RubyLsp::Rails::Addon)
+          addon = ::RubyLsp::Addon.get("Ruby LSP Rails", ">= 0.4.0", "< 0.5") #: as ::RubyLsp::Rails::Addon
           @rails_runner_client = addon.rails_runner_client
           @outgoing_queue << Notification.window_log_message("Activating Tapioca add-on v#{version}")
           @rails_runner_client.register_server_addon(File.expand_path("server_addon.rb", __dir__))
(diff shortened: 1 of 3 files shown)


vinistock:
* **#2304** <a href="https://app.graphite.dev/github/pr/Shopify/tapioca/2304?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/tapioca/2304?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -115,7 +113,8 @@ def workspace_did_change_watched_files(changes)
             next
           end
 
-          entries = T.must(@index).entries_for(change[:uri])
+          entries = @index #: as !nil
+            .entries_for(change[:uri])
```
Morriar: Would this work instead?

```suggestion
          entries = @index&.entries_for(change[:uri])
```

Morriar: Or if the index is expected to be not nil (since we `T.must` it) I'd do `index = @index #: as !nil` at the beginning of the method.

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -207,16 +208,18 @@ def file_updated?(change, path)
 
       #: -> void
       def run_gem_rbi_check
-        gem_rbi_check = RunGemRbiCheck.new(T.must(@global_state).workspace_path)
+        gem_rbi_check = RunGemRbiCheck.new(
+          @global_state #: as !nil
```
Morriar: Let's extract it as a local var like you did earlier with `queue`

lib/ruby_lsp/tapioca/run_gem_rbi_check.rb
```diff
@@ -81,15 +79,16 @@ def generate_gem_rbis
...
-            chdir: project_path,
-          )
+          stdout, stderr, status = Open3 #: as untyped
+            .capture3(
+              "bundle",
+              "exec",
+              "tapioca",
+              "gem",
+              "--lsp_addon",
+              *gems,
```
Morriar, line 89: What do you think of ignoring `7019` error in this project as well?

lib/ruby_lsp/tapioca/addon.rb
vinistock: I went with the local variable 👍

lib/ruby_lsp/tapioca/run_gem_rbi_check.rb
vinistock: I think we should, but not as part of this PR. We haven't migrated the entire of Tapioca to use RBS comments - this is just for the add-on.

As part of the full migration, we should then turn that error off.

PR #2310, opened 2025-06-04
Skip RBS comments in YardDoc listener

KaanOzkan:
### Motivation

YardDoc listener thinks they are method comments and adds it to the RBI when they are already translated to sigs.

### Implementation

I'm trying to naively identify RBS comments through `: ` and `| `

### DIFF
--- a/sorbet/rbi/gems/activesupport@7.1.5.1.rbi
+++ b/sorbet/rbi/gems/activesupport@7.1.5.1.rbi
@@ -18072,8 +18072,6 @@ class Integer < ::Numeric
   def years; end
 end
 
-Integer::GMP_VERSION = T.let(T.unsafe(nil), String)
-
 # source://activesupport//lib/active_support/core_ext/kernel/reporting.rb#3
 module Kernel
   # class_eval on an object acts like +singleton_class.class_eval+.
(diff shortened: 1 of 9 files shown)


lib/tapioca/gem/listeners/yard_doc.rb
```diff
@@ -17,6 +17,7 @@ class YardDoc < Base
           "shareable_constant_value:",
           "rubocop:",
           "@requires_ancestor:",
+          ": ",
```
KaanOzkan: This is used with an `include?` call down below. I want to make sure I'm not ignoring more comments than necessary so added the space. But it's not perfect.

Morriar: I think prefer a fully fledge `rbs_comment?` method we call from `documentation_comments`.

Or at least a comment explaining what is this `: `:
```suggestion
          ": ", # ignore RBS signature comments
```

st0012: Should `| ` be included here too?

KaanOzkan: Good point. Done.

st0012 (approved):

Issue #2087, opened 2024-11-22
Add support for `cursor:` when using `in_batches` on an ActiveRecord PrivateRelation


SamuelDev:
In 8.0 rails added support for a [custom cursor](https://github.com/rails/rails/commit/fd52a754f2c6e234ee46834c63cfde02a368d5c6) when calling `in_batches` on a relation. RBI files generated by tapioca do not currently generate this field.

Rails version: 8.1.0
tapioca version: 0.16.3

SamuelDev:
It is a column name or array of column names. Symbols, possibly strings.

amomchilov:
`T.any(String, Symbol)` might be a good first approximation. If other types are possible, we can widen the type in the future. It looks like it can be either a single value, or an array of values (it's passed into `Array`), so perhaps it should be:

```ruby
T.any(String, Symbol, T::Array[T.any(String, Symbol)])
```

I won't be able to investigate this myself any time soon, but if you confirm that it works and open a PR, I'd be happy to review and merge

PR #2315, opened 2025-06-12
Add support for cursor parameter on ActiveRecord's batch methods in Rails 8.0

lavoiesl:
### Motivation
Support the parameter added in https://github.com/rails/rails/pull/52384

### Implementation
I started by refactoring the code to be DRYer because it was starting to be pretty long.

### Tests
Updated the tests to check for 8.0

🎩 in a different repo that it generates updated signatures and calling `find_each(cursor:)` that it passes type checking.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_relations.rb
+++ b/lib/tapioca/dsl/compilers/active_record_relations.rb
@@ -217,6 +217,16 @@ module Tapioca
           []
         end #: Array[Symbol]
         BATCHES_METHODS = ActiveRecord::Batches.instance_methods(false) #: Array[Symbol]
+        BATCHES_METHODS_PARAMETERS = {
+          start: ["T.untyped", "nil"],
+          finish: ["T.untyped", "nil"],
+          load: ["T.untyped", "false"],
+          batch_size: ["Integer", "1000"],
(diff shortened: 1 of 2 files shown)


lib/tapioca/dsl/compilers/active_record_relations.rb
```diff
@@ -217,6 +217,17 @@ def gather_constants
...
         BATCHES_METHODS = ActiveRecord::Batches.instance_methods(false) #: Array[Symbol]
+        BATCHES_METHODS_PARAMETERS = {
+          start: ["T.untyped", "nil"],
+          finish: ["T.untyped", "nil"],
+          load: ["T.untyped", "false"],
+          batch_size: ["Integer", "1000"],
+          of: ["Integer", "1000"],
+          error_on_ignore: ["T.untyped", "nil"],
+          order: ["Symbol", ":asc"],
+          cursor: ["T.untyped", "primary_key"],
```
lavoiesl, line 228: There is debate at https://github.com/Shopify/tapioca/issues/2087#issuecomment-2499097226 for the right type to use.

I went for untyped for simplicity, but I'm happy to entertain alternatives.

PR #2317, opened 2025-06-12
Do not load DSL compilers before extensions in add-on mode

KaanOzkan:
Extensions may override DSL compilers and should be loaded before so that abstract methods such as `gather_constants` and `decorate` of the original compilers take precedence

These `require` calls were loading the compilers before the extensions which isn't the case for using tapioca from the CLI.

Without the requires the code should still work since `addon.rb` will dispatch `load_compilers_and_extensions` request before `route_dsl` or `fixtures_dsl`.

### DIFF
--- a/lib/ruby_lsp/tapioca/server_addon.rb
+++ b/lib/ruby_lsp/tapioca/server_addon.rb
@@ -2,8 +2,6 @@
 # frozen_string_literal: true
 
 require "tapioca/internal"
-require "tapioca/dsl/compilers/url_helpers"
-require "tapioca/dsl/compilers/active_record_fixtures"
 
 module RubyLsp
   module Tapioca


paracycle:
The changes look good to me, but I don't understand this statement:

> Extensions may override DSL compilers and should be loaded before so that abstract methods such as `gather_constants` and `decorate` of the original compilers take precedence

Which extensions override DSL compilers? They certainly need to be loaded before compilers since they are meant to collect information that compilers will ultimately use, but they really should not be doing any overriding of anything.

paracycle (approved):

lib/ruby_lsp/tapioca/server_addon.rb
```diff
@@ -2,8 +2,6 @@
 # frozen_string_literal: true
 
 require "tapioca/internal"
-require "tapioca/dsl/compilers/url_helpers"
```
Morriar, line 5: Looks like these requires were specifically added to fix something https://github.com/Shopify/tapioca/pull/2165. Are we not reintroducing the same issue?

KaanOzkan: I don't think so, we still have the main code that runs `gather_constants` for both compilers. I think `require` was added since we're used to doing it before referencing a constant but it's not necessary (unless there's a change in how `activate` works). Goal is to have the order:
1. Load this file (`server_addon.rb`)
2. Run `activate` in `addon_rb` which runs the `load_compilers_and_extensions` request
3. `load_compilers_and_extensions` loads all tapioca compilers
4. Routes or Fixture edits reference the compilers that are already loaded.

PR #2320, opened 2025-06-13
Translate abstract methods to RBS comments

Morriar:
Result of running `spoom srb sigs translate`


lib/tapioca/commands/command.rb
```diff
@@ -31,8 +31,9 @@ def run
 
       private
 
-      sig { abstract.void }
-      def execute; end
+      # @abstract
+      #: -> void
+      def execute = raise("Abstract method called")
```
paracycle: Can we make these raise `NotImplementedError`? The exception type was created to mean something different, but most of the community has started using it to signal that a method that is expected to run has no implementation yet.
```suggestion
      def execute = raise NotImplementedError, "Abstract method called"
```

Issue #2208, opened 2025-02-20
Create Tapioca exceptions rather than relying on `Thor::Error`


KaanOzkan:
Reliance of `Thor::Error` and [`exit_on_failure?`](https://github.com/rails/thor/blob/3178667e1727504bf4fb693bf4ac74a5ca6c691e/lib/thor/base.rb#L587) is growing and is being used by Tapioca internals. Instead we can use our own exception and have more control over how to handle it.

drewhoffer:
Hey! I've opened a PR for this guy and would appreciate any feedback anyone can provide (I'm new to working on developer tools). 

One thing I ran into while getting started/into to this codebase was:
When running the tests locally, I had some failing tests due to the `Google::Protobuf::DescriptorPool`. It seems the `build` method isn't available in newer versions of the `protobuf` gem, When i manually installed an older version `gem "google-protobuf", "~> 3.19"` all tests were then green.

If there is anything I could do (open an issue/open a pr) please let me know I'd be happy to help! Might be worth noting somewhere in the contributor docs if others run into the same issue.

Output of tests was:

```

Tapioca
  Dsl
    Compilers
      ProtobufSpec
        Tapioca
          Dsl
            Compilers
              Protobuf
                gather_constants
                  it gathers only classes with Protobuf Module                   ERROR (0.16s)
Minitest::UnexpectedError:         NoMethodError: undefined method `build' for an instance of Google::Protobuf::DescriptorPool
            /tmp/.../lib/cart_pb.rb:6:in `<main>'
            .../lib/ruby/.../bundled_gems.rb:74:in `require'
            .../lib/ruby/.../bundled_gems.rb:74:in `block (2 levels) in replace_require'
            spec/tapioca/dsl/compilers/protobuf_spec.rb:601:in `block (2 levels) in add_proto_file'
            lib/tapioca.rb:20:in `block in silence_warnings'
            .../rubygems/user_interaction.rb:46:in `use_ui'
            lib/tapioca.rb:19:in `silence_warnings'
            spec/tapioca/dsl/compilers/protobuf_spec.rb:601:in `block in add_proto_file'
            <internal:kernel>:90:in `tap'
            spec/tapioca/dsl/compilers/protobuf_spec.rb:595:in `add_proto_file'

                  it gathers no constants if there are no Google::Protobuf classesERROR (0.14s)
Minitest::UnexpectedError:         NoMethodError: undefined method `build' for an instance of Google::Protobuf::DescriptorPool
[...]

PR #2322, opened 2025-06-14
Replace Instances of Thor::Error with Tapioca::Error

drewhoffer:
### Motivation

This PR addresses #2208 which suggested replacing `Thor::Error` with a custom `Tapioca::Error`. 

### Implementation

- I feel my implementation may be naive in thinking that I would simply need to rename `Thor::Error`s to `Tapioca::Error` and am wondering if instead `exit_on_failure?` should be replaced or defined elsewhere.
  - I’m still getting familiar with Tapioca and Thor and I’m not yet confident that my approach fully captures what’s intended around the replacement of `exit_on_failure?`. I’d really appreciate any feedback or pointers on what is expected here.

- Replaced usage of `Thor::Error` with `Tapioca::Error` (which was already defined) across codebase.
- In `CLI` mode we should be exiting (used `exit_on_failure?`)
- In `addon_mode` we should be silently terminating RBI Generation (again using `exit_on_failure?`)
- Updated typo in comment going through some of the protobuf spec

### Tests

- I haven’t added any tests yet. If tests are expected for this change, I’d be happy to add them. I would just need a hand with telling me where they should go. I'm still getting familiar with the codebase and would appreciate any help.

### DIFF
--- a/exe/tapioca
+++ b/exe/tapioca
@@ -27,4 +27,9 @@ end
 
 require_relative "../lib/tapioca/internal"
 
-Tapioca::Cli.start(ARGV)
+begin
+  Tapioca::Cli.start(ARGV)
+rescue Tapioca::Error => e
+  warn(e.message)
+  exit(1)
+end
--- a/lib/tapioca/cli.rb
+++ b/lib/tapioca/cli.rb
@@ -353,7 +353,7 @@ module Tapioca
       default: {}
     def annotations
       if !options[:netrc] && options[:netrc_file]
-        raise Thor::Error, set_color("Options `--no-netrc` and `--netrc-file` can't be used together", :bold, :red)
+        raise Tapioca::Error, set_color("Options `--no-netrc` and `--netrc-file` can't be used together", :bold, :red)
       end
 
       command = Commands::Annotations.new(
--- a/lib/tapioca/commands/abstract_dsl.rb
+++ b/lib/tapioca/commands/abstract_dsl.rb
@@ -160,7 +160,7 @@ module Tapioca
             remove_file(filename) if File.file?(filename)
           end
 
-          raise Thor::Error, ""
+          raise Tapioca::Error, ""
         end
 
         processable_constants
@@ -295,7 +295,7 @@ module Tapioca
             build_error_for_files(cause, diff_for_cause.map(&:first))
           end.join("\n")
 
-          raise Thor::Error, <<~ERROR
+          raise Tapioca::Error, <<~ERROR
             #{set_color("RBI files are out-of-date. In your development environment, please run:", :green)}
               #{set_color("`#{default_command(command)}`", :green, :bold)}
             #{set_color("Once it is complete, be sure to commit and push any changes", :green)}
--- a/lib/tapioca/commands/abstract_gem.rb
+++ b/lib/tapioca/commands/abstract_gem.rb
@@ -201,7 +201,7 @@ module Tapioca
             build_error_for_files(cause, diff_for_cause.map(&:first))
           end.join("\n")
 
-          raise Thor::Error, <<~ERROR
+          raise Tapioca::Error, <<~ERROR
             #{set_color("RBI files are out-of-date. In your development environment, please run:", :green)}
               #{set_color("`#{default_command(command)}`", :green, :bold)}
             #{set_color("Once it is complete, be sure to commit and push any changes", :green)}
--- a/lib/tapioca/commands/annotations.rb
+++ b/lib/tapioca/commands/annotations.rb
@@ -85,7 +85,7 @@ module Tapioca
         end
 
         if indexes.empty?
-          raise Thor::Error, set_color("Can't fetch annotations without sources (no index fetched)", :bold, :red)
+          raise Tapioca::Error, set_color("Can't fetch annotations without sources (no index fetched)", :bold, :red)
         end
 
         indexes
--- a/lib/tapioca/commands/check_shims.rb
+++ b/lib/tapioca/commands/check_shims.rb
@@ -49,7 +49,7 @@ module Tapioca
             result = sorbet("--no-config --print=payload-sources:#{payload_path}")
 
             unless result.status
-              raise Thor::Error, <<~ERROR
+              raise Tapioca::Error, <<~ERROR
                 "Sorbet failed to dump payload"
                 #{result.err}
               ERROR
@@ -87,7 +87,7 @@ module Tapioca
             "\nPlease remove the duplicated definitions from #{@shim_rbi_dir} and #{@todo_rbi_file}", :red
           )
 
-          raise Thor::Error, messages.join("\n")
+          raise Tapioca::Error, messages.join("\n")
         end
 
         say("\nNo duplicates found in shim RBIs", :green)
--- a/lib/tapioca/commands/gem_generate.rb
+++ b/lib/tapioca/commands/gem_generate.rb
@@ -58,7 +58,7 @@ module Tapioca
           if gem.nil?
             next if @lsp_addon
 
-            raise Thor::Error, set_color("Error: Cannot find gem '#{gem_name}'", :red)
+            raise Tapioca::Error, set_color("Error: Cannot find gem '#{gem_name}'", :red)
           end
 
           gems.concat(gem_dependencies(gem)) if @include_dependencies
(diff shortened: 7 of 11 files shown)


exe/tapioca
```diff
@@ -27,4 +27,9 @@ end
 
 require_relative "../lib/tapioca/internal"
 
-Tapioca::Cli.start(ARGV)
+begin
+  Tapioca::Cli.start(ARGV)
+rescue Tapioca::Error => e
+  warn(e.message)
+  exit if Tapioca::Cli.exit_on_failure?
```
KaanOzkan: `exit_on_failure?` is only needed for Thor errors since those are rescued by Thor. That method gives Thor users a way to control if process should exit or silently fail when there is an exception https://github.com/rails/thor/blob/3178667e1727504bf4fb693bf4ac74a5ca6c691e/lib/thor/base.rb#L587



As I mentioned in https://github.com/Shopify/tapioca/issues/2208#issuecomment-2970483363, we can remove the need for `exit_on_failure?` completely. CLI mode (here) should exit every time and the Add-on shouldn't exit at all.

As for testing we can do it manually. You can point to your tapioca branch in a different project, add an early `raise Tapioca::Error` and see how the commands `bin/tapioca gem` or `bundle exec tapioca gem` react. Similarly for add-on mode you'd need a [Ruby LSP](https://github.com/shopify/ruby-lsp) setup, launch editor, ensure tapioca is booted and check that when you hit your debug `raise` call we don't exit the process. For the add-on mode getting logs can be tough, you can write outputs to a file, `File.write("logs.txt", "here", mode: "a")`

written-fresh: Ok great. I took a quick video showing the CLI exiting just to confirm behaviour. 

I tried to get the `server_addon.rb` to blow up to no avail. Unsure if that is really GOOD or if I am missing something.

I was able to execute the [dsl method](https://github.com/Shopify/tapioca/blob/0b4462fa4dc8c584044fbc479489709b8a7cc4e2/lib/ruby_lsp/tapioca/server_addon.rb#L60) by updating a rails model but despite a couple `raise `:boom:` booms` immediately, I was unable to get vscode to show me any indication that the process died (either through the vscode popup or running `watch -n 1 'ps aux | grep ruby-lsp | grep -v grep'`. 

Am I correct in thinking it is related to [ruby lsp rails](https://github.com/Shopify/ruby-lsp-rails/blob/cb9556d454c8bb20a1a73a99c8deb8788a520007/lib/ruby_lsp/ruby_lsp_rails/server.rb#L64)? OR the dsl being a forked process?. I took a look at the output window for the LSP and despite NOT having any rescues (and throwing some exceptions) the main LSP continued and further invocations of the dsl method continued to work (or continued to throw my exception rather)


[Tapioca error CLI](https://github.com/user-attachments/assets/057741ef-b952-41dc-99bb-425311b8c907)

KaanOzkan: > Sorry these may be entry level questions but there is a LOT of reading I've been diving into around all of this tooling 😵.

No worries happy to help.

I'd also expect the error to raise. Make sure it's being hit by using File.write logs. Ruby LSP Rails might be rescuing it but it should be logged in the output tab https://github.com/Shopify/ruby-lsp-rails/blob/cb9556d454c8bb20a1a73a99c8deb8788a520007/lib/ruby_lsp/ruby_lsp_rails/server.rb#L86.

written-fresh: Ok just to confirm. It IS ok that we have a stack trace in the output of the LSP? 

I can make that happen by simply letting the error bubble up. The issue is if we want to have it nicely formatted I don't think we can do that with the current fork implementation. 

I have attached a video of the current behaviour with NO rescue block around the [Cli](https://github.com/Shopify/tapioca/blob/ac1d7a1c9d43f9065fc415c8f89ab20418d93b3a/lib/ruby_lsp/tapioca/server_addon.rb#L72)

As you can see we are getting a stack trace with the `Tapioca::Error`, however tapioca continues to work (with repeated DSL requests), I THINK this is because the forked process is whats blowing up not the main Tapioca process.
[server_addon_exceptions.webm](https://github.com/user-attachments/assets/282eee1b-0921-4266-b314-ca0924c37dca)

With File.writes I placed them around the [`with_notification_wrapper`](https://github.com/Shopify/tapioca/blob/ac1d7a1c9d43f9065fc415c8f89ab20418d93b3a/lib/ruby_lsp/tapioca/server_addon.rb#L54):

```rb

      def with_notification_wrapper(request_name, title, &block)
        with_progress(request_name, title) do
          File.write("addon_logs.txt", "Calling error handler\n", mode: "a")
          with_notification_error_handling(request_name, &block)
          File.write("addon_logs.txt", "We Lived!\n", mode: "a")
        end
      end
```

And after running I can confirm that `"We Lived!"` was hit.

KaanOzkan: > Ok just to confirm. It IS ok that we have a stack trace in the output of the LSP?

Yes we rescue and log errors. As long as the stack trace is there I think it's fine and doesn't have to be a nice format.

Your `raise` is being rescued by Ruby LSP Rails. Important thing is that now we don't run into `exit(1)` after the removal of `def exit_on_failure?` override.

KaanOzkan: I think the PR looks good on a glance. Can you rebase, fix the conflict and clean up the commit history? We don't squash and merge here, I think this PR makes sense as 1 or 2 commits.

written-fresh: Ok @KaanOzkan I think I'm good to go, I appreciate the help!

KaanOzkan:
```suggestion
  exit
```

If we triggered it here we can't be in add-on mode so we don't need this check. Method should also be removed in `cli.rb`.

lib/ruby_lsp/tapioca/server_addon.rb
```diff
@@ -71,7 +71,12 @@ def dsl(constants, *args)
         arguments.push("--workers=1")
         arguments.concat(constants)
 
-        ::Tapioca::Cli.start(arguments)
+        begin
+          ::Tapioca::Cli.start(arguments)
+        rescue ::Tapioca::Error
+          # terminate RBI generation if in addon mode
+          exit if ::Tapioca::Cli.exit_on_failure?
+        end
```
KaanOzkan:
```suggestion
          ::Tapioca::Cli.start(arguments)
```

Opposite of above comment. If we're here we're in add-on mode and don't want to exit. Ruby LSP Rails will rescue any errors for us and output it to the user.

KaanOzkan: With these 2 changes we should be able to remove the `exit_on_failure?` definition in `cli.rb`.

KaanOzkan (commented):
Took a closer look and identified some comments that need to be addressed.

lib/ruby_lsp/tapioca/server_addon.rb
written-fresh: Hey sorry, I think I included one of my early commits with changes I ended up undoing in the rebase :melting_face: Will update now. Good catch!

KaanOzkan:
There are some CI failures, also I think the `Thor::Error` in `lib/tapioca/helpers/config_helper.rb` can be migrated too.

drewhoffer:
If you'd like I can comment the method and just have a note that it was left for this reason and that further investigation would need to be done regarding how to remove it as a dependency? Otherwise I think it's good to go. 

Let me know if you need anything else from my end (screenshots showing behaviour etc. to make your life easier)

Thanks again for the help/patience on this, was a blast to work on/learn from.

lib/ruby_lsp/tapioca/server_addon.rb
```diff
@@ -70,7 +70,6 @@ def dsl(constants, *args)
         arguments.concat(args)
         arguments.push("--workers=1")
         arguments.concat(constants)
-
```
KaanOzkan: I don't think this was intentional

KaanOzkan:
> If you'd like I can comment the method and just have a note that it was left for this reason and that further investigation would need to be done regarding how to remove it as a dependency?

I think it's fine, comment can get outdated, we can rely on the Thor's warning in case we decide to remove it.

Thanks for your contribution 🙂

lib/ruby_lsp/tapioca/server_addon.rb
written-fresh: I'll revert this after work today if needed, I think this got roped in when i added/remvoed the rescue block

KaanOzkan: Yeah can we remove it please, thanks. I didn't want to amend your commit.

PR #2323, opened 2025-06-16
Migrate some left over signatures

Morriar:
### Motivation

`spoom srb sigs translate` now handles signatures inside blocks.

### DIFF
--- a/lib/tapioca/commands/annotations.rb
+++ b/lib/tapioca/commands/annotations.rb
@@ -125,7 +125,7 @@ module Tapioca
         fetched_gems.keys.map(&:name).sort
       end
 
-      sig { params(repo_uris: T::Array[String], gem_info: GemInfo).returns(T::Boolean) }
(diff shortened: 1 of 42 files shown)


lib/tapioca/commands/annotations.rb
```diff
@@ -125,7 +125,7 @@ def fetch_annotations(project_gems)
         fetched_gems.keys.map(&:name).sort
       end
 
-      sig { params(repo_uris: T::Array[String], gem_info: GemInfo).returns(T::Boolean) }
+      #: (Array[String] repo_uris, GemInfo gem_info) -> bool
```
paracycle, line 128: Out of curiosity: Why was this not translated before? It doesn't seem to be inside a block.

paracycle (approved):

lib/tapioca/commands/annotations.rb
Morriar: I _think_ it's a race condition on merge, I've been migrating it and fixing it on the same day:
image

PR #2330, opened 2025-06-20
Skip gem RBI check when sorbet/rbi/gems does not exist

### LINKED ISSUE #2308: Add-on crashes if `sorbet/rbi/gems` doesn't exist
The [gem RBI check](https://github.com/Shopify/tapioca/blob/main/lib/ruby_lsp/tapioca/run_gem_rbi_check.rb) always assumes that the directory will be present, but that's not guaranteed - especially if someone is just setting up their repo with Tapioca.

If the directory is not there, loading the Tapioca add-on crashes.

alexcrocha:
Closes #2308

### Motivation

When setting up a new repo with Tapioca, the `sorbet/rbi/gems` directory doesn't exist yet, and loading the Tapioca add-on crashes. 

### Implementation

Added an early return in `run_gem_rbi_check` to check if the `sorbet/rbi/gems` directory exists before attempting any operations. This ensures the addon gracefully skips the gem RBI check when the directory hasn't been created yet.


lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -209,6 +209,10 @@ def file_updated?(change, path)
       #: -> void
       def run_gem_rbi_check
         state = @global_state #: as !nil
+        gem_rbi_directory = File.join(state.workspace_path, "sorbet", "rbi", "gems")
+
+        return unless File.exist?(gem_rbi_directory)
+
```
paracycle: A few things here:
1. We already have these well known paths as constants on the `Tapioca` object. Please see other usages of it.
2. This is the default value of the output folder for gem RBI generation, but it isn't necessarily the value that is being used by the end user. We should instead be checking the actual value in the config, etc.

paracycle (changes requested):

lib/ruby_lsp/tapioca/addon.rb
alexcrocha: Thank you for highlighting this!

I noticed we were not using the `::Tapioca::DEFAULT_GEM_DIR` constant in `addon.rb` and `run_gem_rbi_check.rb` either, so I've gone ahead and updated those.

We're now also checking the user's config file for any configured gem directory.

Please take a look and let me know if you have more feedback.

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -219,6 +223,24 @@ def run_gem_rbi_check
...
+        if File.exist?(config_file_path)
+          begin
+            config = YAML.load_file(config_file_path, fallback: {})
+            outdir = config.dig("gem", "outdir") || ::Tapioca::DEFAULT_GEM_DIR
+
+            return File.expand_path(outdir, workspace_path)
+          rescue StandardError
+            # Invalid YAML or file read error
+          end
```
alexcrocha: Is this unnecessary?

lib/ruby_lsp/tapioca/addon.rb
```diff
@@ -219,6 +223,24 @@ def run_gem_rbi_check
           queue << Notification.window_log_message(gem_rbi_check.stderr, type: Constant::MessageType::WARNING)
         end
       end
+
+      #: (String) -> String
+      def gem_rbi_directory(workspace_path)
```
paracycle, line 231: I am not sure I understand why this method is needed? The CLI command should be transparently get the correct value of the gem rbi directory as part of the CLI command `options`. That needs to be passed down here, not read the config from scratch.

alexcrocha: This is triggered by the Ruby LSP through the [activate](https://github.com/Shopify/tapioca/blob/98fb088dc945aaa699f0ff8d80a9052c228b8ef7/lib/ruby_lsp/tapioca/addon.rb#L56) method. At this point we haven't shelled out to the Tapioca gem command yet, and the config hasn't been parsed by CLI.

Before we shell out, we need to do some cleanup. When we detect deleted or lingering RBIs (through git commands) we either restore them or delete the new ones, [source](https://github.com/Shopify/tapioca/blob/98fb088dc945aaa699f0ff8d80a9052c228b8ef7/lib/ruby_lsp/tapioca/run_gem_rbi_check.rb#L109).

We can either duplicate the config processing here in the add-on, or move this cleanup to the Tapioca gem command to be triggered optionally in add-on mode.

paracycle: The gem RBI command does that kind of clean up automatically anyway, so I am not even sure what these git operations are trying to do.

I think doing this outside of the gem RBI command is asking for trouble. I don't think we should be doing any config parsing, or relying on constant values to assume anything about where RBI files live, or which ones should be cleaned up, etc.

All that logic already exists. Please tell me what's missing that all this is needed.

alexcrocha: The tapioca gem command only runs when we detect lockfile differences. However, consider this scenario:
- Developer adds/modifies/removes a gem and regenerates RBIs
- Developer reverts the Gemfile changes and runs `bundle install`
- The lockfile returns to its committed state
- LSP restarts due to lockfile change, but finds no diff against the committed version
- `tapioca gem <gems>` doesn't run, leaving orphaned RBI files or missing previously deleted ones

This is where the git operations come in: they clean up orphaned RBIs and restore deleted ones. While I understand the concern about duplicating logic, the add-on and CLI run in separate processes, so we must read the configuration independently.

We did consider the alternative of running `tapioca gem` on every activation to simplify this. However, we decided against it due to the potential negative impact on developer experience. For instance, if a developer pulls `main` and the latest commit didn't include updated gem RBIs, the add-on would immediately generate gem RBIs for files they didn't touch, polluting their workspace. We felt it was better to avoid this.

paracycle: But the gem command can clean up stale RBI files and create RBI files that should exist as well. I still don't understand why this level of detection needs to happen outside of the normal Tapioca gem RBI generation. We should be able to find a way to make the gem RBI generation to be smarter about this.

I don't think we want to try to duplicate the logic of configuration files, etc, outside of that flow. That won't scale.

paracycle: > The tapioca gem command only runs when we detect lockfile differences. However, consider this scenario:
> 
> * Developer adds/modifies/removes a gem and regenerates RBIs
> * Developer reverts the Gemfile changes and runs `bundle install`
> * The lockfile returns to its committed state
> * LSP restarts due to lockfile change, but finds no diff against the committed version
> * `tapioca gem <gems>` doesn't run, leaving orphaned RBI files or missing previously deleted ones

This scenario is exactly why the git based detection is not a great idea in the first place, and not a good reason to add more complexity to this. There is no good way to synchronize the state of RBI files with the state of the lockfile changes.

PR #2335, opened 2025-06-25
Fix single delegated_type

ngan:
### Motivation

When `delegated_type` only has one type, `T.any` is used will error since there's only one type.

### Implementation

Don't use `T.any` when there's only one type.

### Tests
Added a test case.


st0012:
The Ruby head failure is not related to the changes, see: https://github.com/rubygems/rubygems/pull/8791
So I'll merge the PR after other builds pass.

ngan:
thanks Stan!

Issue #1347, opened 2023-01-05
Show a warning for skipped gems?


andyw8:
Tapioca [skips particular gems](https://github.com/Shopify/tapioca/blob/559f365e825e9160896c14041fcef1fc6cbc67c2/lib/tapioca/gemfile.rb#L155-L165), which can be confusing. I ran into this when trying to add RBIs for `debug`.

```
% tapioca gem debug
Requiring all gems to prepare for compiling...  Done

Error: Cannot find gem 'debug'
main % 
```

Perhaps we can output a warning when skipping a particular gem.

drewhoffer:
I'd like to take a look at this if it's still available? Are we just wanting to show an output showing the [IGNORED_GEMS](https://github.com/Shopify/tapioca/blob/32163d328f3edf83883f6f84e9094d9ab141809c/lib/tapioca/gemfile.rb#L122) as output regardless on `tapioca gem`? Or should we show when someone is trying to run `tapioca gem debug` for example.

e.g.

![Image](https://github.com/user-attachments/assets/6b2eb89f-9457-4295-8f5c-1ad78e9e06d6)

KaanOzkan:
> Or should we show when someone is trying to run tapioca gem debug for example

Yes this sounds good and feel free to take it on. I don't think we need to show it unless you specify the ignored gem name as an argument, so nothing on `tapioca gem`.

written-fresh:
Hey, coming back to this after an hour... year... however long away. I'll dust this branch off and see if its still good to open.

PR #2354, opened 2025-07-20
feat: Show list of ignored gems

written-fresh:
### Motivation

This PR addresses #1347 which aims to show a message when trying to run `tapioca gem` on a ignored gem.

### Implementation

I have added a skipped gems property to the cli. 

### Tests

I have now written a test just to confirm something like `gem sorbet` outputs the warning that it was skipped. More than happy to add more if needed.

#### No warning

Showing no warning

#### Warning

image

### DIFF
--- a/lib/tapioca/commands/gem_generate.rb
+++ b/lib/tapioca/commands/gem_generate.rb
@@ -65,7 +65,7 @@ module Tapioca
       def gems_to_generate(gem_names)
         return @bundle.dependencies if gem_names.empty?
 
-        gem_names.each_with_object([]) do |gem_name, gems|
+        (gem_names - @exclude).each_with_object([]) do |gem_name, gems|
           gem = @bundle.gem(gem_name)
 
           if gem.nil?
--- a/lib/tapioca/commands/gem_generate.rb
+++ b/lib/tapioca/commands/gem_generate.rb
@@ -17,7 +17,8 @@ module Tapioca
           halt_upon_load_error: @halt_upon_load_error,
         )
 
-        gem_queue = gems_to_generate(@gem_names).reject { |gem| @exclude.include?(gem.name) }
+        gem_queue = gems_to_generate(@gem_names)
+        gem_queue.reject! { |gem| @exclude.include?(gem.name) }
         anything_done = [
           perform_removals,
           gem_queue.any?,
@@ -28,6 +29,18 @@ module Tapioca
             compile_gem_rbi(gem)
             puts
           end
+        end
+
+        unless @skipped_gems.empty?
+          say("\nNote: Tapioca skipped RBI generation for the following gems because they are ignored by default:", [:yellow, :bold])
+          say(@skipped_gems.join(", "), [:yellow, :bold])
+        end
+
+        user_excluded_gems = @gem_names & @exclude
+
+        unless user_excluded_gems.empty?
+          say("\nNote: Tapioca skipped RBI generation for the following gems because they were excluded:", [:yellow, :bold])
+          say(user_excluded_gems.join(", "), [:yellow, :bold])
         end
 
         if anything_done
@@ -52,15 +65,19 @@ module Tapioca
       def gems_to_generate(gem_names)
         return @bundle.dependencies if gem_names.empty?
 
-        (gem_names - @exclude).each_with_object([]) do |gem_name, gems|
+        gem_names.each_with_object([]) do |gem_name, gems|
           gem = @bundle.gem(gem_name)
 
           if gem.nil?
-            next if @lsp_addon
-
-            raise Tapioca::Error, set_color("Error: Cannot find gem '#{gem_name}'", :red)
+            if @lsp_addon
+              next
+            elsif Gemfile::GemSpec::IGNORED_GEMS.include?(gem_name)
+              @skipped_gems << gem_name
+              next
+            else
+              raise Tapioca::Error, set_color("Error: Cannot find gem '#{gem_name}'", :red)
+            end
           end
-
           gems.concat(gem_dependencies(gem)) if @include_dependencies
           gems << gem
         end
--- a/lib/tapioca/commands/gem_generate.rb
+++ b/lib/tapioca/commands/gem_generate.rb
@@ -17,8 +17,7 @@ module Tapioca
           halt_upon_load_error: @halt_upon_load_error,
         )
 
-        gem_queue = gems_to_generate(@gem_names)
-        gem_queue.reject! { |gem| @exclude.include?(gem.name) }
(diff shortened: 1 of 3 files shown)


spec/tapioca/cli/gem_spec.rb
```diff
@@ -877,6 +877,17 @@ class Secret; end
           assert_success_status(result)
         end
 
+        it "reports explicitly requested ignored gems" do
+          result = @project.tapioca("gem sorbet")
+
+          assert_stdout_includes(result, "Note: Tapioca intentionally ignores the following gems:")
+          assert_stdout_includes(result, "sorbet")
```
KaanOzkan:
```suggestion
          assert_stdout_includes(result, <<~OUT)
            Note: Tapioca intentionally ignores the following gems:
            sorbet
          OUT
```

lib/tapioca/cli.rb
```diff
@@ -321,6 +321,7 @@ def gem(*gems)
         halt_upon_load_error: options[:halt_upon_load_error],
         lsp_addon: options[:lsp_addon],
         verbose: options[:verbose],
+        skipped_gems: [],
```
KaanOzkan: There is no need for this argument if it's always fixed. Can you update `AbstractGem` to not accept it and just set the ivar to empty array.

lib/tapioca/commands/gem_generate.rb
```diff
@@ -44,6 +44,10 @@ def execute
         else
           say("No operations performed, all RBIs are up-to-date.", [:green, :bold])
         end
+        unless @skipped_gems.empty?
+          say("\nNote: Tapioca intentionally ignores the following gems:", [:yellow, :bold])
```
KaanOzkan:
```suggestion
          say("\nNote: Tapioca is skipping gem rbi generation for following gems due to the built-in configuration:", [:yellow, :bold])
```

KaanOzkan: Now that I'm writing it, I think we should also output for gems that are excluded in user config (using `--exclude`)

lib/tapioca/cli.rb
written-fresh: Yeah, good catch!

lib/tapioca/commands/gem_generate.rb
```diff
@@ -17,7 +17,9 @@ def execute
           halt_upon_load_error: @halt_upon_load_error,
         )
 
-        gem_queue = gems_to_generate(@gem_names).reject { |gem| @exclude.include?(gem.name) }
+        gem_queue = gems_to_generate(@gem_names)
+        user_excluded_gems = user_excluded_gem_names(gem_queue)
+        gem_queue.reject! { |gem| @exclude.include?(gem.name) }
```
KaanOzkan:
```suggestion
        user_excluded_gems = @gem_names & @exclude

        gem_queue = gems_to_generate(@gem_names)
        gem_queue.reject! { |gem| @exclude.include?(gem.name) }
```

Simpler

lib/tapioca/commands/gem_generate.rb
KaanOzkan:
```suggestion
   gem_names.each_with_object([]) do |gem_name, gems|
```

lib/tapioca/commands/gem_generate.rb
```diff
@@ -56,16 +66,28 @@ def gems_to_generate(gem_names)
...
         end
       end
 
+      #: (Array[Gemfile::GemSpec] gem_queue) -> Array[String]
+      def user_excluded_gem_names(gem_queue)
+        @exclude.uniq.select do |gem_name|
+          @bundle.gem(gem_name) &&
+            (@gem_names.include?(gem_name) || gem_queue.any? { |gem| gem.name == gem_name })
+        end
+      end
```
KaanOzkan:
```suggestion
```

lib/tapioca/commands/gem_generate.rb
```diff
@@ -44,6 +46,14 @@ def execute
         else
           say("No operations performed, all RBIs are up-to-date.", [:green, :bold])
```
KaanOzkan, line 57: This should be printed at the end, let's move the new prints before this if/else.

lib/tapioca/commands/gem_generate.rb
```diff
@@ -44,6 +46,14 @@ def execute
...
         end
+        unless @skipped_gems.empty?
+          say("\nNote: Tapioca is skipping gem rbi generation for following gems due to the built-in configuration:", [:yellow, :bold])
```
KaanOzkan:
```suggestion
          say("\nNote: Tapioca skipped RBI generation for the following gems because they are ignored by default:", [:yellow, :bold])
```

lib/tapioca/commands/gem_generate.rb
```diff
@@ -44,6 +46,14 @@ def execute
...
+          say(@skipped_gems.join(", "), [:yellow, :bold])
+        end
+        unless user_excluded_gems.empty?
+          say("\nNote: Tapioca is skipping gem rbi generation for following gems due to user configuration:", [:yellow, :bold])
```
KaanOzkan:
```suggestion
          say("\nNote: Tapioca skipped RBI generation for the following gems because they were excluded:", [:yellow, :bold])
```

spec/tapioca/cli/gem_spec.rb
```diff
@@ -877,6 +877,49 @@ class Secret; end
           assert_success_status(result)
         end
 
+        it "reports explicitly requested ignored gems" do
```
KaanOzkan, line 880: Can you also add a test for not reporting excluded gems when we only run `gem` and `gem --all`?

lib/tapioca/commands/gem_generate.rb
KaanOzkan: Also I think we should move `user_excluded_gems` assignment right before it's printed. It's nicer I believe.

KaanOzkan (commented):
Thank you, few more refactors. Also there might be some long lines, it'd be good to run rubocop before pushing.

lib/tapioca/commands/gem_generate.rb
```diff
@@ -65,7 +65,7 @@ def execute
       def gems_to_generate(gem_names)
         return @bundle.dependencies if gem_names.empty?
 
-        gem_names.each_with_object([]) do |gem_name, gems|
+        (gem_names - @exclude).each_with_object([]) do |gem_name, gems|
```
KaanOzkan: CI revealed a failing case where a gem was supplied and excluded, brought back the original behaviour and added a test.

KaanOzkan (approved):

PR #2356, opened 2025-07-21
Generate Active Model attributes in an included `GeneratedAttributeMethods` module

paracycle:
### Motivation

Consumers that define Active Model classes which override any of the attribute models cannot override the signature of the generated attribute methods since we are making it look like the methods are defined on the model class, and not on an included module.

### Implementation

Start generating all attribute methods on a `GeneratedAttributeMethods` module that gets included into the model class.

### Tests

Update the existing tests to match

### DIFF
--- a/lib/tapioca/dsl/compilers/active_model_attributes.rb
+++ b/lib/tapioca/dsl/compilers/active_model_attributes.rb
@@ -26,12 +26,16 @@ module Tapioca
       # # typed: true
       #
       # class Shop
+      #   include GeneratedAttributeMethods
       #
-      #   sig { returns(T.nilable(::String)) }
-      #   def name; end
+      #   module GeneratedAttributeMethods
+      #     sig { returns(T.nilable(::String)) }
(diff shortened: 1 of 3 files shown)


bslobodin (commented):
I'll try to find a bit of time to dig deeper, but a basic test locally ends up with rbis that have a few errors.

- https://github.com/Shopify/trust-battery/compare/bs/test-tapioca
- https://buildkite.com/shopify/trust-battery/builds/10842/steps/canvas?sid=01982e20-a65b-4771-a4e3-4aad33482a48#01982e20-a6dc-4275-9b05-17d8676e276e/156

I also noticed this during `bin/tapioca gem`, not sure if red herring or not,

> Changed strictness of sorbet/rbi/gems/tapioca@0.17.7-8322b4a3836d2efbef293a9963e1cd5b8e2be8fb.rbi to `typed: false` (conflicting with DSL files)

vinistock (approved):

Issue #2361, opened 2025-07-22
Comment parsing errors in generating gem rbis


dduugg:
Tapioca is dropping some comment lines when generating gem rbi shims in the latest release (v0.17.7), as compared with earlier releases. Here is an example using v1.182.0 of the `aws-sdk-s3` gem:

<img width="1276" height="1247" alt="Image" src="https://github.com/user-attachments/assets/3ae18568-36ef-4ddb-8ac4-5aff26e54d0b" />

I don't know which version of tapioca generated the previous rbi file. I suspect the issue involves processing comments with a leading `:` as RBS syntax.

PR #2364, opened 2025-07-23
Fix RBS comment prefix in YardDoc listener

KaanOzkan:
### Motivation

Resolves https://github.com/Shopify/tapioca/issues/2361

### DIFF
--- a/lib/tapioca/gem/listeners/yard_doc.rb
+++ b/lib/tapioca/gem/listeners/yard_doc.rb
@@ -32,7 +32,7 @@ module Tapioca
 
         #: (String line) -> bool
         def rbs_comment?(line)
-          line.strip.start_with?(": ", "| ")
+          line.start_with?(": ", "| ")
         end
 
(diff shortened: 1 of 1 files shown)


lib/tapioca/gem/listeners/yard_doc.rb
```diff
@@ -32,7 +32,7 @@ def initialize(pipeline)
 
         #: (String line) -> bool
         def rbs_comment?(line)
-          line.strip.start_with?(": ", "| ")
+          line.start_with?(":", "|")
```
Morriar: Do we expect some RBS comments to not have the space?

Morriar:
```suggestion
          line.start_with?(": ", "| ")
```

KaanOzkan: No, I removed it by accident, fixed.

PR #2368, opened 2025-08-06
At fix docs

Morriar:
### Motivation

I noticed that our documentation pages were picking up the RBS generic signature comment for compiler classes like here: https://github.com/Shopify/tapioca/blob/main/manual/compiler_activerecordcolumns.md

image

We should ignore these lines when generating the documentation.

### Implementation

Added a custom Yard::DocstringParser that ignores all the lines starting with `#:` or `#|`.

### Tests

See regenerated documentation pages.


paracycle (approved):
Thanks. I'd noticed this as well, but didn't quite know where to go to fix it. ❤️

PR #2387, opened 2025-09-12
Prevent gems from calling `exit` or `abort` during gem load

paracycle:
### Motivation

We have some internal gems that are meant to be used only during development. They have started enforcing that with a check at gem load time, and by calling `exit` to abort the process. This breaks Tapioca gem loading unnecessarily.

### Implementation

We've already had the mechanism for running a block with disabled exits, so I just moved that to top level, and started using it around gem loading as well.

### Tests

Added a test for the new expected behaviour.


lib/tapioca/commands/abstract_gem.rb
```diff
@@ -67,7 +67,7 @@ def compile_gem_rbi(gem)
           reason: "types exported from the `#{gem.name}` gem",
         ) if @file_header
 
-        rbi.root = Runtime::Trackers::Autoload.with_disabled_exits do
+        rbi.root = Tapioca.with_disabled_exits do
```
Morriar, line 70: I think it could have stayed under `Runtime` with other runtime related helpers

paracycle: We can move both of the runtime helper methods under runtime, no problem.

Morriar: https://github.com/Shopify/tapioca/pull/2401

Issue #2393, opened 2025-09-23
Incompatiblity with Sorbet 0.6.12565 under `--enable-experimental-requires-ancestor`


Bo98:
The default `requires_ancestor` (if `--enable-experimental-requires-ancestor` is enabled) has changed upstream: https://github.com/sorbet/sorbet/pull/9345.

This changes means that the `activesupport` RBI causes various incompatibilities:

```
 sorbet/rbi/gems/activerecord@8.0.3.rbi:1105: `ActiveRecord::Associations::AssociationScope::ReflectionProxy` must include `Kernel` (required by `ActiveSupport::Tryable`) https://srb.help/5064
    1105 |class ActiveRecord::Associations::AssociationScope::ReflectionProxy < ::SimpleDelegator
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    sorbet/rbi/gems/activesupport@8.0.3.rbi:14922: required by `ActiveSupport::Tryable` here
       14922 |module ActiveSupport::Tryable
       14923 |  # source://activesupport//lib/active_support/core_ext/object/try.rb#7
       14924 |  def try(*args, **_arg1, &block); end
       14925 |
       14926 |  # source://activesupport//lib/active_support/core_ext/object/try.rb#20
       14927 |  def try!(*args, **_arg1, &block); end
       14928 |end

sorbet/rbi/gems/prism@1.5.1.rbi:22690: `Prism::LexCompat::EndContentToken` must include `Kernel` (required by `ActiveSupport::Tryable`) https://srb.help/5064
       22690 |class Prism::LexCompat::EndContentToken < ::Prism::LexCompat::Token
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    sorbet/rbi/gems/activesupport@8.0.3.rbi:14922: required by `ActiveSupport::Tryable` here
       14922 |module ActiveSupport::Tryable
       14923 |  # source://activesupport//lib/active_support/core_ext/object/try.rb#7
       14924 |  def try(*args, **_arg1, &block); end
       14925 |
       14926 |  # source://activesupport//lib/active_support/core_ext/object/try.rb#20
       14927 |  def try!(*args, **_arg1, &block); end
       14928 |end

[...]

PR #2394, opened 2025-09-25
Do not mixin Kernel for modules

KaanOzkan:
### Motivation

Part of https://github.com/Shopify/tapioca/issues/2393.

Fixes `Kernel` is already included errors by not generating `include ::Kernel` calls in RBIs.
### Implementation

If a module includes Kernel it's to have access to Kernel methods in it's implementation. But in RBIs we don't generate the method bodies so we shouldn't run into isssues.

### DIFF
--- a/lib/tapioca/gem/listeners/mixins.rb
+++ b/lib/tapioca/gem/listeners/mixins.rb
@@ -98,7 +98,13 @@ module Tapioca
           # inherited ancestors, past the location of the constant itself.
           ancestors = Set.new.compare_by_identity.merge(ancestors_of(constant))
 
-          (ancestors - inherited_ancestors).to_a
+          result = (ancestors - inherited_ancestors).to_a
+
+          # Filter out Kernel for modules (but not classes) since modules shouldn't
+          # need to explicitly include Kernel in RBIs
+          result.delete(Kernel) unless Class === constant
+
+          result
         end
       end
     end
--- a/spec/tapioca/gem/pipeline_spec.rb
+++ b/spec/tapioca/gem/pipeline_spec.rb
@@ -1639,7 +1639,6 @@ class Tapioca::Gem::PipelineSpec < Minitest::HooksSpec
         module Baz
           include ::Quux
           include ::Foo
-          include ::Kernel
           extend ::Bar
         end
 


lib/tapioca/gem/listeners/mixins.rb
```diff
@@ -98,7 +98,15 @@ def interesting_ancestors_of(constant)
           # inherited ancestors, past the location of the constant itself.
           ancestors = Set.new.compare_by_identity.merge(ancestors_of(constant))
 
-          (ancestors - inherited_ancestors).to_a
+          result = (ancestors - inherited_ancestors).to_a
+
+          # Filter out Kernel for modules (but not classes) since modules shouldn't
+          # need to explicitly include Kernel in RBIs
+          unless constant.is_a?(Class)
```
paracycle: Let's use:
```suggestion
          unless Class === constant
```
since `is_a?` may be different in `constant`.

lib/tapioca/gem/listeners/mixins.rb
```diff
@@ -98,7 +98,15 @@ def interesting_ancestors_of(constant)
...
+            result = result.reject { |ancestor| ancestor == Kernel }
```
paracycle:
```suggestion
            result = result.delete(Kernel)
```

amomchilov (approved):

lib/tapioca/gem/listeners/mixins.rb
```diff
@@ -98,7 +98,15 @@ def interesting_ancestors_of(constant)
...
           ancestors = Set.new.compare_by_identity.merge(ancestors_of(constant))
 
-          (ancestors - inherited_ancestors).to_a
+          result = (ancestors - inherited_ancestors).to_a
+
+          # Filter out Kernel for modules (but not classes) since modules shouldn't
+          # need to explicitly include Kernel in RBIs
+          unless Class === constant
+            result.delete(Kernel)
+          end
```
paracycle: Nitpick:
```suggestion
          result.delete(Kernel) unless Class === constant
```

paracycle (approved):
Thanks!

PR #2395, opened 2025-09-29
Print currently compiling gems. Add verbose attribute to gem command.

cub8:
### Motivation
Recently in one project I was working on I encountered the following problem when initializing tapioca:
- while compiling gems, the process just halted at some point - I guess it encountered some issues with compiling specific gems
- I had problems with pinpointing exactly WHICH gems Tapioca had problem with - because the pipeline outputs only compiled gems, not ongoing compilations
- This made debugging very hard - finding those problematic gems required me to put debuggers inside library and I don't think that's how it's supposed to work
- Also, it baffled me that `--verbose` flag didn't change anything in output when running `bundle exec tapioca init` or `bundle exec tapioca gems` - I thought it's gonna tell me what's happening in the background or what gems are currently compiled but it looks like this parameter isn't passed to gems command at all
- Eventually, I found those gems that halted tapioca gems compilation process and just excluded them from compilation 

### Implementation
- I've added `@verbose` attribute to `AbstractGem` class which is passed there from cli options
- When `@verbose` is true, it outputs `Currently compiling <gem_name>` before compiling gem
- This make it easier to find in the pipeline output which gems are _being compiled_, but which _weren't compiled_

### Tests
- Two tests for running `bundle exec tapioca gems --verbose` and `bundle exec tapioca gems -V`

### Example
image

### DIFF
--- a/lib/tapioca/cli.rb
+++ b/lib/tapioca/cli.rb
@@ -301,6 +301,7 @@ module Tapioca
         rbi_formatter: rbi_formatter(options),
         halt_upon_load_error: options[:halt_upon_load_error],
         lsp_addon: options[:lsp_addon],
+        verbose: options[:verbose],
       }
 
       command = if verify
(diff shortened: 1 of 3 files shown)


cub8:
I have signed the CLA!

KaanOzkan (approved):
Thanks. I think it makes sense to have.

PR #2399, opened 2025-10-10
feat: Add support for app Sinatra app with ActiveRecord

namtx:
### Motivation

I'm getting an error when trying to generate DSL with `bundle exec tapioca dsl` on Sinatra-based app with ActiveRecord
It complains about `Rails` constant missing
```shell
uninitialized constant Tapioca::Dsl::Pipeline::Rails
```
This PR adds support for a Sinatra app with ActiveRecord.
Task `db:abort_if_pending_migrations` will be loaded from `sinatra/activerecord/rake` instead of `Rails.application.load_tasks`.

### DIFF
--- a/lib/tapioca/dsl/pipeline.rb
+++ b/lib/tapioca/dsl/pipeline.rb
@@ -202,8 +202,13 @@ module Tapioca
         # warns the user, so that they are aware they need to migrate their database
         return if @lsp_addon
         return unless defined?(::Rake)
+        return unless defined?(::Rails) || defined?(::Sinatra)
 
-        Rails.application.load_tasks
+        if defined?(::Rails)
+          Rails.application.load_tasks
+        elsif defined?(::Sinatra) # support Sinatra app with ActiveRecord
+          require "sinatra/activerecord/rake"
+        end
 
         if Rake::Task.task_defined?("db:abort_if_pending_migrations")
           Rake::Task["db:abort_if_pending_migrations"].invoke


lib/tapioca/dsl/pipeline.rb
```diff
@@ -202,8 +202,13 @@ def abort_if_pending_migrations!
...
         return if @lsp_addon
         return unless defined?(::Rake)
+        return unless defined?(::Rails) || defined?(::Sinatra)
 
-        Rails.application.load_tasks
+        if defined?(::Rails)
+          Rails.application.load_tasks
+        elsif defined?(::Sinatra) # support Sinatra app with ActiveRecord
+          require "sinatra/activerecord/rake"
+        end
```
KaanOzkan, line 211: I don't think this will be enough to support Sinatra applications. Tapioca [boots up ](https://github.com/Shopify/tapioca/blob/6bf27f3658e6c3a3b8757bb3079a8b04fbb7d10d/lib/tapioca/loaders/loader.rb#L180)the Rails app. So when a DSL compiler runs we have access to an object space that contains relevant classes for DSL generation (e.g. [descendants of ActiveRecord::Base](https://github.com/Shopify/tapioca/blob/6bf27f3658e6c3a3b8757bb3079a8b04fbb7d10d/lib/tapioca/dsl/compilers/active_record_relations.rb#L189))

Given that it's not a small addition and not used at Shopify, we're unlikely to support Sinatra and maintain it moving forward. However, in theory it should be doable with monkey patching tapioca externally and adding Sinatra loading support.

namtx: agree, 
I came up with monkey patching on my project.
Thanks for explaining.

PR #2400, opened 2025-10-10
Migrate long RBS signature comments to multiline sigs

Morriar:
And reenable `Layout/LineLength`

### DIFF
--- a/spec/tapioca/dsl/compilers/active_record_associations_spec.rb
+++ b/spec/tapioca/dsl/compilers/active_record_associations_spec.rb
@@ -741,7 +741,6 @@ module Tapioca
                     "Cannot generate association `has_many :readers, through: :author` on `Post` since the constant `Reader` does not exist.",
                     "Cannot generate association `has_one :award, through: :blog` on `Post` since the constant `Award` does not exist.",
                   ]
-                  # rubocop:enable Layout/LineLength
 
                   assert_equal(expected_errors, generated_errors)
                 end
@@ -1503,7 +1502,6 @@ module Tapioca
 
                   assert_equal(6, generated_errors.size)
 
-                  # rubocop:disable Layout/LineLength
                   expected_errors = [
                     "Cannot generate association `has_one :author` on `Post` since the constant `Author` does not exist.",
                     "Cannot generate association `has_many :comments` on `Post` since the constant `Comment` does not exist.",
@@ -1512,7 +1510,6 @@ module Tapioca
                     "Cannot generate association `has_many :readers, through: :author` on `Post` since the constant `Reader` does not exist.",
                     "Cannot generate association `has_one :award, through: :blog` on `Post` since the constant `Award` does not exist.",
                   ]
-                  # rubocop:enable Layout/LineLength
 
                   assert_equal(expected_errors, generated_errors)
                 end


tasks/generator_documentation.rake
```diff
@@ -7,19 +7,24 @@ require "tapioca/runtime/reflection"
...
-    #
     # At this point, `content` represents each line of the docstring where
     # the `#` has already been stripped.
-    super(content.sub(/^[:|].*$/, ""))
+
+    # Remove RBS signature comments (lines starting with `#:` or `#|`)
+    content = content.gsub(/^[:|].*$/, "")
+
+    # Remove Rubocop comments (lines starting with `# rubocop:disable` or `# rubocop:enable`)
+    content = content.gsub(/^rubocop:(enable|disable) .*$/, "")
```
vinistock, line 20:
```suggestion
    content.gsub!(/^rubocop:(enable|disable) .*$/, "")
```

Morriar: We can't, the string is frozen

lib/tapioca/dsl/compilers/url_helpers.rb
```diff
@@ -6,6 +6,7 @@
 module Tapioca
   module Dsl
     module Compilers
+      # rubocop:disable Layout/LineLength
```
paracycle: Interesting that the [`AllowURI` attribute](https://docs.rubocop.org/rubocop/cops_layout.html#configurable-attributes-layoutlinelength) doesn't allow the URL in the comment below.

.rubocop.yml
```diff
@@ -19,8 +19,8 @@ Layout/LeadingCommentSpace:
   AllowRBSInlineAnnotation: true
 
 Layout/LineLength:
+  Enabled: true
```
rafaelfranca: We removed it from the styleguide by design. I don't think any of our projects should enable. The reason to remove is that it just create a lot of `rubocop:disable` in the code with no real gain.

lib/tapioca/dsl/compilers/url_helpers.rb
Morriar: It seems the issue is the combination of an URI that contains what looks like a qualified name: https://github.com/rubocop/rubocop/blob/master/lib/rubocop/cop/mixin/line_length_help.rb#L41 and it confuses Rubocop on the line length.

You can repro with this comment:

```rb
# http://XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX#X::X
```

I _think_ it's a bug? Maybe I should report it.

.rubocop.yml
Morriar: I'm on the fence:
* without it we need to manually call out PRs where the line length is abused and it becomes harder to set a clear expectation
* with it, rules is sometime too arbitrary and the `rubocop:disable` are a pain

I can remove it to follow the style guide more closely

rafaelfranca: Most editors wrap long lines those days. Monitors are also larger.

Morriar: Removed it and cleaned the remaining Rubocop disable comments for it 👍

PR #2407, opened 2025-10-16
Drop the `benchmark` dependency

Earlopain:
### Motivation
tapioca currently depends on the `benchmark` gem because of a change to Ruby bundled gems. The functionality is trivial though, and I don't think you have to take on a dependency for that.

### Implementation
Just use `Process.clock_gettime` directly.

### Tests
No behaviour change. I just ran some cli commands and they still print the time in seconds. Probably the output is also already asserted against somewhere.


Morriar (commented):
<deleted>

Earlopain:
How about now?

Morriar (approved):

Morriar:
```
Load Error is: cannot load such file -- benchmark
```

Looks like something needs to be fixed on the tests side.

Earlopain:
Right, some test requires real rails (and activesupport) 7.1.0 which doesn't declare the dependency. I bumped the test to the latest patch release and that should resolve it

amomchilov (approved):

PR #2410, opened 2025-10-18
Simplify test to look for output in stderr and ignore warnings

rwstauner:
Resolves https://github.com/Shopify/tapioca/issues/2414

Upgrading sorbet (which happens in CI when the Gemfile.lock is removed) currently results in a Warning.

Rather than trying to work around whether the warning is missing or present in a section by itself, or present in a section we are looking at, just look for what we want in the whole output without breaking into sections.

This addresses CI failures like https://github.com/Shopify/tapioca/actions/runs/18478106067/job/52647124759?pr=2402#step:6:1546

### DIFF
--- a/spec/tapioca/cli/check_shims_spec.rb
+++ b/spec/tapioca/cli/check_shims_spec.rb
@@ -426,20 +426,17 @@ module Tapioca
           Looking for duplicates...  Done
         OUT
 
-        sections = T.must(result.err).split("\n\n").map(&:strip)
-        duplicates_for_object = sections[-4]&.split("\n")
-        assert_includes(duplicates_for_object, "Duplicated RBI for ::Object:")
-        assert_includes(duplicates_for_object, " * sorbet/rbi/shims/core/object.rbi:1:0-1:17")
+        stderr = T.must(result.err)
+        assert_includes(stderr, "Duplicated RBI for ::Object:")
+        assert_includes(stderr, " * sorbet/rbi/shims/core/object.rbi:1:0-1:17")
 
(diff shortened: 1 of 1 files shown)


spec/tapioca/cli/check_shims_spec.rb
```diff
@@ -427,18 +427,20 @@ def self.some_method_that_is_not_defined_in_the_payload; end
         OUT
 
         sections = T.must(result.err).split("\n\n").map(&:strip)
-        duplicates_for_object = sections[0]&.split("\n")
+        duplicates_for_object = sections[-4]&.split("\n")
```
Morriar: I don't think the sections are actually that important, we could just check that `result.err` contains what we want and avoid all this splitting

rwstauner: Makes sense to me

spec/tapioca/cli/check_shims_spec.rb
```diff
@@ -426,19 +426,21 @@ def self.some_method_that_is_not_defined_in_the_payload; end
           Looking for duplicates...  Done
         OUT
 
-        sections = T.must(result.err).split("\n\n").map(&:strip)
-        duplicates_for_object = sections[0]&.split("\n")
+        stderr = T.must(result.err)
+        duplicates_for_object = stderr
```
Morriar: Let's remove all the unnecessary intermediate variables

rwstauner: Sure.  I did that first but I have this compulsion to produce the smallest possible diff and sometimes I like to leave things like this in case someone wants to change it back to something similar but neither of these are good arguments.

PR #2419, opened 2025-10-28
Fix DSL test failure caused by ActiveModel::SecurePassword change

st0012:
### Motivation

In Rails 8.1, `ActiveModel::SecurePassword` starts to use ActiveSupport's `Numeric#minutes` method (https://github.com/rails/rails/pull/55574) without explicitly requiring that core extension, and that's causing the related test to fail.

Since the behaviour of `ActiveModel::SecurePassword` isn't what that test is for, we can lock related dependencies versions to avoid failures for now.

### DIFF
--- a/spec/tapioca/cli/dsl_spec.rb
+++ b/spec/tapioca/cli/dsl_spec.rb
@@ -1845,7 +1845,9 @@ module Tapioca
         end
 
         it "generates RBIs for ActiveResource containing arbitrary constants, and referenced by file path" do
-          @project.require_real_gem("activeresource")
+          # locking to 6.1.4 to avoid breaking changes in later versions
+          @project.require_real_gem("activeresource", "6.1.4")
+          @project.require_real_gem("activemodel", "8.0.0")
           @project.bundle_install!
           @project.write!("lib/post.rb", <<~RUBY)
             require "active_resource"


spec/tapioca/cli/dsl_spec.rb
```diff
@@ -2856,6 +2863,7 @@ class PostCompiler < Tapioca::Dsl::Compiler
               Tapioca::Dsl::Compilers::ActiveRecordStore                   enabled
               Tapioca::Dsl::Compilers::ActiveSupportConcern                enabled
               Tapioca::Dsl::Compilers::ActiveSupportCurrentAttributes      enabled
+              Tapioca::Dsl::Compilers::ActiveSupportTimeExt                enabled
```
paracycle: why are these here all of a sudden?

spec/tapioca/cli/dsl_spec.rb
```diff
@@ -1878,7 +1878,14 @@ class Post < ActiveResource::Base
             Please review changes and commit them.
           OUT
 
-          assert_empty_stderr(result)
+          # This deprecation warning is added to Rails 8.1
+          # And activeresource is still requiring the deprecated file, so we need to account for this warning
+          assert_stderr_includes_pattern(result, %r{
```
paracycle: So maybe we should check that this is the whole pattern in the error output?

lib/tapioca/dsl/compilers/active_model_secure_password.rb
```diff
@@ -3,6 +3,8 @@
 
 return unless defined?(ActiveModel::SecurePassword)
 
+require "active_support/core_ext/numeric"
```
paracycle: Could we upstream this to Rails? AFAIK, they value each file being complete in terms of its requires.

Also, do we need this require this early? What breaks when we don't do the require?

st0012: [This](https://github.com/rails/rails/blob/main/activemodel/lib/active_model/secure_password.rb#L12) breaks

spec/tapioca/cli/dsl_spec.rb
st0012: I think the `\z` at the end should ensure that?

lib/tapioca/dsl/compilers/active_model_secure_password.rb
jesse-shopify: Since this is the only place that uses `.minutes` would another option be to change that line in Rails in order to remove the dependency on `active_support/core_ext/numeric`?

paracycle: I realize that is the line that breaks, but then how do we get past the `defined?(ActiveModel::SecurePassword)` check? We probably end up with a half-defined `ActiveModel::SecurePassword` as a result anyway.

st0012: I think the upstream should have required the file, yes. But at the same time I'd like to have a green build for now.
I can add comment to suggest that this is temporary and should be removed after Rails cuts a release with our patch.

paracycle: Yeah, we've done that before for other compilers too.

But, my main problem is that I can't see how this solves the original problem. In order to get to this `require`, we need to have gone past the `defined?(ActiveModel::SecurePassword)` check. If we are going past it, the class was defined either fully or partially, but will not be loaded again. Requiring the file that is needed will not fix that error.

paracycle: Looking at the backtrace of the error, it seems like this is working only due to autoloading:
```
        /opt/hostedtoolcache/Ruby/3.2.9/x64/lib/ruby/gems/3.2.0/gems/activemodel-8.1.1/lib/active_model/secure_password.rb:12:in `<module:SecurePassword>': undefined method `minutes' for 15:Integer (NoMethodError)
        
            DEFAULT_RESET_TOKEN_EXPIRES_IN = 15.minutes
                                               ^^^^^^^^
        	from /opt/hostedtoolcache/Ruby/3.2.9/x64/lib/ruby/gems/3.2.0/gems/activemodel-8.1.1/lib/active_model/secure_password.rb:4:in `<module:ActiveModel>'
        	from /opt/hostedtoolcache/Ruby/3.2.9/x64/lib/ruby/gems/3.2.0/gems/activemodel-8.1.1/lib/active_model/secure_password.rb:3:in `<main>'
        	from <internal:/opt/hostedtoolcache/Ruby/3.2.9/x64/lib/ruby/site_ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
        	from <internal:/opt/hostedtoolcache/Ruby/3.2.9/x64/lib/ruby/site_ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
        	from /home/runner/work/tapioca/tapioca/lib/tapioca/dsl/compilers/active_model_secure_password.rb:105:in `gather_constants'
...
```

So, we only actually load the file when we refer to the constant in `gather_constants` of the SecurePassword compiler, and that's why the compiler class can be loaded.

That means we can move the require all the way into the `gather_constants` implementation of the compiler, so that it is as targeted as posssible.

st0012: Thanks for the investigation. I've updated the approach

spec/tapioca/cli/dsl_spec.rb
st0012: Removed with the updated fix

paracycle:
@st0012 Can you just cherry pick this change as the content of this fix for now: https://github.com/Shopify/tapioca/pull/2421/commits/0007cd817fe90fc1c33434f7d92a03387b346ec6

I'll make the real fix in the Rails version upgrade branch.

st0012:
What's blocking #2421's merge?

paracycle:
> What's blocking #2421's merge?

A couple of years of neglect :)

There are a ton of tests that are failing due to Rails changes. I've got most of the them fixed, chasing a couple of stragglers.

st0012:
Closing as https://github.com/Shopify/tapioca/pull/2421 should have fixed the issue

paracycle:
The fix is upstreamed to Rails as: https://github.com/rails/rails/pull/41610

PR #2427, opened 2025-11-11
Add optional CI step to run Sorbet with Prism

thomasmarshall:
This will give us some insight into how feature complete the Prism parser mode for Sorbet is. If this step fails, we should not block the PR from being merged, however it's currently passing.


paracycle (approved):
I am not sure how much confidence this will give us, since this code base isn't that big nor does it use a lot of esoteric Ruby syntax.

What might be a better test is to make Tapioca's internal usage of Sorbet use the Prism parser, and then to compare RBI outputs as a result of that.

PR #2433, opened 2025-11-19
Bump Sorbet, fix `T::Module` violations

amomchilov:
### Motivation

Fixes the errors you get under `type: strict` for not using the generic `T::Module` type.

These changes were mostly automated in separate steps, see the commits for details. 

### Implementation

In the overwhelmingly number of cases, Tapioca's code just deals with "any arbitrary constant", which is `T::Module[T.anything]`.

Are there any places where we can use this new capability to specify more specific module types?

### Tests

Type changes only, nothing to test.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_support_concern.rb
+++ b/lib/tapioca/dsl/compilers/active_support_concern.rb
@@ -92,7 +92,7 @@ module Tapioca
           self.class.dependencies_of(concern)
         end
 
-        #: (?Module concern) -> Array[T::Module[top]]
+        #: (?T::Module[top] concern) -> Array[T::Module[top]]
         def linearized_dependencies(concern = constant)
           # Grab all the dependencies of the concern
           dependencies = dependencies_of(concern)
--- a/lib/tapioca/dsl/pipeline.rb
+++ b/lib/tapioca/dsl/pipeline.rb
@@ -178,7 +178,7 @@ module Tapioca
         # set of constants that are actually in memory with those names.
         filtered_constants = constants_by_name
           .keys
-          .map { |name| T.cast(Runtime::Reflection.constantize(name), Module) }
+          .map { |name| T.cast(Runtime::Reflection.constantize(name), T::Module[T.anything]) }
           .select { |mod| Runtime::Reflection.constant_defined?(mod) }
 
         Set.new.compare_by_identity.merge(filtered_constants)
--- a/lib/tapioca/gem/events.rb
+++ b/lib/tapioca/gem/events.rb
@@ -46,7 +46,7 @@ module Tapioca
       # @override
       #: -> T::Module[top]
       def constant
-        T.cast(@constant, Module)
+        T.cast(@constant, T::Module[T.anything])
       end
 
       #: (String symbol, T::Module[top] constant) -> void
--- a/lib/tapioca/rbi_ext/model.rb
+++ b/lib/tapioca/rbi_ext/model.rb
@@ -5,7 +5,7 @@ module RBI
   class Tree
     extend T::Sig
 
-    #: (::Module constant) ?{ (Scope scope) -> void } -> Scope
+    #: (T::Module[top] constant) ?{ (Scope scope) -> void } -> Scope
     def create_path(constant, &block)
       constant_name = Tapioca::Runtime::Reflection.name_of(constant)
       raise "given constant does not have a name" unless constant_name


vinistock (approved):

lib/tapioca/dsl/compilers/aasm.rb
```diff
@@ -203,9 +203,9 @@ class << self
           extend T::Sig
 
           # @override
-          #: -> T::Enumerable[Module]
+          #: -> T::Enumerable[T::Module[top]]
```
Morriar, line 206: I don't think the `T::` is required on the `Enumerable` as the RBS rewriter will add it. See https://github.com/sorbet/sorbet/blob/master/rbs/TypeToParserNode.cc#L72-L91.

I also think the better approach for this PR would have been to update the list in Sorbet so all the `T::Module` would become unnecessary in RBS signatures.

amomchilov: Agree, and I'll also do that, but I wanted to unblock this sooner

Morriar: https://github.com/sorbet/sorbet/pull/9665

Morriar: Once merged, we should audit all the `T::` remaining in this codebase

amomchilov: Done. I opened a PR up-stack from this one to clean up the `T::`: https://github.com/Shopify/tapioca/pull/2434

lib/tapioca/dsl/compiler.rb
```diff
@@ -21,26 +21,26 @@ class Compiler
...
         # @abstract
-        #: -> T::Enumerable[Module]
+        #: -> T::Enumerable[T::Module[top]]
         def gather_constants = raise NotImplementedError, "Abstract method called"
 
-        #: -> Set[Module]
+        #: -> Set[T::Module[top]]
         def processable_constants
-          @processable_constants ||= T::Set[Module].new.compare_by_identity.merge(gather_constants) #: Set[Module]?
+          @processable_constants ||= T::Set[T::Module[T.anything]].new.compare_by_identity.merge(gather_constants) #: Set[T::Module[top]]?
```
paracycle, line 40: Do we even need the `T::Set` here?
```suggestion
          @processable_constants ||= Set.new.compare_by_identity.merge(gather_constants) #: Set[T::Module[top]]?
```

lib/tapioca/dsl/compilers/aasm.rb
```diff
@@ -203,9 +203,9 @@ class << self
...
           def gather_constants
-            T.cast(ObjectSpace.each_object(::AASM::ClassMethods), T::Enumerable[Module])
+            T.cast(ObjectSpace.each_object(::AASM::ClassMethods), T::Enumerable[T::Module[T.anything]])
```
paracycle, line 208: We might not need these casts anymore, since I think `ObjectSpace.each_object` has the proper return type now. The return type here should already be `T::Enumerator[::AASM::ClassMethods]` which should be compatible with `T::Enumerable[T::Module[T.anything]]`.

amomchilov: Ah there's a few issues here.

If you just remove this "up-ish" cast, you get an error due to `Elem` being covariant instead of contravariant:

> Expected `T::Enumerable[T::Module[T.anything]]` but found `T::Enumerator[AASM::ClassMethods]` for method result type
> 
> Detailed explanation: `AASM::ClassMethods` is not a subtype of `T::Module[T.anything]` for covariant type member `Enumerable::Elem`[7005](https://srb.help/7005)

To out isolate any Enumer*able* vs Enumer*ator* differences, I changed to:

```ruby
# @override
#: -> T::Enumerator[T::Module[::AASM::ClassMethods]]
def gather_constants
  ObjectSpace.each_object(::AASM::ClassMethods)
end
```

> Expected `T::Enumerator[T::Module[T.anything]]` but found `T::Enumerator[AASM::ClassMethods]` for method result type
>
> Detailed explanation: `AASM::ClassMethods` is not a subtype of `T::Module[T.anything]` for covariant type member `Enumerator::Elem`[7005](https://srb.help/7005)

So I thought perhaps I could just narrow the return type, but:

> Expected `T::Enumerator[T::Module[AASM::ClassMethods]]` but found `T::Enumerator[AASM::ClassMethods]` for method result type
> Detailed explanation: `AASM::ClassMethods` is not a subtype of `T::Module[AASM::ClassMethods]` for covariant type member `Enumerator::Elem`[7005](https://srb.help/7005)

FWIW, `::AASM::ClassMethods` _is_ a subtype of `T::Module[::AASM::ClassMethods]`

```ruby
#: -> Module[::AASM::ClassMethods]
def passes
  ::AASM::ClassMethods
end
```

PR #2441, opened 2025-12-02
Fix ActiveRecordRelations compiler insert/upsert methods signatures

vinistock:
### Motivation

Fixes the generated RBI signatures for `ActiveRecord::Relation`'s [insert](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/relation.rb#L664), [insert!](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/relation.rb#L753), [insert_all](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/relation.rb#L743), [insert_all!](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/relation.rb#L810), [upsert](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/relation.rb#L820), and [upsert_all](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/relation.rb#L932) methods.

### Implementation

- Fix `returning` and `unique_by` parameter types
- Add missing `record_timestamps` parameter to all insert/upsert methods
- Add `update_only` and `on_duplicate` parameters to `upsert` and `upsert_all` methods

PR #2446, opened 2025-12-09
Stop generating RBI for insert/upsert related methods

### LINKED ISSUE #2425: Tapioca v0.17.9 generates incorrect signatures for ActiveRecord `#upsert` / `#upsert_all`
#2421 updated the ActiveRecord relations compiler to generate signatures for `#upsert` and `#upsert_all` (as well as many other methods). Howver, the signatures are incorrect:

```ruby
    sig do
      params(
        attributes: Hash,
        returning: T.nilable(T.any(T::Array[Symbol], FalseClass)),
        unique_by: T.nilable(T.any(T::Array[Symbol], Symbol))
      ).returns(ActiveRecord::Result)
    end
    def upsert(attributes, returning: nil, unique_by: nil); end

    sig do
      params(
        attributes: T::Array[Hash],
        returning: T.nilable(T.any(T::Array[Symbol], FalseClass)),
        unique_by: T.nilable(T.any(T::Array[Symbol], Symbol))
      ).returns(ActiveRecord::Result)
    end
    def upsert_all(attributes, returning: nil, unique_by: nil); end
```

There are multiple issues:

* The signatures are missing some valid arguments: `on_duplicate`, `update_only`, and `record_timestamps` (cf. [Rails docs](https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-upsert_all) and [source](https://github.com/rails/rails/blob/v8.1.1/activerecord/lib/active_record/relation.rb#L814-L934))

* `returning` is typed as `T.nilable(T.any(T::Array[Symbol], FalseClass))`, but it can also be a `String` or a `T::Array[String]`. Some examples:

    ```ruby
    Commodity.upsert_all(
      [
        { id: 2, name: "Copper", price: 4.84, created_at: Time.now, updated_at: Time.now  },
        { id: 4, name: "Gold", price: 1380.87, created_at: Time.now, updated_at: Time.now  },
        { id: 8, name: "Steel", price: 1.35, created_at: Time.now, updated_at: Time.now  }
      ],
      returning: Arel.sql("id, (xmax = '0') as inserted, name as new_name") # Arel.sql returns a String
    )

    Commodity.upsert_all(
      [
        { id: 2, name: "Copper", price: 4.84, created_at: Time.now, updated_at: Time.now  },
        { id: 4, name: "Gold", price: 1380.87, created_at: Time.now, updated_at: Time.now  },
        { id: 8, name: "Steel", price: 1.35, created_at: Time.now, updated_at: Time.now  }
      ],
      on_duplicate: Arel.sql("price = GREATEST(commodities.price, EXCLUDED.price)"),
      returning: ["price"] # Need to use String to reference dynamic SQL "columns"
    )
    ```

I'm also curious why these signatures were added to the compiler in the first place. Since they don't reference the model class, it seems like they could be omitted from the compiler and handled via annotations for ActiveRecord instead.

st0012:
### Motivation

These methods' signatures aren't model-specific and don't need to be generated through runtime for every model. We can use rbi-central for them instead.

There are other methods that probably aren't model-specific and can be removed too. But I'd prefer dropping them later as these ones causes the most errors in Core.

Closes #2425 #2441

### Implementation

Just remove associated logic from AR compiler.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_relations.rb
+++ b/lib/tapioca/dsl/compilers/active_record_relations.rb
@@ -604,47 +604,9 @@ module Tapioca
 
         #: -> void
         def create_association_relation_methods
-          returning_type = "T.nilable(T.any(T::Array[Symbol], FalseClass))"
-          unique_by_type = "T.nilable(T.any(T::Array[Symbol], Symbol))"
-
-          ASSOCIATION_METHODS.each do |method_name|
-            case method_name
-            when :insert_all, :insert_all!, :upsert_all
-              parameters = [
-                create_param("attributes", type: "T::Array[Hash]"),
-                create_kw_opt_param("returning", type: returning_type, default: "nil"),
-              ]
-
-              # Bang methods don't have the `unique_by` parameter
-              unless bang_method?(method_name)
-                parameters << create_kw_opt_param("unique_by", type: unique_by_type, default: "nil")
-              end
-
-              association_relation_methods_module.create_method(
-                method_name.to_s,
-                parameters: parameters,
-                return_type: "ActiveRecord::Result",
-              )
-            when :insert, :insert!, :upsert
-              parameters = [
-                create_param("attributes", type: "Hash"),
-                create_kw_opt_param("returning", type: returning_type, default: "nil"),
-              ]
-
-              # Bang methods don't have the `unique_by` parameter
-              unless bang_method?(method_name)
-                parameters << create_kw_opt_param("unique_by", type: unique_by_type, default: "nil")
-              end
-
-              association_relation_methods_module.create_method(
-                method_name.to_s,
-                parameters: parameters,
-                return_type: "ActiveRecord::Result",
-              )
-            when :proxy_association
-              # skip - private method
-            end
-          end
+          # skips insert/upsert methods - these methods' signatures aren't model-specific and don't need to be generated dynamically
+          # also skips proxy_association method - it's a private method
+          # but there could be other association methods that we need to generate
         end
 
         #: -> void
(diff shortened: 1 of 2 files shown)


lib/tapioca/dsl/compilers/active_record_relations.rb
```diff
@@ -602,51 +601,6 @@ def create_relation_methods
           end
         end
 
-        #: -> void
-        def create_association_relation_methods
```
paracycle, line 606: I would still like to keep this method, even if it is empty now, since there could be methods on association relation that don't exist on normal relations that we might want to cater for.

lib/tapioca/dsl/compilers/active_record_relations.rb
```diff
@@ -177,7 +177,6 @@ def decorate
           create_classes_and_includes
           create_common_methods
           create_relation_methods
-          create_association_relation_methods
```
paracycle, line 180: As stated above, I'd rather keep this call even if it is a no-op.

lib/tapioca/dsl/compilers/active_record_relations.rb
```diff
@@ -602,51 +601,6 @@ def create_relation_methods
...
         end
 
-        #: -> void
-        def create_association_relation_methods
-          returning_type = "T.nilable(T.any(T::Array[Symbol], FalseClass))"
-          unique_by_type = "T.nilable(T.any(T::Array[Symbol], Symbol))"
-
-          ASSOCIATION_METHODS.each do |method_name|
-            case method_name
-            when :insert_all, :insert_all!, :upsert_all
```
paracycle, line 612: It might be good to do the same treatment as in `create_common_methods` and make it explicit why we are not generating these association methods here.

paracycle (approved):

PR #2450, opened 2025-12-16
Add compiler for Rails.env methods

bdewater-thatch:
### Motivation
Why spend 10 seconds manually typing up shims if I can spend 10 minutes writing a compiler? Across all Tapioca users the math checks out: https://xkcd.com/1205/

### Implementation
I figured DEFAULT_ENVIRONMENTS can be skipped since https://github.com/Shopify/rbi-central/blob/7d98e7310d9ac5c6d34a48c05fb74f75dfc2d4ef/rbi/annotations/activesupport.rbi#L8-L21 already covers these.

### DIFF
--- a/spec/rails_spec_helper.rb
+++ b/spec/rails_spec_helper.rb
@@ -6,6 +6,15 @@ module Tapioca
     extend Tapioca::Helpers::Test::Content
 
     class << self
+      #: (String path) -> void
+      def define_fake_rails_app(path)
+        base_folder = Pathname.new(path)
+        config_class = Struct.new(:root)
+        config = config_class.new(base_folder)
+        app_class = Struct.new(:config)
+        Rails.application = app_class.new(config)
+      end
+
       def load_active_storage
         add_ruby_file("application.rb", <<~RUBY)
           ENV["DATABASE_URL"] = "sqlite3::memory:"
--- a/spec/tapioca/dsl/compilers/active_support_environment_inquirer_spec.rb
+++ b/spec/tapioca/dsl/compilers/active_support_environment_inquirer_spec.rb
@@ -19,7 +19,7 @@ module Tapioca
           describe "with a Rails app" do
             before do
               require "rails"
-              define_fake_rails_app
+              Tapioca::RailsSpecHelper.define_fake_rails_app(tmp_path("lib"))
               add_default_environment_files
             end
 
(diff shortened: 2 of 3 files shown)


spec/tapioca/dsl/compilers/active_support_environment_inquirer_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+                assert_equal(expected, rbi_for("ActiveSupport::EnvironmentInquirer"))
+              end
+            end
+          end
+        end
+
+        private
+
+        #: -> void
+        def define_fake_rails_app
```
KaanOzkan: This could be moved to a helper (like `spec/rails_spec_helper.rb`) and be shared with the fixtures_spec usage.

bdewater-thatch:
@KaanOzkan I addressed your feedback, PTAL

lib/tapioca/dsl/compilers/active_support_environment_inquirer.rb
```diff
@@ -0,0 +1,64 @@
...
+      # end
+      # ~~~
+      #: [ConstantType = singleton(::ActiveSupport::EnvironmentInquirer)]
+      class ActiveSupportEnvironmentInquirer < Compiler
+        extend T::Sig
+
+        # @override
+        #: -> void
+        def decorate
+          envs = Rails.root.join("config", "environments").glob("*.rb").map { |f| f.basename(".rb").to_s }.sort
```
amomchilov:
```suggestion
          envs = Rails.root.glob("config/environments/*.rb").map { |f| f.basename(".rb").to_s }.sort
```

spec/rails_spec_helper.rb
```diff
@@ -6,6 +6,15 @@ module RailsSpecHelper
     extend Tapioca::Helpers::Test::Content
 
     class << self
+      #: -> void
+      def define_fake_rails_app
+        base_folder = Pathname.new(tmp_path("lib"))
+        config_class = Struct.new(:root)
+        config = config_class.new(base_folder)
+        app_class = Struct.new(:config)
+        Rails.application = app_class.new(config)
```
amomchilov: Can we use real objects here?


```suggestion
        base_folder = Pathname.new(tmp_path("lib"))
        config = Rails::Application::Configuration.new(base_folder)
        app_class = Class.new(Rails::Application)
        Rails.application = app_class.new(config)
```

amomchilov (approved):
Very thorough, thanks for contributing this!

Two small requests and I'm good to merge

spec/rails_spec_helper.rb
bdewater-thatch: This changes `Rails.root` to return path of the Tapioca repo instead of `tmp_path` (as used by `add_content_file`), e.g. on my machine `#<Pathname:/Users/bdewater/src/github.com/thatch-health/tapioca>`, so the test breaks.

PR #2468, opened 2026-01-02
Fix binpath in CI error messages

amomchilov:
### Motivation

Fix the command suggested by this error message in Tapioca's CI:

```diff
bundle exec exe/tapioca gem --verify
Checking for out-of-date RBIs...

RBI files are out-of-date. In your development environment, please run:
-  `bin/tapioca gem`
+  `exe/tapioca gem`
Once it is complete, be sure to commit and push any changes

Reason:
  File(s) changed:
  - sorbet/rbi/gems/action_text-trix@2.1.16.rbi
  - sorbet/rbi/gems/bigdecimal@4.0.1.rbi
  - sorbet/rbi/gems/concurrent-ruby@1.3.6.rbi
  - sorbet/rbi/gems/i18n@1.14.8.rbi
  - sorbet/rbi/gems/loofah@2.25.0.rbi
  - sorbet/rbi/gems/rack-test@2.2.0.rbi
  - sorbet/rbi/gems/rackup@2.3.1.rbi
  - sorbet/rbi/gems/rails-dom-testing@2.3.0.rbi
  - sorbet/rbi/gems/rails-html-sanitizer@1.6.2.rbi
  - sorbet/rbi/gems/rdoc@7.0.3.rbi
  - sorbet/rbi/gems/uri@1.1.1.rbi
  - sorbet/rbi/gems/zeitwerk@2.7.4.rbi
```

### Implementation

I just conditionally change the value of `BINARY_FILE` on startup, but that involves an extra IO op to stat `Bundler.root + "tapioca.gemspec"`.

Another idea I had was to symlink from `bin/tapioca` to `exe/tapioca`

### Tests

Not tested :D

### DIFF
--- a/lib/tapioca.rb
+++ b/lib/tapioca.rb
@@ -17,7 +17,6 @@ module Tapioca
   TAPIOCA_DIR = "#{SORBET_DIR}/tapioca" #: String
   TAPIOCA_CONFIG_FILE = "#{TAPIOCA_DIR}/config.yml" #: String
 
-  BINARY_FILE = "bin/tapioca" #: String
   DEFAULT_POSTREQUIRE_FILE = "#{TAPIOCA_DIR}/require.rb" #: String
   DEFAULT_RBI_DIR = "#{SORBET_DIR}/rbi" #: String
   DEFAULT_DSL_DIR = "#{DEFAULT_RBI_DIR}/dsl" #: String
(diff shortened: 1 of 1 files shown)


lib/tapioca.rb
amomchilov, line 1: On a related note, should we change the output to include the `bundle exec` prefix?

That's defined by `default_command`:

https://github.com/Shopify/tapioca/blob/6df809efdf1706f727627d6f19a2e3559296d1c3/lib/tapioca/commands/command.rb#L35-L38

KaanOzkan: I use `bundle exec tapioca` and didn't run into this issue but I assume `bin/tapioca` was picked instead of `bundle exec` for a reason so I think your current fix is better.

KaanOzkan (approved):

paracycle (changes requested):
Instead of all this, let's run `bundle binstub tapioca` and commit the `bin/tapioca` file. 😉 

Using `tapioca` in the `tapioca` repo should be the same as using it in another application.

amomchilov:
Superseded by #2479

PR #2469, opened 2026-01-02
Silence expected warning about `__id__` definition

amomchilov:
### Motivation

Removes this line noise from CI logs:

```
/home/runner/work/tapioca/tapioca/spec/tapioca/runtime/reflection_spec.rb:55: warning: redefining '__id__' may cause serious problems
```

### Implementation

Hooks in the `Warning` callback chain, raises and exception, and ignores it if it's the one we expected.

### DIFF
--- a/spec/tapioca/runtime/reflection_spec.rb
+++ b/spec/tapioca/runtime/reflection_spec.rb
@@ -52,8 +52,18 @@ module Tapioca
         ::String
       end
 
-      def __id__ # rubocop:disable Naming/MethodName
-        1
+      begin
+        ::Warning.raise_warnings_as_exceptions do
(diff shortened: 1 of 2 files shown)


spec/tapioca/runtime/reflection_spec.rb
Morriar, line 116: Do we even need this test?

Morriar, line 135: Could we not override `__id__` and just test the id against the real one?


```suggestion
          refute_equal(foo.__id__, Runtime::Reflection.object_id_of(foo))
```

paracycle: I think the test is useful since it would error if someone removed the `bind_call` versions of the various look-ups.

PR #2480, opened 2026-01-06
Switch to Package URI for source location comments

paracycle:
### Motivation

Follow up to https://github.com/Shopify/ruby-lsp/pull/3807

### Implementation

- Vendor `package_url.rb`
- Allow `:` characters in `subpath` to not be URL encoded
- Regenerate all RBI files to update source location comments
- ???
- Profit ($$$)

### Tests

Update existing tests

### DIFF
--- a/sorbet/rbi/shims/erb.rbi
+++ b/sorbet/rbi/shims/erb.rbi
@@ -1,0 +1,3 @@
+# typed: true
+
+module ERB::Escape; end
--- a/sorbet/rbi/shims/set.rbi
+++ b/sorbet/rbi/shims/set.rbi
@@ -1,0 +1,4 @@
+# typed: true
+
+module Set::SubclassCompatible; end
+module Set::SubclassCompatible::ClassMethods; end


lib/tapioca/helpers/package_url.rb
```diff
@@ -0,0 +1,416 @@
+# typed: false
```
Morriar, line 1: Would it be easier to import the gem under a `vendor/` directory so it can be updated using the command line rather than copying files?

Maybe we should also avoid re-namespacing it under `Tapioca`?

paracycle: > Would it be easier to import the gem under a `vendor/` directory so it can be updated using the command line rather than copying files?

The upstream file has has only 2 changes in the last 5 years, so I doubt this will be a problem for us going forward.

> Maybe we should also avoid re-namespacing it under `Tapioca`?

I _do_ want to re-namespace it under Tapioca. If the upstream gem gets loaded along with Tapioca, not changing the namespace will result in various constant redefinition warnings.

paracycle: Forgot to include a link to file changes upstream: https://github.com/package-url/packageurl-ruby/commits/main/lib/package_url.rb

PR #2482, opened 2026-01-07
Support RBS signatures in check-shims command

sambostock:
### Motivation

The `check-shims` command detects duplicate definitions between shim files and generated RBIs. It correctly ignores duplicates when the shim adds a `sig` block (since the shim is adding type information, not just duplicating). However, it doesn't recognize RBS comment syntax (`#:`) as a signature, so shims using RBS syntax are incorrectly flagged as duplicates.

With Sorbet's [RBS support](https://sorbet.org/docs/rbs-support), users can now write type signatures using RBS comments instead of `sig` blocks. The `check-shims` command should treat these equivalently.

### Implementation

Modified `has_duplicated_methods_and_attrs?` in `RBIFilesHelper` to check for `RBI::RBSComment` instances in a node's `comments` array, in addition to checking the `sigs` array. The logic mirrors the existing signature handling:

- A shim with an RBS comment is not considered a duplicate of an untyped method (it's adding type info)
- A shim with a different RBS signature is not a duplicate (it's refining the type)
- A shim with an identical RBS signature is flagged as a duplicate

Added a small helper method `extract_rbs_comments` to extract RBS comments from a node.

### Tests

Added 3 new tests mirroring the existing `sig` block tests:
- `ignores duplicates that have an RBS signature`
- `ignores duplicates that have a different RBS signature`
- `detects duplicates that have the same RBS signature`

### DIFF
--- a/lib/tapioca/helpers/rbi_files_helper.rb
+++ b/lib/tapioca/helpers/rbi_files_helper.rb
@@ -239,6 +239,7 @@ module Tapioca
             return true if shim_or_todo_prop.sigs.any? { |sig| node.sigs.include?(sig) }
 
             # Another prop has the same RBS comment, we have a duplicate
+            # Note: can't use `intersect?` here because RBI::RBSComment doesn't implement `eql?`/`hash`
             other_rbs_comments = extract_rbs_comments(node)
             return true if shim_rbs_comments.any? { |rbs| other_rbs_comments.include?(rbs) }
           end
--- a/lib/tapioca/helpers/rbi_files_helper.rb
+++ b/lib/tapioca/helpers/rbi_files_helper.rb
@@ -217,7 +217,9 @@ module Tapioca
         shims_or_todos_props.each do |shim_or_todo_prop|
           other_nodes = extract_methods_and_attrs(nodes) - [shim_or_todo_prop]
 
-          if shim_or_todo_prop.sigs.empty?
+          shim_rbs_comments = extract_rbs_comments(shim_or_todo_prop)
+
+          if shim_or_todo_prop.sigs.empty? && shim_rbs_comments.empty?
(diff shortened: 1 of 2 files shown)


lib/tapioca/helpers/rbi_files_helper.rb
```diff
@@ -235,6 +237,10 @@ def has_duplicated_methods_and_attrs?(nodes, shims_or_todos)
           other_nodes.each do |node|
             # Another prop has the same sig, we have a duplicate
             return true if shim_or_todo_prop.sigs.any? { |sig| node.sigs.include?(sig) }
+
+            # Another prop has the same RBS comment, we have a duplicate
+            other_rbs_comments = extract_rbs_comments(node)
+            return true if shim_rbs_comments.any? { |rbs| other_rbs_comments.include?(rbs) }
```
amomchilov, line 244:
```suggestion
            return true if shim_rbs_comments.intersect?(other_rbs_comments)
```

amomchilov (approved):

lib/tapioca/helpers/rbi_files_helper.rb
sambostock: Unfortunately `intersect?` won't work here because `RBI::RBSComment` implements `==` but not `eql?`/`hash`:

```ruby
c1 = RBI::RBSComment.new('() -> void')
c2 = RBI::RBSComment.new('() -> void')

c1 == c2           # => true
[c1].include?(c2)  # => true  (uses ==)
[c1].intersect?([c2])  # => false (uses eql?/hash)
```

I've added a comment to explain this.

egiurleo: @sambostock Not blocking this PR, but could we make `RBI::RBSComment` implement `.eql?` or is there something preventing that?

egiurleo (approved):

PR #2489, opened 2026-01-15
Use type builder to generate DSL RBI files

paracycle:


paracycle:
> What do you think of first doing a refactor to allow compilers to use 2 type builders, one for RBI, one for RBS.

I don't understand why we need 2 type builders, though. I thought the whole point of the type builder abstraction was to be able to generate RBI syntax or RBS syntax from the same types, and I also got some of that working as well.

PR #2491, opened 2026-01-16
Upgrade minitest from 5.27.0 to 6.0.1 for supported Rails (8.1, main)

larouxn:
### Motivation
Came across this `gem "minitest", "< 6.0"` lock in `Gemfile` and decided to remove it because Rails has since added support for Minitest 6 in Rails 8.1.2 which we upgraded to in https://github.com/Shopify/tapioca/pull/2487. Even Rails 8.0.x now supports Minitest 6 but currently only in 8-0-stable branch as there hasn't been a release yet, hence the conditional.

Follow up to https://github.com/Shopify/tapioca/pull/2458

References
- https://github.com/minitest/minitest/issues/1045
- https://github.com/rails/rails/releases/tag/v8.1.2
- https://github.com/Shopify/tapioca/pull/2487
  - With this merged we can remove the Rails bump from this PR. :)

### Implementation
Deleted Minitest version lock and then `bundle update minitest --conservative`. ᕕ( ᐛ )ᕗ

### Tests :tophat: 

Hit _some_ errors. Maybe flake or perhaps an issue with my machine. Would like to see in CI.

image


larouxn:
Alrighty, I have some possible fixes for minitest 6 compatibility (https://github.com/Shopify/tapioca/pull/2491/changes/e6f0167aea822e53a7eeb706837897010b905b04). Have tested locally and only the protobuf spec fails 12 times on both Rails 8.1 and Rails 8.0, so likely my machine as all other tests pass. All that said, ideally we wouldn't need this many conditionals. Thus, if we want to wait for the following, that's fine by me.
- Rails 8.0's next release (8.0.5?), which will likely include the fix for minitest 6 support
  - https://github.com/rails/rails/commit/ec62932ee7d31e0ef870e61c2d7de2c3efe3faa6
- minitest-reporters next release, which will likely include fixes for minitest 6 support
  - https://github.com/minitest-reporters/minitest-reporters/issues/368

larouxn:
Alrighty, I have some fixes for minitest 6 compatibility (https://github.com/Shopify/tapioca/pull/2491/changes/e6f0167aea822e53a7eeb706837897010b905b04). Tested locally and only the protobuf spec fails, same 12 failures on both Rails 8.1 and Rails 8.0, and all other tests pass so likely cause is my machine. Additionally ran this branch through CI on my fork and [everything passed](https://github.com/larouxn/tapioca/actions/runs/21156447648), aside from 3 failures and 1 error on Rails main for all Ruby versions ([example](https://github.com/larouxn/tapioca/actions/runs/21156447648/job/60842150455#step:6:138)) re: Rails main being a moving target. ([quick fix for reference](https://github.com/Shopify/tapioca/commit/bf3b093e598d154b6c107e4082fe1acfd84e584b?diff=split))

That said, ideally we wouldn't need so many conditionals. If we want to wait for the following so we don't, that's fine.
- Rails 8.0's next release (8.0.5?), which will likely include the fix for minitest 6 support
  - https://github.com/rails/rails/commit/ec62932ee7d31e0ef870e61c2d7de2c3efe3faa6
- minitest-reporters next release, which will likely include fixes for minitest 6 support
  - https://github.com/minitest-reporters/minitest-reporters/issues/368

Issue #343, opened 2021-06-01
Generate gems RBIs for vendored gems


Morriar:
Some projects use vendored gems, for example under the `bundler/` directory.

Such gems don't appear in the Gemfile and thus Tapioca doesn't generate any RBI for them.

Since we generally exclude the vendored directory from type-checking, we're getting missing constants and methods.

We should find a way to make Tapioca generate RBI files for vendored gems.

dirceu:
From Ufuk:

> It looks like Tapioca does the right thing about vendored gems because we explicitly check for it: https://github.com/Shopify/tapioca/blob/3528c25b308fafbc9149f1f693529100f9dcc6fd/lib/tapioca/helpers/gem_helper.rb#L16-L19
> 
> On the flip side, we should not be processing any gem that "lives" inside the repo, since it will be type-checked along with the repo itself anyway

I'm gonna go ahead and close this issue; we can reopen if we find a specific use case to replicate the problem.

PR #2492, opened 2026-01-16
Generate RBIs for vendored gems

alex-tan:
### Motivation

Addresses #343

We have _some_ vendored gems in `vendor/gems`. Note that this is not our BUNDLE_PATH. We just point to some gems by using the `path:` option in the Gemfile. Without this change tapioca doesn't compile rbis for these gems.

### Implementation

There was an ignore condition that ignored any gems in the app dir. This removes it.

### Tests

Added

### DIFF
--- a/spec/tapioca/gemfile_spec.rb
+++ b/spec/tapioca/gemfile_spec.rb
@@ -137,6 +137,14 @@ module Tapioca
       RbConfig::CONFIG["rubylibprefix"] = original_rubylibprefix
     end
 
+    it "ignore? returns false for vendored gems in the app directory" do
+      # Create a vendored gem within the project directory
+      foo_gem = mock_gem("foo", "0.0.1", path: "#{@project.absolute_path}/vendor/gems")
+      foo_spec = make_spec(foo_gem)
+
+      refute(foo_spec.ignore?(@project.absolute_path))
+    end
+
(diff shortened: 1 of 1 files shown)


lib/tapioca/gemfile.rb
```diff
@@ -150,7 +150,7 @@ def ==(other)
 
       #: (String gemfile_dir) -> bool
       def ignore?(gemfile_dir)
-        gem_ignored? || gem_in_app_dir?(gemfile_dir, full_gem_path)
```
paracycle, line 153: I am afraid this won't be enough since the condition exists for a specific reason: In our core monolith at Shopify we have what we call "in-repo gems" which are gems that exist inside folders under the app path. Since they are under the app path, they already get type-checked by Sorbet, so we should not be generating extra RBI files for them as well. For all practical purposes, the code in those paths in our Core monolith is a part of the monolith even though it gets loaded as a gem.

Your case is similar but somewhat different. You load some actual gems from the `vendor` folder inside your app path, so you want to generate RBIs for them, but removing this condition will break our use-case.

alex-tan: Ah I see. I tried adding typechecking to these gems and was able to work through the issues. Thanks 👍

PR #2496, opened 2026-01-21
Use nix to manage development environment

rafaelfranca:
This will allow us to have the same environment as the rest of the company.

### DIFF
--- a/dev.yml
+++ b/dev.yml
@@ -1,10 +1,14 @@
 name: tapioca
 
+nix: true
 type: ruby
 
 up:
   - ruby
   - bundler
+  - packages:
+      - libyaml
+      - zlib
 
 env:
   SRB_SKIP_GEM_RBIS: 'true'


Morriar (approved):

dev.yml
```diff
@@ -1,10 +1,14 @@
...
 
+nix: true
 type: ruby
 
 up:
   - ruby
   - bundler
+  - packages:
+      - libyaml
+      - zlib
```
paracycle, line 11: Are these needed because `ruby` depends on them? If so, isn't that something that should be baked into the requirements of the `ruby` derivation?

rafaelfranca: `libyaml` because of `psych` and `zlib` because of `faraday-gzip` that depends on the `zlib` gem. If those gems weren't being required in the Gemfile.lock those packages would not be needed, so no, I don't think it should be in the ruby derivation.

paracycle: But `psych` and `zlib` are both default gems, so they should be coming with a Ruby installation, and their dependencies should also be pulled in by default, no? (I am not sure if the `zlib` gem has a dependency on the `zlib` package, but `psych` should definitely be pulling `libyaml` in)

rafaelfranca: `libyaml` and `zlib` as just needed to build Ruby. After it is built you don't need it anymore. You only need again if you aren't using the Ruby default gems and is building it from source again, which is the case here, no? Now, why rubygems is building from source and not using the default gem I don't know. Maybe different versions?

paracycle: Sorry, I don't mean to hold up this PR for this, we can discuss async as well, but my understanding of the Nix store was that dependencies of things that you installed persisted in the store, so I was expecting at least `libyaml` to be present after installing `ruby`. But, maybe my assumptions are wrong.

rafaelfranca: Yeah, that is what I thought. The derivation has it as build packages. After it is built it is gone https://github.com/bobvanderlinden/nixpkgs-ruby/blob/master/ruby/package-fn.nix#L76-L79.

Maybe we should add to our own ruby flake, but we don't have one right now, AFAIK

PR #2497, opened 2026-01-21
Make ActiveRecord delete/destroy_by argument a rest parameters

caleb:
### Motivation
The `delete_by` and `destroy_by` ActiveRecord methods take multiple arguments (like `where`) instead of a single argument.

### Implementation
I just replaced `create_param` with `create_rest_param` and updated the tests

### Tests
I updated the tests to expect the rest argument instead of a single argument


caleb:
@rafaelfranca Ha! Yeah, it's there now.

paracycle (approved):
Looks good to me, thanks for fixing this.

PR #2522, opened 2026-03-04
Pre-load AR schemas and disconnect before forking DSL workers

paracycle:
### Motivation

With multiple workers, we've been hitting connection limits on CI when verifying DSL RBI files. This PR attempts to fix that problem.

### Implementation

Pre-load the schema for all ActiveRecord models in the parent process before forking parallel workers, so the schema cache is inherited via copy-on-write. Then clear all database connections so forked children don't inherit stale connections from the parent. Together, this should ensure that child processes do not make any database connections, since most of our introspection should just be hitting the schema.

### Tests

No new tests.

### DIFF
--- a/lib/tapioca/dsl/pipeline.rb
+++ b/lib/tapioca/dsl/pipeline.rb
@@ -73,6 +73,21 @@ module Tapioca
 
         if defined?(::ActiveRecord::Base) && constants_to_process.any? { |c| ::ActiveRecord::Base > c }
           abort_if_pending_migrations!
+
+          # Pre-load the schema for all AR models we're about to process. This populates the schema cache in the
+          # parent process so that forked workers inherit it via copy-on-write and don't each need to establish their
+          # own database connection just to load schema information.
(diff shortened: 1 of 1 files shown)


lib/tapioca/dsl/pipeline.rb
```diff
@@ -73,6 +73,21 @@ def run(&blk)
...
+            next unless ::ActiveRecord::Base > c
+
+            ar_model = c #: as singleton(::ActiveRecord::Base)
+            ar_model.load_schema
+          end
+
+          # Disconnect all database connections before forking workers. The migration check and schema pre-loading
+          # above may have established connections; clearing them ensures forked children don't inherit stale
+          # connections from the parent, avoiding exceeding connection limits (e.g. Semian ticket counts).
+          ::ActiveRecord::Base.connection_handler.clear_all_connections!
```
Morriar, line 90: Do we need this even with `@number_of_workers == 1`?

PR #2524, opened 2026-03-05
Use Rubydex to document symbols instead of YARD

Morriar:
This PR replaces YARD/yard-sorbet with [Rubydex](https://github.com/Shopify/rubydex) for generating documentation in RBI files.

* Add `rubydex` gem dependency
* Replace the `YardDoc` listener with a new `Documentation` listener backed by Rubydex's static index

The impact on performance is pretty nice. `tapioca gems --all` on Tapioca itself is now 10x faster and requires 4x less memory.

image


sorbet/rbi/gems/aasm@5.5.2.rbi
```diff
@@ -107,18 +105,20 @@ class AASM::Base
...
+  # [1] options (or nil)
+  # or
+  # [0] state
+  # [1..] state
   #
   # pkg:gem/aasm#lib/aasm/base.rb:90
   def state(*args); end
 
-  # Returns the value of attribute state_machine.
-  #
```
amomchilov: Any explanation for these changes?

lib/tapioca/gem/pipeline.rb
```diff
@@ -31,7 +31,10 @@ def initialize(
         @events = [] #: Array[Gem::Event]
 
         @payload_symbols = Static::SymbolLoader.payload_symbols #: Set[String]
-        @bootstrap_symbols = load_bootstrap_symbols(@gem) #: Set[String]
+        gem_graph = Static::SymbolLoader.graph_from_paths(@gem.files) #: Rubydex::Graph
+        gem_symbols = gem_graph.declarations.map(&:name).to_set
```
paracycle: This part of this PR feels out of scope for just switching the document parser backend. Here, we are also replacing the tool that does the gem symbol indexing, but I am not sure if we have full feature parity at that level yet.

How confident are we that this won't cause actual RBI generation problems in the wild? Can we not leave this alone, and make the `Documentation` class do a reindexing with Rubydex for documentation only?

lib/tapioca/static/symbol_loader.rb
```diff
@@ -38,16 +46,12 @@ def engine_symbols(gem)
...
           Set.new
         end
 
-        #: (Gemfile::GemSpec gem) -> Set[String]
-        def gem_symbols(gem)
-          symbols_from_paths(gem.files)
-        end
-
         #: (Array[Pathname] paths) -> Set[String]
         def symbols_from_paths(paths)
```
paracycle, line 60: Is there any code left that's still using this method?

lib/tapioca/gem/listeners/documentation.rb
```diff
@@ -0,0 +1,95 @@
...
+          # only under the instance method name.
+          if declaration.nil? && name.include?("::<")
+            declaration = @gem_graph[name.sub(/::<[^>]+>#/, "#")]
+          end
+          return [] unless declaration
+
+          # Use comments from the first definition that has comments to avoid duplicates
+          # when a method is defined in multiple places with identical documentation.
+          comments = declaration.definitions.lazy.map(&:comments).find { |c| c.to_a.any? }
+          return [] unless comments
```
vinistock: I think I would merge the documentation from all definitions and use `uniq!` to avoid duplicates. There are legitimate cases where people document definitions for the same thing separately.


```suggestion
          comments = declaration.definitions.flat_map(&:comments)
          comments.uniq!
          return comments if comments.empty?
```

lib/tapioca/gem/pipeline.rb
Morriar: Removed for now but I think we should put it back in a separate PR

lib/tapioca/static/symbol_loader.rb
Morriar: This was a valid suggestion until we remove the symbols bootstraping

paracycle (approved):

lib/tapioca/gem/listeners/documentation.rb
```diff
@@ -0,0 +1,94 @@
...
+          comments = declaration.definitions.flat_map(&:comments)
+          comments.uniq!
+          return [] if comments.empty?
+
+          lines = comments
+            .map { |comment| comment.string.gsub(/^#+ ?/, "") }
+            .reject { |line| IGNORED_COMMENTS.any? { |comment| line.include?(comment) } || rbs_comment?(line) }
+
+          # Strip leading and trailing blank lines, matching YARD's behavior
+          lines = lines.drop_while(&:empty?).reverse.drop_while(&:empty?).reverse
```
paracycle: I think this would be way more performant:

```suggestion
          lines = lines.reverse_each.drop_while(&:empty?).reverse_each.drop_while(&:empty?)
```
since it won't actually reverse the array elements twice, just iterate from the end and then from the beginning.

PR #2527, opened 2026-03-08
Remove ruby < 3.2 `attached_object` workaround

Earlopain:
### Motivation
Tapioca now requires ruby >= 3.2 so this is dead code.

### Implementation
Just move it back inline like it was before.

### Tests
No behaviour change


lib/tapioca/runtime/reflection.rb
```diff
@@ -215,6 +205,12 @@ def resolve_loc(locations)
         SourceLocation.from_loc([file, resolved_loc.lineno])
       end
 
+      #: (Class singleton_class) -> T::Module[top]?
+      def attached_class_of(singleton_class)
+        result = singleton_class.attached_object
+        Module === result ? result : nil
+      end
```
KaanOzkan: Nit: Could we move this to right after `superclass_of` above? I think it fits better.

lib/tapioca/runtime/reflection.rb
```diff
@@ -215,6 +205,12 @@ def resolve_loc(locations)
         SourceLocation.from_loc([file, resolved_loc.lineno])
       end
 
+      #: (Class singleton_class) -> T::Module[top]?
```
KaanOzkan:
```suggestion
      #: (Class[top] singleton_class) -> T::Module[top]?
```

Earlopain: Nice, thanks for the help

PR #2535, opened 2026-03-16
Migrate release workflow to use trusted publisher

st0012:
### Motivation

Replace the current release workflow with the trusted publisher pattern used by other repos like `Shopify/spoom` and `Shopify/rbi`. The new workflow publishes the gem to RubyGems.org via OIDC (no API key needed) and then creates a GitHub release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,27 +1,52 @@
-name: Release
+name: Release gem on RubyGems.org and create GitHub release
 
 on:
   push:
     tags:
-      - v*
+      - 'v[0-9]+.[0-9]+.[0-9]+'
 
 jobs:
-  publish:
+  release_gem:
+    name: Release gem on RubyGems.org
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -1,27 +1,52 @@
...
+    permissions:
+      contents: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Create GitHub release
+        run: |
+          tag_name="$(git describe --tags --abbrev=0)"
```
paracycle, line 49: Hmm, since this action will be triggered with a `tag` push, I wonder if there is no way to read the tag that triggered the push instead of trying to find it again in the repository.

paracycle: Ah, it is actually what we've been using all along: `github.ref` or `github.ref_name`

PR #2537, opened 2026-03-18
Avoid duplicating definition of ActiveRecord::Relation.destroy_all

reu:
### Motivation

`destroy_all` is defined both as a hardcoded common method (line 612) and dynamically via the `RELATION_METHODS` loop (line 862), since `ActiveRecord::Relation` includes `destroy_all` as an instance method. This results in a duplicated `def destroy_all; end` signature in the generated RBI files.

### Implementation

Added a guard (`unless RELATION_METHODS.include? :destroy_all`) to the hardcoded `destroy_all` definition so it is only generated when `destroy_all` is not already part of `RELATION_METHODS`. When ActiveRecord includes it in `Relation.instance_methods(false)`, the dynamic loop handles it and the hardcoded one is skipped, avoiding the duplicate.

Updated the expected output in the spec to remove the duplicated `destroy_all` entries.

### Tests

Updated existing specs to reflect the corrected (non-duplicated) output.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_relations.rb
+++ b/lib/tapioca/dsl/compilers/active_record_relations.rb
@@ -612,7 +612,7 @@ module Tapioca
           create_common_method(
             "destroy_all",
             return_type: "T::Array[#{constant_name}]",
-          )
+          ) unless RELATION_METHODS.include? :destroy_all
 
           FINDER_METHODS.each do |method_name|
             case method_name
--- a/spec/tapioca/dsl/compilers/active_record_relations_spec.rb
+++ b/spec/tapioca/dsl/compilers/active_record_relations_spec.rb
@@ -147,9 +147,6 @@ module Tapioca
                     sig { returns(T::Array[::Post]) }
                     def destroy_all; end
 
-                    sig { returns(T::Array[::Post]) }
-                    def destroy_all; end
-
                     sig { params(args: T.untyped).returns(T::Array[::Post]) }
                     def destroy_by(*args); end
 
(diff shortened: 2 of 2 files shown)


lib/tapioca/dsl/compilers/active_record_relations.rb
```diff
@@ -612,7 +612,7 @@ def create_common_methods
           create_common_method(
             "destroy_all",
             return_type: "T::Array[#{constant_name}]",
-          )
+          ) unless RELATION_METHODS.include? :destroy_all
```
KaanOzkan, line 615: I think it's better to remove this completely, looks like it's been [there](https://github.com/rails/rails/blame/ef95886ef996db20ddc05f2d1c8c7342b05121e6/activerecord/lib/active_record/relation.rb#L1023) since Rails 3. So I don't see how this `include?` will fail for Tapioca's `Rails > 8.0` requirement.

```
create_common_method(
  "destroy_all",
  return_type: "T::Array[#{constant_name}]",
)
```

KaanOzkan (commented):
Good find @reu, sorry for the delay, I must've missed the notification.

lib/tapioca/dsl/compilers/active_record_relations.rb
Morriar: Since there's already a `when :delete_all, :destroy_all` case in the `RELATION_METHODS` loop (line 939) that generates the exact same signature, I wonder if we could just remove the hardcoded definition entirely instead of guarding it?

```suggestion

```

`destroy_all` has been on `ActiveRecord::Relation` for a long time, so the dynamic path should always handle it. It would avoid leaving a dead code path behind.

Issue #2542, opened 2026-03-23
Generated RBI signatures missing `type_parameters` declaration


dejmedus:
When a compiler (ex. [ActiveSupportCurrentAttributes](https://github.com/Shopify/tapioca/blob/d27f9cb4395d7d79a5b0dd657a5a44ec7d1cb7a3/lib/tapioca/dsl/compilers/active_support_current_attributes.rb#L86)) uses [create_method_from_def](https://github.com/Shopify/tapioca/blob/daa4f71223e267dda073233cbb37647934409710/lib/tapioca/dsl/compiler.rb#L141), the generated RBI signature may be missing its `type_parameters` declaration ([which sorbet considers invalid](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0A%0A%23%20This%20is%20invalid%20%0Asig%20%7B%20params%28t%3A%20T.untyped%2C%20blk%3A%20T.proc.returns%28T.type_parameter%28%3AT%29%29%29.returns%28T.type_parameter%28%3AT%29%29%20%7D%0Adef%20with_type_bad%28t%2C%20%26blk%29%0A%20%20blk.call%0Aend%0A%0A%23%20This%20is%20valid%20%E2%80%94%20declares%20with%20type_parameters%0Asig%20%7B%20type_parameters%28%3AT%29.params%28t%3A%20T.untyped%2C%20blk%3A%20T.proc.returns%28T.type_parameter%28%3AT%29%29%29.returns%28T.type_parameter%28%3AT%29%29%20%7D%0Adef%20with_type_good%28t%2C%20%26blk%29%0A%20%20blk.call%0Aend)). Example:

```ruby
class Current < ActiveSupport::CurrentAttributes
    sig { type_parameters(:T).params(t: T.untyped, blk: T.proc.returns(T.type_parameter(:T))).returns(T.type_parameter(:T)) }
    def with_type(t, &blk)
      # ...
    end
```

generates this RBI:

```ruby
# missing type_parameters(:T)
sig { params(t: T.untyped, blk: T.proc.returns(T.type_parameter(:T))).returns(T.type_parameter(:T)) }
def with_type(t, &blk); end
```

We'll need to pass any type params through `create_method_from_def` and `create_method` so that `RBI::Sig.new` will have access to them

PR #2550, opened 2026-03-24
Include type_parameters declarations in signatures

dejmedus:
- closes https://github.com/Shopify/tapioca/issues/2542

### Motivation

Previously, when a compiler used [create_method_from_def](https://github.com/Shopify/tapioca/blob/daa4f71223e267dda073233cbb37647934409710/lib/tapioca/dsl/compiler.rb#L141) (ex. [ActiveSupportCurrentAttributes](https://github.com/Shopify/tapioca/blob/d27f9cb4395d7d79a5b0dd657a5a44ec7d1cb7a3/lib/tapioca/dsl/compilers/active_support_current_attributes.rb#L86)) any method signatures would be generated without `type_parameters` declarations. This created [invalid](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0A%0A%23%20This%20is%20invalid%20%0Asig%20%7B%20params%28t%3A%20T.untyped%2C%20blk%3A%20T.proc.returns%28T.type_parameter%28%3AT%29%29%29.returns%28T.type_parameter%28%3AT%29%29%20%7D%0Adef%20with_type_bad%28t%2C%20%26blk%29%0A%20%20blk.call%0Aend%0A%0A%23%20This%20is%20valid%20%E2%80%94%20declares%20with%20type_parameters%0Asig%20%7B%20type_parameters%28%3AT%29.params%28t%3A%20T.untyped%2C%20blk%3A%20T.proc.returns%28T.type_parameter%28%3AT%29%29%29.returns%28T.type_parameter%28%3AT%29%29%20%7D%0Adef%20with_type_good%28t%2C%20%26blk%29%0A%20%20blk.call%0Aend) signatures when type params were used in the method sig

```diff
- sig { params(a: T.type_parameter(:U)).returns(T.type_parameter(:U)) }
+ sig { type_parameters(:U).params(a: T.type_parameter(:U)).returns(T.type_parameter(:U)) }
```

### Implementation

This PR moves out `TYPE_PARAMETER_MATCHER` and the type param extraction logic from gem `sorbet_signatures.rb` file to `rbi_helper.rb` so that they can also be used by dsl. This way `create_method_from_def` can pass any type params to `create_method` so that `RBI::Sig.new` will be able to include them in the generated signature

### Tests

- in `compiler_spec.rb` `compiles a class with methods that have signatures`:
   - updated a case to add `type_parameters` declaration
   - added new case to test more complex type param use
- added a method with type params to `active_support_current_attributes_spec.rb` `copies types from instance methods to class methods`

### DIFF
--- a/lib/tapioca/helpers/rbi_helper.rb
+++ b/lib/tapioca/helpers/rbi_helper.rb
@@ -94,6 +94,13 @@ module Tapioca
       end
     end
 
+    TYPE_PARAMETER_MATCHER = /T\.type_parameter\(:?([[:word:]]+)\)/
+
+    #: (Array[String] type_strings) -> Array[String]
+    def extract_type_parameters(type_strings)
+      type_strings.join(", ").scan(TYPE_PARAMETER_MATCHER).flatten.uniq
+    end
+
     #: (String name) -> bool
     def valid_method_name?(name)
       Prism.parse_success?("def self.#{name}(a); end")
--- a/lib/tapioca/dsl/compiler.rb
+++ b/lib/tapioca/dsl/compiler.rb
@@ -139,11 +139,16 @@ module Tapioca
 
       #: (RBI::Scope scope, (Method | UnboundMethod) method_def, ?class_method: bool) -> void
       def create_method_from_def(scope, method_def, class_method: false)
+        parameters = compile_method_parameters_to_rbi(method_def)
+        return_type = compile_method_return_type_to_rbi(method_def)
+        type_params = extract_type_parameters(parameters.map(&:type).push(return_type))
+
         scope.create_method(
           method_def.name.to_s,
-          parameters: compile_method_parameters_to_rbi(method_def),
-          return_type: compile_method_return_type_to_rbi(method_def),
+          parameters: parameters,
+          return_type: return_type,
           class_method: class_method,
+          type_params: type_params,
         )
       end
 
--- a/lib/tapioca/gem/listeners/sorbet_signatures.rb
+++ b/lib/tapioca/gem/listeners/sorbet_signatures.rb
@@ -8,8 +8,6 @@ module Tapioca
         include Runtime::Reflection
         include RBIHelper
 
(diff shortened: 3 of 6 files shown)


lib/tapioca/helpers/rbi_helper.rb
```diff
@@ -94,6 +94,13 @@ def as_non_nilable_type(type)
       end
     end
 
+    TYPE_PARAMETER_MATCHER = /T\.type_parameter\(:?([[:word:]]+)\)/
+
+    #: (Array[String] type_strings) -> Array[String]
+    def extract_type_parameters(type_strings)
+      type_strings.join(", ").scan(TYPE_PARAMETER_MATCHER).flatten.uniq
+    end
```
paracycle, line 102: I'd missed this PR earlier, so sorry for the delayed review, but it seems like this functionality should be a part of the inner working of `create_method` or `RBI::Sig` construction, since all the inputs to this method seem to always be `params` and `return_type`.

I suggest:
1. Reverting the signature of `create_method` to remove the `type_params` parameter
2. Adding a public method to `RBI::Sig` for doing this calculation and for setting `type_params`. Something like `set_type_params_from_params`. Btw, we don't need to pass `return_type` to that calculation since there is no way that we can have a type param only in the out position, that means the signature is broken in the first place.
3. Making `create_method` method call this method on the created sig after constructing it.

dejmedus: Just saw you beat me to this, thank you! (+ thank you for the detailed feedback)

> Btw, we don't need to pass return_type to that calculation since there is no way that we can have a type param only in the out position, that means the signature is broken in the first place.

Ah, that makes sense

paracycle: I still left the return type in, just in case, so that we don't end up generating broken signatures, even if the signature itself would be meaningless like that.

PR #2566, opened 2026-03-31
Encapsulate `type_param` building in `create_method`

paracycle:
### Motivation

Addressing the comment: https://github.com/Shopify/tapioca/pull/2550#discussion_r3002615512

### Implementation

We can completely encapsulate the `type_param` building inside of `create_method` since we have all the data that we need to build that already passed into the method.

### Tests

No added tests

### DIFF
--- a/lib/tapioca/gem/listeners/sorbet_signatures.rb
+++ b/lib/tapioca/gem/listeners/sorbet_signatures.rb
@@ -40,7 +40,7 @@ module Tapioca
           sig.return_type = return_type
           @pipeline.push_symbol(return_type)
 
-          sig.type_params.concat(extract_type_parameters(parameter_types.values.map(&:to_s).push(return_type)))
+          sig.type_params.concat(extract_type_parameters(parameter_types.values.map(&:to_s).append(return_type)))
 
           case signature.mode
           when "abstract"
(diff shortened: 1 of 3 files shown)


lib/tapioca/gem/listeners/sorbet_signatures.rb
```diff
@@ -40,7 +40,7 @@ def compile_signature(signature, parameters)
           sig.return_type = return_type
           @pipeline.push_symbol(return_type)
 
-          sig.type_params.concat(extract_type_parameters(parameter_types.values.map(&:to_s).push(return_type)))
+          sig.type_params.concat(extract_type_parameters(parameter_types.values.map(&:to_s).append(return_type)))
```
dejmedus, line 43: Question for my notes: is `.append` always preferred over `.push`?

dejmedus (commented):
Thank you! This is much cleaner than me passing everything down

lib/tapioca/gem/listeners/sorbet_signatures.rb
paracycle: Not really, I think this is more about the "intent" of the code, at least for me.

For me `push` is for appending an element to the end of an array, where as `append` is for adding extra elements to the array. For the former, I am not sure if the return value would be the array itself of the elements that are pushed to it, but for the latter I am more confident that the array will be returned.

PR #2567, opened 2026-04-01
Use Prism in Sorbet calls (if enabled)

amomchilov:
### Motivation  
  
We should use the Prism mode in repos that have it enabled in Sorbet (with `--parse=prism` in their `sorbet/config`),

### Implementation  
  
Added a simple little`SorbetConfig` class which parses the config file, and exposes nicer APIs for reading its various values.  
  
What's the right place to put this? Perhaps in Spoom? 

### Performance

Shaves off ~1.5 seconds (5%) on the Shopify core monolith.

```
$ hyperfine --warmup 5 --runs 20 \
  -n "legacy" '~/sorbet-master --no-config --cache-dir= --parser=original --print=symbol-table-json --quiet --stop-after=namer .' \
  -n "prism"  '~/sorbet-master --no-config --cache-dir= --parser=prism    --print=symbol-table-json --quiet --stop-after=namer .'

Benchmark 1: "legacy, no cache"
  Time (mean ± σ):     28.100 s ±  0.795 s    [User: 64.862 s, System: 18.303 s]
  Range (min … max):   27.020 s … 29.405 s    20 runs

Benchmark 2: "prism, no cache"
  Time (mean ± σ):     26.880 s ±  0.494 s    [User: 44.113 s, System: 18.467 s]
  Range (min … max):   25.809 s … 27.619 s    20 runs

Summary
  "prism" ran
    1.05 ± 0.04 times faster than "legacy"
```

### Tests

Added.

PR #2568, opened 2026-04-01
Use Sorbet cache (if enabled)

amomchilov:
### Motivation  
  
If a repo has enabled the Sorbet cache, we should use it.

### Implementation  
  
Builds upon the `SorbetCache` #2568.

### Performance

Shaves off around ~3s on the Shopify core monolith.

The cache stores the desugared ASTs, after the parsing, RBS rewriting, and desugaring is done. If there's a cache hit, no parsing happens, so that's why you see the legacy and Prism parsers have the same time. So this mitigates some of the benefit of #2567, but Prism is still helpful when there's a cache miss.

```
hyperfine --warmup 5 --runs 20 \
  -n "prism, with cache"  '~/sorbet-master --no-config --cache-dir=tmp/sorbet_cache --parser=prism    --print=symbol-table-json --quiet --stop-after=namer .' \
  -n "legacy, with cache" '~/sorbet-master --no-config --cache-dir=tmp/sorbet_cache --parser=original --print=symbol-table-json --quiet --stop-after=namer .' \
  -n "legacy, no cache"   '~/sorbet-master --no-config --cache-dir=                 --parser=original --print=symbol-table-json --quiet --stop-after=namer .' \

  -n "prism, no cache"    '~/sorbet-master --no-config --cache-dir=                 --parser=prism    --print=symbol-table-json --quiet --stop-after=namer .'

Benchmark 1: "prism, with cache"
  Time (mean ± σ):     24.932 s ±  0.687 s    [User: 23.678 s, System: 18.433 s]
  Range (min … max):   23.812 s … 26.870 s    20 runs
  
Benchmark 2: "legacy, with cache"
  Time (mean ± σ):     25.031 s ±  0.354 s    [User: 23.550 s, System: 18.331 s]
  Range (min … max):   24.518 s … 25.917 s    20 runs

Benchmark 3: "prism, no cache"
  Time (mean ± σ):     26.880 s ±  0.494 s    [User: 44.113 s, System: 18.467 s]
  Range (min … max):   25.809 s … 27.619 s    20 runs
  
Benchmark 4: "legacy, no cache"
  Time (mean ± σ):     28.100 s ±  0.795 s    [User: 64.862 s, System: 18.303 s]
  Range (min … max):   27.020 s … 29.405 s    20 runs


Summary
  "prism, with cache" ran
    1.00 ± 0.03 times faster than "legacy, with cache"
    1.08 ± 0.04 times faster than "prism, no cache"
    1.13 ± 0.04 times faster than "legacy, no cache"
```

### Tests

Added.

### DIFF
--- a/lib/tapioca/helpers/sorbet_helper.rb
+++ b/lib/tapioca/helpers/sorbet_helper.rb
@@ -39,15 +39,24 @@ module Tapioca
           end.to_h #: Hash[String, String | bool | nil]
 
           new(
+            # `--cache-dir=` disables the cache, modeled here with a `nil` value
+            cache_dir: if (value = options["--cache-dir"]).is_a?(String)
+                         value.empty? ? nil : value
+                       end,
+
             parser: options["--parser"] == "prism" ? :prism : :original,
           )
         end
       end
 
-      #: (parser: Symbol) -> void
-      def initialize(parser:)
+      #: (cache_dir: String?, parser: Symbol) -> void
+      def initialize(cache_dir:, parser:)
+        @cache_dir = cache_dir #: String?
         @parser = parser #: Symbol
       end
+
+      #: String?
+      attr_reader :cache_dir
 
       #: Symbol
       attr_reader :parser
@@ -64,6 +73,10 @@ module Tapioca
 
     #: (*String sorbet_args) -> Spoom::ExecResult
     def sorbet(*sorbet_args)
+      if sorbet_config.cache_dir
+        sorbet_args << "--cache-dir=#{sorbet_config.cache_dir}"
+      end
+
       if sorbet_config.parse_with_prism?
         sorbet_args << "--parser=prism"
       end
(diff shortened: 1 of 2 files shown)


lib/tapioca/helpers/sorbet_helper.rb
```diff
@@ -39,16 +39,25 @@ def parse(content)
           end.to_h #: Hash[String, String | bool | nil]
 
           new(
+            # `--cache-dir=` disables the cache, modeled here with a `nil` value
+            cache_dir: if (value = options["--cache-dir"]).is_a?(String)
+                         value.empty? ? nil : value
+                       end,
```
KaanOzkan, line 45: I think this would be more readable outside of `new`, was there a specific reason to have it here?

Also should this use `shellescape`?

amomchilov: Sure, I extracted local variables  
  
> Also should this use `shellescape`?  
  
I don't think so. The C++ code consumes these values directly, not mediated by a shell.

Morriar: > Sure, I extracted local variables

Did you?

Are we missing a push?

KaanOzkan: Looks like it. LGTM after that.

PR #2569, opened 2026-04-01
Use RBS in Sorbet calls (if enabled)

amomchilov:


Morriar (commented):
Why do we need this?

paracycle:
> Why do we need this?

I have the same question. I don't understand the motivation here.

PR #2571, opened 2026-04-01
Fix `assert_raises` calls

amomchilov:
### Motivation

These calls are trying to pass a String/Regex as a matcher, but it's actually just interpreted as an exception type to try to rescue. It's confusing the base Minitest version of [`assert_raises(*exp)`](https://www.rubydoc.info/gems/minitest/Minitest/Assertions#assert_raises-instance_method) for the augment Rails [`assert_raises(*exp, match:)`](https://api.rubyonrails.org/classes/ActiveSupport/Testing/Assertions.html#method-i-assert_raises).

### DIFF
--- a/spec/tapioca/helpers/sorbet_helper_spec.rb
+++ b/spec/tapioca/helpers/sorbet_helper_spec.rb
@@ -28,7 +28,7 @@ class Tapioca::SorbetHelperSpec < Minitest::Spec
     end
 
     it "raises for an unknown feature check" do
-      assert_raises do
+      assert_raises(RuntimeError) do
         sorbet_supports?(:unknown_feature_name)
       end
     end


spec/tapioca/dsl/compilers/json_api_client_resource_spec.rb
```diff
@@ -178,9 +178,9 @@ class Post < JsonApiClient::Resource
               # We let the generation raise, the user should not define an association without a corresponding class.
               # The association will be unusable anyway.
 
-              assert_raises(NameError, /uninitialized constant Post::User/) do
-                rbi_for(:Post)
-              end
+              e = assert_raises(NameError) { rbi_for(:Post) }
+
+              assert_match("uninitialized constant Post::User", e.message)
```
Morriar: Looks like `assert_match` will create a regexp anyway: 

```rb
    pattern = case(pattern)
      when String
        Regexp.new(Regexp.escape(pattern))
      else
        pattern
    end
```

Wouldn't passing a literal regexp with `/pattern/` be _slightly_ faster and more memory efficient?

amomchilov: AHaha i saw this change and had an inkling you might flag this. Won't someone think of the bytes?!

If it needs escaping, I'll keep the string format, because it's much easier to read without the backslashes (e.g. `/sig \{ returns\(void\)\) \}/`)

PR #2573, opened 2026-04-01
Allow `refactor` label to meet CI requirement

amomchilov:
### Motivation

> PR is missing at least one of the following labels: documentation, bugfix, chore, enhancement, ignore-for-release, security, sorbet, dependencies

It's pretty reasonable to have a PR that's just labelled [refactor](https://github.com/Shopify/tapioca/labels?q=refactor). Right now, the CI would force you to tag it with something else like `chore` or `enhancement`, which aren't necessarily a good fit.

### DIFF
--- a/bin/check_pr_labels
+++ b/bin/check_pr_labels
@@ -27,7 +27,13 @@ labels = JSON.parse(arg)["labels"].map { |label| label["name"] }
 matching_labels = labels & REQUIRED_LABELS
 
 if matching_labels.empty?
-  warn("PR is missing at least one of the following labels: #{REQUIRED_LABELS.join(", ")}")
+  warn <<~MSG
+    Tapioca has an automated release process which generates the release notes based on PRs labels.
+    (See https://github.com/Shopify/tapioca/blob/main/.github/release.yml)
+
+    Please ask a Shopify employee to label your PR with at least one of the following labels:
+    #{REQUIRED_LABELS.map { |label| " - #{label}" }.join("\n")}
+  MSG
   exit(1)
 end
 


bin/check_pr_labels
```diff
@@ -8,6 +8,7 @@ REQUIRED_LABELS = [
   "bugfix",
   "chore",
   "enhancement",
+  "refactor",
```
paracycle, line 11: This on its own isn't enough, we need to also classify the tag to feed into one of the release categories here: https://github.com/Shopify/tapioca/blob/main/.github/release.yml#L10-L23

Actually, maybe, instead of manually maintaining this list, we should just check against the union of labels for all the categories

Morriar: We could continue using `chore` for this as it doesn't bring much value for end users

paracycle: I can also see us adding `refactor` to the `exclude` list.

amomchilov: I like adding it to the exclude list.

bin/check_pr_labels
```diff
@@ -26,7 +27,13 @@ labels = JSON.parse(arg)["labels"].map { |label| label["name"] }
 matching_labels = labels & REQUIRED_LABELS
 
 if matching_labels.empty?
-  warn("PR is missing at least one of the following labels: #{REQUIRED_LABELS.join(", ")}")
+  warn <<~MSG
```
amomchilov, line 30: This error will always be triggered on community contributed PRs, whose authors wouldn't have the access level necessary to label it themselves.

We talked about this with the team a while ago, and didn't have a good solution. For now, let's at least give more clear instruction.

amomchilov: cc @ufuk @Morriar

PR #2585, opened 2026-04-06
Patch `Void` type handling in Sorbet runtime for correct return types

paracycle:
Superseedes #2579 

### Motivation

Since https://github.com/sorbet/sorbet/pull/10064 was merged, for any sig that declares `.void.checked(:tests)`, the runtime will treat the return type as `T.anything` instead of `void` and thus break Tapioca's signature generation for methods that use this pattern.

### Implementation

This commit adds a patch to the Sorbet runtime to ensure that the return type is correctly treated as `void` in these cases, allowing Tapioca to generate accurate signatures.

If we ever see a `void` call, we store a boolean flag to later on revert the `void` to `T.anything` replacement that's done in the `finalize!` method of the `DeclBuilder` class, which allows Tapioca to generate proper return types for methods

### Tests

Added a new test for this behaviour.

### DIFF
--- a/lib/tapioca/sorbet_ext/void_patch.rb
+++ b/lib/tapioca/sorbet_ext/void_patch.rb
@@ -1,0 +1,29 @@
+# typed: true
+# frozen_string_literal: true
+
+# If Signature has `effective_return_type`, then `return_type` always returns the correct type.
+# Ref: https://github.com/sorbet/sorbet/pull/10121
+return if T::Private::Methods::Signature.method_defined?(:effective_return_type)
+
+module T
+  module Private
+    module Methods
+      module DeclBuilderPatch
+        def void
+          super.tap do
+            @_real_returns_is_void = true
+          end
+        end
+
+        def finalize!
+          super.tap do
+            #: self as untyped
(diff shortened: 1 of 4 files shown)


lib/tapioca/sorbet_ext/void_patch.rb
```diff
@@ -0,0 +1,25 @@
+# typed: true
+# frozen_string_literal: true
+
+module T
+  module Private
+    module Methods
+      module DeclBuilderPatch
+        def void
+          super.tap do
+            @_real_returns_is_void = true
```
jez, line 14: fwiw, I think we could move this strategy into sorbet-runtime itself, and avoid having to patch it here. I think that adding one more `true`/`false` instance variable to the `Signature` data structure would not cause an increase in memory usage because of the way that Ruby buckets objects with N instance variables.

paracycle: Sure, something like this I assume: https://github.com/sorbet/sorbet/pull/10121

I think I'll still merge this and then, later on, make the patching conditional on the existence of `Signature#real_return_type` method.

PR #2592, opened 2026-04-10
Display line diffs in `dsl --verify` error messages

dejmedus:
### Motivation

When `tapioca dsl --verify` detects out of date RBI files, the error message lists which files were added, removed, or changed. We could make it easier to visualize these changes by including the line diff output in the error message

<details><summary>Output example</summary>
tapioca-verify-diff
</details> 

### Implementation

This PR diffs updated files against the originals during RBI verification and then includes the output in the error message when the diff is 250 lines or less, or whatever value is provided to `--max-diff-lines` flag

### Tests

- updated `verify` tests to include line diff
- added a test for multi file changes and changes longer than max diff lines
- added tests for diff behaviour


lib/tapioca/commands/abstract_dsl.rb
```diff
@@ -305,16 +331,20 @@ def build_error_for_files(cause, files)
...
           say("Nothing to do, all RBIs are up-to-date.")
         else
           reasons = diff.group_by(&:last).sort.map do |cause, diff_for_cause|
             build_error_for_files(cause, diff_for_cause.map(&:first))
           end.join("\n")
 
-          raise Tapioca::Error, <<~ERROR
+          file_diff = if diff_output.lines.count.between?(1, 250)
+            "#{set_color("Diff:", :red)}\n#{diff_output.chomp}"
+          end
```
KaanOzkan, line 337: Do you think it's useful to output a message when the diff is longer than 250 lines and explain we're omitting it.

lib/tapioca/commands/abstract_dsl.rb
```diff
@@ -323,6 +353,8 @@ def report_diff_and_exit_if_out_of_date(diff, command)
 
             #{set_color("Reason:", :red)}
             #{reasons}
+
+            #{file_diff}
```
KaanOzkan: We should prevent this from printing a blank line if it's nil. You could prepend `\n\n` to `file_diff` above and then do `#{reasons}#{file_diff}`. 

Another solution is doing 2 `chomp` calls. It might be better, I'm not sure.

lib/tapioca/commands/abstract_dsl.rb
```diff
@@ -305,16 +331,20 @@ def build_error_for_files(cause, files)
...
+      def report_diff_and_exit_if_out_of_date(diff, diff_output, command)
         if diff.empty?
           say("Nothing to do, all RBIs are up-to-date.")
         else
           reasons = diff.group_by(&:last).sort.map do |cause, diff_for_cause|
             build_error_for_files(cause, diff_for_cause.map(&:first))
           end.join("\n")
 
-          raise Tapioca::Error, <<~ERROR
+          file_diff = if diff_output.lines.count.between?(1, 250)
```
KaanOzkan: Should this 250 limit be a CLI option instead and default to 250? `dsl --verify --max-diff-lines=1000`

lib/tapioca/commands/abstract_dsl.rb
```diff
@@ -291,9 +296,30 @@ def verify_dsl_rbi(tmp_dir:)
...
+          "--label",
+          filename.to_s,
+          old_path.to_s,
+          new_path.to_s,
+        )
+        raise stderr.chomp unless [0, 1].include?(status.exitstatus)
+
+        stdout
+      rescue => e
+        say_error("Failed to create #{filename} diff. #{e.message}", :yellow)
```
KaanOzkan: Why did you opt for yellow and not red? I think red is more consistent with rest of our errors.

spec/tapioca/cli/dsl_spec.rb
```diff
@@ -1934,13 +1934,28 @@ def perform(foo, bar)
...
         it "does nothing and returns exit status 0 with no changes" do
           @project.tapioca("dsl")
           result = @project.tapioca("dsl --verify")
 
-          assert_stdout_includes(result, <<~OUT)
+          assert_stdout_equals(<<~OUT, result)
+            Loading DSL extension classes... Done
+            Loading Rails application... Done
+            Loading DSL compiler classes... Done
+            Checking for out-of-date RBIs...
```
KaanOzkan, line 2138: Did you make this change to be more consistent with other tests? I assume it's not needed.

spec/tapioca/commands/abstract_dsl_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+  module Commands
+    class AbstractDslSpec < Minitest::Spec
+      include Tapioca::Helpers::Test::Isolation
+
+      describe "#file_diff" do
+        before do
+          @command = DslVerify.new(
+            requested_constants: [],
+            requested_paths: [],
+            outpath: Pathname.new("tmp"),
```
KaanOzkan: Does `DslVerify` write something to the file system? I think other `DslVerify` tests run on a temporary directory that's cleaned up but this is not an integration test so it shouldn't use that, it's more costly.

If it is writing something we should clean up after, maybe we can create `@tmp = Dir.mktmpdir` in the `before` block, use that as the outpath and then clean it up in an `after` block.

lib/tapioca/commands/abstract_dsl.rb
dejmedus: 👍🏼 I swapped `chomp` for [`rstrip`](https://docs.ruby-lang.org/en/4.0/String.html)

lib/tapioca/commands/abstract_dsl.rb
dejmedus: Great idea 👌🏼

lib/tapioca/commands/abstract_dsl.rb
dejmedus: I looked through [`say_error` uses](https://github.com/search?q=repo%3AShopify%2Ftapioca%20say_error&type=code) and it seemed like `:yellow` was used for warnings and `:red` for errors. I leaned towards this being a warning. Makes sense, changed to `:red` :)

spec/tapioca/cli/dsl_spec.rb
dejmedus: This was so I could assert `stdout_equals` rather than `stdout_includes` to ensure there isn't a stray `Diff:`

spec/tapioca/commands/abstract_dsl_spec.rb
dejmedus: If im understanding correctly, `outpath` is required to create `DslVerify` but the command isn't being executed here, I just need an instance to pull `file_diff` from. I changed it to `/dev/null` to make it clearer that its unused. tbh I wouldn't be that against removing this test file if we don't find it helpful, file diff is being tested by verify tests also

dejmedus: Actually ignore above, I moved out `file_diff` to [rbi_files_helpers](https://github.com/Shopify/tapioca/pull/2592/changes#diff-07ca5fecd052292c34a1f7dba8e5e08db82bf38a83c78af5e670a39625e6ed20R142) and the [tests](https://github.com/Shopify/tapioca/pull/2592/changes#diff-d8a1e4bf89fa47867ecbdefdbd6038768ee3cdf3ee20c6da615d42fcbbcc1d1fR23) look much nicer now :)

lib/tapioca/commands/abstract_dsl.rb
```diff
@@ -305,16 +312,25 @@ def build_error_for_files(cause, files)
...
         else
           reasons = diff.group_by(&:last).sort.map do |cause, diff_for_cause|
             build_error_for_files(cause, diff_for_cause.map(&:first))
           end.join("\n")
 
-          raise Tapioca::Error, <<~ERROR
+          diff_lines = diff_output.lines.count
+          file_diff = if diff_lines.between?(1, @max_diff_lines)
+            "#{set_color("Diff:", :red)}\n#{diff_output.chomp}"
+          elsif diff_lines > @max_diff_lines
```
KaanOzkan: Instead of skipping it completely should we display the diff **until** `@max_diff_lines`?

lib/tapioca/helpers/rbi_files_helper.rb
```diff
@@ -137,6 +139,26 @@ def validate_rbi_files(command:, gem_dir:, dsl_dir:, auto_strictness:, gems: [],
...
+        "diff",
+        "-u",
+        "--label",
+        filename.to_s,
+        "--label",
+        filename.to_s,
+        old_path.to_s,
+        new_path.to_s,
+      )
+      Kernel.raise stderr.chomp unless [0, 1].include?(status.exitstatus)
```
KaanOzkan: Using exceptions for control flow is frowned upon and slow. So we shouldn't rely on it when we can just call `say_error` here instead. If Open3 itself can raise while running keep the `rescue` block too.


```suggestion
      say_error("Failed to create #{filename} diff. #{e.message}", :red) unless [0, 1].include?(status.exitstatus)
```

lib/tapioca/commands/abstract_dsl.rb
```diff
@@ -240,9 +243,9 @@ def compile_dsl_rbi(constant_name, rbi, outpath: @outpath, quiet: false)
 
       #: (Pathname dir) -> void
       def perform_dsl_verification(dir)
-        diff = verify_dsl_rbi(tmp_dir: dir)
+        diff, diff_output = verify_dsl_rbi(tmp_dir: dir)
```
KaanOzkan: I don't love the tuple return here and I also forgot about `report_diff_and_exit_if_out_of_date`. What do you think about moving the calculation of diff output to `report_diff_and_exit_if_out_of_date`? It's the same class but it'll move some logic away from `perform_dsl_verification`/`verify_dsl_rbi` and make it easier to read. Something like this:

```ruby
  #: (Hash[String, Symbol] diff, tmp_dir: Pathname, command: Symbol) -> void
  def report_diff_and_exit_if_out_of_date(diff, tmp_dir:, command:)
    if diff.empty?
      say("Nothing to do, all RBIs are up-to-date.")
      return
    end

    reasons = diff.group_by(&:last).sort.map do |cause, diff_for_cause|
      build_error_for_files(cause, diff_for_cause.map(&:first))
    end.join("\n")

    diff_output = build_diff_output(diff, tmp_dir)
    diff_lines  = diff_output.count("\n")

    diff_section =
      if diff_lines.between?(1, @max_diff_lines)
        "#{set_color("Diff:", :red)}\n#{diff_output.chomp}"
      elsif diff_lines > @max_diff_lines
        set_color("Diff not displayed as it exceeds #{@max_diff_lines} lines", :red)
      else
        ""
      end

    raise Tapioca::Error, <<~ERROR.rstrip
      #{set_color("RBI files are out-of-date. In your development environment, please run:", :green)}
        #{set_color("`#{default_command(command)}`", :green, :bold)}
      #{set_color("Once it is complete, be sure to commit and push any changes", :green)}
      If you don't observe any changes after running the command locally, ensure your database is in a good
      state e.g. run `bin/rails db:reset`

      #{set_color("Reason:", :red)}
      #{reasons}

      #{diff_section}
    ERROR
  end

  #: (Hash[String, Symbol] diff, Pathname tmp_dir) -> String
  def build_diff_output(diff, tmp_dir)
    out = String.new
    line_count = 0

    diff.each do |file, status|
      chunk = case status
      when :added   then file_diff(file, File::NULL, tmp_dir / file)
      when :removed then file_diff(file, @outpath / file, File::NULL)
      when :changed then file_diff(file, @outpath / file, tmp_dir / file)
      end

      out << chunk
      line_count += chunk.count("\n")
      break if line_count > @max_diff_lines
    end

    out
  end
```

spec/tapioca/helpers/rbi_files_helper_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+# the including class's initialize method to expect Thor command args, which trips
+# up Minitest's test creation. Since we just want to test the helper methods, we
+# can bypass Thor's initialize method and create tests directly
+module TestFriendlyThor
+  include Thor::Base
+
+  def initialize(name)
+    Minitest::Runnable.instance_method(:initialize).bind(self).call(name)
+  end
+end
```
KaanOzkan: Instead of this could we get away with defining our own `say_error` method? It's simpler if that's all we need.

KaanOzkan (commented):
Sorry I'm suggesting a refactor that I didn't noticed before.

lib/tapioca/helpers/rbi_files_helper.rb
dejmedus: > Using exceptions for control flow is frowned upon and slow

Noted, thank you !

lib/tapioca/commands/abstract_dsl.rb
dejmedus: Agree, changed. Thanks!

spec/tapioca/helpers/rbi_files_helper_spec.rb
dejmedus: We could, but I believe we'd have to turn off type check in the file since [files helpers require Thor as ancestor](https://github.com/Shopify/tapioca/blob/2d3279f5ddbc7c391056060381c368d5429d5a14/lib/tapioca/helpers/rbi_files_helper.rb#L5)

KaanOzkan: Hmm yeah I guess we could also extract `file_diff` to another file and remove the `say_error` dependency. Curious if anyone else has a better suggestion, worth asking on slack.

lib/tapioca/helpers/rbi_files_helper.rb
```diff
@@ -137,6 +139,30 @@ def validate_rbi_files(command:, gem_dir:, dsl_dir:, auto_strictness:, gems: [],
...
 
+    #: (String filename, String old_path, String new_path) -> String
+    def file_diff(filename, old_path, new_path)
+      stdout, stderr, status = Open3.capture3(
+        "diff",
+        "-u",
+        "--label",
+        filename,
+        "--label",
+        filename,
```
amomchilov: We can give more useful labels, rather than the same file name in both cases, like:

https://github.com/Shopify/tapioca/blob/85ec455af080b87962569f88dfbb1d397c4150dc/spec/tapioca/cli/dsl_spec.rb#L2123-L2138


```suggestion
        "--label",
        "Current #{filename}",
        "--label",
        "Correct #{filename} (After running `bin/tapioca dsl`)",
```

lib/tapioca/helpers/rbi_files_helper.rb
```diff
@@ -137,6 +139,30 @@ def validate_rbi_files(command:, gem_dir:, dsl_dir:, auto_strictness:, gems: [],
...
+        old_path,
+        new_path,
+      )
```
amomchilov: You can also use the `=` form to keep the `--label` flag name with its value, and you can interleave labels and file paths:


```suggestion
      stdout, stderr, status = Open3.capture3(
        "diff",
        "-u",
        "--label=\"Current #{filename}\"",
        old_path,
        "--label=\"Correct #{filename} (After running `bin/tapioca dsl`)\"",
        new_path,
      )
```

Example:

```console
$ diff -u --label="before" <(echo "hello") --label "after" <(echo "world")
--- before
+++ after
@@ -1 +1 @@
-hello
+world
```

lib/tapioca/helpers/rbi_files_helper.rb
```diff
@@ -137,6 +139,30 @@ def validate_rbi_files(command:, gem_dir:, dsl_dir:, auto_strictness:, gems: [],
       Kernel.raise Tapioca::Error, error_messages.join("\n") if parse_errors.any?
     end
 
+    #: (String filename, String old_path, String new_path) -> String
```
amomchilov:
```suggestion
    #: (String filename, Pathname | String old_path, Pathname | String new_path) -> String
```

lib/tapioca/commands/abstract_dsl.rb
```diff
@@ -305,26 +308,67 @@ def build_error_for_files(cause, files)
...
+
+      #: (Hash[String, Symbol] diff, Pathname tmp_dir) -> String
+      def build_diff_output(diff, tmp_dir)
+        out = String.new
+        line_count = 0
+
+        diff.each do |file, status|
+          filename = file.to_s
+          old_path = (@outpath / file).to_s
+          new_path = (tmp_dir / file).to_s
```
amomchilov: Move the `to_s` calls into `file_diff`

lib/tapioca/commands/abstract_dsl.rb
```diff
@@ -305,26 +308,67 @@ def build_error_for_files(cause, files)
...
+
+          chunk = case status
+          when :added then file_diff(filename, File::NULL, new_path)
+          when :removed then file_diff(filename, old_path, File::NULL)
+          when :changed then file_diff(filename, old_path, new_path)
```
amomchilov: Just to confirm, does this ever call `file_diff` at most once per path? Or is it possible `diff` contains multiple entries pertaining to the same file?

lib/tapioca/helpers/rbi_files_helper.rb
dejmedus: 👌🏼 💯

lib/tapioca/helpers/rbi_files_helper.rb
dejmedus: Nice !

lib/tapioca/helpers/rbi_files_helper.rb
dejmedus: Updated! I also realized that filename should actually be a `Pathname`, so I [changed that throughout](https://github.com/Shopify/tapioca/pull/2592/commits/70de51e8c9d035c9e2e1b74afe5946c63c2e22d0)  as well

lib/tapioca/commands/abstract_dsl.rb
dejmedus: Yes, each file can only have one status (`added`, `removed`, or `changed`)

lib/tapioca/helpers/rbi_files_helper.rb
```diff
@@ -137,6 +139,29 @@ def validate_rbi_files(command:, gem_dir:, dsl_dir:, auto_strictness:, gems: [],
...
+        new_path.to_s,
+      )
+
+      unless [0, 1].include?(status.exitstatus)
+        say_error("Failed to create #{filename} diff. #{stderr.chomp}", :red)
+        return ""
+      end
+
+      stdout
+    rescue => e
```
amomchilov: What are we aiming to rescue here, exactly?

spec/tapioca/cli/dsl_spec.rb
```diff
@@ -2173,6 +2188,29 @@ def perform(foo, bar)
             Reason:
               File(s) removed:
               - sorbet/rbi/dsl/post.rbi
+
+            Diff:
+            --- Current post.rbi
+            +++ Correct post.rbi (After running `bin/tapioca dsl`)
```
amomchilov: Nice.

spec/tapioca/cli/dsl_spec.rb
```diff
@@ -2173,6 +2188,29 @@ def perform(foo, bar)
...
+            @@ -1,18 +0,0 @@
+            -# typed: true
+            -
+            -# DO NOT EDIT MANUALLY
+            -# This is an autogenerated file for dynamic methods in `Post`.
+            -# Please instead update this file by running `bin/tapioca dsl Post`.
```
amomchilov, line 2212: Oooo can we reuse whatever logic is used to generate this `bin/tapioca dsl Post`, and use that in the `+++` label above?

spec/tapioca/cli/dsl_spec.rb
```diff
@@ -2262,10 +2321,162 @@ class Post
             Reason:
               File(s) changed:
               - sorbet/rbi/dsl/post.rbi
+
+            Diff:
+            --- Current post.rbi
+            +++ Correct post.rbi (After running `bin/tapioca dsl`)
+            @@ -10,6 +10,12 @@
+            #{" "}
```
amomchilov, line 2341: intentional spaces? Kinda weird

spec/tapioca/helpers/rbi_files_helper_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+  include Tapioca::SorbetHelper
+  include Tapioca::RBIFilesHelper
+
+  describe "#file_diff" do
+    it "returns diff when files differ" do
+      Dir.mktmpdir do |dir|
+        a = File.join(dir, "a")
+        b = File.join(dir, "b")
+        File.write(a, "line1\nline2\n")
+        File.write(b, "line1\nline3\n")
```
amomchilov:
```suggestion
      Pathname.mktmpdir do |dir|
        a = dir /  "a"
        b = dir /  "b"
        a.write("line1\nline2\n")
        b.write("line1\nline3\n")
```

amomchilov: Likewise below

lib/tapioca/helpers/rbi_files_helper.rb
dejmedus: I narrowed this to `SystemCallError`. The thought was that if for some reason we couldn't execute diff we'd continue, though maybe thats overkill?

spec/tapioca/cli/dsl_spec.rb
dejmedus: This would be awesome! Currently we don't have easy access to the constant by the time we're here, but I'll work on a separate PR for this

spec/tapioca/helpers/rbi_files_helper_spec.rb
dejmedus: It [seems like](https://rubyapi.org/3.2/o/pathname#method-i-mkdir) we don't get `Pathname.mktmpdir` until Ruby [3.4](https://rubyapi.org/3.4/o/pathname#method-c-mktmpdir) and [gemspec min is 3.2](https://github.com/Shopify/tapioca/blob/main/tapioca.gemspec#L43), but I just wrapped it

spec/tapioca/helpers/rbi_files_helper_spec.rb
dejmedus: > extract file_diff to another file and remove the say_error dependency

Done:) I moved `file_diff` to its own file and replaced `say_error` with a `warn`

lib/tapioca/commands/abstract_dsl.rb
```diff
@@ -310,26 +314,66 @@ def build_error_for_files(cause, files)
...
+
+        diff_output = build_diff_output(diff, tmp_dir)
+        diff_lines = diff_output.count("\n")
+
+        diff_section =
+          if diff_lines.between?(1, @max_diff_lines)
+            "#{set_color("Diff:", :red)}\n#{diff_output.chomp}"
+          elsif diff_lines > @max_diff_lines
+            truncated_output = diff_output.lines.first(@max_diff_lines).join
+            "#{set_color("Diff truncated to #{@max_diff_lines} lines:", :red)}\n#{truncated_output.rstrip}"
```
KaanOzkan: Could mention the new CLI flag here.

lib/tapioca/cli.rb
```diff
@@ -176,6 +180,7 @@ def dsl(*constant_or_paths)
         halt_upon_load_error: options[:halt_upon_load_error],
         compiler_options: options[:compiler_options],
         lsp_addon: self.class.addon_mode,
+        max_diff_lines: options[:max_diff_lines],
```
KaanOzkan, line 187: We should check that it's a positive number just in case. It'd error later on anyways but I think it's good to validate user input at this layer.

spec/tapioca/cli/dsl_spec.rb
dejmedus: > Kinda weird

Agree, `diff -u` uses single space as the "unchanged" marker

lib/tapioca/commands/abstract_dsl.rb
dejmedus: Good idea!

<details><summary>message</summary>
flag-mention
</details>

KaanOzkan (approved):

spec/tapioca/helpers/file_helper_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+        b = dir / "b"
+        a.write("line1\nline2\n")
+        b.write("line1\nline3\n")
+
+        result = file_diff(FILENAME, a, b)
+
+        assert_includes(result, "--- Current x.rbi")
+        assert_includes(result, "+++ Correct x.rbi (After running `bin/tapioca dsl`)")
+        assert_includes(result, "-line2")
+        assert_includes(result, "+line3")
```
amomchilov: Any reason not to assert over the entire output? That'll ensure the order of these snippets is correct

lib/tapioca/commands/abstract_dsl.rb
```diff
@@ -310,26 +314,66 @@ def build_error_for_files(cause, files)
...
+          chunk = case status
+          when :added then file_diff(file, File::NULL, new_path)
+          when :removed then file_diff(file, old_path, File::NULL)
+          when :changed then file_diff(file, old_path, new_path)
+          else ""
+          end
+
+          out << chunk
+          line_count += chunk.count("\n")
+          break if line_count > @max_diff_lines
```
amomchilov: Minor bug: this can permit too many lines. Imagine the limit is 250, and there's 3 chunks of 200 each.

The first one will be let through, but so would the second (without truncation). The `line_count` will hit 400, so we would break before the third chunk get's concatenated in, but we won't trim off the 150 excess lines of the second chunk.

amomchilov (approved):

lib/tapioca/commands/abstract_dsl.rb
dejmedus: Truncation and the this was truncated message output were happening in [`report_diff`](https://github.com/Shopify/tapioca/pull/2592/changes#diff-eac3a2b55717bc0a9dc122f9aa1ec0928d809b68e82e0a08c3cfd7ffae97970cR335). The intention with `break if line_count > @max_diff_lines` was just to not loop longer than necessary

I can see how that was odd/confusing tho, so I've moved truncation into `build_diff` and now return both the output and whether it was truncated

lib/tapioca/helpers/file_helper.rb
```diff
@@ -0,0 +1,42 @@
...
+  module FileHelper
+    #: (Pathname filename, Pathname | String old_path, Pathname | String new_path) -> String
+    def file_diff(filename, old_path, new_path)
+      filename = filename.to_s
+      stdout, stderr, status = Open3.capture3(
+        "diff",
+        "-u",
+        "--label=Current #{filename}",
+        old_path.to_s,
+        "--label=Correct #{filename} (After running `bin/tapioca dsl`)",
```
Morriar:
```suggestion
        "--label=Expected #{filename} (After running `bin/tapioca dsl`)",
```

lib/tapioca/helpers/file_helper.rb
```diff
@@ -0,0 +1,42 @@
...
+        new_path.to_s,
+      )
+
+      unless [0, 1].include?(status.exitstatus)
+        error_msg("Failed to create #{filename} diff. #{stderr.chomp}")
+        return ""
```
Morriar: I'd return `nil` if it couldn't be computed

PR #2603, opened 2026-04-21
Use `rest_type`/`keyrest_type` for a Sorbet signature

KaanOzkan:
### Motivation

`sorbet-runtime` 0.6.13149 (sorbet/sorbet#10177) removed `@has_rest` and `@has_keyrest` from `T::Private::Methods::Signature`, since they were redundant with the `@rest_type`/`@keyrest_type` ivars (nil when absent).

### Implementation

Update to pre-existing variables `rest_type`/`keyrest_type`, no Sorbet version bump is necessary.
### Tests

Already tested

### DIFF
--- a/lib/tapioca/dsl/compiler.rb
+++ b/lib/tapioca/dsl/compiler.rb
@@ -124,10 +124,12 @@ module Tapioca
         signature.kwarg_types.each { |_, kwarg_type| params << kwarg_type.to_s }
 
         # rest parameter type
-        params << signature.rest_type.to_s if signature.has_rest
+        rest_type = signature.rest_type
+        params << rest_type.to_s if rest_type
 
         # keyrest parameter type
-        params << signature.keyrest_type.to_s if signature.has_keyrest
+        keyrest_type = signature.keyrest_type
+        params << keyrest_type.to_s if keyrest_type
 
         # special case `.void` in a proc
         unless signature.block_name.nil?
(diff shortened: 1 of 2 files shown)


lib/tapioca/dsl/compiler.rb
```diff
@@ -124,10 +124,10 @@ def parameters_types_from_signature(method_def, signature)
         signature.kwarg_types.each { |_, kwarg_type| params << kwarg_type.to_s }
 
         # rest parameter type
-        params << signature.rest_type.to_s if signature.has_rest
+        params << signature.rest_type.to_s if signature.rest_type
```
vinistock: Nit
```suggestion
        rest_type = signature.rest_type
        params << rest_type.to_s if rest_type
```

lib/tapioca/dsl/compiler.rb
```diff
@@ -124,10 +124,10 @@ def parameters_types_from_signature(method_def, signature)
...
 
         # keyrest parameter type
-        params << signature.keyrest_type.to_s if signature.has_keyrest
+        params << signature.keyrest_type.to_s if signature.keyrest_type
```
vinistock: Nit
```suggestion
        keyrest_type = signature.keyrest_type
        params << keyrest_type.to_s if keyrest_type
```

vinistock (approved):

lib/tapioca/dsl/compiler.rb
KaanOzkan: Ah yes, we don't need two invocations. Thanks.

alexcrocha (approved):

PR #2604, opened 2026-04-22
Lift `sorbet-static-and-runtime` upper version cap and bump minimum version

KaanOzkan:
To be merged after https://github.com/Shopify/tapioca/pull/2603

The upper cap `<= 0.6.13118` was added in 11c4ab93 for a bug in sorbet's `--print=missing-constants` output (reversed multi-segment paths like `::Y::X` instead of `::X::Y`), which caused incorrect `todo.rbi` generation.

That bug was fixed in sorbet/sorbet#10132, first released in sorbet-runtime 0.6.13114. Verified empirically:

- 0.6.13110 (pre-fix): `::X::Y.foo` -> `::Y::X` (reversed)
- 0.6.13114 (first release with fix): `::X::Y.foo` -> `::X::Y` (correct)

Remove the upper cap and raise the lower bound to `>= 0.6.13114` to exclude the known-buggy release window (0.6.13097..0.6.13110).

Second commit fixes a type checking error due to a new Sorbet update: https://github.com/Shopify/tapioca/actions/runs/24794894694/job/72561922226

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -22,7 +22,7 @@ PATH
       rbi (>= 0.3.7)
       require-hooks (>= 0.2.2)
       rubydex (>= 0.1.0.beta10)
-      sorbet-static-and-runtime (>= 0.6.12698, <= 0.6.13118)
+      sorbet-static-and-runtime (>= 0.6.13114)
       spoom (>= 1.7.9)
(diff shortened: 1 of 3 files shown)


KaanOzkan:
Closing for https://github.com/Shopify/tapioca/pull/2605

PR #2607, opened 2026-04-23
Bootstrap gem symbols from Rubydex graph

Morriar:
## Summary

Extracted from #2524. Instead of shelling out to Sorbet to get the symbol table for a gem's files, we use the Rubydex graph which is already available and faster. The engine symbols path also switches to Rubydex for consistency.

* Replace Sorbet's symbol table (`symbols_from_paths`) with Rubydex graph for bootstrapping gem symbols in the pipeline
* The gem graph is now always built (not conditional on `include_doc`), and reused for both symbol bootstrapping and documentation
* Remove `gem_symbols` from `SymbolLoader` and `load_bootstrap_symbols` from `Pipeline` since Rubydex handles both cases

### DIFF
--- a/lib/tapioca/static/symbol_loader.rb
+++ b/lib/tapioca/static/symbol_loader.rb
@@ -27,11 +27,6 @@ module Tapioca
         end
 
         #: (Gemfile::GemSpec gem) -> Set[String]
-        def gem_symbols(gem)
-          symbols_from_paths(gem.files)
-        end
-
-        #: (Gemfile::GemSpec gem) -> Set[String]
         def engine_symbols(gem)
           gem_engine = engines.find do |engine|
             gem.full_gem_path == engine.config.root.to_s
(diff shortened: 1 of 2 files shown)


lib/tapioca/static/symbol_loader.rb
```diff
@@ -26,9 +26,12 @@ def graph_from_paths(paths)
...
 
-        #: (Gemfile::GemSpec gem) -> Set[String]
-        def gem_symbols(gem)
-          symbols_from_paths(gem.files)
+        #: (Rubydex::Graph graph) -> Set[String]
+        def symbols_from_graph(graph)
+          namespaces = graph.declarations.grep(Rubydex::Namespace).map(&:name)
+          constants = graph.declarations.grep(Rubydex::Constant).map(&:name)
+          aliases = graph.declarations.grep(Rubydex::ConstantAlias).map(&:name)
+          (namespaces + constants + aliases).to_set
```
vinistock: It seems we're separating 3 collections just to immediately merge them back. How about something like this?
```suggestion         
          graph.declarations.select do |decl|
            decl.is_a?(Rubydex::Namespace) ||
              decl.is_a?(Rubydex::Constant) ||
              decl.is_a?(Rubydex::ConstantAlias)
            end
          end.to_set
```

vinistock (approved):
Do we still need the `include_doc` flag? If memory serves, we added that because including docs was super slow, but with Rubydex the problem no longer exists.

Maybe we can clean up that flag

PR #2608, opened 2026-04-23
Use Rubydex graph for constants_from_requested_paths

Morriar:
## Summary

Extracted from #2524. Instead of shelling out to Sorbet to get the symbol table for requested paths, we use the Rubydex graph which is already available and faster.

* Replace `symbols_from_paths` (Sorbet shell-out) with `graph_from_paths` (Rubydex) in `constants_from_requested_paths`
* Remove the now-unused `symbols_from_paths` method from `SymbolLoader`


vinistock (approved):
It looks like there's some engine filtering that's failing, but I think it's related to the base branch and not this one

st0012 (approved):

Issue #2589, opened 2026-04-08
Tapioca v0.18.0 generates invalid gem RBI for stripe gem


olivier-thatch:
Hi all,

I'm upgrading our codebase from Sorbet 0.5 and Tapioca 0.17.9 to Sorbet 0.6 and Tapioca 0.19.0.

After force regenerating all gem RBIs, Sorbet reports 24418 (!) errors in `sorbet/rbi/gems/stripe@19.0.0.rbi`.

At a glance, it appears that the signatures for all write accessors throughout the stripe gem are invalid:
```
sorbet/rbi/gems/stripe@19.0.0.rbi:4316: Malformed sig. Type not specified for parameter value https://srb.help/5003
    4316 |  def expand=(value); end
                        ^^^^^
    sorbet/rbi/gems/stripe@19.0.0.rbi:4315: Add it to the signature's params here
    4315 |  sig { params(_expand: T.nilable(T::Array[String])).returns(T.nilable(T::Array[String])) }
```

The stripe gem provides its own RBI signatures, and the signatures for write accessors use `_name` as the argument name, e.g.:
```ruby
    sig { params(_expand: T.nilable(T::Array[String])).returns(T.nilable(T::Array[String])) }
    def expand=(_expand); end
```

(from https://github.com/stripe/stripe-ruby/blob/v19.0.0/rbi/stripe/params/account_capability_list_params.rbi#L10-L11)

However, for some reason Tapioca forcefully replaces the argument name with `value` in the method definition, but not in the signature:
```ruby
  # pkg:gem/stripe#lib/stripe/params/account_capability_list_params.rb:7
  sig { params(_expand: T.nilable(T::Array[String])).returns(T.nilable(T::Array[String])) }
  def expand=(value); end
```

At first I thought this might be related to #2248, but this appears to be a different issue. AFAICT, this is caused by a behavior change in Tapioca v0.18.0. When I force-regenerate the RBI with Tapioca v0.17.9, everything's fine.

I can work around it by not regenerating the RBI for the stripe gem for the time being, but this only works until we need to upgrade the stripe gem.

KaanOzkan:
Can you confirm if this is introduced by `v0.18.0`?

olivier-thatch:
> Can you confirm if this is introduced by `v0.18.0`?

Yes, I just double-checked and can confirm.

[...]

PR #2611, opened 2026-04-28
Normalize setter sig param names from exported gem RBIs

dduugg:
### Motivation

Some gems (notably [stripe-ruby](https://github.com/stripe/stripe-ruby)) ship exported RBIs where setter sigs use the attribute name with a leading underscore (`_expand`, `_payment_intent`, etc.) as the parameter name. However, stripe overrides `attr_accessor` with `define_method("#{name}=") { |value| ... }`, so the runtime setter parameter is named `value`.

After tapioca's `Keep::LEFT` merge, the method def comes from runtime introspection (`value`) but the sig is pulled from the exported RBI (`_expand`). Sorbet rejects this sig/def mismatch with error 5003, producing type errors in codebases that depend on the stripe gem.

### Implementation

After `merge_trees`, the merged tree is walked by a new `SetterSigParamNormalizer` visitor. For each setter method (name ends with `=`) with a single `ReqParam`, any attached sig param whose name differs from the method def's param name is replaced with a new `SigParam` using the method def's param name. The type annotation from the exported RBI is preserved — only the parameter name is normalized.

`SetterSigParamNormalizer` is an `RBI::Visitor` subclass defined inside `Commands::AbstractGem` and marked `private_constant`.

### Tests

Added an integration test that creates a mock gem with a `define_method`-based setter (runtime parameter `value`) and a bundled RBI whose sig uses `_expand`, then asserts the generated RBI has `params(value: ...)` in the sig.

### DIFF
--- a/lib/tapioca/commands/abstract_gem.rb
+++ b/lib/tapioca/commands/abstract_gem.rb
@@ -283,10 +283,55 @@ module Tapioca
         end
 
         file.root = RBI::Rewriters::Merge.merge_trees(file.root, tree, keep: RBI::Rewriters::Merge::Keep::LEFT)
+        normalize_setter_sig_params(file.root)
       rescue RBI::ParseError => e
         say_error("\n\n  RBIs exported by `#{gem.name}` contain errors and can't be used:", :yellow)
         say_error("Cause: #{e.message} (#{e.location})")
       end
+
+      # After merging a gem's exported RBI (Keep::LEFT), setter sigs from the exported RBI may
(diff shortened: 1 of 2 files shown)


KaanOzkan (commented):
Looks to be same as https://github.com/Shopify/tapioca/issues/2589 which should be fixed in `rbi`. I'll release a new version.

EDIT: Done, try `v0.3.11` for the `rbi` gem

dduugg:
Thanks, @KaanOzkan !

PR #2616, opened 2026-05-07
Skip RBS rewrite when no markers are present

mattkubej:
## Summary

Avoid running Spoom's RBS-to-Sorbet-sig translator for typed Ruby files that cannot contain runtime-rewriteable RBS syntax.

Tapioca currently calls `Spoom::Sorbet::Translate.rbs_comments_to_sorbet_sigs` for every typed Ruby file loaded through the RBS rewriter. A large portion of typed files in Core do not contain RBS signatures or supported RBS annotations, so the translator often parses source only to return it unchanged.

This keeps the existing `typed:` gate and adds a conservative marker gate. Translation now runs only when source contains one of:

- `#:`
- `#|`
- `# @abstract`
- `# @interface`
- `# @sealed`
- `# @final`
- `# @requires_ancestor:`
- `# @override`
- `# @overridable`
- `# @without_runtime`

False positives are safe because they still use the existing translator. False negatives would be unsafe, so the annotation list is intentionally aligned with Spoom's currently supported runtime-rewrite annotations.

### DIFF
--- a/lib/tapioca/rbs/rewriter.rb
+++ b/lib/tapioca/rbs/rewriter.rb
@@ -62,6 +62,50 @@ end
 
 require "require-hooks/setup"
 
+module Tapioca
+  module RBS
+    module Rewriter
+      TYPED_FILE_PATTERN = /^\s*#\s*typed: (ignore|false|true|strict|strong|__STDLIB_INTERNAL)/
+      # Markers consumed by Spoom's RBS comment rewriter beyond `#:`/`#|`.
+      # Keep in sync with Spoom: missing markers skip rewrites; extra markers only add safe false positives.
+      RBS_ANNOTATION_MARKERS = [
+        "# @abstract",
+        "# @interface",
+        "# @sealed",
+        "# @final",
+        "# @requires_ancestor:",
+        "# @override",
+        "# @overridable",
+        "# @without_runtime",
+      ].freeze #: Array[String]
+      RBS_REWRITE_PATTERN = Regexp.union(["#:", "#|"] + RBS_ANNOTATION_MARKERS).freeze #: Regexp
+
+      class << self
+        #: (String source) -> bool
+        def typed_file?(source)
+          source.match?(TYPED_FILE_PATTERN)
+        end
+
+        #: (String source) -> bool
+        def possible_rbs_runtime_rewrite_syntax?(source)
+          source.match?(RBS_REWRITE_PATTERN)
+        end
+
+        #: (untyped path, String source) -> String?
+        def rewrite(path, source)
+          return unless typed_file?(source)
+          return source unless possible_rbs_runtime_rewrite_syntax?(source)
+
+          Spoom::Sorbet::Translate.rbs_comments_to_sorbet_sigs(source, file: path)
+        rescue Spoom::Sorbet::Translate::Error
+          # If we can't translate the RBS comments back into Sorbet's signatures, we just skip the file.
+          source
+        end
+      end
+    end
+  end
+end
+
 # We need to include `T::Sig` very early to make sure that the `sig` method is available since gems using RBS comments
 # are unlikely to include `T::Sig` in their own classes.
 Module.include(T::Sig)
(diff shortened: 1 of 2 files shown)


lib/tapioca/rbs/rewriter.rb
```diff
@@ -36,6 +36,52 @@ def load_iseq(path)
...
+          source.match?(TYPED_FILE_PATTERN)
+        end
+
+        #: (String source) -> bool
+        def possible_rbs_runtime_rewrite_syntax?(source)
+          return true if source.include?("#:") || source.include?("#|")
+          return false unless source.include?("# @")
+
+          RBS_ANNOTATION_MARKERS.any? { |marker| source.include?(marker) }
+        end
```
paracycle, line 92: This seems to be scanning the source string multiple times, can we turn the search into a `Regexp.union` and do a single scan through the `source`?

paracycle: I think something like:
```ruby
SEARCH_REGEXP = Regexp.union(TYPED_FILE_PATTERN, Regexp.escape("#:"), Regexp.escape("#|"), *RBS_ANNOTATION_MARKERS.map { Regexp.escape(it) })

def should_rewrite?
  source.index(SEARCH_REGEXP)
end
```
should give you the decision to rewrite or not with a single pass.

mattkubej: Good call on the redundant scans. I just pushed an amend.

However, I kept `typed_file?` as a separate stage rather than folding into the union. A single union would introduce OR semantics, so a typed file with no markers would match and I think we'd want to avoid the expensive Spoom translator execution in this case.

paracycle: Ah, right, I forgot that the `typed` check wasn't being OR'ed. Good call.

lib/tapioca/rbs/rewriter.rb
```diff
@@ -62,6 +62,50 @@ def setup(**_kwargs)
...
+        "# @abstract",
+        "# @interface",
+        "# @sealed",
+        "# @final",
+        "# @requires_ancestor:",
+        "# @override",
+        "# @overridable",
+        "# @without_runtime",
+      ].freeze #: Array[String]
+      RBS_REWRITE_PATTERN = Regexp.union(["#:", "#|"] + RBS_ANNOTATION_MARKERS).freeze #: Regexp
```
paracycle, line 81: Are you sure that none of these have special Regexp characters, or does `Regexp.union` do the right thing? I am surprised that we don't need `Regexp.escape` here.

paracycle (approved):
Looks good to me, but have one question

lib/tapioca/rbs/rewriter.rb
mattkubej: Just sanity checked this. Looks like `Regexp.union` runs each String arg through `Regexp.escape` before joining.

Looking at the Ruby implementation, it looks like `Regexp.escape` and `Regexp.quote` share the same C function.
- https://github.com/ruby/ruby/blob/master/re.c#L4874-L4875
   - `quote` and `escape` map to `rb_reg_s_quote` 
- https://github.com/ruby/ruby/blob/master/re.c#L4385
   -  union calls `rb_reg_s_quote` for string case

Check to show union's source against manually escaped and joined version:
```ruby
markers = ["#:", "#|", "# @abstract", "# @interface", "# @sealed", "# @final", "# @requires_ancestor:", "# @override", "# @overridable", "# @without_runtime"]
pattern = Regexp.union(markers)
pattern.source == markers.map { |m| Regexp.escape(m) }.join("|")
# => true
pattern.source
# => "\#:|\#\||\#\ @abstract|\#\ @interface|\#\ @sealed|\#\ @final|\#\ @requires_ancestor:|\#\ @override|\#\ @overridable|\#\ @without_runtime"
```

Documentation doesn't explicitly state this behavior, but the example seems to highlight this escaping behavior as well: https://docs.ruby-lang.org/en/3.4/Regexp.html#method-c-union

Morriar (approved):

lib/tapioca/rbs/rewriter.rb
amomchilov: heh I had the same suggestion here: https://github.com/shop/world/pull/675510/changes#r3220064805

Is that other pull request just a PoC that's replaced by this one?

lib/tapioca/rbs/rewriter.rb
```diff
@@ -62,6 +62,50 @@ def setup(**_kwargs)
...
+        def possible_rbs_runtime_rewrite_syntax?(source)
+          source.match?(RBS_REWRITE_PATTERN)
+        end
+
+        #: (untyped path, String source) -> String?
+        def rewrite(path, source)
+          return unless typed_file?(source)
+          return source unless possible_rbs_runtime_rewrite_syntax?(source)
+
+          Spoom::Sorbet::Translate.rbs_comments_to_sorbet_sigs(source, file: path)
```
amomchilov, line 99: I don't see any reason why somebody might want `Spoom::Sorbet::Translate.rbs_comments_to_sorbet_sigs` _without_ your early-return optimization. Would it be better to move this optimization into Spoom itself?

amomchilov:
Thank you! We'll move this logic right into Spoom https://github.com/Shopify/spoom/issues/916

PR #2617, opened 2026-05-08
Cache RBS rewrites with bootsnap & add `dsl --only-bootsnap-rbs-cache`

KaanOzkan:
## Motivation

`tapioca dsl` rewrites RBS comments to Sorbet `sig {}` blocks on every loaded file via `require-hooks`. On Shopify's monolith this dominates wall time (locally ~16 min). Caching the post-rewrite bytecode in bootsnap drops it to ~4 min (4.2x).

## Implementation

- Opt-in via `TAPIOCA_RBS_CACHE=1`. The rewriter sets up bootsnap with a dedicated cache directory (default `tmp/cache/bootsnap-tapioca-rbs`, override via `TAPIOCA_BOOTSNAP_CACHE_DIR`) and forwards `BOOTSNAP_READONLY` to bootsnap's `readonly:` arg. After setup, `Tapioca::RBS::BootsnapGuard` is prepended on `Bootsnap.singleton_class` so the host's own `Bootsnap.setup` raises instead of silently clobbering the cache dir.

- `--only-bootsnap-rbs-cache` populates the cache and exits before any RBI work (CI prime/consumer pattern, mutually exclusive with `--verify`). `bootsnap` stays an optional runtime dep; a Sorbet shim covers the static surface.

## Tests

CLI tests in `spec/tapioca/cli/dsl_spec.rb` cover the flag's early-exit, the `--verify` mutex, and the guard's raise.

### DIFF
--- a/lib/tapioca/commands/dsl_generate.rb
+++ b/lib/tapioca/commands/dsl_generate.rb
@@ -4,12 +4,23 @@
 module Tapioca
   module Commands
     class DslGenerate < AbstractDsl
+      #: (?only_bootsnap_rbs_cache: bool, **untyped) -> void
+      def initialize(only_bootsnap_rbs_cache: false, **kwargs)
+        @only_bootsnap_rbs_cache = only_bootsnap_rbs_cache
+        super(**T.unsafe(kwargs))
+      end
+
       private
 
       # @override
       #: -> void
       def execute
         load_application
+
+        if @only_bootsnap_rbs_cache
+          say("Bootsnap RBS cache populated, exiting before RBI generation.", :green)
+          return
+        end
 
         say("Compiling DSL RBI files...")
         say("")
--- a/lib/tapioca/rbs/rewriter.rb
+++ b/lib/tapioca/rbs/rewriter.rb
@@ -1,40 +1,38 @@
 # typed: strict
 # frozen_string_literal: true
 
-require "require-hooks/setup"
-
 # This code rewrites RBS comments back into Sorbet's signatures as the files are being loaded.
 # This will allow `sorbet-runtime` to wrap the methods as if they were originally written with the `sig{}` blocks.
 # This will in turn allow Tapioca to use this signatures to generate typed RBI files.
 
-begin
-  # When in a `bootsnap` environment, files are loaded from the cache and won't trigger the `source_transform` method.
-  # The `require-hooks` gem comes with a `bootsnap` mode that will disable the `bootsnap/compile_cache/iseq` caching.
-  # Sadly, we're way to early in the boot process to use it as bootsnap won't be loaded yet and the `require-hooks`
-  # setup won't pick it up.
-  #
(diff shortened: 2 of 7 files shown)


lib/tapioca/commands/dsl_generate.rb
```diff
@@ -11,6 +11,11 @@ class DslGenerate < AbstractDsl
       def execute
         load_application
 
+        if @only_bootsnap_rbs_cache
+          say("Bootsnap RBS cache populated, exiting before RBI generation.", :green)
```
Morriar, line 21: If someone runs `--only-bootsnap-rbs-cache` without setting `TAPIOCA_RBS_CACHE=1`, they'll get "Bootsnap RBS cache populated" but nothing was actually cached. Should we warn here when the env var isn't set?

```suggestion
        if @only_bootsnap_rbs_cache
          unless ENV["TAPIOCA_RBS_CACHE"] == "1"
            say_error("Warning: --only-bootsnap-rbs-cache has no effect without TAPIOCA_RBS_CACHE=1", :yellow)
          end
          say("Bootsnap RBS cache populated, exiting before RBI generation.", :green)
          return
        end
```

lib/tapioca/rbs/rewriter.rb
```diff
@@ -1,41 +1,36 @@
...
-        end
-      end
-    end
+# When TAPIOCA_RBS_CACHE=1, set up bootsnap with a dedicated cache directory
+# and load require-hooks so the RBS-rewritten iseqs get cached. Subsequent
+# runs read the rewritten iseq directly and skip the rewrite.
+#
+# The host app must also skip its own Bootsnap.setup under this flag,
+# otherwise it'll override our settings during Rails boot.
+if ENV["TAPIOCA_RBS_CACHE"] == "1"
```
KaanOzkan, line 41: Rewriter is loaded before any CLI parsing so went with a ENV variable.

lib/tapioca/rbs/rewriter.rb
```diff
@@ -1,41 +1,70 @@
...
+#
+# After our setup, BootsnapGuard is prepended so the host application can't
+# replace our cache directory.
+if ENV["TAPIOCA_RBS_CACHE"] == "1"
+  begin
+    require "bootsnap"
+    # Respect BOOTSNAP_READONLY for consumers reading a pre-populated cache
+    # (e.g. a CI prime step).
+    readonly_env = ENV["BOOTSNAP_READONLY"]
+    readonly = !readonly_env.nil? && !["0", "false"].include?(readonly_env)
```
amomchilov: I suggest copying Bootsnap's logic for this directly: https://github.com/rails/bootsnap/blob/ff7f97c77123113474fea6ae46459fea5c92ec9f/lib/bootsnap.rb#L190-L193

amomchilov (approved):

PR #2618, opened 2026-05-08
Fix nullability of GraphQL scalar types when passed as required arguments

### LINKED ISSUE #2337: Tapioca v0.17.5 doesn't handle typing of output of custom GraphQL Scalars correctly
Just noticed what I believe to be a bug when updating to the latest version of the Gem in a project that also uses the [`graphql-ruby`](https://github.com/rmosolgo/graphql-ruby) gem and where we have some custom scalar types. While the latest version does a great job of locating the correct type, I'm finding that for input arguments that reference the scalar types, the generated methods in the rbi are always nullable.

I've traced the reason down to [this line](https://github.com/Shopify/tapioca/blob/bfd05549107ddfc2c8efb232e89653221f7f032b/lib/tapioca/dsl/helpers/graphql_type_helper.rb#L66), and the underlying issue is that the `coerce_result` of any scalar type has to handle the `nil` case, but in cases where we have a required input argument the value of the argument will never be null. So the type should only have a `T.nilable(...)` if the input argument is not required.

joeytepp:
### Motivation

Fixes #2337

Currently, the generated rbis for a GraphQL mutation that contains an argument that uses a scalar type will always treat the argument as nullable, even when it is required. This PR fixes it so that required arguments are treated as non-nullable.

### Implementation
Updated `GraphQLTypeHelper` to use `RBIHelper.as_non_nilable_type` in the current branch of logic.

### Tests
Added a new test in `spec/tapioca/dsl/compilers/graphql_mutation_spec.rb`

### DIFF
--- a/lib/tapioca/dsl/helpers/graphql_type_helper.rb
+++ b/lib/tapioca/dsl/helpers/graphql_type_helper.rb
@@ -80,7 +80,7 @@ module Tapioca
             signature = Runtime::Reflection.signature_of(method)
             return_type = signature&.return_type
 
-            valid_return_type?(return_type) ? return_type.to_s : "T.untyped"
+            valid_return_type?(return_type) ? RBIHelper.as_non_nilable_type(return_type.to_s) : "T.untyped"
           when GraphQL::Schema::InputObject.singleton_class
             type_for_constant(unwrapped_type)
           when Module
--- a/lib/tapioca/dsl/helpers/graphql_type_helper.rb
+++ b/lib/tapioca/dsl/helpers/graphql_type_helper.rb
@@ -80,6 +80,8 @@ module Tapioca
             signature = Runtime::Reflection.signature_of(method)
             return_type = signature&.return_type
 
+            # Wrap as non-nilable for required arguments. `coerce_input` supports both
+            # required and optional; optional arguments are re-wrapped below based on `type.non_null?`
             valid_return_type?(return_type) ? RBIHelper.as_non_nilable_type(return_type.to_s) : "T.untyped"
           when GraphQL::Schema::InputObject.singleton_class
             type_for_constant(unwrapped_type)
--- a/spec/tapioca/dsl/compilers/graphql_mutation_spec.rb
+++ b/spec/tapioca/dsl/compilers/graphql_mutation_spec.rb
@@ -361,6 +361,45 @@ module Tapioca
               assert_equal(expected, rbi_for(:CreateComment))
             end
 
+            it "generates correct RBI for custom scalars whose coerce_input returns a nilable type" do
+              add_ruby_file("create_comment.rb", <<~RUBY)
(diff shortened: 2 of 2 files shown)


joeytepp:
I have signed the CLA!

lib/tapioca/dsl/helpers/graphql_type_helper.rb
```diff
@@ -80,7 +80,7 @@ def type_for(type, ignore_nilable_wrapper: false, prepare_method: nil)
             signature = Runtime::Reflection.signature_of(method)
             return_type = signature&.return_type
 
-            valid_return_type?(return_type) ? return_type.to_s : "T.untyped"
+            valid_return_type?(return_type) ? RBIHelper.as_non_nilable_type(return_type.to_s) : "T.untyped"
```
KaanOzkan, line 85: Can you give context on why we can assume `coerce_input` can't return nil?

joeytepp: `coerce_input` has to handle both required and optional arguments, and for an optional argument it still gets called with an input of `nil`. So when we have a required argument the signature should no longer have a `T.nilable(...)` since we know the output won't be `nil`.

KaanOzkan: Let's capture this in a comment since there's quite a bit going on, something like
```
# Wrap as non-nilable for required arguments. `coerce_input` supports both
# required and optional; optional arguments are re-wrapped below based on `type.non_null?`
```

joeytepp: Great, I've updated it with the comment as suggested.

KaanOzkan (approved):

lib/tapioca/dsl/helpers/graphql_type_helper.rb
```diff
@@ -80,7 +80,9 @@ def type_for(type, ignore_nilable_wrapper: false, prepare_method: nil)
             signature = Runtime::Reflection.signature_of(method)
             return_type = signature&.return_type
 
-            valid_return_type?(return_type) ? return_type.to_s : "T.untyped"
+            # Wrap as non-nilable for required arguments. `coerce_input` supports both
+            # required and optional; optional arguments are re-wrapped below based on `type.non_null?`
+            valid_return_type?(return_type) ? RBIHelper.as_non_nilable_type(return_type.to_s) : "T.untyped"
```
amomchilov, line 85: I wonder if we should be using `T::Utils.unwrap_nilable` on the runtime type objects in cases like this, instead of String processing

Thoughts @KaanOzkan ?

KaanOzkan: Good call, it's much faster, opened https://github.com/Shopify/tapioca/pull/2625

PR #2624, opened 2026-05-13
Warn when `--only-bootsnap-rbs-cache` runs without `TAPIOCA_RBS_CACHE=1`

KaanOzkan:
## Summary

- Addresses [PR #2617 review comment](https://github.com/Shopify/tapioca/pull/2617#discussion_r3209320189) from @Morriar.
- Without `TAPIOCA_RBS_CACHE=1` set, `bootsnap/setup` isn't required at boot and no RBS cache gets populated. Previously `tapioca dsl --only-bootsnap-rbs-cache` would still print "Bootsnap RBS cache populated", misleading users into thinking it had done useful work.
- Now emits a warning to stderr in that case so the misuse is visible.

## Test plan

- [x] New spec: `warns when --only-bootsnap-rbs-cache is set without TAPIOCA_RBS_CACHE=1`
- [x] Existing happy-path spec updated to set `TAPIOCA_RBS_CACHE=1` and assert bootsnap stats appear on stderr (confirming bootsnap actually ran)

### DIFF
--- a/lib/tapioca/commands/dsl_generate.rb
+++ b/lib/tapioca/commands/dsl_generate.rb
@@ -18,7 +18,11 @@ module Tapioca
         load_application
 
         if @only_bootsnap_rbs_cache
-          say("Bootsnap RBS cache populated, exiting before RBI generation.", :green)
+          if ENV["TAPIOCA_RBS_CACHE"] == "1"
+            say("Bootsnap RBS cache populated, exiting before RBI generation.", :green)
+          else
+            say_error("Warning: --only-bootsnap-rbs-cache requires TAPIOCA_RBS_CACHE=1 to populate the cache", :yellow)
+          end
           return
         end
 
--- a/spec/tapioca/cli/dsl_spec.rb
+++ b/spec/tapioca/cli/dsl_spec.rb
@@ -669,13 +669,34 @@ module Tapioca
             end
           RB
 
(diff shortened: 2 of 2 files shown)


lib/tapioca/commands/dsl_generate.rb
```diff
@@ -18,6 +18,9 @@ def execute
         load_application
 
         if @only_bootsnap_rbs_cache
+          unless ENV["TAPIOCA_RBS_CACHE"] == "1"
+            say_error("Warning: --only-bootsnap-rbs-cache has no effect without TAPIOCA_RBS_CACHE=1", :yellow)
+          end
           say("Bootsnap RBS cache populated, exiting before RBI generation.", :green)
```
alexcrocha: If I am understanding correctly, shouldn't the two messages live in different branches?

```suggestion
          if ENV["TAPIOCA_RBS_CACHE"] == "1"
             say("Bootsnap RBS cache populated, exiting before RBI generation.", :green)
          else
             say_error("Warning: --only-bootsnap-rbs-cache has no effect without TAPIOCA_RBS_CACHE=1", :yellow)
          end
```

alexcrocha (dismissed):

lib/tapioca/commands/dsl_generate.rb
KaanOzkan: Good catch.

alexcrocha (approved):
👍

PR #2625, opened 2026-05-15
Use `T::Utils.unwrap_nilable` for scalar `coerce_input` return types

KaanOzkan:
### Motivation

`type_for` was stripping the `T.nilable(...)` wrapper from a custom scalar's `coerce_input` return type by formatting the runtime type to a string and matching a regex. Operating on the runtime type tree is both simpler and ~8x faster on the common `T.nilable(SimpleType)` case (measured: 15.3ms vs 1.8ms across 1000 calls).

Output is identical for every input the spec covers, including the new nilable-scalar case added in #2618.

### Implementation

`Use T::Utils.unwrap_nilable || return_type` instead. `unwrap_nilable` returns nil if the type isn't nilable.

### DIFF
--- a/lib/tapioca/dsl/helpers/graphql_type_helper.rb
+++ b/lib/tapioca/dsl/helpers/graphql_type_helper.rb
@@ -82,7 +82,7 @@ module Tapioca
 
             # Wrap as non-nilable for required arguments. `coerce_input` supports both
             # required and optional; optional arguments are re-wrapped below based on `type.non_null?`
-            valid_return_type?(return_type) ? RBIHelper.as_non_nilable_type(return_type.to_s) : "T.untyped"
+            valid_return_type?(return_type) ? (T::Utils.unwrap_nilable(return_type) || return_type).to_s : "T.untyped"
           when GraphQL::Schema::InputObject.singleton_class
             type_for_constant(unwrapped_type)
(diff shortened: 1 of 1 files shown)


paracycle (approved):
I hadn't even realized that `return_type` was already a runtime type. This makes a lot of sense!

amomchilov:
@paracycle There's actually several places where we `to_s` a runtime type, only to later try to regex our way into figuring out things back out about it.

e.g.

https://github.com/Shopify/tapioca/blob/ab86175ba385e2f5f58b2d8bdf3f25d30816d904/lib/tapioca/dsl/helpers/active_model_type_helper.rb#L11-L22

https://github.com/Shopify/tapioca/blob/ab86175ba385e2f5f58b2d8bdf3f25d30816d904/lib/tapioca/dsl/compilers/active_model_attributes.rb#L123-L124

I considered changing it, but the repo deals with types as strings a fair bit. I think it would more clear (and generally faster) to operate on type values, and stringify them later.

What do you think?

paracycle:
> What do you think?

Overall, this is what I think: https://github.com/Shopify/tapioca/pull/2489

If we can move to our own type builder syntax, then we can also start generating RBS syntax if we wanted to.

PR #2630, opened 2026-05-21
Restore `Bootsnap` fallback for RBS rewriting

KaanOzkan:
Fixes a bug on unreleased `main` that @dejmedus pointed out in the new rewriter implementation when the ENV variable is omitted.

### Motivation

PR #2617 moved the old `load_iseq` bypass behind `TAPIOCA_RBS_CACHE=1`, so hosts that call `Bootsnap.setup` without the opt-in cache bypass `RequireHooks` and lose RBS comment signatures. Restore the default bypass after `require-hooks` setup while keeping the opt-in cache path intact.

### Implementation

Bring back the overload from https://github.com/Shopify/tapioca/blob/3fe6bb63d76a086e09e57db87c5d1cdf8092bea7/lib/tapioca/rbs/rewriter.rb#L23-L37
### Tests

Included


spec/tapioca/cli/dsl_spec.rb
```diff
@@ -680,6 +680,65 @@ class Post
           assert_success_status(result)
         end
 
+        it "preserves RBS comment rewriting when the host sets up Bootsnap without TAPIOCA_RBS_CACHE" do
```
KaanOzkan, line 683: Figured the test is worth it given last time this bug went through. It uses a custom compiler so that we don't boot rails but an existing compiler could also show this issue.

lib/tapioca/rbs/rewriter.rb
```diff
@@ -62,6 +64,34 @@ def setup(**_kwargs)
 
 require "require-hooks/setup"
 
+unless rbs_cache_enabled
```
KaanOzkan: Used `if` and `unless` so that `require "require-hooks/setup"` call isn't duplicated.

lib/tapioca/rbs/rewriter.rb
```diff
@@ -62,6 +64,34 @@ def setup(**_kwargs)
...
+    require "bootsnap"
+    require "bootsnap/compile_cache/iseq"
+
+    module Bootsnap
+      module CompileCache
+        module ISeq
+          module InstructionSequenceMixin
+            #: (String) -> RubyVM::InstructionSequence
+            def load_iseq(path)
+              super if defined?(super)
```
amomchilov: Maybe just clarify:

```suggestion
              super if defined?(super) # Disable Bootsnaps' hook, but trigger any others
```

lib/tapioca/rbs/rewriter.rb
amomchilov: Eh, I think a single `if`/`else` with a repeated `require "require-hooks/setup"` would be more clear, but that's just me

amomchilov (approved):

lib/tapioca/rbs/rewriter.rb
paracycle: Does the order of requires matter here?

amomchilov (approved):

lib/tapioca/rbs/rewriter.rb
KaanOzkan: > Does the order of requires matter here?

Yes, [require-hooks/setup](https://github.com/ruby-next/require-hooks/blob/v0.4.0/lib/require-hooks/setup.rb) chooses a mode based on bootsnap being loaded or not. In TAPIOCA_RBS_CACHE (`if`) we want it after loading bootsnap. In `else`, we want it before bootsnap is loaded.

PR #2632, opened 2026-05-25
Add Rails engine support to UrlHelpers compiler

### LINKED ISSUE #474: Make `UrlHelpers` generator Rails Engine aware
It turns out that Rails Engines have their own router and the controllers, etc inside a Rails Engine get an engine specific `url_helper` module included, not the `Rails.application` one.

In order to be able to properly represent this, we need to do the following:
- Loop over all the Rails engines inside the application
- For each engine:
  - Grab the helpers modules `engine.routes.named_routes.url_helpers_module` and `engine.routes.named_routes.path_helpers_module` and give them engine specific names, like `XXX::Engine::GeneratedUrlHelpersModule`.
  - Discover all the named modules that include those url helper modules and gather them as constants to decorate.
  - In `decorate`, create module definitions for the app and engine url helpers and add includes for them for the modules that include them.

## Test Case

```ruby
module Article
  class Engine < ::Rails::Engine
    isolate_namespace Article

    routes.draw do
      resource :articles
    end
  end
end

class Application < Rails::Application
  routes.draw do
    resource :index
    mount Article::Engine, at: "/", as: "articles"
  end
end
```
should create:
```ruby
# generated_path_helpers_module.rbi
# typed: strong

module GeneratedPathHelpersModule
  include ::ActionDispatch::Routing::UrlFor
  include ::ActionDispatch::Routing::PolymorphicRoutes

  sig { params(args: T.untyped).returns(String) }
  def edit_index_path(*args); end

  sig { params(args: T.untyped).returns(String) }
  def index_path(*args); end

  sig { params(args: T.untyped).returns(String) }
  def new_index_path(*args); end

  sig { params(args: T.untyped).returns(String) }
  def articles_path(*args); end
end
```
```ruby
# generated_url_helpers_module.rbi
# typed: strong

module GeneratedUrlHelpersModule
  include ::ActionDispatch::Routing::UrlFor
  include ::ActionDispatch::Routing::PolymorphicRoutes

  sig { params(args: T.untyped).returns(String) }
  def edit_index_url(*args); end

  sig { params(args: T.untyped).returns(String) }
  def index_url(*args); end

  sig { params(args: T.untyped).returns(String) }
  def new_index_url(*args); end

  sig { params(args: T.untyped).returns(String) }
  def articles_url(*args); end
end
```
```ruby
# article/engine/generated_path_helpers_module.rbi
# typed: strong

module Article::Engine::GeneratedPathHelpersModule
  include ::ActionDispatch::Routing::UrlFor
  include ::ActionDispatch::Routing::PolymorphicRoutes

  sig { params(args: T.untyped).returns(String) }
  def edit_article_path(*args); end

  sig { params(args: T.untyped).returns(String) }
  def article_path(*args); end

  sig { params(args: T.untyped).returns(String) }
  def new_article_path(*args); end

  sig { params(args: T.untyped).returns(String) }
  def articles_path(*args); end
end
```
```ruby
# article/engine/generated_url_helpers_module.rbi
# typed: strong

module Article::Engine::GeneratedUrlHelpersModule
  include ::ActionDispatch::Routing::UrlFor
  include ::ActionDispatch::Routing::PolymorphicRoutes

  sig { params(args: T.untyped).returns(String) }
  def edit_article_url(*args); end

  sig { params(args: T.untyped).returns(String) }
  def article_url(*args); end

  sig { params(args: T.untyped).returns(String) }
  def new_article_url(*args); end

  sig { params(args: T.untyped).returns(String) }
  def articles_url(*args); end
end
```
and the corresponding includes to the modules that include the above helpers.

bdewater-thatch:
### Motivation
Discover Rails engines during URL helper generation and register their engine-scoped path and URL helper modules. Fixes https://github.com/Shopify/tapioca/issues/474

Generate mounted helper RBIs for engine mount points, including typed routes proxy classes that include the engine helper modules. Add mounted helper mixins only for classes/modules that actually receive Rails' mounted helpers, and cover the new behavior with regression tests.

### Implementation
AI assisted with manual refactoring + testing on our own app.

### DIFF
--- a/lib/tapioca/dsl/compilers/url_helpers.rb
+++ b/lib/tapioca/dsl/compilers/url_helpers.rb
@@ -7,7 +7,22 @@ module Tapioca
   module Dsl
     module Compilers
       # `Tapioca::Dsl::Compilers::UrlHelpers` generates RBI files for classes that include or extend
-      # [`Rails.application.routes.url_helpers`](https://api.rubyonrails.org/v5.1.7/classes/ActionDispatch/Routing/UrlFor.html#module-ActionDispatch::Routing::UrlFor-label-URL+generation+for+named+routes).
+      # [`Rails.application.routes.url_helpers`](https://api.rubyonrails.org/classes/ActionDispatch/Routing/UrlFor.html#module-ActionDispatch::Routing::UrlFor-label-URL+generation+for+named+routes).
+      #
+      # The compiler registers generated constants to represent the Rails route helper modules:
+      #
+      # 1. `GeneratedPathHelpersModule` holds the main application's path helpers, such as `post_path`.
+      #
+      # 2. `GeneratedUrlHelpersModule` holds the main application's URL helpers, such as `post_url`.
+      #
+      # 3. `GeneratedMountedHelpers` is a synthetic module for mounted application and engine helpers, such as
+      # `main_app` and `articles`. Rails exposes these helpers through an anonymous dynamic module, so the compiler creates
+      # a named RBI module that can be included or extended by classes that receive mounted helpers at runtime. It is
+      # only generated for applications that mount an engine that defines its own routes.
+      #
+      # For mounted engines, the compiler also registers engine-scoped `GeneratedPathHelpersModule` and
+      # `GeneratedUrlHelpersModule` constants. Mounted engine helper methods return a synthetic
+      # `GeneratedRoutesProxy` subclass that includes those engine-scoped helper modules.
       #
       # For example, with the following setup:
       #
@@ -16,6 +31,8 @@ module Tapioca
       # class Application < Rails::Application
       #   routes.draw do
       #     resource :index
+      #
+      #     mount Blog::Engine, at: "/blog", as: "articles"
       #   end
       # end
       # ~~~
@@ -42,6 +59,9 @@ module Tapioca
       #   include ActionDispatch::Routing::UrlFor
       #
       #   sig { params(args: T.untyped).returns(String) }
+      #   def articles_path(*args); end
+      #
+      #   sig { params(args: T.untyped).returns(String) }
       #   def edit_index_path(*args); end
       #
       #   sig { params(args: T.untyped).returns(String) }
@@ -58,6 +78,9 @@ module Tapioca
       # module GeneratedUrlHelpersModule
       #   include ActionDispatch::Routing::PolymorphicRoutes
       #   include ActionDispatch::Routing::UrlFor
+      #
+      #   sig { params(args: T.untyped).returns(String) }
+      #   def articles_url(*args); end
       #
       #   sig { params(args: T.untyped).returns(String) }
       #   def edit_index_url(*args); end
@@ -78,6 +101,86 @@ module Tapioca
       #   include GeneratedUrlHelpersModule
       # end
       # ~~~
+      #
+      # ~~~rb
+      # # blog/config/routes.rb
+      # Blog::Engine.routes.draw do
+      #   resources :posts
+      # end
+      # ~~~
+      #
+      # ~~~rbi
+      # # blog/engine/generated_path_helpers_module.rbi
+      # # typed: true
+      # module Blog::Engine::GeneratedPathHelpersModule
+      #   include ActionDispatch::Routing::PolymorphicRoutes
+      #   include ActionDispatch::Routing::UrlFor
+      #
+      #   sig { params(args: T.untyped).returns(String) }
+      #   def edit_post_path(*args); end
+      #
+      #   sig { params(args: T.untyped).returns(String) }
+      #   def new_post_path(*args); end
+      #
+      #   sig { params(args: T.untyped).returns(String) }
+      #   def post_path(*args); end
+      #
+      #   sig { params(args: T.untyped).returns(String) }
+      #   def posts_path(*args); end
+      # end
+      # ~~~
+      #
+      # ~~~rbi
+      # # blog/engine/generated_url_helpers_module.rbi
+      # # typed: true
+      # module Blog::Engine::GeneratedUrlHelpersModule
+      #   include ActionDispatch::Routing::PolymorphicRoutes
+      #   include ActionDispatch::Routing::UrlFor
+      #
+      #   sig { params(args: T.untyped).returns(String) }
+      #   def edit_post_url(*args); end
+      #
+      #   sig { params(args: T.untyped).returns(String) }
+      #   def new_post_url(*args); end
+      #
+      #   sig { params(args: T.untyped).returns(String) }
+      #   def post_url(*args); end
+      #
+      #   sig { params(args: T.untyped).returns(String) }
+      #   def posts_url(*args); end
+      # end
+      # ~~~
+      #
+      # ~~~rbi
+      # # generated_mounted_helpers.rbi
+      # # typed: true
+      # module GeneratedMountedHelpers
+      #   sig { returns(Blog::Engine::GeneratedRoutesProxy) }
+      #   def articles; end
+      #
+      #   sig { returns(GeneratedRoutesProxy) }
+      #   def main_app; end
+      # end
+      # ~~~
+      #
+      # ~~~rbi
+      # # generated_routes_proxy.rbi
+      # # typed: true
+      # class GeneratedRoutesProxy < ::ActionDispatch::Routing::RoutesProxy
+      #   include GeneratedPathHelpersModule
+      #   include GeneratedUrlHelpersModule
+      # end
+      # ~~~
+      #
+      # ~~~rbi
+      # # blog/engine/generated_routes_proxy.rbi
+      # # typed: true
+      # class Blog::Engine::GeneratedRoutesProxy < ::ActionDispatch::Routing::RoutesProxy
+      #   include Blog::Engine::GeneratedPathHelpersModule
+      #   include Blog::Engine::GeneratedUrlHelpersModule
+      # end
+      # ~~~
+      #
       #: [ConstantType = Module[top]]
       class UrlHelpers < Compiler
         # @override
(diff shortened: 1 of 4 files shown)


lib/tapioca/dsl/compilers/url_helpers.rb
```diff
@@ -141,10 +171,79 @@ def gather_non_discoverable_includers
             end.freeze
           end
 
+          #: (Module[top] mounted_helpers_module) -> Array[Module[top]]
+          def setup_engine_helper_modules(mounted_helpers_module)
```
KaanOzkan:
```suggestion
          def register_engine_route_helpers(mounted_helpers_module)
```

lib/tapioca/dsl/compilers/url_helpers.rb
```diff
@@ -141,10 +171,79 @@ def gather_non_discoverable_includers
...
+            engine_helper_modules
+          end
+
+          # Map mount names to engine classes by inspecting mounted_helpers methods.
+          # Rails' mounted_helpers methods call `_routes_context` on `self`, so we
+          # create a minimal context object that satisfies that interface. This lets
+          # us instantiate the RoutesProxy and read its @routes to match back to an
+          # engine's RouteSet.
+          #: (Module[top] mounted_helpers_module, Hash[untyped, untyped] routes_to_engine) -> void
+          def register_engine_mount_names(mounted_helpers_module, routes_to_engine)
```
KaanOzkan:
```suggestion
          def register_mounted_engine_helpers(mounted_helpers_module, routes_to_engine)
```

lib/tapioca/dsl/compilers/url_helpers.rb
```diff
@@ -141,10 +171,79 @@ def gather_non_discoverable_includers
...
+            # the public one would fail), but record the public name for RBI output.
+            mounted_helpers_module.instance_methods(false).each do |method_name|
+              next if method_name == :main_app
+              # Only process the public methods (non-underscore-prefixed)
+              next if method_name.start_with?("_")
+
+              private_name = :"_#{method_name}"
+              next unless mounted_helpers_module.instance_methods(false).include?(private_name)
+
+              begin
```
KaanOzkan: I don't think this is tested, could you add a test case for `mount Blog::Engine, at: "/blog", as: "articles"` to show that mapping uses `articles`.

lib/tapioca/dsl/compilers/url_helpers.rb
KaanOzkan, line 185: We should update this class documentation with the new functionality, will need to regenerate docs after (`rake generate_dsl_documentation`)

lib/tapioca/dsl/compilers/url_helpers.rb
```diff
@@ -106,28 +109,55 @@ def gather_constants
 
             url_helpers_module = Rails.application.routes.named_routes.url_helpers_module
             path_helpers_module = Rails.application.routes.named_routes.path_helpers_module
+            mounted_helpers_module = Rails.application.routes.mounted_helpers
 
             Object.const_set(:GeneratedUrlHelpersModule, url_helpers_module)
```
KaanOzkan, line 224: Now that there's 3, I think we should add a comment briefly explaining what the 3 constants are for, especially mention why `GeneratedMountedHelpers` is synthetic.

lib/tapioca/dsl/compilers/url_helpers.rb
```diff
@@ -178,6 +277,107 @@ def generate_module_for(root, constant)
           end
         end
 
+        #: (RBI::Tree root) -> void
+        def generate_mounted_helpers_module(root)
```
KaanOzkan, line 376: Since this is a long function I think it'll be good to have a comment that contains a simple `.rb` to `.rbi` generation that it's responsible for.

KaanOzkan (commented):
Implementation looks fine to me but as someone without much context on the compiler it was difficult to wrap my head around it. I have some suggestions on how to improve maintainability. Curious to hear if you have more ideas.

bdewater-thatch:
@KaanOzkan addressed your feedback, PTAL

lib/tapioca/dsl/compilers/url_helpers.rb
```diff
@@ -9,6 +9,21 @@ module Compilers
...
       # [`Rails.application.routes.url_helpers`](https://api.rubyonrails.org/v5.1.7/classes/ActionDispatch/Routing/UrlFor.html#module-ActionDispatch::Routing::UrlFor-label-URL+generation+for+named+routes).
       #
+      # The compiler registers generated constants to represent the Rails route helper modules:
+      #
+      # 1. `GeneratedPathHelpersModule` holds the main application's path helpers, such as `post_path`.
+      #
+      # 2. `GeneratedUrlHelpersModule` holds the main application's URL helpers, such as `post_url`.
+      #
+      # 3. `GeneratedMountedHelpers` is a synthetic module for mounted application and engine helpers, such
+      # as `main_app` and `blog`. Rails exposes these helpers through an anonymous dynamic module, so the
```
KaanOzkan: It'd be good to define `blog` here too with something like "`blog` (from `mount Blog::Engine at: "/blog"`)"

KaanOzkan (approved):
Implementation looks good to me, @paracycle are you able to take a look as well? PR is adding significant complexity to add engine support and I want to make sure it's correct.

lib/tapioca/dsl/compilers/url_helpers.rb
```diff
@@ -106,28 +124,55 @@ def gather_constants
 
             url_helpers_module = Rails.application.routes.named_routes.url_helpers_module
             path_helpers_module = Rails.application.routes.named_routes.path_helpers_module
+            mounted_helpers_module = Rails.application.routes.mounted_helpers
 
             Object.const_set(:GeneratedUrlHelpersModule, url_helpers_module)
             Object.const_set(:GeneratedPathHelpersModule, path_helpers_module)
+            Object.const_set(:GeneratedMountedHelpers, Module.new)
```
amomchilov: Can we make this conditional, so it doesn't exist on Rails applications with no mounted engines?

lib/tapioca/dsl/compilers/url_helpers.rb
```diff
@@ -106,28 +124,55 @@ def gather_constants
...
+
+            # Build engine registry: { mount_name => engine_class }
+            @engine_mount_names = T.let(
+              {},
+              T.nilable(T::Hash[Symbol, T.class_of(::Rails::Engine)]),
+            )
```
amomchilov: This ivar has a very strange lifetime

It's set here in `@gather_constants`, read from `@engine_mount_names`, and set only from `register_mounted_engine_helpers`. Do we need an ivar for this at all? Can `register_mounted_engine_helpers` return a Hash of the mounted engine helpers it found?

lib/tapioca/dsl/compilers/url_helpers.rb
```diff
@@ -141,10 +186,79 @@ def gather_non_discoverable_includers
...
+            register_mounted_engine_helpers(mounted_helpers_module, routes_to_engine)
+
+            engine_helper_modules
+          end
+
+          # Map mount names to engine classes by inspecting mounted_helpers methods.
+          # Rails' mounted_helpers methods call `_routes_context` on `self`, so we
+          # create a minimal context object that satisfies that interface. This lets
+          # us instantiate the RoutesProxy and read its @routes to match back to an
+          # engine's RouteSet.
```
amomchilov: Is there really no public API for this? 🤔

lib/tapioca/dsl/compilers/url_helpers.rb
```diff
@@ -106,28 +124,55 @@ def gather_constants
...
+            constants
+              .concat(NON_DISCOVERABLE_INCLUDERS)
+              .push(GeneratedUrlHelpersModule, GeneratedPathHelpersModule)
+              .push(GeneratedMountedHelpers)
+              .concat(engine_helper_modules)
+          end
+
+          #: -> Hash[Symbol, singleton(::Rails::Engine)]
+          def engine_mount_names
+            @engine_mount_names || {}
```
amomchilov: When can this actually be nil?

amomchilov (commented):
This is a great feature idea, but I'm finding the code kinda hard to follow/review, especially because so many of these concepts have such similar names.

Could you please take a human pass over this and try to see if anything could be simplified? Some documentation on the helper methods would also help

lib/tapioca/dsl/compilers/url_helpers.rb
```diff
@@ -141,10 +186,79 @@ def gather_non_discoverable_includers
...
+
+              private_name = :"_#{method_name}"
+              next unless mounted_helpers_module.instance_methods(false).include?(private_name)
+
+              begin
+                proxy = mounted_helpers_module.instance_method(private_name).bind_call(context)
+                engine_routes = proxy.instance_variable_get(:@routes)
+                engine_class = routes_to_engine[engine_routes]
+                T.must(@engine_mount_names)[method_name] = engine_class if engine_class
+              rescue
```
amomchilov: Can you please tighten this rescue clause up? What are we expecting? `NameError`?

lib/tapioca/dsl/compilers/url_helpers.rb
amomchilov: I think you might be able to make this class-level ivar declaration on the class level (not inside a method), and it won't have to be nilable. Then the `T.must`, and `|| {}` won't be necessary. Then you can replace the `def engine_mount_names` with `attr_reader :engine_mount_names` on the singleton class

amomchilov:
@bdewater-thatch All of my comments were marked resolved, but aren't actually addressed

lib/tapioca/dsl/compilers/url_helpers.rb
```diff
@@ -106,28 +124,55 @@ def gather_constants
...
+              T.nilable(T::Hash[Symbol, T.class_of(::Rails::Engine)]),
+            )
+            engine_helper_modules = register_engine_route_helpers(mounted_helpers_module)
 
             constants = all_modules.select do |mod|
               next unless name_of(mod)
 
               # Fast-path to quickly disqualify most cases
-              next false unless url_helpers_module > mod ||
+              has_helpers = url_helpers_module > mod ||
```
Morriar: Should we also account for classes that include `Rails.application.routes.mounted_helpers` directly?

Rails exposes this as a public module for callers that only want `main_app` / mounted engine proxies. With the current fast-path, those constants won't be gathered because they don't include the path/url helper modules, so the compiler never emits `include GeneratedMountedHelpers` for them.

I think we need to include `mounted_helpers_module` in both this fast-path and the `includes_helper?` checks, with a regression test like:

```ruby
class MyClass
  include Rails.application.routes.mounted_helpers
end
```

lib/tapioca/dsl/compilers/url_helpers.rb
```diff
@@ -178,6 +293,127 @@ def generate_module_for(root, constant)
...
+        # ~~~
+        #: (RBI::Tree root) -> void
+        def generate_mounted_helpers_module(root)
+          engine_mount_names = self.class.engine_mount_names
+
+          root.create_module("GeneratedMountedHelpers") do |mod|
+            # main_app always returns a plain RoutesProxy
+            mod.create_method(
+              "main_app",
+              return_type: "ActionDispatch::Routing::RoutesProxy",
```
Morriar: I wonder if `main_app` should return a generated proxy too.

With `RoutesProxy` as the return type, `main_app` exists but `main_app.index_path` / `main_app.index_url` still won't typecheck because the gem RBI only exposes the dynamic `method_missing` behavior. Could we mirror the engine proxies here with a synthetic main-app proxy that includes `GeneratedPathHelpersModule` and `GeneratedUrlHelpersModule`?

bdewater-thatch:
@amomchilov my bad, I had an agent rebase and didn't double check the results before re-requesting review. Fixed now.

amomchilov:
@bdewater-thatch looks good! that ivar really cleaned up nice.

PR #2633, opened 2026-05-25
Simplify DSL constant filtering

franciskafieh:
### Motivation

`Tapioca::Dsl::Pipeline#filter_anonymous_and_reloaded_constants` currently filters anonymous constants and also reconciles multiple objects with the same constant name by re-constantizing every gathered name.

That duplicate-name reconciliation was added for an old Rails reload/initializer edge case. Since that behavior is no longer a blessed Rails usage, remove the extra work from the default DSL pipeline. If this still turns out to be needed for a real app, we can restore it behind an explicit compatibility flag.

### Implementation

- Replace the duplicate-name grouping/re-constantization pass with an in-place anonymous-constant filter.
- Rename the private helper to `filter_anonymous_constants` now that it no longer handles reloaded constants.
- Update the reloaded-constant test to encode the new expected behavior: reloaded constants are treated as regular named constants and are not re-resolved or warned about.

### Tests

- `bin/test spec/tapioca/cli/dsl_spec.rb`
- `bin/typecheck`
- `bin/style`
- `bin/test`

### Benchmarks

on an internal repo with 3500+ files, `bin/tapioca dsl --verify` runs 3% faster (N=5)


lib/tapioca/dsl/pipeline.rb
```diff
@@ -154,32 +154,10 @@ def gather_constants(requested_constants, requested_paths, skipped_constants)
...
-        filtered_constants = constants_by_name
-          .keys
-          .map { |name| T.cast(Runtime::Reflection.constantize(name), T::Module[T.anything]) }
-          .select { |mod| Runtime::Reflection.constant_defined?(mod) }
-
-        Set.new.compare_by_identity.merge(filtered_constants)
+      def filter_anonymous_constants(constants)
+        Set.new.compare_by_identity.merge(
+          constants.select { |constant| Runtime::Reflection.name_of(constant) },
+        )
```
paracycle: I think we can mutate the `constants` that are passed in using:
```suggestion
        constants.select! { |constant| Runtime::Reflection.name_of(constant) }
```
which should be less allocations and much faster.

spec/tapioca/cli/dsl_spec.rb
```diff
@@ -1418,74 +1418,6 @@ class Post
           assert_success_status(result)
         end
 
-        it "must warn about reloaded constants and process only the newest one" do
```
paracycle, line 1421: Instead of deleting this test, can we encode what the expected behaviour should be?

franciskafieh: for sure! working on that rn. i wanted this branch to be pushed up as a draft but Pi did not want to do that 🙃

lib/tapioca/dsl/pipeline.rb
paracycle: Actually, sorry, let's use `keep_if` which guarantees to mutate and return `self`. `select!` returns `nil` if no modifications were made:
```ruby
irb> Set[:a, :b, :c].select! { it }
#=> nil

irb> Set[:a, :b, :c].keep_if { it }
#=> Set[:a, :b, :c]

irb> Set[:a, :b, :c, nil].keep_if { it }
#=> Set[:a, :b, :c]
```

paracycle (approved):

spec/tapioca/cli/dsl_spec.rb
franciskafieh: [added](https://github.com/Shopify/tapioca/pull/2633/changes#diff-7584a244af28b63c63f38aab96ddd4e0411f1322275665c9556e7324e03080acR1421) ✅

KaanOzkan (approved):

spec/tapioca/cli/dsl_spec.rb
```diff
@@ -1477,10 +1474,10 @@ class Post
 
               module SmartPropertiesGeneratedMethods
                 sig { returns(T.nilable(::String)) }
-                def body; end
+                def title; end
```
paracycle: Hmm, I am not sure if what methods get generated is quite deterministic: https://github.com/Shopify/tapioca/actions/runs/26416952346/job/77763377644?pr=2633#step:6:141

franciskafieh: i don't think this check is necessary and, as you mention, is non-deterministic, so i have removed it. i think [asserting the `force` line exists](https://github.com/Shopify/tapioca/pull/2633/changes#diff-7584a244af28b63c63f38aab96ddd4e0411f1322275665c9556e7324e03080acR1450) is enough. let me know if this checks out!

PR #2634, opened 2026-05-25
Add 'breaking-change' label to required label list

paracycle:
### Motivation

We should allow PRs that are labelled `breaking-change`.


franciskafieh (approved):
thank you!

Issue #1861, opened 2024-04-10
Tapioca automatically converts nested style to compact style, causing 'Unable to resolve constant $class' error in RBI


hiko1129:
sorbet version: 0.5.11332
tapioca version: 0.13.1
ruby version: 3.2.2

I encountered an issue when using Tapioca to generate RBI files for a gem dependency. The generated RBI file contains an `Unable to resolve constant $class` error.

Here's a minimal reproduction of the issue:

In repo X, I have the following files:
lib/x.rb
```
module X
  class Foo
  end

  class Bar
    def foo
      Foo.new
    end
  end
end
```
rbi/defs.rbi
```
# typed: strong

module X
  class Foo
  end

  class Bar
    sig { returns(Foo) }
    def foo
    end
  end
end
```

In repo Y, which depends on the gem from repo X, I have the following files:
Gemfile
```
gem "x"
```

After running bin/tapioca gem, the generated RBI file at sorbet/rbi/gems/x.rbi looks like this:
```
module X; end

class X::Foo
end

class X::Bar
  sig { returns(Foo) } # Unable to resolve constant `Foo`
  def foo
  end
end
```

Suspect file: [pipeline.rb on GitHub](https://github.com/Shopify/tapioca/blob/main/lib/tapioca/gem/pipeline.rb)

KaanOzkan:
Thanks for the report. This is definitely annoying but I think the implementation will be too complex with too much overhead that it might not be worth it. Currently the exported gem rbis functionality is quiet simple. We parse the contents of `rbi/` and [merge it](https://github.com/Shopify/tapioca/blob/2df42ebb793c0f81868835f6f2a173da6d0f5f23/lib/tapioca/commands/abstract_gem.rb#L307) with generated RBI definitions. So we don't do any introspection for the contents of the `rbi/` folder. For this to be automated we'd have to find the fully qualified name of each constant referenced in a sig and modify the tree.

A better solution could be encouraging gem authors to use fully qualified names while writing signatures. We could have a cop in https://github.com/shopify/rubocop-sorbet to do that but it's hard to enforce adoption of `rubocop-sorbet`.

PR #2637, opened 2026-05-26
Fully qualify type constants in ActiveRecordDelegatedTypes RBI

dduugg:
### Motivation

Refs #1861.

`delegated_type :role, types: %w[Foo Bar]` causes the `ActiveRecordDelegatedTypes` compiler to write the type strings verbatim into the generated RBI:

```rb
class Namespace::Owner
  module GeneratedDelegatedTypeMethods
    sig { returns(T.nilable(Foo)) }
    def foo; end
  end
end
```

Sorbet's lexical nesting for `class Namespace::Owner` is just `Namespace::Owner`, not `Namespace` + `Namespace::Owner`. So a bare `Foo` reference inside the nested `GeneratedDelegatedTypeMethods` module never reaches `Namespace::Foo`, even when that's the only place the constant is defined. The DSL file fails to type-check with `Unable to resolve constant`, and users end up papering over it with `bin/tapioca todo` shims or by hand-editing the generated RBI.

This is the same root cause documented in #1861 (compact-style class declaration losing intermediate namespaces from lexical scope), specialized to the `delegated_type` case where the symptoms are most visible because the user-supplied type strings flow straight through.

### Implementation

Resolve each type string through `ActiveRecord::Base#compute_type` — the same namespace-aware lookup Rails uses for STI and polymorphic associations — and emit the constant's qualified name via `qualified_name_of`. Each type is resolved exactly once per role, and a `(raw, qualified)` pair is threaded through both `populate_*` helpers.

The output is now consistent with the associations compiler, which already emits `::Foo` form for the same reason.

- `T.nilable(Foo)` → `T.nilable(::Foo)` for already-passing models.
- `T.nilable(Foo)` → `T.nilable(::Namespace::Foo)` for the namespaced case that was previously broken. Types resolving to a constant outside the parent namespace (e.g. `Content::Entry` referencing `Shared::Message`) are also handled.
- Private-constant children are fixed by `compute_type`'s `safe_constantize` candidate walking.

#### Failure modes

When a type entry can't be resolved, the compiler records an `add_error` diagnostic (mirroring the `MissingConstantError` pattern in `active_record_associations.rb`) and substitutes `T.untyped`. This surfaces the problem to the user instead of silently emitting an unresolvable bare constant. Three failure modes are handled:

- **`NameError`** — the canonical "constant doesn't exist" case.
- **`LoadError`** — `compute_type` calls `constantize` on `::`-prefixed names, which can fail through the autoloader. Rescuing this prevents a broken autoload path from aborting the entire compile run.
- **Anonymous class** (nil from `qualified_name_of`) — treated the same as an unresolved constant rather than silently falling back to the raw user-supplied string.

When some — but not all — types in a `delegated_type` call are unresolvable, the `build_<role>` return type collapses to `T.untyped` rather than emitting `T.any(::Foo, T.untyped)`, since the union is equivalent to `T.untyped` in Sorbet and the per-type error has already been recorded.

### Tests

- Updated the three existing `decorate` tests to expect the new `::Foo` form (matches the associations compiler convention).
- Added `it "generates RBI file with fully-qualified type names for namespaced types"` covering `class Content::Entry` with `Content::Message` / `Content::Comment` — types in the same namespace as the parent.
- Added `it "qualifies module-prefixed type names that resolve outside the parent namespace"` covering `class Content::Entry` with `Shared::Message` / `Shared::Comment` — the canonical #1861 cross-namespace case.
- Added `it "emits T.untyped and an error when a type cannot be resolved"` covering the single-type `NameError` path; uses `expect_dsl_compiler_errors!` and asserts on `generated_errors`.
- Added `it "collapses the build return type to T.untyped when only some types resolve"` covering the mixed-resolvability `T.any` collapse.
- Added `it "qualifies a type string that already has a leading \`::\` without doubling it"` covering the leading-`::` edge case (verifies we don't emit `::::Foo`).

All 12 tests in `active_record_delegated_types_spec.rb` pass, all 134 tests across `active_record_*_spec.rb` pass, `bin/style` and `bin/typecheck` are clean.


lib/tapioca/dsl/compilers/active_record_delegated_types.rb
```diff
@@ -77,8 +77,9 @@ def decorate
               constant.__tapioca_delegated_types.each do |role, data|
                 types = data.fetch(:types)
                 options = data.fetch(:options, {})
-                populate_role_accessors(mod, role, types)
-                populate_type_helpers(mod, role, types, options)
+                type_pairs = types.map { |type| [type, qualified_type_name(type, role)] }
```
KaanOzkan: I think something like `ResolvedType = Struct.new(:raw_name, :qualified_name, keyword_init: true)` is better here, wdyt?

spec/tapioca/dsl/compilers/active_record_delegated_types_spec.rb
```diff
@@ -258,6 +258,327 @@ def message_id; end
 
               assert_equal(expected, rbi_for(:Entry))
             end
+
+            it "generates RBI file with fully-qualified type names for namespaced types" do
```
KaanOzkan:
```suggestion
            it "generates RBI file with fully-qualified names for unqualified types" do
```

KaanOzkan: I see that we're introducing 5 tests, I think we can reduce the LOC and combine it into 2 tests, one for success one for error using multiple delegated_types, wdyt?

lib/tapioca/dsl/compilers/active_record_delegated_types.rb
dduugg: Done in 606618b4. Introduced `ResolvedType = Struct.new(:raw_name, :qualified_name, keyword_init: true)` and threaded it through `decorate` → `populate_role_accessors`/`populate_type_helpers`/`populate_type_helper`/`build_return_type` instead of the `[raw, qualified]` tuples. Reads much better, thanks.

spec/tapioca/dsl/compilers/active_record_delegated_types_spec.rb
dduugg: Folded this into the consolidation below — the namespaced test is now part of a single success spec named "generates RBI file with fully-qualified names for unqualified, namespaced, and root-prefixed types", which captures the unqualified case you were pointing at along with the namespaced and `::`-prefixed ones.

dduugg: Done in 606618b4 — collapsed the five new specs into two:

- A success spec using one `Content::Entry` model with two `delegated_type` roles, covering all three resolution cases in a single RBI: unqualified-into-parent-namespace (`Message` → `::Content::Message`), cross-namespace module-prefixed (`Shared::Message`), and root-prefixed (`::Toplevel`, which also guards against `::::` doubling).
- An error spec using one `Entry` model with two roles, covering both the wholly-unresolvable collapse (`%w[Phantom]`) and the partially-resolvable collapse (`%w[Message Ghost]`), asserting both recorded errors.

All 9 specs pass, `bin/typecheck` and `bin/style` clean.

KaanOzkan (approved):

PR #2642, opened 2026-06-07
Skip stale RBI file removal when `--only` is set

### LINKED ISSUE #1612: `bin/tapioca dsl --only` should not purge existing directory
I believe `--only` is mostly used as a debugging tool to test specific compiler output. However, its output is hard to understand since tapioca purges the files that aren't generated and creates a large diff https://github.com/Shopify/tapioca/blob/5a9d90c9b6f65681609cd38a84595bd3639646b7/lib/tapioca/commands/abstract_dsl.rb#L88

We could skip the purging if `--only` is present. Am I missing any valid use cases where `--only` will be used permanently by a repo? In that case we could have a separate flag to skip purging.

Vein05:
### Motivation

Fixes #1612.

When `tapioca dsl --only` runs, only the specified compilers execute. The purge logic in `generate_dsl_rbi_files` computes `existing_rbi_files - generated_files` and treats the result as stale. Since only a subset of compilers ran, RBI files from every other compiler end up in the purge set and get deleted, producing a large misleading diff.

### Implementation

Guard the purge computation in `abstract_dsl.rb` with an `@only.any?` check. When `--only` is present, the method returns an empty `Set` and logs a message explaining that stale RBI removal was skipped. The guard is placed after `pipeline.run` so file generation still executes normally (including for `dsl --only --verify`).

### Tests

Added a new test (`"does not remove existing RBI files when using --only"`) that:
1. Generates RBI files for two compilers via a full `dsl` run
2. Re-runs with `--only SidekiqWorker`
3. Asserts both RBI files still exist and no "Removing stale RBI files" message appears

Updated four existing `--only` tests to include the new skip message in their expected stdout.

I added `say("Skipping stale RBI removal because `--only` is set.", :yellow)` instead of `"Skipping purging because of --only flag" ` as that felt more relevant here. 

This will be my first time working with production ruby, and I may have skipped something important. I'm open to suggestions. Thanks! 

All 69 tests pass (0 failures, 1 pre-existing protobuf error). `bin/typecheck` passes with no errors.

### DIFF
--- a/spec/tapioca/cli/dsl_spec.rb
+++ b/spec/tapioca/cli/dsl_spec.rb
@@ -1311,6 +1311,7 @@ module Tapioca
             Compiling DSL RBI files...
 
                   create  sorbet/rbi/dsl/job.rbi
+            Skipping stale RBI removal because `--only` is set.
 
             Done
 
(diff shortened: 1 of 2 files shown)


Vein05:
I have signed the CLA!

spec/tapioca/cli/dsl_spec.rb
```diff
@@ -1352,6 +1353,42 @@ module FooModule; end
...
+              include Sidekiq::Worker
+              def perform(foo, bar)
+              end
+            end
+          RB
+
+          @project.tapioca("dsl")
+
+          assert_project_file_exist("sorbet/rbi/dsl/post.rbi")
+          assert_project_file_exist("sorbet/rbi/dsl/job.rbi")
```
KaanOzkan: Could you write the files directly using `@project.write` instead of running tapioca dsl? It'd be faster and the file contents doesn't really matter.

Vein05: Oh yeah, that makes sense. Got it! added.

KaanOzkan (approved):

PR #2649, opened 2026-06-16
Fix Sorbet types for serialized + encrypted ActiveRecord columns

### LINKED ISSUE #2631: Incorrect return type for ActiveRecord encrypted hashes
Bit of an edge case, but:

```ruby
class MyModel < ApplicationRecord
  serialize :secret_hash, type:Hash, coder: JSON
  encrypts :secret_hash
end
```

results in the following signatures:
```ruby
class MyModel
    sig { returns(::String) }
    def secret_hash; end

    sig { params(value: ::String).returns(::String) }
    def secret_hash=(value); end
end
```

which makes sense since `secret_hash` is a string in the DB schema (due to being encrypted), but the type for the accessors should be `T::Hash[T.untyped, T.untyped]`.

I can take a stab at a PR if this is something that would be accepted into Tapioca.

olivier-thatch:
## Summary

When an ActiveRecord column is both `serialize`d and `encrypts`ed:

```ruby
class MyModel < ApplicationRecord
  serialize :secret_hash, type: Hash, coder: JSON
  encrypts :secret_hash
end
```

the `ActiveRecordColumns` DSL compiler generated `::String` accessors instead of `T::Hash[T.untyped, T.untyped]`, leaking the encrypted-at-rest DB representation rather than the Ruby type users work with.

Two gaps combined to cause this (in `lib/tapioca/dsl/helpers/active_record_column_type_helper.rb`):

1. **Encryption branch never reached the serialized handling.** `encrypts` wraps the prior type in `ActiveRecord::Encryption::EncryptedAttributeType`, whose `cast_type` is the underlying `ActiveRecord::Type::Serialized`. The encryption branch reflected only on the outer type and hardcoded a `::String` fallback, ignoring `cast_type`. It now recurses into `cast_type` when reflection is inconclusive — routing serialized columns through the existing serialized handler while still resolving to `::String` for plain encrypted string/text columns.

2. **`serialized_column_type` only recognized the YAML coder.** A `type: Hash, coder: JSON` declaration produces an `ActiveRecord::Coders::ColumnSerializer` coder (which carries `object_class == Hash`), not a `YAMLColumn`. The coder match was generalized to `ColumnSerializer` (the superclass of `YAMLColumn`), so `type:` declarations made with non-YAML coders are honored — both with and without `encrypts`. Bare `coder: JSON` (no `type:`) still resolves to `T.untyped`.

`not_nilable_serialized_column?` now also unwraps the encryption layer so nullable encrypted-serialized columns get the same non-nilable getter as their non-encrypted counterparts.

Closes #2631.

## Test plan

- [x] `bin/test spec/tapioca/dsl/compilers/active_record_columns_spec.rb` — 27 tests pass, including a new "serialized columns that are also encrypted" test (Hash + Array variants) and an extended non-encrypted `type: Hash, coder: JSON` assertion
- [x] `bin/typecheck` — no errors
- [x] `bundle exec rubocop` on changed files — no offenses

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
+++ b/lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
@@ -123,9 +123,11 @@ module Tapioca
             # Reflect to see if `ActiveModel::Type::Value` is being used first.
             getter_type = Tapioca::Dsl::Helpers::ActiveModelTypeHelper.type_for(column_type)
 
-            # Fallback to String as `ActiveRecord::Encryption::EncryptedAttributeType` inherits from
-            # `ActiveRecord::Type::Text` which inherits from `ActiveModel::Type::String`.
-            return "::String" if getter_type == "T.untyped"
+            # Otherwise resolve the wrapped `cast_type` (e.g. a serialized column), which `encrypts`
+            # preserves. This naturally falls back to `::String` for plain encrypted string/text columns.
+            if getter_type == "T.untyped" && column_type.respond_to?(:cast_type)
+              return type_for_activerecord_value(column_type.cast_type, column_nullability:)
+            end
 
             as_non_nilable_if_persisted_and_not_nullable(getter_type, column_nullability:)
           when ActiveRecord::Type::String
@@ -237,7 +239,7 @@ module Tapioca
         #: (ActiveRecord::Type::Serialized column_type) -> String
         def serialized_column_type(column_type)
           case column_type.coder
-          when ActiveRecord::Coders::YAMLColumn
+          when ActiveRecord::Coders::ColumnSerializer
             case column_type.coder.object_class
             when Array.singleton_class
               "T::Array[T.untyped]"
@@ -253,8 +255,11 @@ module Tapioca
 
         #: (untyped column_type) -> bool
         def not_nilable_serialized_column?(column_type)
+          if defined?(ActiveRecord::Encryption) && ActiveRecord::Encryption::EncryptedAttributeType === column_type
+            column_type = column_type.cast_type
+          end
           return false unless column_type.is_a?(ActiveRecord::Type::Serialized)
-          return false unless column_type.coder.is_a?(ActiveRecord::Coders::YAMLColumn)
+          return false unless column_type.coder.is_a?(ActiveRecord::Coders::ColumnSerializer)
 
           [Array.singleton_class, Hash.singleton_class].include?(column_type.coder.object_class.singleton_class)
         end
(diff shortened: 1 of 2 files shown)


lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
```diff
@@ -123,9 +123,11 @@ def type_for_activerecord_value(column_type, column_nullability:)
...
             getter_type = Tapioca::Dsl::Helpers::ActiveModelTypeHelper.type_for(column_type)
 
-            # Fallback to String as `ActiveRecord::Encryption::EncryptedAttributeType` inherits from
-            # `ActiveRecord::Type::Text` which inherits from `ActiveModel::Type::String`.
-            return "::String" if getter_type == "T.untyped"
+            # Otherwise resolve the wrapped `cast_type` (e.g. a serialized column), which `encrypts`
+            # preserves. This naturally falls back to `::String` for plain encrypted string/text columns.
+            if getter_type == "T.untyped" && column_type.respond_to?(:cast_type)
+              return type_for_activerecord_value(column_type.cast_type, column_nullability:)
+            end
```
paracycle, line 130: I am not sure why this would be needed, since `ActiveModelTypeHelper.type_for` is explicitly doing the `cast_type` (and `deserialize`, `cast`) signature checks for type discovery, as long as they exist. If we want to extend that logic so that it does runtime type discovery, then that needs to live inside the method, not here.

lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
```diff
@@ -237,7 +239,7 @@ def enum_setter_type(column_type)
         #: (ActiveRecord::Type::Serialized column_type) -> String
         def serialized_column_type(column_type)
           case column_type.coder
-          when ActiveRecord::Coders::YAMLColumn
```
paracycle, line 240: This might still be needed to support older versions of Rails. Can you please find when this coder name changed?

lib/tapioca/dsl/helpers/active_record_column_type_helper.rb
```diff
@@ -253,8 +255,11 @@ def serialized_column_type(column_type)
 
         #: (untyped column_type) -> bool
         def not_nilable_serialized_column?(column_type)
+          if defined?(ActiveRecord::Encryption) && ActiveRecord::Encryption::EncryptedAttributeType === column_type
+            column_type = column_type.cast_type
+          end
```
paracycle, line 260: Again, if we will be doing the `cast_type` runtime type discovery, then we should do it for all column types, and not just for the encryption one.

PR #2653, opened 2026-06-23
Handle payload superclass redefinition errors during gem RBI validation

### LINKED ISSUE #1834: Automatically resolve superclass redefinition errors while generating gem RBIs
We already typecheck resulting RBIs to convert strictness to false: https://github.com/Shopify/tapioca/blob/ba2951e75fbea6bad8b4ea0641ea3df38eb9e60d/lib/tapioca/helpers/rbi_files_helper.rb#L86

Similarly if we run into a superclass redefinition error we can populate `sorbet/config` with `--suppress-payload-superclass-redefinition-for=ConstantName` to prevent typecheck errors.

We probably will want to update the payload as well so this should still inform the user about the mismatch that's happening instead of fixing it silently.

zeeshankhan-05:
### Motivation

Fixes #1834.

When Tapioca validates generated gem RBIs, some definitions can redefine a Sorbet payload class with a different superclass. Sorbet reports this as error `5012` and recommends adding `--suppress-payload-superclass-redefinition-for=ConstantName` to `sorbet/config`. Tapioca already auto-adjusts RBI strictness for error `4010`, but it did not handle `5012` or inform users about the mismatch.

Closes #1834

### Implementation

- Keep the existing `--stop-after namer` validation pass for error `4010` auto-strictness behavior.
- When `auto_strictness` is enabled, run a second full Sorbet pass to detect resolver-level error `5012` (the namer pass alone does not surface this error).
- Parse the affected constant from Sorbet error output and append `--suppress-payload-superclass-redefinition-for=ConstantName` to `sorbet/config` when missing.
- Use exact-line idempotency (`config.lines(chomp: true).include?(flag)`) so similar suppressions for other constants do not block the correct entry.
- Preserve existing `sorbet/config` content when appending (no stripping of trailing whitespace or blank lines).
- Print a user-facing message when a suppression is added or when it is already present, so the mismatch is not fixed silently.

### Tests

Added an integration test in the `describe "strictness"` block of `spec/tapioca/cli/gem_spec.rb` (following maintainer guidance on #1834). The test exercises `tapioca gem`, verifies the suppression is written to `sorbet/config`, verifies user-facing output on first and second runs, and confirms idempotency.

**Ran locally on the clean branch (1 commit ahead of `upstream/main`):**

- `bin/test spec/tapioca/cli/gem_spec.rb` — 67 tests, 494 assertions, 0 failures, 0 errors, 1 skip
- `bin/test spec/tapioca/cli/dsl_spec.rb` — 69 tests, 335 assertions, 0 failures, 0 errors, 0 skips
- `bin/typecheck` — pass
- `bin/style` — pass

An earlier full-suite run (`bin/test`) completed with 794 tests, 0 failures, 2 errors, and 2 skips. The two errors were unrelated `RubyLsp::Tapioca::Addon` `URI::InvalidURIError` failures caused by spaces in the local directory path; they are not related to this change.

### DIFF
--- a/lib/tapioca/helpers/rbi_files_helper.rb
+++ b/lib/tapioca/helpers/rbi_files_helper.rb
@@ -129,23 +129,12 @@ module Tapioca
         ERR
       end
 
-      if auto_strictness
-        redef_errors = errors.select { |error| error.code == 4010 }
-        update_gem_rbis_strictnesses(redef_errors, gem_dir) if redef_errors.any?
-
-        payload_superclass_errors = errors.select do |error|
-          error.more.any? { |line| line.include?(SUPPRESS_PAYLOAD_SUPERCLASS_REDEFINITION_FLAG) }
-        end
-        if payload_superclass_errors.any?
-          update_sorbet_config_for_payload_superclass_redefinitions(payload_superclass_errors)
-        end
-      end
+      handled_errors = apply_validation_fixes(errors, gem_dir: gem_dir, auto_strictness: auto_strictness)
 
       Kernel.raise Tapioca::Error, error_messages.join("\n") if parse_errors.any?
 
-      unhandled_errors = errors.reject do |error|
-        auto_fixable_validation_error?(error, gem_dir: gem_dir, dsl_dir: dsl_dir)
-      end
+      unhandled_errors = errors - handled_errors
+      unhandled_errors.reject! { |error| ignored_validation_error?(error, gem_dir: gem_dir, dsl_dir: dsl_dir) }
 
       if unhandled_errors.empty?
         say("  No errors found\n\n", [:green, :bold])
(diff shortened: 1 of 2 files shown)


zeeshankhan-05:
Hi @KaanOzkan. I appreciate the help with #1834 about using the `describe "strictness"` block in `spec/tapioca/cli/gem_spec.rb` and verifying suppression, user messaging, and idempotency. I followed that approach in this PR. Whenever you have some time, would you be able to review my PR?

zeeshankhan-05:
I have signed the CLA!

lib/tapioca/helpers/rbi_files_helper.rb
```diff
@@ -86,7 +86,20 @@ def validate_rbi_files(command:, gem_dir:, dsl_dir:, auto_strictness:, gems: [],
 
       errors = Spoom::Sorbet::Errors::Parser.parse_string(res.err || "")
 
-      if errors.empty?
+      payload_superclass_errors = T.let([], T::Array[Spoom::Sorbet::Errors::Error])
+      if auto_strictness
+        payload_superclass_res = sorbet(
```
KaanOzkan: Instead of 2 calls to Sorbet, I think we just need to use `--stop-after resolver` instead of `namer` and we can do one call to look at errors. There should also be a early return if there are no errors so that we don't parse the result twice.

lib/tapioca/helpers/rbi_files_helper.rb
```diff
@@ -258,6 +272,64 @@ def extract_methods_and_attrs(nodes)
...
+
+      nil
+    end
+
+    #: (String constant) -> void
+    def add_payload_superclass_suppression_to_config(constant)
+      flag = "#{SUPPRESS_PAYLOAD_SUPERCLASS_REDEFINITION_FLAG}=#{constant}"
+      config_path = Tapioca::SORBET_CONFIG_FILE
+      config = File.exist?(config_path) ? File.read(config_path) : ""
+      added = !config.lines(chomp: true).include?(flag)
```
KaanOzkan:
```suggestion
      flag_already_present = !config.lines(chomp: true).include?(flag)
```

We can also simplify the branching in the method. For example, you can print in the first conditional instead of creating a separate conditional for printing.

lib/tapioca/helpers/rbi_files_helper.rb
```diff
@@ -86,7 +86,20 @@ def validate_rbi_files(command:, gem_dir:, dsl_dir:, auto_strictness:, gems: [],
...
+      if auto_strictness
+        payload_superclass_res = sorbet(
+          "--no-config",
+          "--error-url-base=#{error_url_base}",
+          dsl_dir,
+          gem_dir,
+        )
+        payload_superclass_errors = Spoom::Sorbet::Errors::Parser
+          .parse_string(payload_superclass_res.err || "")
+          .select { |error| error.code == 5012 }
```
KaanOzkan: 5012 errors isn't enough of an indicator for payload superclass mismatch as it's used for any parent/subclass mismatch. You can use `error.more` directly here instead.

We need a new test for this too:

```rb
it "must not add a payload suppression for non-payload superclass redefinitions" do
  @project.write!("sorbet/rbi/dsl/non_payload_superclass_conflict.rbi", <<~RBI)
    # typed: true

    class NonPayloadSuperclassConflict < ::Object
    end
  RBI

  @project.write!("sorbet/rbi/gems/bar@0.3.0.rbi", <<~RBI)
    # typed: true

    class NonPayloadSuperclassConflict < ::String
    end
  RBI

  result = @project.tapioca("gem foo")

  refute_includes(
    @project.read("sorbet/config"),
    "--suppress-payload-superclass-redefinition-for=NonPayloadSuperclassConflict",
  )

  assert_empty_stderr(result)
  assert_success_status(result)
end

```

zeeshankhan-05:
I appreciate the feedback! I think I should have addressed all of your concerns and added some extra testing to ensure things are running properly with those updates

lib/tapioca/helpers/rbi_files_helper.rb
```diff
@@ -258,6 +263,57 @@ def extract_methods_and_attrs(nodes)
...
+      config_path = Tapioca::SORBET_CONFIG_FILE
+      config = File.exist?(config_path) ? File.read(config_path) : ""
+      flag_already_present = config.lines(chomp: true).include?(flag)
+
+      if flag_already_present
+        say(
+          "\n  Payload superclass of `#{constant}` was redefined; `#{flag}` is already in sorbet/config",
+          [:yellow, :bold],
+        )
+      else
```
KaanOzkan: We can remove this `else` by making the above `if` return early:

```rb
        say(
          "\n  Payload superclass of `#{constant}` was redefined; `#{flag}` is already in sorbet/config\n",
          [:yellow, :bold],
        )
        return
```
Note the extra `\n` at the end of the string.

KaanOzkan: Rest of the PR looks good to me. Also triggered CI.

zeeshankhan-05:
Sounds good! Just made another commit to address that, should be ready to merge now (hopefully)

KaanOzkan:
Added a commit to not hide this behind the `auto-strictness` flag.

KaanOzkan (approved):

PR #2657, opened 2026-06-24
Fix generic type registry cache collisions

st0012:
## Problems

- Generic type clones were cached only by generated name, so a redefined constant with the same name could reuse the stale clone from the previous constant.
- Newer Sorbet casts can take a module fast path that bypasses Tapioca's existing generic type coercion patch.
- This behavioral fix was previously mixed into #2648, making it harder to review separately from the dependency bump.

## Solutions

- Cache generic type clones by the original constant identity, then by generated name.
- Preserve Tapioca generic type overrides when Sorbet casts use the module fast path.
- Add regression coverage for redefined generic constants and generic interface casts.


KaanOzkan (approved):
lgtm 😛

PR #2667, opened 2026-07-07
Infer attr_accessor writer signatures

Morriar:
## Summary

This PR teaches gem RBI generation to infer signatures for writers generated by `attr_accessor`.

When code uses:

```ruby
sig { returns(Integer) }
attr_accessor :foo
```

`sorbet-runtime` attaches the pending signature to the generated reader (`foo`) only. The generated writer (`foo=`) has no runtime signature (https://github.com/sorbet/sorbet/issues/7520), so Tapioca previously emitted it as untyped:

```rbi
sig { returns(::Integer) }
def foo; end

def foo=(_arg0); end
```

With this change, Tapioca infers the writer signature from the matching reader signature:

```rbi
sig { returns(::Integer) }
def foo; end

sig { params(foo: ::Integer).returns(::Integer) }
def foo=(foo); end
```

## Details

The inference is intentionally conservative:

- It only applies when the writer has no runtime signature.
- The method must look like an attr-style writer:
    - method name ends with `=`
    - exactly one unnamed required parameter
- A matching reader must exist on the same constant.
- The reader must have a simple zero-argument signature.
- The writer and reader’s original source locations must match, ensuring we only infer signatures for methods generated by the same `attr_accessor` call and avoid separately-defined or manually-defined writer methods.

### DIFF
--- a/lib/tapioca/gem/listeners/methods.rb
+++ b/lib/tapioca/gem/listeners/methods.rb
@@ -69,6 +69,7 @@ module Tapioca
 
           begin
             signature = signature_of!(method)
+            signature ||= inferred_attr_writer_signature(method, constant)
             method = signature.method if signature #: UnboundMethod
 
             case @pipeline.method_definition_in_gem(method.name, constant)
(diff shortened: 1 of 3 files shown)


vinistock (approved):

lib/tapioca/gem/listeners/methods.rb
```diff
@@ -192,6 +193,60 @@ def method_names_by_visibility(mod)
...
+          return if reader_signature.rest_type
+          return if reader_signature.keyrest_type
+          return if reader_signature.block_type
+          return unless same_source_location?(writer_method, reader_signature.method)
+
+          T::Private::Methods::Signature.new(
+            method: writer_method,
+            method_name: writer_method.name,
+            raw_arg_types: { reader_method.name => reader_signature.return_type },
+            raw_return_type: T::Private::Types::Void::Private::INSTANCE,
```
paracycle: Technically writers return the same type as the type of the argument, so that `a = b = 42` works, so this is more correct as:

```suggestion
            raw_return_type: reader_signature.return_type,
```

spec/tapioca/gem/pipeline_spec.rb
```diff
@@ -1933,6 +1933,61 @@ def foo; end
...
+
+          sig { params(c: ::Symbol).void }
+          def c=(c); end
+        end
+      RBI
+
+      assert_equal(output, compile)
+    end
+
+    it "does not infer attr_accessor writer signatures for manually defined writers" do
```
paracycle, line 1991: Can we also add a test for separate reader and writer defined for the same attr?
```ruby
sig { returns(Integer) }
attr_reader :foo

attr_writer :foo
```
should not be generating a signature for the writer.

lib/tapioca/gem/listeners/methods.rb
```diff
@@ -192,6 +193,60 @@ def method_names_by_visibility(mod)
...
+
+        #: (UnboundMethod method, Module[top] constant) -> UnboundMethod?
+        def attr_reader_for_writer(method, constant)
+          method_name = method.name.to_s
+          return unless method_name.end_with?("=")
+          return unless method.parameters == [[:req]]
+
+          reader_method = constant.instance_method(method_name.delete_suffix("=").to_sym)
+          return unless method_owned_by_constant?(reader_method, constant)
+
```
paracycle: I think we can move the `same_source_location?` call here, so that we can return early if the reader (that we just found) and the writer are not defined by the same line of code.

Morriar:
Handled the review comments:

* inferred writer signatures now return the reader type instead of `void`
* moved the source-location check into reader matching
* added coverage for separate `attr_reader`/`attr_writer` declarations

paracycle (approved):

amomchilov:
This looks like a pretty big whole in Sorbet runtime. It might be better to fix it there, than to try to cope in Tapioca after-the-fact  
  
https://github.com/sorbet/sorbet/issues/7520

PR #2669, opened 2026-07-13
Reset Bootsnap RBS cache on lockfile changes & remove cache priming for CI

KaanOzkan:
### Motivation

1. Tapioca could reuse stale RBS rewrite entries after `tapioca` or one of its dependencies changed because Bootsnap does not include those dependencies in its cache key.
2. `--only-bootsnap-rbs-cache` was added to prime the cache in CI, but sharing Bootsnap caches across CI builds is unreliable and the read-only path complicates cache setup for little benefit.

### Implementation

1. Store the current `Gemfile.lock` digest in Tapioca’s dedicated Bootsnap cache. When the digest changes, delete the Bootsnap payload so the current run rebuilds it.
2. Remove `--only-bootsnap-rbs-cache`, read-only cache handling, and their documentation. Keep the cache focused on writable local runs.

### Tests

Added coverage that changing `Gemfile.lock` resets the existing Bootsnap payload.

### DIFF
--- a/lib/tapioca/rbs/rewriter.rb
+++ b/lib/tapioca/rbs/rewriter.rb
@@ -30,7 +30,7 @@ module Tapioca
       end
     end
 
-    module BootsnapSetup
+    module BootsnapIntegration
       class << self
         extend T::Sig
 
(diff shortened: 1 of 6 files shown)


lib/tapioca/rbs/rewriter.rb
```diff
@@ -29,32 +31,41 @@ def setup(**_kwargs)
...
+    end
   end
 end
 
-# When TAPIOCA_RBS_CACHE=1, set up bootsnap with a dedicated cache directory
-# and load require-hooks so the RBS-rewritten iseqs get cached. Subsequent
-# runs read the rewritten iseq directly and skip the rewrite.
-#
-# After our setup, BootsnapGuard is prepended so the host application can't
-# replace our cache directory.
```
alexcrocha, line 40: nit: I found this comment useful to better understand `BootsnapGuard`. I'd keep it, but up to you

lib/tapioca/rbs/rewriter.rb
```diff
@@ -1,6 +1,8 @@
 # typed: strict
 # frozen_string_literal: true
 
+require "tapioca/rbs/bootsnap_cache"
```
alexcrocha: We are only needing this inside the `if ENV["TAPIOCA_RBS_CACHE"] == "1"` branch, as far as I can tell. Should we move this `require` there and avoid requiring `bundle`, `digest`, and `fileutils` by default?

alexcrocha (approved):
Left a couple of comments, but overall looks good to me 👍

lib/tapioca/rbs/rewriter.rb
KaanOzkan: Thanks, I kept it.

PR #2672, opened 2026-07-14
Make aasm compiler v6.0 aware

### LINKED ISSUE #2671: aasm 6.0 compiler emits non-existent event methods for namespaced state machines
The built-in `Tapioca::Dsl::Compilers::AASM` generates method definitions that do not exist at runtime for namespaced state machines when using `aasm` >= 6.0: https://github.com/Shopify/tapioca/blob/7b7e4d893cccf71354a0c0934dd2d079d6335752/lib/tapioca/dsl/compilers/aasm.rb#L98-L111

This causes issues with the introduction of https://github.com/aasm/aasm/pull/880, since the plain methods are no longer defined. From that pull request:

> This is a breaking change: plain-named event methods (sell!, may_sell?) are no longer defined when a namespace is active. Update all call sites to use the namespaced form.

Only the namespaced form exists, but the compiler still emits the plain forms unconditionally even though none exist at runtime or test time. (I discovered this during a Dependabot update that caused unit tests to fail.)

## Environment

- `tapioca`: 0.19.1
- `aasm`: 6.0.0
- `sorbet-static`: 0.6.13329

## Repro

```ruby
class Light
  include AASM

  aasm :power, namespace: :power do
    state :off, initial: true
    state :on

    event :turn_on do
      transitions from: :off, to: :on
    end
  end
end
```

Runtime behavior under aasm 6.0.0:

```ruby
Light.new.respond_to?(:turn_on!)           # => false   (plain form removed in aasm 6)
Light.new.respond_to?(:turn_on_power!)     # => true    (namespaced form)
Light.new.respond_to?(:may_turn_on?)       # => false
Light.new.respond_to?(:may_turn_on_power?) # => true
```

Generated RBI:

```ruby
class Light
  def turn_on(*opts, &block); end                     # phantom
  def turn_on!(*opts, &block); end                    # phantom
  def turn_on_without_validation!(*opts, &block); end # phantom
  def may_turn_on?; end                               # phantom

  def turn_on_power(*opts, &block); end               # real
  def turn_on_power!(*opts, &block); end              #
  def may_turn_on_power?; end                         # real
end
```

## Notes

I'm going to try to test a patch in our internal use cases to ensure compatibility and will try to submit an upstream pull request. Also related: #2666 (has no updates to the compiler)

allan-thatch:
### Motivation

Fixes #2671. Also see ["Migrating from AASM version 5 to 6"](https://github.com/aasm/aasm/blob/master/README_FROM_VERSION_5_TO_6.md#namespaced-event-methods-no-longer-define-plain-named-methods).

### Implementation

The aasm compiler now (a) withholds plain methods when namespaced *and* on v6 and (b) adds the namespaced `_without_validation!` method when on v6.

### Tests

The compiler test only accommodates the >=v6.0 path (see #2672 for the discussion)

### DIFF
--- a/spec/tapioca/dsl/compilers/aasm_spec.rb
+++ b/spec/tapioca/dsl/compilers/aasm_spec.rb
@@ -359,16 +359,7 @@ module Tapioca
                   def foo_sleeping?; end
 
                   sig { returns(T::Boolean) }
-                  def may_run?; end
-
-                  sig { returns(T::Boolean) }
                   def may_run_foo?; end
-
-                  sig { params(opts: T.untyped, block: T.nilable(T.proc.void)).returns(T.untyped) }
-                  def run(*opts, &block); end
-
-                  sig { params(opts: T.untyped, block: T.nilable(T.proc.void)).returns(T.untyped) }
-                  def run!(*opts, &block); end
 
                   sig { params(opts: T.untyped, block: T.nilable(T.proc.void)).returns(T.untyped) }
                   def run_foo(*opts, &block); end
@@ -377,7 +368,7 @@ module Tapioca
                   def run_foo!(*opts, &block); end
 
                   sig { params(opts: T.untyped, block: T.nilable(T.proc.void)).returns(T.untyped) }
-                  def run_without_validation!(*opts, &block); end
+                  def run_foo_without_validation!(*opts, &block); end
 
                   class << self
                     sig { params(args: T.untyped, block: T.nilable(T.proc.bind(PrivateAASMMachine).void)).returns(PrivateAASMMachine) }
--- a/spec/tapioca/dsl/compilers/aasm_spec.rb
+++ b/spec/tapioca/dsl/compilers/aasm_spec.rb
@@ -14,6 +14,17 @@ module Tapioca
             require "aasm"
           end
 
+          # Runs the block with `AASM::VERSION` stubbed, so both the aasm < 6 and
+          # aasm >= 6 code paths can be exercised regardless of the installed version.
+          #: (String version) { -> void } -> void
+          def with_aasm_version(version, &block)
+            original = ::AASM::VERSION
+            ::Tapioca::Runtime.silence_warnings { ::AASM.const_set(:VERSION, version) }
+            block.call
+          ensure
+            ::Tapioca::Runtime.silence_warnings { ::AASM.const_set(:VERSION, original) }
+          end
+
           describe "initialize" do
             it "gathers no constants if there are no classes that include AASM" do
               assert_empty(gathered_constants)
@@ -345,7 +356,7 @@ module Tapioca
                 end
               RUBY
 
-              expected = template(<<~RBI)
+              rbi = <<~RBI
                 # typed: strong
 
                 class StateMachine
@@ -476,7 +487,14 @@ module Tapioca
                 end
               RBI
 
-              assert_equal(expected, rbi_for(:StateMachine))
+              # Assert against the currently installed aasm version.
+              assert_equal(template(rbi), rbi_for(:StateMachine))
+
+              # Also exercise the aasm < 6 code path, which no longer matches the
+              # installed version once aasm >= 6 is in use.
+              with_aasm_version("5.5.2") do
+                assert_equal(template(rbi), rbi_for(:StateMachine))
+              end
             end
           end
         end
--- a/spec/tapioca/dsl/compilers/aasm_spec.rb
+++ b/spec/tapioca/dsl/compilers/aasm_spec.rb
@@ -345,7 +345,7 @@ module Tapioca
                 end
               RUBY
 
-              expected = <<~RBI
+              expected = template(<<~RBI)
                 # typed: strong
 
                 class StateMachine
@@ -358,6 +358,7 @@ module Tapioca
                   sig { returns(T::Boolean) }
                   def foo_sleeping?; end
 
+                <% if aasm_version(">= 6") %>
                   sig { returns(T::Boolean) }
                   def may_run_foo?; end
 
@@ -369,6 +370,28 @@ module Tapioca
 
                   sig { params(opts: T.untyped, block: T.nilable(T.proc.void)).returns(T.untyped) }
                   def run_foo_without_validation!(*opts, &block); end
+                <% else %>
+                  sig { returns(T::Boolean) }
+                  def may_run?; end
+
+                  sig { returns(T::Boolean) }
+                  def may_run_foo?; end
+
+                  sig { params(opts: T.untyped, block: T.nilable(T.proc.void)).returns(T.untyped) }
+                  def run(*opts, &block); end
+
+                  sig { params(opts: T.untyped, block: T.nilable(T.proc.void)).returns(T.untyped) }
+                  def run!(*opts, &block); end
+
+                  sig { params(opts: T.untyped, block: T.nilable(T.proc.void)).returns(T.untyped) }
+                  def run_foo(*opts, &block); end
+
+                  sig { params(opts: T.untyped, block: T.nilable(T.proc.void)).returns(T.untyped) }
+                  def run_foo!(*opts, &block); end
+
+                  sig { params(opts: T.untyped, block: T.nilable(T.proc.void)).returns(T.untyped) }
+                  def run_without_validation!(*opts, &block); end
+                <% end %>
 
                   class << self
                     sig { params(args: T.untyped, block: T.nilable(T.proc.bind(PrivateAASMMachine).void)).returns(PrivateAASMMachine) }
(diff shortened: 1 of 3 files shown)


allan-thatch:
I have signed the CLA!

spec/tapioca/dsl/compilers/aasm_spec.rb
```diff
@@ -358,26 +358,17 @@ def foo_running?; end
...
-
                   sig { params(opts: T.untyped, block: T.nilable(T.proc.void)).returns(T.untyped) }
                   def run_foo(*opts, &block); end
 
                   sig { params(opts: T.untyped, block: T.nilable(T.proc.void)).returns(T.untyped) }
                   def run_foo!(*opts, &block); end
 
                   sig { params(opts: T.untyped, block: T.nilable(T.proc.void)).returns(T.untyped) }
-                  def run_without_validation!(*opts, &block); end
+                  def run_foo_without_validation!(*opts, &block); end
```
KaanOzkan, line 371: Thanks for the PR. Since this version is brand new, I'd prefer if we kept older tests, we have a similar pattern in [here](https://github.com/Shopify/tapioca/blob/68ea7e6f5b5008c052307673ca167c31ae52d67e/spec/tapioca/dsl/compilers/active_record_relations_spec.rb#L176-L184). It is then executed as ERB in [here](https://github.com/Shopify/tapioca/blob/68ea7e6f5b5008c052307673ca167c31ae52d67e/lib/tapioca/helpers/test/template.rb#L15-L23). 

I think you can define a similar helper in this file:
```rb
def aasm_version(selector)
  ::Gem::Requirement.new(selector).satisfied_by?(::Gem::Version.new(::AASM::VERSION))
end
```

allan-thatch: Ah, perfect! Thanks for linking those. Updated!

allan-thatch: I'm realizing now that there's no tooling currently (like Rails in the CI matrix) that actually installs and drives the test suite for the older version of `aasm`. I'll push up a commit to stub this out in tests vs. adding another variant to the test matrix (this seemed heavy, and not quite on the same level as Ruby/Rails). I've included sample implementations for both below, so LMK if you'd prefer the other!

<details>
<summary>Test matrix</summary>

```patch
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index cc6424c9..7f487af6 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,7 @@ jobs:
         ruby: ["3.2", "3.3", "3.4", "4.0","head"]
         rails: ["8.0", "current", "main"]
         rubygems: ["4.0.10"]
+        aasm: ["current"]
         exclude:
           - ruby: "3.2"
             rails: "main"
@@ -52,9 +53,13 @@ jobs:
             experimental: true
           - ruby: "head"
             experimental: true
-    name: Ruby ${{ matrix.ruby }} - Rails ${{ matrix.rails }} - RubyGems ${{ matrix.rubygems }}
+          - ruby: "3.4"
+            rails: "current"
+            aasm: "5.5"
+    name: Ruby ${{ matrix.ruby }} - Rails ${{ matrix.rails }} - RubyGems ${{ matrix.rubygems }} - AASM ${{ matrix.aasm }}
     env:
       RAILS_VERSION: ${{ matrix.rails }}
+      AASM_VERSION: ${{ matrix.aasm }}
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
       - name: Install protobuf
diff --git a/Gemfile b/Gemfile
index 43b343c3..83e695fa 100644
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,8 @@ gemspec
 CURRENT_RAILS_VERSION = "8.1"
 rails_version = ENV.fetch("RAILS_VERSION", CURRENT_RAILS_VERSION)
 
+aasm_version = ENV.fetch("AASM_VERSION", "current")
+
 gem "minitest"
 gem "minitest-hooks"
 gem "minitest-reporters"
@@ -52,7 +54,11 @@ group :development, :test do
   gem "sidekiq"
   gem "nokogiri"
   gem "config"
-  gem "aasm"
+  if aasm_version == "current"
+    gem "aasm"
+  else
+    gem "aasm", "~> #{aasm_version}"
+  end
   gem "bcrypt"
   gem "xpath"
   gem "kredis"
```
</details>

<details>
<summary>Test stub/helper</summary>

```patch
diff --git a/spec/tapioca/dsl/compilers/aasm_spec.rb b/spec/tapioca/dsl/compilers/aasm_spec.rb
index 52aebda2..d6df99e2 100644
--- a/spec/tapioca/dsl/compilers/aasm_spec.rb
+++ b/spec/tapioca/dsl/compilers/aasm_spec.rb
@@ -14,6 +14,17 @@ module Tapioca
             require "aasm"
           end
 
+          # Runs the block with `AASM::VERSION` stubbed, so both the aasm < 6 and
+          # aasm >= 6 code paths can be exercised regardless of the installed version.
+          #: (String version) { -> void } -> void
+          def with_aasm_version(version, &block)
+            original = ::AASM::VERSION
+            ::Tapioca::Runtime.silence_warnings { ::AASM.const_set(:VERSION, version) }
+            block.call
+          ensure
+            ::Tapioca::Runtime.silence_warnings { ::AASM.const_set(:VERSION, original) }
+          end
+
           describe "initialize" do
             it "gathers no constants if there are no classes that include AASM" do
               assert_empty(gathered_constants)
@@ -345,7 +356,7 @@ module Tapioca
                 end
               RUBY
 
-              expected = template(<<~RBI)
+              rbi = <<~RBI
                 # typed: strong
 
                 class StateMachine
@@ -476,7 +487,14 @@ module Tapioca
                 end
               RBI
 
-              assert_equal(expected, rbi_for(:StateMachine))
+              # Assert against the currently installed aasm version.
+              assert_equal(template(rbi), rbi_for(:StateMachine))
+
+              # Also exercise the aasm < 6 code path, which no longer matches the
+              # installed version once aasm >= 6 is in use.
+              with_aasm_version("5.5.2") do
+                assert_equal(template(rbi), rbi_for(:StateMachine))
+              end
             end
           end
         end
```
</details>

KaanOzkan: Ah my bad. I'm leaning towards not adding a new pipeline just for this, let's keep the support for older `aasm` versions but skip the tests. Sorry about the extra work, I forgot.

spec/tapioca/dsl/compilers/aasm_spec.rb
```diff
@@ -358,6 +369,19 @@ def foo_running?; end
                   sig { returns(T::Boolean) }
                   def foo_sleeping?; end
 
+                <% if aasm_version(">= 6") %>
```
KaanOzkan:
```suggestion
```

spec/tapioca/dsl/compilers/aasm_spec.rb
```diff
@@ -358,6 +369,19 @@ def foo_running?; end
...
+
+                  sig { params(opts: T.untyped, block: T.nilable(T.proc.void)).returns(T.untyped) }
+                  def run_foo(*opts, &block); end
+
+                  sig { params(opts: T.untyped, block: T.nilable(T.proc.void)).returns(T.untyped) }
+                  def run_foo!(*opts, &block); end
+
+                  sig { params(opts: T.untyped, block: T.nilable(T.proc.void)).returns(T.untyped) }
+                  def run_foo_without_validation!(*opts, &block); end
+                <% else %>
```
KaanOzkan: We can remove the whole branch
```suggestion
```

spec/tapioca/dsl/compilers/aasm_spec.rb
```diff
@@ -462,7 +487,14 @@ def success(&block); end
...
 
-              assert_equal(expected, rbi_for(:StateMachine))
+              # Assert against the currently installed aasm version.
+              assert_equal(template(rbi), rbi_for(:StateMachine))
+
+              # Also exercise the aasm < 6 code path, which no longer matches the
+              # installed version once aasm >= 6 is in use.
+              with_aasm_version("5.5.2") do
+                assert_equal(template(rbi), rbi_for(:StateMachine))
+              end
```
KaanOzkan:
```suggestion
              assert_equal(expected, rbi_for(:StateMachine))
```

spec/tapioca/dsl/compilers/aasm_spec.rb
allan-thatch: All good, I think I can just revert the last two commits.

KaanOzkan (approved):

PR #2673, opened 2026-07-15
Fix generic type validation for T::Props and composite members

### LINKED ISSUE #2130: Tapioca DSL crashes when using `T::Struct` containing generic members
# Summary

When running `tapioca dsl` in a Rails app that

* Has runtime checks (that deliberately crash the app) enabled
* Contains a `T::Struct` with a member whose type is a generic interface
* Instantiates that struct during app startup with a concrete implementation of the generic interface

Tapioca will crash while trying to create the struct because of a type mismatch error between the interface and the implementation.

I created a repository containing what seems to be the absolute minimum configuration necessary to activate this bug. You can find it here: https://github.com/amanda-mitchell/tapioca-bug-repro

# Workarounds

Although I'd love to see this get a long term fix, I'd also like to document some potential workarounds for folks who might run across this in the meantime.

## Use `setter_validate`

`T::Struct` contains an undocumented option for member declarations called `setter_validate`, and this is expected to a proc that performs validation side effects. Providing an empty (no-op) lambda will force the implementation of `T::Struct` onto an alternate code path that does not activate the bug.

## Disable runtime checks while running Tapioca

If you have some code that can detect when Tapioca is running and set your implementation of `T::Configuration.call_validation_error_handler` to something that no-ops, you won't encounter the crash.

dduugg:
### Motivation

Fixes #2130.

When a constant is typed with a Tapioca generic type (e.g. a generic interface `Foo[Bar]`) as a `T::Props`/`T::Struct` member, instantiating or setting that member raises a spurious `TypeError`, even though the value is an instance of the underlying constant. This happens whether the member is typed with the generic directly (`const :x, Foo[Bar]`) or nested inside a composite type (`T::Hash[..., Foo[Bar]]`, `T.all(Foo[Bar], ...)`), and it blocks `bin/tapioca dsl` from loading applications that use generic types in prop/struct members.

### Implementation

`GenericTypeRegistry::GenericType` overrides `valid?` to validate against the underlying constant (`@underlying_type`) instead of the generic *clone* it wraps as `raw_type`. Previously it subclassed `T::Types::Simple`, but `Simple` validates with `obj.is_a?(raw_type)` and `sorbet-runtime` has fast paths keyed on `is_a?(T::Types::Simple)` that read `raw_type` directly and bypass `valid?`/`recursively_valid?` entirely:

- `T::Props::Private::SetterFactory` uses `simple_non_nil_proc`/`simple_nilable_proc` for `Simple` types, checking `val.is_a?(raw_type)` inline. This is what a bare `const :x, Foo[Bar]` (or `T.nilable(Foo[Bar])`) member hits.
- `T.nilable` union construction takes a `SimplePairUnion` fast path for `Simple` members.

Because `raw_type` is the clone rather than the underlying constant, every one of these rejects a valid instance. Subclassing `T::Types::Base` instead removes all of those fast paths, so all validation routes through the `valid?` override. `Base` is more minimal than `Simple`, so the pieces it does not provide are added: `name`, `build_type`, and `subtype_of_single?` (which returns `true`, matching the always-true `<=` we already define on the clone in `create_generic_type`).

This complements #2657, which fixed the `T.let`/`T.cast` coercion path but did not cover the setter/composite validation paths.

### Tests

Added tests to `spec/tapioca/runtime/generic_type_registry_spec.rb` covering both a bare `T::Struct` member (`const :member, SampleGenericInterface[Object]`) and a composite member (`T::Hash[Symbol, SampleGenericInterface[Object]]`), each constructed with an implementation instance. Both fail with a `TypeError` before this change and pass after.

### DIFF
--- a/lib/tapioca/runtime/generic_type_registry.rb
+++ b/lib/tapioca/runtime/generic_type_registry.rb
@@ -25,10 +25,9 @@ module Tapioca
 
       @type_variables = {}.compare_by_identity #: Hash[Module[top], Array[TypeVariableModule]]
 
-      # Subclasses `T::Types::Base`, not `T::Types::Simple`: `Simple` fast paths
-      # (e.g. `T::Props::Private::SetterFactory`) check `raw_type` directly, but our
-      # `raw_type` is the clone, so they reject instances of the underlying constant.
-      # `Base` routes all validation through the `valid?` override below.
+      # Subclasses `T::Types::Base` rather than `T::Types::Simple` so that all
+      # validation goes through the `valid?` override below. `Simple` fast paths
+      # check `raw_type` directly, which for us is the clone, not the underlying type.
       class GenericType < T::Types::Base
         #: Module[top]
         attr_reader :raw_type
--- a/lib/tapioca/runtime/generic_type_registry.rb
+++ b/lib/tapioca/runtime/generic_type_registry.rb
@@ -25,12 +25,10 @@ module Tapioca
 
       @type_variables = {}.compare_by_identity #: Hash[Module[top], Array[TypeVariableModule]]
 
-      # Subclasses `T::Types::Base` rather than `T::Types::Simple` on purpose. Our
-      # `raw_type` is the generic *clone*, not the underlying constant, so the
-      # `obj.is_a?(raw_type)` checks that `Simple` and its `is_a?(Simple)` fast paths
-      # (e.g. `T::Props::Private::SetterFactory`) perform against `raw_type` would
-      # reject instances of the underlying constant. `Base` routes all validation
-      # through our `valid?` override instead.
+      # Subclasses `T::Types::Base`, not `T::Types::Simple`: `Simple` fast paths
+      # (e.g. `T::Props::Private::SetterFactory`) check `raw_type` directly, but our
+      # `raw_type` is the clone, so they reject instances of the underlying constant.
+      # `Base` routes all validation through the `valid?` override below.
       class GenericType < T::Types::Base
         #: Module[top]
         attr_reader :raw_type
@@ -61,9 +59,7 @@ module Tapioca
           nil
         end
 
-        # We are not implementing a runtime type checker, so it is enough for
-        # covariance/contravariance checks to pass, matching the always-true `<=`
-        # we define on the generic clone in `create_generic_type`.
+        # Matches the always-true `<=` we define on the clone in `create_generic_type`.
         #: (T::Types::Base type) -> bool
         private def subtype_of_single?(type)
           true
--- a/lib/tapioca/runtime/generic_type_registry.rb
+++ b/lib/tapioca/runtime/generic_type_registry.rb
@@ -38,6 +38,15 @@ module Tapioca
         def valid?(obj)
           obj.is_a?(@underlying_type)
         end
+
+        # `T::Types::Base#recursively_valid?` checks the clone `raw_type` instead of
+        # routing through `valid?`, so delegate to keep composite (`T.all`, `T::Hash`)
+        # and `T::Props` member validation consistent with the `valid?` override above.
+        # @override
+        #: (untyped obj) -> bool
+        def recursively_valid?(obj)
+          valid?(obj)
+        end
       end
 
       class << self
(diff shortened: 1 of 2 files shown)


KaanOzkan (commented):
My agent found a case where `sorbet-runtime` takes a [different code path](https://github.com/sorbet/sorbet/blob/bcf50834ef51c8b3f0362b231de82333b5b4826d/gems/sorbet-runtime/lib/types/props/private/setter_factory.rb#L49) for `T::Types::Simple` and bypasses `valid?` or `recursively_valid?` checks.

```rb
require "tapioca/internal"

module I
  extend T::Generic
  interface!
  E = type_member
end

class C
  include I
end

class S < T::Struct
  const :x, I[Object]
end

S.new(x: C.new)
```
results in `TypeError: Parameter 'x': Can't set S.x to #<C:0x0000000127c91288> (instance of C) - need a I[::Object]`


Maybe we shouldn't inherit from `T::Types::Simple` 🤔 , wdyt?

dduugg:
Good catch, you're right, and I've taken your suggestion.

The bare `T::Struct` member hits `SetterFactory`'s `is_a?(T::Types::Simple)` fast path, which reads `raw_type` directly and skips both `valid?` and `recursively_valid?`. Since our `raw_type` is the generic *clone* rather than the underlying constant, `val.is_a?(raw_type)` is false for an instance of the constant, so the overrides never got a chance to run. (`T.nilable(...)` union construction has the same Simple-keyed fast path.)

So instead of overriding `recursively_valid?`, `GenericType` now subclasses `T::Types::Base`. That removes the `is_a?(Simple)` fast paths entirely and routes every validation path through the `valid?` override. I filled in the pieces `Simple` had provided (`name`, `build_type`, and `subtype_of_single?`, which just returns `true`, matching the always-true `<=` we define on the clone) and dropped the now-unnecessary `recursively_valid?` override.

Added a spec for your exact repro (bare generic member) alongside the composite one. `bin/typecheck` and the generic specs are green.

lib/tapioca/runtime/generic_type_registry.rb
```diff
@@ -25,11 +25,19 @@ module GenericTypeRegistry
 
       @type_variables = {}.compare_by_identity #: Hash[Module[top], Array[TypeVariableModule]]
 
-      class GenericType < T::Types::Simple
+      # Subclasses `T::Types::Base`, not `T::Types::Simple`: `Simple` fast paths
+      # (e.g. `T::Props::Private::SetterFactory`) check `raw_type` directly, but our
+      # `raw_type` is the clone, so they reject instances of the underlying constant.
+      # `Base` routes all validation through the `valid?` override below.
```
KaanOzkan: I think this comment can be improved by changing the language and making it more concise, or be removed completely.

lib/tapioca/runtime/generic_type_registry.rb
```diff
@@ -38,6 +46,24 @@ def initialize(raw_type, underlying_type)
...
+        #: -> nil
+        def build_type
+          nil
+        end
+
+        # Matches the always-true `<=` we define on the clone in `create_generic_type`.
+        #: (T::Types::Base type) -> bool
+        private def subtype_of_single?(type)
+          true
+        end
```
KaanOzkan, line 65: I guess it's okay to allow everything while running `tapioca dsl`? There's a similar comment on `create_generic_type` below.

lib/tapioca/runtime/generic_type_registry.rb
dduugg: Good call. Tightened it to 3 lines and dropped the `SetterFactory` name-drop in ce867cc. I kept a short version since "why `Base` and not `Simple`" is the crux of this thread.

lib/tapioca/runtime/generic_type_registry.rb
dduugg: Right, the full rationale already lives on `create_generic_type`'s `<=` ("we act like all subtype checks pass", for serialization rather than runtime enforcement), so I left this as a one-line cross-reference to it instead of duplicating. And yes, allowing everything is fine here since `tapioca dsl` only serializes sigs, it doesn't enforce them at runtime.

KaanOzkan (approved):

PR #2675, opened 2026-07-17
Avoid errors for empty DSL matches in add-on mode

### LINKED ISSUE #2348: Trying to generate DSLs with add-on for a non-processable constant errors
When saving a file that contains a constant that cannot be processed by any DSL compilers, the add-on errors. The error is rescued and nothing actually happens, but it prints to the output tab and looks like something crashed.

It would be nice to just return early without printing any errors.

```
~/tapioca/lib/tapioca/dsl/pipeline.rb:63:in 'Tapioca::Dsl::Pipeline#run': Tapioca::Error
  from ~/tapioca/lib/tapioca/commands/abstract_dsl.rb:64:in 'Tapioca::Commands::AbstractDsl#generate_dsl_rbi_files'
  from ~/tapioca/lib/tapioca/commands/dsl_generate.rb:17:in 'Tapioca::Commands::DslGenerate#execute'
  from ~/tapioca/lib/tapioca/commands/command.rb:25:in 'block in Tapioca::Commands::Command#run'
```

zeeshankhan-05:
## Summary

- Return early when the add-on finds no processable DSL for a constant
- Avoid logging a `Tapioca::Error` in the editor
- Keep the normal CLI error behavior unchanged

Fixes #2348

## Test plan

- Added an add-on test using an existing `Foo` class with no processable DSL
- Confirmed the test fails without the fix and passes with it
- Confirmed the CLI empty-match tests still pass

### DIFF
--- a/spec/tapioca/addon_spec.rb
+++ b/spec/tapioca/addon_spec.rb
@@ -235,7 +235,7 @@ module RubyLsp
           notification.params&.message.to_s
         end
 
-        message.include?("Tapioca::Error") || message.include?("failed with StandardError")
+        message.include?("Tapioca::Error") || message.match?(/\ARequest .* failed\b/)
       end
     end
   end
--- a/spec/tapioca/addon_spec.rb
+++ b/spec/tapioca/addon_spec.rb
@@ -28,6 +28,43 @@ module RubyLsp
         FileUtils.rm_rf("spec/dummy/sorbet/rbi")
       end
 
+      it "does not error when a constant has no processable DSL" do
+        create_client
+
+        # Foo exists but has no processable DSL. dsl requests are forked, so wait on a
+        # subsequent successful generation and keep draining logs briefly afterward to
+        # avoid missing a concurrent Foo error notification.
+        @client.delegate_notification(
+          server_addon_name: "Tapioca",
+          request_name: "dsl",
+          constants: ["Foo"],
+        )
+        @client.delegate_notification(
+          server_addon_name: "Tapioca",
+          request_name: "dsl",
+          constants: ["NotifyUserJob"],
+        )
+        wait_until_exists("spec/dummy/sorbet/rbi/dsl/notify_user_job.rbi")
+
+        error_logs = []
+        deadline = Process.clock_gettime(Process::CLOCK_MONOTONIC) + 1.0
+        loop do
+          error_logs.concat(drain_log_messages.select do |message|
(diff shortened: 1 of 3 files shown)


spec/tapioca/addon_spec.rb
```diff
@@ -28,6 +28,43 @@ class AddonSpec < Minitest::HooksSpec
...
+
+        error_logs = []
+        deadline = Process.clock_gettime(Process::CLOCK_MONOTONIC) + 1.0
+        loop do
+          error_logs.concat(drain_log_messages.select do |message|
+            message.include?("Tapioca::Error") || message.include?("failed with StandardError")
+          end)
+          break if Process.clock_gettime(Process::CLOCK_MONOTONIC) >= deadline
+
+          sleep(0.1)
```
amomchilov: This is guaranteed to be flakey if the GitHub action runner is overcommited. Is there some other wait we could do?

amomchilov:
Good improvement though!

spec/tapioca/addon_spec.rb
zeeshankhan-05: Ah, you're right. I'll replace the timed poll with a wait on the dsl request's work-done-progress "end" event. So since the error log is always sent before "end", then blocking until "end" makes the check deterministic. I'll enable `window.workDoneProgress` on the test client

zeeshankhan-05: Now waiting on the dsl work-done-progress "end" event instead of a timed poll. Errors are always logged before "end", so the check is deterministic.

spec/tapioca/addon_spec.rb
```diff
@@ -182,6 +204,39 @@ def wait_until_exists(path)
...
+      end
+
+      def error_message?(notification)
+        message = if notification.is_a?(Hash)
+          notification.dig(:params, :message).to_s
+        else
+          notification.params&.message.to_s
+        end
+
+        message.include?("Tapioca::Error") || message.include?("failed with StandardError")
```
KaanOzkan: I think this is a bit restrictive for the future. What about a more general:
`message.include?("Tapioca::Error") || message.match?(/\ARequest .* failed\b/)`

zeeshankhan-05: Yeah, you're right. I just updated to match the broader `Request … failed` shape so we don’t miss non-`StandardError` failures

KaanOzkan (approved):

PR #2687, opened 2026-07-28
Preserve anonymous parameters in generated RBIs

Morriar:
Ruby reflects anonymous splat, keyword splat, and block parameters using pseudo-names:

```ruby
def foo(*, **, &); end

method(:foo).parameters
# => [[:rest, :*], [:keyrest, :**], [:block, :&]]
```

Tapioca was treating those names as invalid parameter names and replacing them with fallback names like `_arg0`, `_arg1`, and `_arg2`. That caused two issues:

* Methods were generated with named parameters instead of preserving the original anonymous syntax.
* RBS-translated signatures could lose parameter types because Sorbet keeps the signature types under the original names: `"*"`, `"**"`, and `"&"`.

For example, this could generate invalid RBI like:

```ruby
sig { params(_arg0: , _arg1: , _arg2: ).void }
def foo(*_arg0, **_arg1, &_arg2); end
```

This change preserves anonymous parameter names for signature lookup, while rendering them as anonymous RBI parameters:

```ruby
sig { params("*": ::Integer, "**": ::String, "&": T.proc.void).void }
def foo(*, **, &); end
```

It also bumps the minimum `rbi` dependency to `0.4.1`, since older versions don’t support anonymous `RestParam`, `KwRestParam`, and `BlockParam` nodes represented with `nil` names.

### DIFF
--- a/lib/tapioca/gem/listeners/methods.rb
+++ b/lib/tapioca/gem/listeners/methods.rb
@@ -126,10 +126,13 @@ module Tapioca
               end
             end
 
-            # Sanitize param names
-            name = fallback_arg_name unless valid_parameter_name?(name)
+            # Sanitize param names, except for anonymous splat, keyword splat,
+            # and block parameters. Ruby reflects those as `:*`, `:**`, and `:&`,
+            # and Sorbet signatures use the same names to store their types.
+            anonymous_parameter = anonymous_parameter_name?(type, name)
+            name = fallback_arg_name unless anonymous_parameter || valid_parameter_name?(name)
 
-            [type, name]
+            [type, name, anonymous_parameter]
           end
 
           rbi_method = RBI::Method.new(
@@ -138,26 +141,27 @@ module Tapioca
             visibility: visibility,
           )
 
-          sanitized_parameters.each do |type, name|
+          sanitized_parameters.each do |type, name, anonymous_parameter|
             case type
             when :req
               rbi_method << RBI::ReqParam.new(name)
             when :opt
               rbi_method << RBI::OptParam.new(name, "T.unsafe(nil)")
             when :rest
-              rbi_method << RBI::RestParam.new(name)
+              rbi_method << RBI::RestParam.new(anonymous_parameter ? nil : name)
             when :keyreq
               rbi_method << RBI::KwParam.new(name)
             when :key
               rbi_method << RBI::KwOptParam.new(name, "T.unsafe(nil)")
             when :keyrest
-              rbi_method << RBI::KwRestParam.new(name)
+              rbi_method << RBI::KwRestParam.new(anonymous_parameter ? nil : name)
             when :block
-              rbi_method << RBI::BlockParam.new(name)
+              rbi_method << RBI::BlockParam.new(anonymous_parameter ? nil : name)
             end
           end
 
-          @pipeline.push_method(symbol_name, constant, method, rbi_method, signature, sanitized_parameters)
+          parameters_for_signature = sanitized_parameters.map { |type, name, _anonymous_parameter| [type, name] }
+          @pipeline.push_method(symbol_name, constant, method, rbi_method, signature, parameters_for_signature)
           tree << rbi_method
         end
 
@@ -251,6 +255,20 @@ module Tapioca
         def same_source_location?(method, other_method)
           source_location = method.source_location
           !!source_location && source_location == other_method.source_location
+        end
+
+        #: (Symbol type, String name) -> bool
+        def anonymous_parameter_name?(type, name)
+          case type
+          when :rest
+            name == "*"
+          when :keyrest
+            name == "**"
+          when :block
+            name == "&"
+          else
+            false
+          end
         end
 
         #: (Module[top] constant, String method_name) -> bool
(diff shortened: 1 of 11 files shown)


lib/tapioca/gem/listeners/methods.rb
```diff
@@ -125,10 +125,13 @@ def compile_method(tree, symbol_name, constant, method, visibility = RBI::Public
...
-            # Sanitize param names
-            name = fallback_arg_name unless valid_parameter_name?(name)
+            # Sanitize param names, except for anonymous splat, keyword splat,
+            # and block parameters. Ruby reflects those as `:*`, `:**`, and `:&`,
+            # and Sorbet signatures use the same names to store their types.
+            anonymous_parameter = anonymous_parameter_name?(type, name)
+            name = fallback_arg_name unless anonymous_parameter || valid_parameter_name?(name)
 
-            [type, name]
+            [type, name, anonymous_parameter]
```
amomchilov:
```suggestion
            is_anonymous_parameter = anonymous_parameter_name?(type, name)
            name = fallback_arg_name unless is_anonymous_parameter || valid_parameter_name?(name)

            [type, name, is_anonymous_parameter]
```

lib/tapioca/gem/listeners/methods.rb
```diff
@@ -137,26 +140,27 @@ def compile_method(tree, symbol_name, constant, method, visibility = RBI::Public
             visibility: visibility,
           )
 
-          sanitized_parameters.each do |type, name|
+          sanitized_parameters.each do |type, name, anonymous_parameter|
```
amomchilov:
```suggestion
          sanitized_parameters.each do |type, name, is_anonymous_parameter|
```

amomchilov: Do we need to thread the 3rd parameter through like this?  
  
Could we instead modify the definition of `name` above, where `anonymous_parameter` is first set?

amomchilov: rc

Morriar: We need to pass something because the method param needs to be generated with a `nil` name but the signature param lookup needs the `"*"` name.

I change the tuple to use both names instead of passing the anonymous flag 👍

lib/tapioca/gem/listeners/methods.rb
```diff
@@ -253,6 +258,20 @@ def same_source_location?(method, other_method)
...
+          when :rest
+            name == "*"
+          when :keyrest
+            name == "**"
+          when :block
+            name == "&"
+          else
+            false
+          end
+        end
```
KaanOzkan: I think we have to fix this for DSL RBIs too in https://github.com/Shopify/tapioca/blob/5e9c26a7311a30a06de6b23040fbee127a10a614/lib/tapioca/dsl/compiler.rb#L156

Morriar: Good point. Fixed in [1a1f4cc](https://github.com/Shopify/tapioca/pull/2687/commits/1a1f4cc5dfdbf5d257cf8ae195e0b5ae12d273f0).

I moved the anonymous parameter check into `RBIHelper` and reused it from the DSL compiler so `create_method_from_def` preserves anonymous `*`, `**`, and `&` parameters too.

KaanOzkan (approved):
LGTM, just requires an expectation update

stadia:
Confirming this in the wild, in case a real-world data point helps with prioritisation.

**Versions:** tapioca 0.19.2, rbi 0.4.3, sorbet-static-and-runtime 0.6.13427, Ruby 3.4, Rails 8.

A gem we consume defines callback macros in a Rails concern using anonymous forwarding:

```ruby
module ClassMethods
  def on_like_received(method_name = nil, **, &)
    register_social_callback(:on_like_received, method_name, **, &)
  end

  def on_delete_requested(*)
    set_callback(:on_delete_requested, *)
  end

  private

  # named params — generates a correct sig
  def register_social_callback(callback_name, method_name = nil, **options, &block)
    ...
  end
end
```

`tapioca gem <name>` emits:

```ruby
sig { params(method_name: T.nilable(::Symbol), _arg1: , _arg2: ).returns(T.untyped) }
```

and `srb tc` then fails on the generated RBI itself:

```
sorbet/rbi/gems/<gem>@0.9.0-....rbi:1364: Malformed type declaration. Unknown type syntax. Expected a ClassName or T.<func>
    1364 |  sig { params(method_name: T.nilable(::Symbol), _arg1: , _arg2: ).returns(T.untyped) }
```

13 errors in that app, all of them inside generated gem RBIs — no user code involved.

Two details that match your diagnosis and might be useful as test cases:

- It is specifically the *anonymous* parameters that lose their type. In the snippet above, `method_name` keeps `T.nilable(::Symbol)` in the same signature, and `register_social_callback` — same file, same module, but with named `**options` / `&block` — generates a perfectly valid sig. So the failure is per-parameter, not per-method.
- It affects both the positional (`*`) and the keyword+block (`**`, `&`) forms, and it applies to private methods too, since those are emitted into the RBI as well.

**Workaround** for anyone hitting this before the fix lands: post-process the generated RBI, replacing the empty type `_arg\d+: ` with `_arg\d+: T.untyped`. Worth noting that `tapioca gem --verify` compares the *set* of RBI files against `Gemfile.lock` rather than their contents, so patched files do not break that CI check.

We briefly considered asking the gem to rename its parameters, but `def m(**, &)` is valid, idiomatic Ruby 3.2+, so patching the consumer's RBI layer is clearly the right side to absorb this on. Would be great to see #2687 land.

PR #2688, opened 2026-07-30
Support per-gem documentation exclusions

st0012:
### Motivation

Some gems, such as Stripe, can generate 10–20 MB RBI files. Omitting their source comments can substantially reduce those files, while disabling documentation globally also removes useful comments from smaller gem RBIs.

### Implementation

- Allow `gem.doc` to accept an `exclude` list.
- Generate RBIs normally for excluded gems while omitting their source comments.
- Preserve the existing Boolean and command-line documentation options.
- Document and validate the new configuration.

### Tests

Added configuration-validation and end-to-end gem-generation coverage.

### DIFF
--- a/lib/tapioca/cli.rb
+++ b/lib/tapioca/cli.rb
@@ -302,6 +302,9 @@ module Tapioca
         raise MalformattedArgumentError, "Option '--verify' must be provided without any other arguments" if verify
       end
 
+      doc = options[:doc]
+      doc_exclude = doc.is_a?(Hash) ? doc.fetch("exclude", []) : []
+
       command_args = {
         gem_names: all ? [] : gems,
         exclude: options[:exclude],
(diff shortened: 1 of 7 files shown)


lib/tapioca/cli.rb
```diff
@@ -311,7 +314,8 @@ def gem(*gems)
         typed_overrides: options[:typed_overrides],
         outpath: Pathname.new(options[:outdir]),
         file_header: options[:file_header],
-        include_doc: options[:doc],
+        include_doc: doc != false,
+        doc_exclude: doc_exclude,
```
KaanOzkan, line 318: I like that this hash, true or false setup reads better in the `config.yml` but I wonder if it's confusing since this defaults to `doc: true` and `exclude: foo`. Wdyt?
```yml
gem:
  doc:
    exclude:
    - foo
```

Compared to having
```yml
gem:
  doc: true
  exclude_doc:
  - foo
```

lib/tapioca/cli.rb
```diff
@@ -302,6 +302,9 @@ def gem(*gems)
         raise MalformattedArgumentError, "Option '--verify' must be provided without any other arguments" if verify
       end
 
+      doc = options[:doc]
+      doc_exclude = doc.is_a?(Hash) ? doc.fetch("exclude", []) : []
```
KaanOzkan, line 306: We have a `--config` flag which complicates things, imagine someone doing `tapioca gem foo --doc --config sorbet/tapioca/config.yml` and the config has exclude. In that case I think we should instead respect the `exclude` flag. We also need a test for this behaviour where they are used together.

PR #2689, opened 2026-07-30
Share reflected method parameter compilation

Morriar:
Centralize reflected parameter name sanitization and `RBI::Param` construction in `RBIHelper` and reuse the helper from DSL and gem method compilation.

Use RBI 0.4.3 to preserve reflected `**nil` parameters as `RBI::NoKwParam`, with regression coverage.

### DIFF
--- a/lib/tapioca/helpers/rbi_helper.rb
+++ b/lib/tapioca/helpers/rbi_helper.rb
@@ -101,14 +101,43 @@ module Tapioca
       type_strings.join(", ").scan(TYPE_PARAMETER_MATCHER).flatten.uniq
     end
 
+    #: (Symbol type, String? name, Integer index) -> [RBI::Param, String]
+    def create_method_parameter(type, name, index)
+      name = "_arg#{index}" unless valid_parameter_name?(name)
+      name = T.must(name)
+
+      parameter = case type
+      when :req
+        RBI::ReqParam.new(name)
+      when :opt
+        RBI::OptParam.new(name, "T.unsafe(nil)")
+      when :rest
+        RBI::RestParam.new(name)
+      when :keyreq
+        RBI::KwParam.new(name)
+      when :key
+        RBI::KwOptParam.new(name, "T.unsafe(nil)")
+      when :keyrest
+        RBI::KwRestParam.new(name)
+      when :block
+        RBI::BlockParam.new(name)
+      when :nokey
+        RBI::NoKwParam.new
+      else
+        Kernel.raise "Unknown type `#{type}`."
+      end
+
+      [parameter, name]
+    end
+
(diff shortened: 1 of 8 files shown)


vinistock (approved):

lib/tapioca/helpers/rbi_helper.rb
```diff
@@ -101,6 +101,34 @@ def extract_type_parameters(type_strings)
...
+      when :keyreq
+        RBI::KwParam.new(name)
+      when :key
+        RBI::KwOptParam.new(name, "T.unsafe(nil)")
+      when :keyrest
+        RBI::KwRestParam.new(name)
+      when :block
+        RBI::BlockParam.new(name)
+      else
+        Kernel.raise "Unknown type `#{type}`."
```
KaanOzkan, line 127: TIL, agent found that `:nokey` is a valid parameter kind for: `def foo(*args, **nil)`. Apparently it's used to reject kwargs. It's not supported  in `rbi` but Sorbet doesn't support it either so not raising might be enough.

paracycle: There will also be a `:noblock` version of it in Ruby 4.1: https://github.com/ruby/ruby/commit/99674185d63162eadcded3d2623a181fddf2bbb3

```ruby
def foo(*args, &nil) = 42

method(:foo).parameters # => [[:rest, :args], [:noblock]]
```

lib/tapioca/helpers/rbi_helper.rb
```diff
@@ -101,6 +101,34 @@ def extract_type_parameters(type_strings)
       type_strings.join(", ").scan(TYPE_PARAMETER_MATCHER).flatten.uniq
     end
 
+    #: (Symbol type, String? name, Integer index) -> [RBI::Param, String]
+    def create_method_parameter(type, name, index)
+      fallback_name = "_arg#{index}"
+      name ||= fallback_name
+      name = fallback_name unless valid_parameter_name?(name)
```
paracycle: It seems we are allocating a `fallback_name` unnecessarily for each parameter here. Can we also refactor `valid_parameter_name?` to accept a `String?` and reject `nil` values, so that we can do:

```suggestion
      name = "_arg#{index}" unless valid_parameter_name?(name)
```

This way the new string allocation will only happen for invalid or missing parameter names

paracycle (approved):
Nice refactor, thank you!

lib/tapioca/helpers/rbi_helper.rb
Morriar: Added support to `rbi`: https://github.com/Shopify/rbi/pull/639

PR #2695, opened 2026-08-04
Handle all annotation source request failures

KaanOzkan:
### Motivation

Ran into this issue in this CI run: https://github.com/Shopify/tapioca/actions/runs/30581510365/job/91002633500?pr=2669#step:6:151
### Implementation

Treat any error while fetching an HTTP annotation source as an unavailable source. I briefly looked into exception classes that can be raised and there were too many, I think the risk of calling `say_http_error` with a non network related exception in this function is low so I went with this approach instead of adding more exception classes.

### DIFF
--- a/lib/tapioca/commands/annotations.rb
+++ b/lib/tapioca/commands/annotations.rb
@@ -187,7 +187,7 @@ module Tapioca
           say_http_error(path, repo_uri, message: response.class)
           nil
         end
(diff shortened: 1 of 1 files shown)


paracycle (approved):
Approach seems reasonable to me

Morriar (approved):

PR #2696, opened 2026-08-08
Fix generic ActiveJob signatures

### LINKED ISSUE #2233: Tapioca::Dsl::Compilers::ActiveJob creates an invalid RBI for generic job classes
Given classes like this

```
class Operation
  abstract!

  sig { abstract.void }
  def save!; end
end

class OperationJob < Active::Job
    extend T::Sig
    extend T::Generic
    abstract!
    OperationType = type_member { { upper: Operations::Model } }

    sig { returns(OperationType) }
    attr_reader :operation

    sig { abstract.returns(T::Class[OperationType]) }
    def operation_type; end

    def perform(**args)
      @operation = operation_type.new(**args)
      @operation.save!
    end
end
```

The compiler produces output like this
```
class. OperationJob
  class << self
    sig do
      params(
        args: T.untyped,
        block: T.nilable(T.proc.params(job: OperationJob).void)
      ).returns(T.any(OperationJob, FalseClass))
    end
    def perform_later(**args, &block); end

    sig { params(args: T.untyped).returns(T.untyped) }
    def perform_now(**args); end
  end
end
```

This fails type checking with the error `Malformed type declaration. Generic class without type arguments`.

Locally we worked around this by monkey patching the compiler to skip generics by calling 
```
def decorate
  return if constant.is_a?(T::Generic)

  # continue with existing code
end
```

Should this change be upstreamed? If so I'm happy to submit a PR

iamhaseebn:
### Motivation

Generic ActiveJob subclasses currently generate bare job class references in `perform_later` signatures. Sorbet requires generic classes to include type arguments, so the generated RBI fails type checking.

Fixes #2233.

### Implementation

- Detect generic ActiveJob subclasses using `T::Generic === constant`.
- Look up their non-fixed type members and use `T.untyped` for each generated type argument.
- Use the parameterized job type for both the callback parameter and the `perform_later` return type.
- Preserve existing signatures for non-generic jobs.

### Tests

- Added regression coverage for jobs with one type member and multiple type members.
- Ran `bundle exec bin/test spec/tapioca/dsl/compilers/active_job_spec.rb`.
- Ran the complete suite: 821 tests, 3,190 assertions, 0 failures, 0 errors, 2 skips.
- Ran `bundle exec bin/typecheck` and `bundle exec bin/style`.
- Ran documentation, README, gem RBI, and shim verification checks.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_job.rb
+++ b/lib/tapioca/dsl/compilers/active_job.rb
@@ -47,7 +47,7 @@ module Tapioca
 
           root.create_path(constant) do |job|
             method = constant.instance_method(:perform)
-            constant_name = name_of(constant)
+            constant_name = type_name_of(constant) #: as !nil
             parameters = compile_method_parameters_to_rbi(method)
             return_type = compile_method_return_type_to_rbi(method)
 
@@ -69,7 +69,27 @@ module Tapioca
 
         private
 
-        #: (Array[RBI::TypedParam] parameters, String? constant_name) -> Array[RBI::TypedParam]
+        # Resolves a constant into a valid Sorbet type reference,
+        # applying `T.untyped` for any unfixed generic type variables.
+        #
+        # @example
+        #   type_name_of(StandardJob) # => "::StandardJob"
+        #   type_name_of(GenericJob)  # => "::SomeModule::GenericJob[T.untyped]"
+        #: (Module[top] constant) -> String?
+        def type_name_of(constant)
+          type_name = qualified_name_of(constant)
+          return type_name if !type_name || type_name.end_with?("]")
+
+          type_variables = Runtime::GenericTypeRegistry.lookup_type_variables(constant)
+          return type_name unless type_variables
+
+          type_variables = type_variables.reject(&:fixed?)
+          return type_name if type_variables.empty?
+
+          "#{type_name}[#{type_variables.map { "T.untyped" }.join(", ")}]"
+        end
+
+        #: (Array[RBI::TypedParam] parameters, String constant_name) -> Array[RBI::TypedParam]
         def perform_later_parameters(parameters, constant_name)
           if ::Gem::Requirement.new(">= 7.0").satisfied_by?(::ActiveJob.gem_version)
             parameters.reject! { |typed_param| RBI::BlockParam === typed_param.param }
--- a/spec/tapioca/dsl/compilers/active_job_spec.rb
+++ b/spec/tapioca/dsl/compilers/active_job_spec.rb
@@ -73,7 +73,7 @@ module Tapioca
 
                 class NotifyJob
                   class << self
-                    sig { params(user_id: T.untyped, block: T.nilable(T.proc.params(job: NotifyJob).void)).returns(T.any(NotifyJob, FalseClass)) }
+                    sig { params(user_id: T.untyped, block: T.nilable(T.proc.params(job: ::NotifyJob).void)).returns(T.any(::NotifyJob, FalseClass)) }
                     def perform_later(user_id, &block); end
 
                     sig { params(user_id: T.untyped).returns(T.untyped) }
@@ -100,7 +100,7 @@ module Tapioca
 
                 class NotifyJob
                   class << self
-                    sig { params(user_id: ::Integer, block: T.nilable(T.proc.params(job: NotifyJob).void)).returns(T.any(NotifyJob, FalseClass)) }
+                    sig { params(user_id: ::Integer, block: T.nilable(T.proc.params(job: ::NotifyJob).void)).returns(T.any(::NotifyJob, FalseClass)) }
                     def perform_later(user_id, &block); end
 
                     sig { params(user_id: ::Integer).void }
@@ -109,6 +109,41 @@ module Tapioca
                 end
               RBI
               assert_equal(expected, rbi_for(:NotifyJob))
+            end
+
+            it "generates correct RBI file for a generic job" do
+              add_ruby_file("job.rb", <<~RUBY)
+                require "active_record"
+
+                class GenericJob < ActiveJob::Base
+                  extend T::Sig
+                  extend T::Generic
+
+                  Input = type_member
+                  Fixed = type_member { { fixed: String } }
+                  Output = type_member { { upper: ActiveRecord::Base } }
+
+                  sig { params(value: Input).returns(Output) }
+                  def perform(value)
+                    raise NotImplementedError
+                  end
+                end
+              RUBY
+
+              expected = template(<<~RBI)
+                # typed: strong
+
+                class GenericJob
+                  class << self
+                    sig { params(value: Input, block: T.nilable(T.proc.params(job: ::GenericJob[T.untyped, T.untyped]).void)).returns(T.any(::GenericJob[T.untyped, T.untyped], FalseClass)) }
+                    def perform_later(value, &block); end
+
+                    sig { params(value: Input).returns(Output) }
+                    def perform_now(value); end
+                  end
+                end
+              RBI
+              assert_equal(expected, rbi_for(:GenericJob))
             end
 
             it "generates correct RBI file for subclass with block argument" do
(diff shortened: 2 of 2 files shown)


iamhaseebn:
I have signed the CLA!

lib/tapioca/dsl/compilers/active_job.rb
```diff
@@ -69,13 +70,26 @@ def decorate
 
         private
 
-        #: (Array[RBI::TypedParam] parameters, String? constant_name) -> Array[RBI::TypedParam]
-        def perform_later_parameters(parameters, constant_name)
+        #: (String constant_name) -> String
+        def generic_job_type(constant_name)
```
amomchilov: Let's document this

```suggestion
        # Resolves the constant name into a valid Sorbet type reference, 
        # applying `T.untyped` for any unfixed generic type variables.
        #
        # @example
        #   generic_job_type("StandardJob") # => "StandardJob"
        #   generic_job_type("GenericJob")  # => "GenericJob[T.untyped]"
        #: (String constant_name) -> String
        def generic_job_type(constant_name)
```

iamhaseebn: Thanks! I've added documentation for `generic_job_type` in 85766fad, including examples for standard and generic jobs. The focused ActiveJob spec, RuboCop, docs generation, and typecheck all pass.

lib/tapioca/dsl/compilers/active_job.rb
```diff
@@ -47,14 +47,15 @@ def decorate
 
           root.create_path(constant) do |job|
             method = constant.instance_method(:perform)
-            constant_name = name_of(constant)
+            constant_name = T.must(name_of(constant))
```
paracycle:
```suggestion
            constant_name = name_of(constant) #: as !nil
```

lib/tapioca/dsl/compilers/active_job.rb
```diff
@@ -69,13 +70,26 @@ def decorate
...
+          return constant_name unless T::Generic === constant
+
+          type_variables = Runtime::GenericTypeRegistry.lookup_type_variables(constant)
+          return constant_name unless type_variables
+
+          type_arguments = type_variables.reject(&:fixed?).map { "T.untyped" }
+          return constant_name if type_arguments.empty?
+
+          "#{constant_name}[#{type_arguments.join(", ")}]"
+        end
```
paracycle: It would be nicer to have:
```suggestion
        #: (Module[top] constant) -> String?
        def generic_name_of(constant)
          type_name = name_of(constant)
          return type_name if !type_name || type_name.end_with?("]")

          type_variables = Runtime::GenericTypeRegistry.lookup_type_variables(constant)
          return type_name unless type_variables

          type_variables = type_variables.reject(&:fixed?)
          return type_name if type_variables.empty?

          "#{type_name}[#{type_variables.map { "T.untyped" }.join(", ")}]"
        end
```

As a matter of fact, I think we should make `name_of` implement this out of the box, so that it works in all contexts with no changes, but not sure what else that might affect.

lib/tapioca/dsl/compilers/active_job.rb
iamhaseebn: Updated in 03e7ba6b to use the RBS non-nil assertion.

lib/tapioca/dsl/compilers/active_job.rb
iamhaseebn: Updated in 03e7ba6b to derive the name in generic_name_of. I kept this scoped to the ActiveJob compiler for now because changing the shared name_of could affect other compiler contexts.

lib/tapioca/dsl/compilers/active_job.rb
amomchilov: Hey @iamhaseebn, that doc got lost in one of your subsequent commits.

Can you add it back, and tidy up the git history? I think most (all?) of this makes sense to squash down to just one commit.

iamhaseebn: Good catch, thanks - I've restored the documentation and squashed this down to one signed commit.

lib/tapioca/dsl/compilers/active_job.rb
```diff
@@ -69,7 +69,27 @@ def decorate
 
         private
 
-        #: (Array[RBI::TypedParam] parameters, String? constant_name) -> Array[RBI::TypedParam]
+        # Resolves a constant name into a valid Sorbet type reference,
```
paracycle:
```suggestion
        # Resolves a constant into a valid Sorbet type reference,
```

lib/tapioca/dsl/compilers/active_job.rb
```diff
@@ -69,7 +69,27 @@ def decorate
...
+        # applying `T.untyped` for any unfixed generic type variables.
+        #
+        # @example
+        #   generic_name_of(StandardJob) # => "StandardJob"
+        #   generic_name_of(GenericJob)  # => "GenericJob[T.untyped]"
+        #: (Module[top] constant) -> String?
+        def generic_name_of(constant)
```
paracycle:
```suggestion
        #   type_name_of(StandardJob) # => "::StandardJob"
        #   type_name_of(GenericJob)  # => "::SomeModule::GenericJob[T.untyped]"
        #: (Module[top] constant) -> String?
        def type_name_of(constant)
```

lib/tapioca/dsl/compilers/active_job.rb
```diff
@@ -69,7 +69,27 @@ def decorate
...
+          type_name = name_of(constant)
```
paracycle:
```suggestion
          type_name = qualified_name_of(constant)
```

spec/tapioca/dsl/compilers/active_job_spec.rb
```diff
@@ -111,6 +111,69 @@ def perform_now(user_id); end
               assert_equal(expected, rbi_for(:NotifyJob))
             end
 
+            it "generates correct RBI file for a generic job" do
```
paracycle, line 114: I don't think this test adds much value given the test below exists. Can we just name the test below `"generates correct RBI file for a generic job"` instead and remove this?

spec/tapioca/dsl/compilers/active_job_spec.rb
```diff
@@ -111,6 +111,69 @@ def perform_now(user_id); end
...
+            end
+
+            it "generates correct RBI file for a job with multiple type members" do
+              add_ruby_file("job.rb", <<~RUBY)
+                class GenericJob < ActiveJob::Base
+                  extend T::Sig
+                  extend T::Generic
+
+                  Input = type_member
+                  Output = type_member
```
paracycle: Testing with multiple kinds of type members would be great:
```suggestion
                  Input = type_member
                  Fixed = type_member { { fixed: String } }
                  Output = type_member { { upper: ActiveRecord::Base } }
```

paracycle (commented):
Sorry, I realize `type_name_of` is a better name for the method, since it returns the proper Sorbet type name, instead of the constant name. Also, it is much safer to make it call `qualified_name_of` so that the names are anchored properly.

I think after this, I am good to merge. Thank you!

lib/tapioca/dsl/compilers/active_job.rb
iamhaseebn: Updated.

lib/tapioca/dsl/compilers/active_job.rb
iamhaseebn: Updated.

lib/tapioca/dsl/compilers/active_job.rb
iamhaseebn: Updated.

spec/tapioca/dsl/compilers/active_job_spec.rb
iamhaseebn: Updated.

spec/tapioca/dsl/compilers/active_job_spec.rb
iamhaseebn: Updated.

paracycle (approved):
Thank you for the contribution and for the multiple round of reviews.

PR #2698, opened 2026-08-17
Bump `sqlite3` to `v2.9.6`

KaanOzkan:
### Motivation

Resolves a minor security alert

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -393,12 +393,12 @@ GEM
       concurrent-ruby (~> 1.1)
       logger
       rack (>= 2.2.4, < 4)
-    sqlite3 (2.9.0-aarch64-linux-gnu)
-    sqlite3 (2.9.0-aarch64-linux-musl)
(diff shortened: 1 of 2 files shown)


sorbet/tapioca/config.yml
```diff
@@ -1,6 +1,7 @@
 ---
 gem:
-  exclude: []
+  exclude:
+    - mini_portile2
```
KaanOzkan: Package proxy has the precompiled `sqlite3` `v2.9.6` gem for ARM macOS but not Linux 🤷 . Bundler therefore uses the precompiled gem locally and the generic source gem in Linux CI, where `mini_portile2` is activated. This causes Tapioca to generate different RBI sets across platforms and fail `gem --verify`.

I excluded `mini_portile2` because it is only a transitive build dependency and we do not use its API.

jesse-shopify (approved):

Morriar (approved):

PR #2701, opened 2026-08-18
Skip ActiveSupport deprecation proxies when gathering DSL constants

### LINKED ISSUE #2463: Rails 8.1 deprecation warnings for LoadInterlockAwareMonitor during DSL generation
## Description

When running `tapioca dsl` with Rails 8.1, five deprecation warnings are emitted:

```
DEPRECATION WARNING: ActiveSupport::Concurrency::LoadInterlockAwareMonitor is deprecated and will be removed in Rails 9.0. Use Monitor directly instead, as the loading interlock is no longer used.
```

## Root Cause

Rails 8.1 deprecated `ActiveSupport::Concurrency::LoadInterlockAwareMonitor` by replacing it with a `DeprecatedConstantProxy`. This proxy emits a deprecation warning whenever the constant is accessed - even for inspection.

During DSL RBI generation, Tapioca's compilers iterate over all loaded modules and constants via `all_modules.select { ... }` in `gather_constants` methods. When these compilers encounter `LoadInterlockAwareMonitor`, simply accessing the constant triggers the deprecation warning.

The warnings originate from these compilers:
- `Tapioca::Dsl::Compilers::ActiveSupportConcern` (lines 73-76)
- `Tapioca::Dsl::Compilers::MixedInClassAttributes` (line 72-73)
- `Tapioca::Dsl::Compilers::UrlHelpers` (lines 119-125)

## Reproduction Steps

1. Create a Rails 8.1 application with Sorbet/Tapioca
2. Run `bin/tapioca dsl`
3. Observe the deprecation warnings in output

## Expected Behavior

`tapioca dsl` should run without emitting deprecation warnings for deprecated constant proxies that are merely being enumerated.

## Environment

- Tapioca version: 0.17.10
- Rails version: 8.1.1
- Ruby version: 3.4.7

## Possible Solutions

1. **Filter out DeprecatedConstantProxy constants** - In the `gather_constants` methods, check if a constant is a `DeprecatedConstantProxy` before accessing it for inspection
2. **Silence deprecations during constant enumeration** - Wrap the constant enumeration in `ActiveSupport.deprecator.silence { ... }`

## Workaround

As a temporary workaround, projects can add a custom deprecation behavior in a Rails initializer:

```ruby
original_behaviors = ActiveSupport.deprecator.behavior.dup
ActiveSupport.deprecator.behavior = -> (message, callstack, deprecator) do
  return if message.to_s.include?("LoadInterlockAwareMonitor")

  original_behaviors.each do |behavior|
    behavior.call(message, callstack, deprecator)
  end
end
```

## Related

- Rails PR that deprecated the class: https://github.com/rails/rails/pull/53452

Hashim1999164:
Fixes Shopify/tapioca#2463

Rails 8.1 replaced LoadInterlockAwareMonitor with a DeprecatedConstantProxy. DSL compilers walk every loaded module and call methods such as singleton_class on each one. Those calls go through method_missing on the proxy and print deprecation warnings even though Tapioca is only enumerating ObjectSpace.

This change drops DeprecatedConstantProxy modules from all_modules so every DSL compiler skips them. Kernel.class is used to recognize the proxy without triggering the warning.

The ActiveSupportConcern gather_constants spec covers the Rails 8.1 LoadInterlockAwareMonitor proxy.

### DIFF
--- a/lib/tapioca/dsl/compiler.rb
+++ b/lib/tapioca/dsl/compiler.rb
@@ -75,7 +75,19 @@ module Tapioca
             @@requested_constants.grep(Module)
           else
             ObjectSpace.each_object(Module).to_a
-          end.freeze #: Enumerable[Module[top]]?
+          end.reject { |mod| deprecated_constant_proxy?(mod) }.freeze #: Enumerable[Module[top]]?
+        end
+
+        # Rails 8.1+ wraps deprecated constants in DeprecatedConstantProxy, which
+        # undefines most instance methods and warns from method_missing. Inspecting
+        # those modules during DSL discovery (is_a?, singleton_class, etc.) emits
+        # deprecation warnings even though Tapioca is only enumerating ObjectSpace.
+        # class_of uses Kernel#class so we can recognize the proxy without warning.
+        #: (Module[top] mod) -> bool
+        def deprecated_constant_proxy?(mod)
+          proxy_class_name = name_of(class_of(mod))
+          proxy_class_name == "ActiveSupport::Deprecation::DeprecatedConstantProxy" ||
+            proxy_class_name == "ActiveSupport::DeprecatedConstantProxy"
         end
       end
 
--- a/lib/tapioca/dsl/compiler.rb
+++ b/lib/tapioca/dsl/compiler.rb
@@ -74,8 +74,8 @@ module Tapioca
           @all_modules ||= if @@requested_constants.any?
             @@requested_constants.grep(Module)
           else
-            ObjectSpace.each_object(Module).to_a
-          end.reject { |mod| deprecated_constant_proxy?(mod) }.freeze #: Enumerable[Module[top]]?
+            ObjectSpace.each_object(Module).reject { |mod| deprecated_constant_proxy?(mod) }.to_a
+          end.freeze #: Enumerable[Module[top]]?
         end
 
         # Rails 8.1+ wraps deprecated constants in DeprecatedConstantProxy, which
--- a/spec/tapioca/dsl/compilers/active_support_concern_spec.rb
+++ b/spec/tapioca/dsl/compilers/active_support_concern_spec.rb
@@ -22,12 +22,14 @@ module Tapioca
               ActiveSupport.deprecator.behavior = ->(message, *) { warnings << message.to_s }
 
               begin
-                constants = gathered_constants
+                proxy = ActiveSupport::Concurrency.const_get(:LoadInterlockAwareMonitor, false)
+                gathered_constants
+                all_modules = Tapioca::Dsl::Compilers::ActiveSupportConcern.send(:all_modules)
               ensure
                 ActiveSupport.deprecator.behavior = previous_behavior
               end
 
-              refute_includes(constants, "ActiveSupport::Concurrency::LoadInterlockAwareMonitor")
+              refute(all_modules.any? { |mod| Tapioca::Runtime::Reflection.are_equal?(mod, proxy) })
               assert_empty(warnings.grep(/LoadInterlockAwareMonitor/))
             end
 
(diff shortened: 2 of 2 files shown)


spec/tapioca/dsl/compilers/active_support_concern_spec.rb
```diff
@@ -14,6 +14,23 @@ def before_setup
...
+              previous_behavior = ActiveSupport.deprecator.behavior
+              ActiveSupport.deprecator.behavior = ->(message, *) { warnings << message.to_s }
+
+              begin
+                constants = gathered_constants
+              ensure
+                ActiveSupport.deprecator.behavior = previous_behavior
+              end
+
+              refute_includes(constants, "ActiveSupport::Concurrency::LoadInterlockAwareMonitor")
```
KaanOzkan: I didn't dive into it but agent suggested this diff instead and claimed that current `refute_includes` would pass even without this change because the name is delegated and becomes `Monitor`, it might be better to do the diff below. Also can you ensure the final test fails without your change?

```
- constants = gathered_constants
+ proxy = ActiveSupport::Concurrency.const_get(:LoadInterlockAwareMonitor, false)
+ gathered_constants
+ all_modules = Tapioca::Dsl::Compilers::ActiveSupportConcern.send(:all_modules)
...
- refute_includes(constants, "ActiveSupport::Concurrency::LoadInterlockAwareMonitor")
+ refute(all_modules.any? { |mod| Tapioca::Runtime::Reflection.are_equal?(mod, proxy) })
  assert_empty(warnings.grep(/LoadInterlockAwareMonitor/))
```

lib/tapioca/dsl/compiler.rb
```diff
@@ -75,7 +75,19 @@ def all_modules
             @@requested_constants.grep(Module)
           else
             ObjectSpace.each_object(Module).to_a
-          end.freeze #: Enumerable[Module[top]]?
+          end.reject { |mod| deprecated_constant_proxy?(mod) }.freeze #: Enumerable[Module[top]]?
```
KaanOzkan: I think it's better to abide by the requested constants even if they result in warnings. Can you move this reject above?

`ObjectSpace.each_object(Module).reject { |mod| deprecated_constant_proxy?(mod) }`

KaanOzkan (commented):
Thank you for the fix. Minor suggestions.

Hashim1999164:
Moved the deprecation proxy skip onto the ObjectSpace path only, so requested constants still make it through. The spec now looks up the proxy by identity instead of the delegated name.

lib/tapioca/dsl/compiler.rb
Hashim1999164: Moved. Requested constants stay in the set even if they warn. ObjectSpace discovery is the only path that skips the proxies.

spec/tapioca/dsl/compilers/active_support_concern_spec.rb
Hashim1999164: Updated. The spec now loads the proxy with inherit false and checks identity against all_modules so the assertion fails if the skip is missing.

spec/tapioca/dsl/compilers/active_support_concern_spec.rb
```diff
@@ -14,6 +14,27 @@ def before_setup
...
+                )
+                gathered_constants
+                all_modules = Tapioca::Dsl::Compilers::ActiveSupportConcern.send(:all_modules)
+              ensure
+                ActiveSupport.deprecator.behavior = previous_behavior
+              end
+
+              refute(all_modules.any? { |mod| Tapioca::Runtime::Reflection.are_equal?(mod, proxy) })
+              assert_empty(warnings.grep(/FakeDeprecatedConstant/))
+            end
```
KaanOzkan, line 38: I updated this test to register a fake constant as deprecated since these things might change in future Rails versions.

Hashim1999164:
Linters failed on RSpec/RemoveConst in the deprecation proxy spec. Added the same rubocop disable the other specs use for that cleanup.

KaanOzkan (approved):

PR #2706, opened 2026-08-25
Support finding signatures through prepended methods

### LINKED ISSUE #2705: Prepending a module can cause `tapioca dsl` to produce an untyped signature
A prepended module can cause Tapioca to not find the signature for a method, leading to it generating an untyped signature. Worse yet, this can cause a difference between how `tapioca dsl` and `tapioca dsl Foo` work:

1. `tapioca dsl` does an eager load, which can load the Ruby file which prepends the module onto `Foo`. The RBI will have an untyped signature.
2. `tapioca dsl Foo` doesn't eager load, so it might not load the file which has the prepended module. The RBI will have a typed signature.

# Cause

This stems from the behaviour of T::Utils.signature_for_method`, which looks up a signature for a given owner_id/method_name pair. If a prepended module doesn't have a sig, it'll return `nil` (even if the method it overrides _does_ have a sig). 

```ruby
require "sorbet-runtime"

class Foo
  extend T::Sig
  
  sig { returns(String) }
  def foo = "original"
end

p T::Utils.signature_for_instance_method(Foo, :foo) # => the real sig

module M
  def foo = "prepended"
end

Foo.prepend(M)

p T::Utils.signature_for_instance_method(Foo, :foo) # => nil
```

# Potential solutions

## Keep searching until we find a sig

```ruby
#: (UnboundMethod) -> [UnboundMethod, T::Private::Methods::Signature]
def first_signature(method)
  while method
    signature = T::Utils.signature_for_method(method)
    return [method, signature] if signature
    
    method = method.super_method
  end
  
  nil
end

method, sig = first_signature(Foo.instance_method(:foo))
p sig
```

This might be desirable in this case, but cause weirdness in other cases (e.g. when an method override doesn't have a sig, but its method parameter declaration differs from that of its parent method). Needs investigation.

KaanOzkan:
### Motivation

Resolves https://github.com/Shopify/tapioca/issues/2705
### Implementation

Find signatures hidden by `prepend` by walking `super_method` through prepended ancestor positions and stopping at the first ordinary implementation.

This requires a few things:

1. Require call sites to pass the constant or receiver from which the method was resolved (`lookup_from`) because a wrapper-owned method does not retain that scope.
2. Infer prepended positions by comparing the lookup scope’s flattened ancestor chain with each ancestor’s own chain because Ruby does not expose this information.
3. Repeat the lookup once more because lazy evaluation of a `sig` may register it under the wrapper that was already inspected.

For example in this case `lookup_from` lets us identify `Wrapper` as prepended, so we inspect `Child#foo`, and stop before incorrectly falling through to `Parent#foo`.

```rb
class Parent
  #: (Integer) -> Integer
  def foo(value) = value
end

module Wrapper
  def foo(...) = super
end

class Child < Parent
  prepend Wrapper

  #: (String) -> String
  def foo(value) = value
end

method = Child.instance_method(:foo)

method.owner                         # => Wrapper
method.super_method.owner            # => Child
method.super_method.super_method.owner # => Parent

signature_of(method, lookup_from: Child) # => (String) -> String
```

### Tests

TBD on Core

### DIFF
--- a/lib/tapioca/runtime/reflection.rb
+++ b/lib/tapioca/runtime/reflection.rb
@@ -19,7 +19,9 @@ module Tapioca
       PUBLIC_INSTANCE_METHODS_METHOD = Module.instance_method(:public_instance_methods) #: UnboundMethod
       PROTECTED_INSTANCE_METHODS_METHOD = Module.instance_method(:protected_instance_methods) #: UnboundMethod
       PRIVATE_INSTANCE_METHODS_METHOD = Module.instance_method(:private_instance_methods) #: UnboundMethod
+      INSTANCE_METHOD_METHOD = Module.instance_method(:instance_method) #: UnboundMethod
       METHOD_METHOD = Kernel.instance_method(:method) #: UnboundMethod
+      METHOD_RECEIVER_METHOD = Method.instance_method(:receiver) #: UnboundMethod
       UNDEFINED_CONSTANT = Module.new.freeze #: Module[top]
 
       REQUIRED_FROM_LABELS = ["<top (required)>", "<main>", "<compiled>"].freeze #: Array[String]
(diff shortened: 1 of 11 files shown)


lib/tapioca/runtime/reflection.rb
```diff
@@ -125,7 +125,32 @@ def qualified_name_of(constant)
...
+        #   end
+        #   class Example
+        #     sig { void }
+        #     def foo; end
+        #     prepend Wrapper
+        #   end
+        # Looking up the original `Example#foo` evaluates its `sig`. Because `Wrapper`
+        # was prepended, Ruby now resolves `Example#foo` to `Wrapper#foo`, so Sorbet
+        # stores the signature for `Wrapper#foo`. The first pass causes the store; the
+        # second pass finds `Example#foo`'s signature on `Wrapper#foo`.
```
amomchilov: This comment is too hard-coded to the prepend case, but this implementation doesn't do anything to limit to prepended modules (and it's surprisingly difficult to identify them)

lib/tapioca/runtime/reflection.rb
```diff
@@ -125,7 +125,32 @@ def qualified_name_of(constant)
...
+          current_method = method #: (UnboundMethod | Method)?
+
+          while current_method
+            signature = T::Utils.signature_for_method(current_method)
+            return signature if signature
+
+            current_method = current_method.super_method
+          end
+
+          return nil if pass.zero? && !T::Utils.signature_for_method(method)
```
amomchilov: hmm?

lib/tapioca/runtime/reflection.rb
KaanOzkan: Added a precursor to the comment, wdyt?

lib/tapioca/runtime/reflection.rb
KaanOzkan: I wanted to make it faster in the normal case and return early, new version should be easier to understand

lib/tapioca/runtime/reflection.rb
```diff
@@ -125,7 +125,36 @@ def qualified_name_of(constant)
...
+        needs_second_pass = false #: bool
+        2.times do
+          current_method = method #: (UnboundMethod | Method)?
+
+          while current_method
+            needs_second_pass ||= T::Private::Methods.has_sig_block_for_method(current_method)
+            signature = T::Utils.signature_for_method(current_method)
+            return signature if signature
+
+            current_method = current_method.super_method
```
Morriar: I wonder if we should avoid walking past the method owner unless we're still resolving prepended methods.

As written, this also changes the regular override case where the child method has no sig but the parent method does:

```ruby
class Parent
  extend T::Sig

  sig { params(value: String).returns(String) }
  def foo(value) = value
end

class Child < Parent
  def foo(value, extra = nil) = value
end
```

`signature_of(Child.instance_method(:foo))` now returns `Parent#foo`'s signature, so Tapioca can generate RBI using the parent's arity/types for an untyped override. Let's keep the super lookup constrained to the prepend-wrapper path, or at least add a regression test that ordinary unsigned overrides still return `nil`.

lib/tapioca/runtime/reflection.rb
jesse-shopify: I also think this comment could focus more on the purpose of the method rather than describing the implementation. Not a blocker.

jesse-shopify (dismissed):

KaanOzkan:
This is more difficult than I originally thought because walking the entire `super_method` chain to find a signature also affects ordinary overrides:
```rb
class Parent
  #: (String) -> String
  def foo(value) = value
end

class Child < Parent
  def foo(value, suffix) = ""
end

foo = Child.instance_method(:foo)

# Before: signature_of(foo) # => nil
# Current PR: signature_of(foo) # => Parent#foo's signature (incorrect)
```

I’m looking for a clean way to support prepended methods without changing this behavior.

PR #2707, opened 2026-08-28
Include ActiveRecord::TestFixtures in the fixtures RBI

mokevnin:
### Motivation

Rails mixes `ActiveRecord::TestFixtures` into `ActiveSupport::TestCase` from the [`:active_support_test_case` load hook in `rails/test_help.rb`](https://github.com/rails/rails/blob/main/railties/lib/rails/test_help.rb). That file is required from an app's `test_helper.rb`, so the hook never runs during RBI generation, the include never makes it into the gem RBI, and Sorbet doesn't see the class methods the module contributes through `mixes_in_class_methods` — most visibly `fixtures`:

```ruby
class PostTest < ActiveSupport::TestCase
  fixtures :all # Method `fixtures` does not exist on `T.class_of(PostTest)` (7003)
end
```

Today every Rails app using fixtures shims this by hand (a `sorbet/rbi/shims` entry, or a local DSL compiler that emits nothing but the include). `ActiveRecordFixtures` already targets `ActiveSupport::TestCase` and already knows Rails is loaded, so it's the natural place to make the include explicit.

### Implementation

One line in `decorate` — `mod.create_include("ActiveRecord::TestFixtures")` — inside the existing `create_path`.

The early return now sits above the `select!` that drops fixture sets whose model constant can't be resolved, so the two cases it used to conflate are separated: an app with no fixture files at all still generates no RBI, while an app that has fixture files but no matching models gets an RBI with just the include (it writes `fixtures :all` too).

### Tests

`spec/tapioca/dsl/compilers/active_record_fixtures_spec.rb` updated for the extra line, plus a new case for fixtures with no associated model at all; the whole file passes. `bin/typecheck` and `rubocop` are clean, and `manual/` is regenerated via `bin/docs`.

### DIFF
--- a/lib/tapioca/dsl/compilers/active_record_fixtures.rb
+++ b/lib/tapioca/dsl/compilers/active_record_fixtures.rb
@@ -37,10 +37,10 @@ module Tapioca
       # ~~~
       #
       # The `include` is generated because Rails mixes `ActiveRecord::TestFixtures` into
-      # `ActiveSupport::TestCase` from a `:active_support_test_case` load hook in `rails/test_help.rb`,
-      # which is only required from an app's `test_helper.rb` and therefore never runs during RBI
-      # generation. Without it, Sorbet does not see the class methods the module contributes via
-      # `mixes_in_class_methods`, such as `fixtures`.
+      # `ActiveSupport::TestCase` through the `:active_support_test_case` load hook in
+      # `rails/test_help.rb`. Since RBI generation does not load an app's test helper, this runtime
+      # include is not captured. Without it, Sorbet does not see the class methods the module
+      # contributes via `mixes_in_class_methods`, such as `fixtures`.
       #: [ConstantType = singleton(ActiveSupport::TestCase)]
       class ActiveRecordFixtures < Compiler
         MISSING = Object.new
@@ -53,8 +53,6 @@ module Tapioca
           else
             method_names_from_eager_fixture_loader
           end
-
-          return if method_names.empty?
 
           method_names.select! { |name| fixture_class_mapping_from_fixture_files[name] != MISSING }
 
--- a/lib/tapioca/dsl/compilers/active_record_fixtures.rb
+++ b/lib/tapioca/dsl/compilers/active_record_fixtures.rb
@@ -37,9 +37,10 @@ module Tapioca
       # ~~~
       #
       # The `include` is generated because Rails mixes `ActiveRecord::TestFixtures` into
-      # `ActiveSupport::TestCase` from the `active_record.test_fixtures` railtie initializer, which
-      # does not run during gem RBI generation. Without it, Sorbet does not see the class methods
-      # the module contributes via `mixes_in_class_methods`, such as `fixtures`.
+      # `ActiveSupport::TestCase` from a `:active_support_test_case` load hook in `rails/test_help.rb`,
+      # which is only required from an app's `test_helper.rb` and therefore never runs during RBI
+      # generation. Without it, Sorbet does not see the class methods the module contributes via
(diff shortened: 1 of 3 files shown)


lib/tapioca/dsl/compilers/active_record_fixtures.rb
```diff
@@ -50,6 +57,8 @@ def decorate
           return if method_names.empty?
 
           root.create_path(constant) do |mod|
+            mod.create_include("ActiveRecord::TestFixtures")
```
Morriar: I think we need to avoid gating this include on the filtered fixture method list. The include is what makes the `fixtures` class method visible, and that method is still valid when we can't generate typed fixture accessors. For example, an app with only model-less/raw fixture files will get names from `fixtures(:all)`, then we filter them out as `MISSING` and return before emitting this include. That leaves the original `fixtures :all` error unfixed for that shape of app.

Could we keep the no-fixture-files behavior by returning before the `MISSING` filter, then always creating this path for any fixture names?

lib/tapioca/dsl/compilers/active_record_fixtures.rb
```diff
@@ -26,13 +26,20 @@ module Compilers
...
       #   sig { returns(T::Array[Post]) }                                                          #   No names: returns an Array of all fixtures
       #   sig { params(fixture_name: T.any(String, Symbol)).returns(Post) }                        #   One name: returns the requested fixture
       #   sig { params(fixture_name: T.any(String, Symbol), other_fixtures: T.any(String, Symbol)) # Many names: returns an Array of the requested fixtures
       #   .returns(T::Array[Post]) }
       #   def posts(fixture_name = nil, *other_fixtures); end
       # end
       # ~~~
+      #
+      # The `include` is generated because Rails mixes `ActiveRecord::TestFixtures` into
+      # `ActiveSupport::TestCase` from the `active_record.test_fixtures` railtie initializer, which
```
KaanOzkan: I couldn't find that initializer, I think it's being included through this load hook https://github.com/rails/rails/blob/4130768a1b0d95da640ac792920e54988cf2d12f/railties/lib/rails/test_help.rb#L15-L24. Can we change the wording here?

lib/tapioca/dsl/compilers/active_record_fixtures.rb
KaanOzkan, line 60: Is it okay for this `include` to be skipped by the early return above?

KaanOzkan (commented):
Thanks, this is a unique change but I think it's beneficial to have it in the compiler as you said.

lib/tapioca/dsl/compilers/active_record_fixtures.rb
mokevnin: You're right, there is no such initializer — thanks for digging. It's the `:active_support_test_case` load hook in `rails/test_help.rb`, which an app requires from `test_helper.rb`, so it never runs during RBI generation. Reworded here, in `manual/` and in the PR description.

lib/tapioca/dsl/compilers/active_record_fixtures.rb
mokevnin: No, it isn't — good catch. The `select!` above the early return drops fixture sets whose model constant can't be resolved, so an app that has fixture files but no matching models writes `fixtures :all` and still got no include.

Moved the early return above the `select!`, which separates the two cases it was conflating: no fixture files at all keeps generating nothing, everything else gets the include. Added a spec for fixtures with no associated model, which now yields an RBI with just the include.

KaanOzkan (approved):

PR #2709, opened 2026-09-01
Evaluate lazy signatures hidden by prepended methods

### LINKED ISSUE #2710: Filtered DSL generation misses lazy signatures hidden by `prepend`
Related to #2705, but caused by the reverse ordering. Here the wrapper is already prepended while the original method's `sig` remains unevaluated, causing filtered generation to miss a signature that full generation finds.

`tapioca dsl Foo` can generate an untyped RBI while `tapioca dsl` generates a typed RBI when an already-loaded prepended method hides an unevaluated `sig`.

For example:

```ruby
require "sorbet-runtime"

module Wrapper
  def call(*args)
    super
  end
end

class Foo
  extend T::Sig

  sig { params(value: String).returns(String) }
  def call(value)
    value
  end

  prepend Wrapper
end

method = Foo.instance_method(:call)

method.owner
# => Wrapper

T::Utils.signature_for_method(method)
# => nil

T::Utils.run_all_sig_blocks

T::Utils.signature_for_method(method)
# => signature with (String) -> String
```

Filtered DSL generation does not call `T::Utils.run_all_sig_blocks`, so a DSL compiler inspecting `Foo.instance_method(:call)` sees the wrapper's parameters and no signature:

```ruby
sig { params(args: T.untyped).returns(T.untyped) }
def call(*args); end
```

Unfiltered generation evaluates all pending signatures first, allowing Sorbet Runtime to associate the signature with the currently resolved wrapper:

```ruby
sig { params(value: String).returns(String) }
def call(value); end
```

Both commands should generate the typed definition.

KaanOzkan:
### Motivation
Fixes https://github.com/Shopify/tapioca/issues/2710
 
When `prepend` places a wrapper before a method whose `sig` remains
pending, looking up the wrapper returns no signature. 

```rb
module Wrapper
  def call(...) = super
end

class Foo
  extend T::Sig

  sig { params(value: String).returns(String) }
  def call(value) = value

  prepend Wrapper
end

method = Foo.instance_method(:call)

# Before: signature_of(method) # => nil
# After:  signature_of(method) # => Foo#call signature
```
### Implementation

Query the underlying method to evaluate the pending `sig` using `T::Utils.signature_forMethod`, then query the wrapper again after `sorbet-runtime` associates the signature with it.
### Tests

On top of automated tests, tested on Core

### DIFF
--- a/lib/tapioca/runtime/reflection.rb
+++ b/lib/tapioca/runtime/reflection.rb
@@ -125,7 +125,24 @@ module Tapioca
 
       #: ((UnboundMethod | Method) method) -> untyped
       def signature_of!(method)
-        T::Utils.signature_for_method(method)
+        signature = T::Utils.signature_for_method(method)
+        return signature if signature
+
+        # Walk the super method chain to evaluate pending signatures that may have been hidden by `prepend`.
+        # During that walk, Sorbet may register a signature on the requested wrapper, so check it again.
+        # See https://github.com/Shopify/tapioca/issues/2710 for details.
+        current_method = method.super_method #: (UnboundMethod | Method)?
+        while current_method
+          # force evaluation of the signature
+          _ = T::Utils.signature_for_method(current_method)
+
+          evaluated_signature = T::Utils.signature_for_method(method)
+          return evaluated_signature if evaluated_signature
+
+          current_method = current_method.super_method
+        end
+
+        nil
       rescue LoadError, StandardError
         Kernel.raise SignatureBlockError
       end
--- a/spec/tapioca/cli/dsl_spec.rb
+++ b/spec/tapioca/cli/dsl_spec.rb
@@ -801,6 +801,61 @@ module Tapioca
           assert_success_status(result)
         end
 
+        it "finds lazy signatures hidden by prepended methods" do
+          @project.write!("lib/wrapped_method.rb", <<~RB)
+            # typed: strict
+
+            module MethodWrapper
+              def call(*args)
+                super
+              end
+            end
+
+            class WrappedMethod
+              #: (String) -> String
+              def call(value)
+                value
+              end
+
+              prepend MethodWrapper
+            end
+          RB
+
+          @project.write!("lib/compilers/wrapped_method_compiler.rb", <<~RB)
+            require "wrapped_method"
+
+            class WrappedMethodCompiler < Tapioca::Dsl::Compiler
+              extend T::Sig
+              extend T::Generic
+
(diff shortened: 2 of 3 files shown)


lib/tapioca/runtime/reflection.rb
```diff
@@ -125,7 +125,23 @@ def qualified_name_of(constant)
...
-        T::Utils.signature_for_method(method)
+        signature = T::Utils.signature_for_method(method)
+        return signature if signature
+
+        # Walk the super method chain to evaluate pending signatures that may have been hidden by `prepend`.
+        # During that walk, Sorbet may register a signature on the requested wrapper, so check it again.
+        # See https://github.com/Shopify/tapioca/issues/2710 for details.
+        current_method = method.super_method #: (UnboundMethod | Method)?
+        while current_method
+          T::Utils.signature_for_method(current_method)
```
Morriar: Could we avoid letting ordinary ancestor sig failures affect an untyped override here?

The check below prevents us from returning a parent signature, but this call still executes the parent pending `sig`. That means `signature_of!` can now raise/report an unrelated parent sig error for a method that has no signature of its own:

```rb
class Parent
  extend T::Sig
  sig { params(value: MissingConstant).returns(String) }
  def call(value); value; end
end

class Child < Parent
  def call(value); value; end
end

Runtime::Reflection.signature_of!(Child.instance_method(:call)) # now raises
```

This used to return `nil` because `Child#call` is untyped. Let's either narrow the fallback to the prepended-wrapper case or ignore ancestor sig failures unless re-checking `method` actually produced a signature.

vinistock:
Out of curiosity, what happens if the prepended method doesn't match the arity of the super method?

```ruby
module Wrapper
  def foo(a, b, c)
    super(a + b, c)
  end
end

class Bar
  prepend Wrapper

  #: (Integer, Integer) -> Integer
  def foo(a, b)
    a + b
  end
end
```

KaanOzkan:
@vinistock in your exact repro sorbet-runtime raises `You're trying to replace 'foo' on 'Bar', but that method exists in a prepended module (Wrapper), which we don't currently support.` but if I move the `prepend` to the end we generate the RBI with 2 parameters.

vinistock (approved):

amomchilov (approved):

lib/tapioca/runtime/reflection.rb
amomchilov: Maybe to be more explicit:

```suggestion
          # force evaluation of the signature
          _ = T::Utils.signature_for_method(current_method)
```

PR #2711, opened 2026-09-01
Add gem extensions

vinistock:
### Motivation

Similar to DSL extensions, it is useful to be able to patch certain behaviour to ensure accurate gem RBI generation. Gems can expose extensions that get required ahead of time, ensuring that they patch with Tapioca compatibility in mind.

The README and test included provide an example we hit recently that can be easily fixed with an extension.

### Implementation

Very similar to DSL extensions. We use `Gem.find_files` to detect gem extensions contributed by gems and load them ahead of time, before generation starts.

### Tests

Added a test.


lib/tapioca/loaders/gem.rb
```diff
@@ -49,6 +50,19 @@ def initialize(bundle:, prerequire:, postrequire:, default_command:, halt_upon_l
...
       end
 
+      #: -> void
+      def load_gem_extensions
+        say("Loading gem extension classes... ")
+
+        # Extensions are loaded before the bundle is required so that they can patch the gems
+        # they apply to as those gems are being loaded.
+        ::Gem.find_files("tapioca/gem/extensions/*.rb").each do |extension|
+          require File.expand_path(extension)
```
Morriar, line 59: I wonder if we should also allow project-local files to be loaded as extensions?

README.md
```diff
@@ -345,6 +346,49 @@ This option can be used in CI to make sure the RBI files are *up-to-date* and en
 
 Tapioca will import any signatures found in the `rbi/` folder of a given gem and combine them with the RBIs it generates. This is useful when a gem doesn't want to depend on `sorbet-runtime` but still wants to provide type safety to users during static checks. Note that the `rbi/` folder needs to be included in the gem release using the `.gemspec` file. Applications can choose not to import these signatures using the `--no-exported-gem-rbis` flag.
 
+#### Writing custom gem extensions
+
+Similar to [DSL extensions](#writing-custom-dsl-extensions), it is sometimes useful to patch and modify certain behavior from gems to ensure that RBIs are generated correctly. One example is generic types that implement their own `[]` method.
+
+```ruby
+module Patch
```
Morriar: Can we use a less niche example?

lib/tapioca/loaders/gem.rb
Morriar: Oh wait that's what `prerequire` is for right? I wonder if we should merge both concepts 🤔

lib/tapioca/loaders/gem.rb
```diff
@@ -27,6 +27,7 @@ def load_application(bundle:, prerequire:, postrequire:, default_command:, halt_
       # @override
       #: -> void
       def load
+        load_gem_extensions
```
Morriar: I wonder if this should happen _after_ we load the `prerequire`?

I think we need to make this work:

```ruby
@project.write!("config/prerequire.rb", <<~RUBY)
  PREREQUIRE_LOADED = true
RUBY

foo = mock_gem("foo", "0.0.1") do
  write!("lib/foo.rb", "module Foo; end")

  write!("lib/tapioca/gem/extensions/foo.rb", <<~RUBY)
    raise "prerequire was not loaded" unless defined?(PREREQUIRE_LOADED)
    require "foo"
  RUBY
end

@project.require_mock_gem(foo)
@project.bundle_install!

result = @project.tapioca(
  "gem foo --prerequire config/prerequire.rb",
)

assert_empty_stderr(result)
assert_success_status(result)
```

It seems that would be the same order than DSL extensions: project local first, then gem provided extensions:  https://github.com/Shopify/tapioca/blob/c64298ca683c0593072b2602a51d52f596326644/lib/tapioca/loaders/dsl.rb#L76-L82

README.md
vinistock: I changed the example. Let me know what you think.

lib/tapioca/loaders/gem.rb
vinistock: Fixing this, I ended up noticing that we have `Gem < Loader` and `DSL < Loader`, but some of the `Gem` specific behaviour was in the parent class (load_bundle).

I moved that method into `Gem`, so we can control the order and load extensions after prerequires.

lib/tapioca/loaders/gem.rb
vinistock: I agree that it ends up being the same concept, but with different names. I'm not sure how we should approach it though.

Simply removing `prerequire` and turning it into gem extensions would be a breaking change. We could also support both for some time or accept the distinct names for a bit until we're ready to make a breaking change.

WDYT?

lib/tapioca/loaders/gem.rb
```diff
@@ -49,6 +49,33 @@ def initialize(bundle:, prerequire:, postrequire:, default_command:, halt_upon_l
         @halt_upon_load_error = halt_upon_load_error
       end
 
+      #: -> void
+      def load_gem_extensions
+        say("Loading gem extension classes... ")
+
+        # Extensions are loaded before the bundle is required so that they can patch the gems
+        # they apply to as those gems are being loaded.
+        ::Gem.find_files("tapioca/gem/extensions/*.rb").each do |extension|
```
Morriar, line 62: Should we skip extensions from excluded gems here? Right now an excluded gem can still run code during `tapioca gem` if it ships `lib/tapioca/gem/extensions/*.rb`, because we scan and require the extension before `require_bundle` has a chance to honor the exclude. That weakens `--exclude` as the side-effect escape hatch.

Maybe we can resolve the extension path back to its `Gemfile::GemSpec` and skip it when the owning gem is excluded.

spec/tapioca/cli/gem_spec.rb
```diff
@@ -2298,6 +2298,82 @@ class Application < Rails::Application
...
+
+      describe "custom extensions" do
+        after do
+          project.write_gemfile!(project.tapioca_gemfile)
+          @project.require_default_gems
+          @project.remove!("sorbet/rbi")
+          @project.remove!("../gems")
+        end
+
+        it "loads extensions" do
```
Morriar: Let's add a regression for the load order too. This proves extensions are loaded eventually, but it would still pass if they were loaded after `gemfile.require_bundle`.

The important contract from the loader/docs is `prerequire` -> gem extensions -> bundle, especially because `prerequire` is the escape hatch for setting up extension dependencies.

KaanOzkan (approved):

README.md
Morriar: I like it 👍

lib/tapioca/loaders/gem.rb
Morriar: I think we should make the gem extensions behave the same way than DSL ones and allow both project local and gem provided extensions.

Once we have this we can start deprecating `prerequire` saying to switch to the new gem extension feature.

lib/tapioca/loaders/gem.rb
vinistock: Yes, we should indeed.

lib/tapioca/loaders/gem.rb
vinistock: Agreed. The latest commit has the changes.

PR #2714, opened 2026-09-10
Create parameter type list based on position instead of name

vinistock:
### Motivation

I was testing out the latest Tapioca main and noticed that we are producing invalid syntax in RBIs for anonymous keyword rest or block parameters (see tests).

We are currently outputting this:

```ruby
# Note the missing type after the parameter name, which is a syntax error
sig { arg_1: }.void
def foo(**); end
```

### Implementation

The essence of the issue is that we create a hash of parameter types, which we lookup based on names. This only works if the parameter actually has a name.

The idea of the implementation is to rely on the fact that Sorbet errors if parameters are out of order and process them based on their position instead. This allows us to at least produce syntactically valid RBIs under the scenarios.

### Tests

Added tests.

### DIFF
--- a/spec/tapioca/gem/pipeline_spec.rb
+++ b/spec/tapioca/gem/pipeline_spec.rb
@@ -5012,5 +5012,90 @@ class Tapioca::Gem::PipelineSpec < Minitest::HooksSpec
 
       assert_equal(output, compile(include_doc: true))
     end
+
+    it "compiles signatures for methods with anonymous block parameters" do
+      add_ruby_file("foo.rb", <<~RUBY)
+        # typed: true
+
+        class Foo
+          #: (String path) { -> void } -> void
+          def anon_block(path, &); end
+        end
+      RUBY
+
+      output = template(<<~RBI)
+        class Foo
+          sig { params(path: ::String, _arg1: T.proc.void).void }
+          def anon_block(path, &_arg1); end
+        end
+      RBI
+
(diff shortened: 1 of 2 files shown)


spec/tapioca/gem/pipeline_spec.rb
```diff
@@ -5012,5 +5012,90 @@ def multi_param_method(zebra, alpha, beta); end
...
+      output = template(<<~RBI)
+        class Foo
+          sig { params(path: T.untyped, _arg1: T.untyped).returns(T.untyped) }
+          def bar(path, **_arg1); end
+        end
+      RBI
+
+      assert_equal(output, compile)
+    ensure
+      T::Configuration.instance_variable_set(:@sig_validation_error_handler, nil)
```
Morriar: Can we restore the previous handler instead of always resetting this global state to `nil`? If the test setup, or another example, installs a handler before this runs, this example will silently clear it.

Something like:

```ruby
previous_handler = T::Configuration.instance_variable_get(:@sig_validation_error_handler)
T::Configuration.sig_validation_error_handler = ->(*) {}
# ...
ensure
  T::Configuration.instance_variable_set(:@sig_validation_error_handler, previous_handler)
end
```

KaanOzkan (approved):

vinistock:
Actually, we should push #2687 forward instead.

PR #2718, opened 2026-09-18
Fix method signatures being overwritten by prepended modules

### LINKED ISSUE #2072: Tapioca does not properly attribute prepended methods
I haven't verified this yet, but @paracycle told me that if one gem defines a method, and another gem overrides that method via a prepend, the prepended method will be incorrectly written to both gem RBIs.

Example:

```
# in gem foo
class Foo
  def foo; end
end
```

```
# in gem bar
module Bar
  def foo; end
end

Foo.prepend(Bar)
```

Tapioca will write `bar`'s `foo` method to `foo`'s gem RBIs, which is fine if they have the same signature, but not great if the prepend changes the method signature.

DaneHarrison:
### Motivation
Fixes #2072 - When a module is prepended onto a class, Tapioca generates the RBI using the prepended module's signature instead of its own.

<br/>

### Implementation

`compile_method` receives methods obtained with `Module#instance_method`, which follows the ancestor chain. When a
module is prepended to a class and defines the method, `instance_method` returns the prepended module's `UnboundMethod` rather than the class's own, so `compile_method` was compiling the wrong method.

- `method_owned_by_constant?` becomes `method_defined_by_constant`. It walks `super_method` and returns the method the
  constant itself defines.

- New `signature_defined_by_constant` returns the signature declared on that method. A signature is only accepted
  if `signature.method.owner` matches the method's owner.

- The attr_accessor writer inference resolves the constant's own reader first, so it works in either evaluation order.

<br/>

### Further Considerations:
- If both the class and the prepended module have a sig for the same name, Sorbet keeps only one per key. Whichever block runs last wins - see [wrap_method_if_needed](https://github.com/sorbet/sorbet/blob/0eac3406feb048ffc0014132bd1990ae3fcae69e/gems/sorbet-runtime/lib/types/private/methods/call_validation.rb#L19) and [unwrap_method](https://github.com/sorbet/sorbet/blob/0eac3406feb048ffc0014132bd1990ae3fcae69e/gems/sorbet-runtime/lib/types/private/methods/_methods.rb#L613)
- There are other callers to `signature_of` i.e under `dsl`, I suspect those will need to be updated as well
  - Let me know if you'd like me to do a follow up PR on this

<br/>

### Tests
- No sig: a prepended module in front of the method, and multiple prepended modules
- Cross-gem prepend matching the issue, checking both gems' RBIs
- Sig with a prepended module that has different parameters
- Only the prepended module has a sig (the class's method must not take it)
- Sig evaluated before the prepend, and between two prepends
- attr_accessor writer inference with the reader's sig evaluated before and after the prepend
- A prepended module's sig that fails to load is reported once, against the module
- Skipped: both the class and the prepended module have a sig (Sorbet keeps only one)


DaneHarrison:
I have signed the CLA!

lib/tapioca/gem/listeners/methods.rb
```diff
@@ -218,9 +221,8 @@ def attr_reader_for_writer(method, constant)
           reader_method = T.let(constant.instance_method(method_name.delete_suffix("=").to_sym), UnboundMethod)
           reader_method = original_method(reader_method)
           return unless same_source_location?(method, reader_method)
-          return unless method_owned_by_constant?(reader_method, constant)
 
-          reader_method
+          method_defined_by_constant(reader_method, constant)
```
Morriar: Let's move the ownership resolution before the source-location check. With a prepended `attr_accessor`, `constant.instance_method(:foo)` still returns the prepended reader, so this compares the class writer against the prepended reader and skips inferred writer signatures even though the class reader is available through `super_method`.

```suggestion
          reader_method = T.let(constant.instance_method(method_name.delete_suffix("=").to_sym), UnboundMethod)
          reader_method = original_method(reader_method)
          reader_method = method_defined_by_constant(reader_method, constant)
          return unless reader_method
          return unless same_source_location?(method, reader_method)

          reader_method
```

Can we add a regression test for `sig { returns(...) } attr_accessor` hidden behind a prepended module too?

lib/tapioca/gem/listeners/methods.rb
```diff
@@ -65,10 +65,12 @@ def compile_directly_owned_methods(
         #| ) -> void
         def compile_method(tree, symbol_name, constant, method, visibility = RBI::Public.new)
           return unless method
-          return unless method_owned_by_constant?(method, constant)
+
+          method = method_defined_by_constant(method, constant)
```
KaanOzkan: Agent brought up that the method visibility can also be incorrectly attributed to the prepended method, could you add a test and a fix for this?

```rb
it "uses the constant's own method visibility instead of the prepended method's visibility" do
  add_ruby_file("foo.rb", <<~RUBY)
    module Foo
      def bar(x); end
    end

    class Baz
      prepend Foo

      private

      def bar; end
    end
  RUBY

  output = template(<<~RBI)
    class Baz
      include ::Foo

      private

      def bar; end
    end

    module Foo
      def bar(x); end
    end
  RBI

  assert_equal(output, compile)
end
```

spec/tapioca/gem/pipeline_spec.rb
```diff
@@ -1847,6 +1847,410 @@ def bar; end
...
+        end
+      RBI
+
+      assert_equal(output, compile)
+    end
+
+    it "compiles each method with its own sig when both the class and the prepended module have one" do
+      # Sorbet files the class's sig under the prepended module's method, which is also where the module's
+      # own sig is filed, so whichever sig is evaluated last overwrites the other one.
+      skip "Sorbet keeps only one of the two signatures"
```
KaanOzkan: I'm in support of removing this test, I don't see it being changed in the future.

lib/tapioca/gem/listeners/methods.rb
```diff
@@ -215,12 +245,16 @@ def attr_reader_for_writer(method, constant)
...
+          # otherwise be looked at instead, and its source location would never match the writer's.
+          reader_method = method_defined_by_constant(resolved_reader, constant)
+          return unless reader_method
+
+          reader_method = original_method(reader_method, constant)
           return unless same_source_location?(method, reader_method)
-          return unless method_owned_by_constant?(reader_method, constant)
 
-          reader_method
+          method_defined_by_constant(reader_method, constant)
```
KaanOzkan: Is this second call necessary?

spec/tapioca/gem/pipeline_spec.rb
```diff
@@ -1847,6 +1847,410 @@ def bar; end
       assert_equal(output, compile)
     end
 
+    it "compiles a method using its own signature, not the signature of a module prepended in front of it" do
```
KaanOzkan: This and the next test shouldn't say "signature" IMO since it's not Sorbet signatures. Maybe `method definition`?

spec/tapioca/gem/pipeline_spec.rb
```diff
@@ -1847,6 +1847,410 @@ def bar; end
...
+          include ::B
+
+          def bar; end
+        end
+      RBI
+
+      assert_equal(output, compile)
+    end
+
+    it "compiles a method with a sig using its own signature when the prepended module has different parameters" do
```
KaanOzkan: Could you add a singleton equivalent for this?

```rb
it "compiles a singleton method using its own signature through a prepended module" do
  add_ruby_file("foo.rb", <<~RUBY)
    module Wrapper
      def bar(x, y); end
    end

    class Baz
      extend T::Sig

      sig { params(x: Integer).returns(Integer) }
      def self.bar(x)
        x
      end

      singleton_class.prepend(Wrapper)
    end
  RUBY

  output = template(<<~RBI)
    class Baz
      extend ::Wrapper

      class << self
        sig { params(x: ::Integer).returns(::Integer) }
        def bar(x); end
      end
    end

    module Wrapper
      def bar(x, y); end
    end
  RBI

  assert_equal(output, compile)
end
```

KaanOzkan (commented):
I tried to do this for [DSL generation](https://github.com/Shopify/tapioca/pull/2706) but it was too complex. I think this is acceptable.

